PR #38, opened 2018-05-02
continue to output files

### LINKED ISSUE #17: Do not exit on errors
If we call `bat file1 file2 file3`, we should still output `file3` if `file2` is erroneous.

The exit code should still be nonzero.

As another example, see #27

rleungx:
Closes #17.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -382,11 +382,23 @@ fn run() -> Result<()> {
                 let mut exit = false;
                 for file in files {
                     let line_changes = get_git_diff(&file.to_string());
-                    print_file(&options, theme, &assets.syntax_set, file, &line_changes)
-                        .unwrap_or_else(|e| {
-                            exit = true;
-                            eprintln!("{}: {}: {}", Red.paint("[bat error]"), file, e);
-                        });
+                    let res = print_file(&options, theme, &assets.syntax_set, file, &line_changes);
+                    if let Err(error) = res {
+                        if match error {
+                            Error(ErrorKind::Io(ref io_error), _)
+                                if io_error.kind() == io::ErrorKind::BrokenPipe =>
+                            {
+                                true
+                            }
+                            _ => {
+                                exit = true;
+                                eprintln!("{}: {}: {}", Red.paint("[bat error]"), file, error);
+                                false
+                            }
+                        } {
+                            return Err(error);
+                        }
+                    }
                 }
                 if exit {
                     process::exit(1);
--- a/src/main.rs
+++ b/src/main.rs
@@ -379,9 +379,17 @@ fn run() -> Result<()> {
             })?;
 
             if let Some(files) = app_matches.values_of("FILE") {
+                let mut exit = false;
                 for file in files {
                     let line_changes = get_git_diff(&file.to_string());
-                    print_file(&options, theme, &assets.syntax_set, file, &line_changes)?;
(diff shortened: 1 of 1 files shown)


sharkdp:
Welcome, and thank you very much!

There is at least one possible error (inside `print_file`) where we still want to exit immediately (io::Error with kind = BrokenPipe). Could we somehow filter for that and pass it down to `run`?

src/main.rs
```diff
@@ -379,9 +379,29 @@ fn run() -> Result<()> {
             })?;
 
             if let Some(files) = app_matches.values_of("FILE") {
+                let mut exit = false;
                 for file in files {
                     let line_changes = get_git_diff(&file.to_string());
-                    print_file(&options, theme, &assets.syntax_set, file, &line_changes)?;
+                    let res = print_file(&options, theme, &assets.syntax_set, file, &line_changes);
+                    if let Err(error) = res {
+                        if match error {
```
sharkdp: This looks correct, but the logic is quite convoluted and a bit hard to understand. The `match` statement looks good but if would be great if we could get rid of the additional `if` and the `exit` variable`.

I know it's not easy due to borrowing issues (I shortly tried myself yesterday :smile:), but it would be great if we could clean this up a little bit. If you don't want to invest more time, I can also merge this as-is and work on this later.

pitkley: My take on this:

```diff
diff --git a/src/main.rs b/src/main.rs
index e5a2bb2..95ee25d 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -383,21 +383,15 @@ fn run() -> Result<()> {
                 for file in files {
                     let line_changes = get_git_diff(&file.to_string());
                     let res = print_file(&options, theme, &assets.syntax_set, file, &line_changes);
-                    if let Err(error) = res {
-                        if match error {
-                            Error(ErrorKind::Io(ref io_error), _)
-                                if io_error.kind() == io::ErrorKind::BrokenPipe =>
-                            {
-                                true
-                            }
-                            _ => {
-                                exit = true;
-                                eprintln!("{}: {}: {}", Red.paint("[bat error]"), file, error);
-                                false
-                            }
-                        } {
-                            return Err(error);
+
+                    match res {
+                        Err(Error(ErrorKind::Io(ref io_error), _))
+                            if io_error.kind() == io::ErrorKind::BrokenPipe => {}
+                        Err(error) => {
+                            exit = true;
+                            eprintln!("{}: {}: {}", Red.paint("[bat error]"), file, error);
                         }
+                        _ => {}
                     }
                 }
                 if exit {
```

I don't think having the `exit` variable is bad. Working around that would probably require returning a list of/the error to `main`, which I don't think is really worth it.

rleungx: @pitkley But it seems that it won't exit immediately when the BrokenPipe error happens, right?

pitkley: @rleungx ah, I misunderstood your code on that part. That will probably make the change more difficult, due to borrowing issues. Not having a good alternative then, after all.

sharkdp: Ok, let's try to merge this as-is (after the conflicts have been resolved). @rleungx Sorry for the delay. Let me know, if you need any help.

sharkdp: Ok, thank you for your support. I will look into this in the next days.

sharkdp:
Closed in favor of #136. @rleungx Thank you very much.

Issue #2, opened 2018-04-21
Add support for reading from stdin?


sharkdp:
Also, allow `-` as a path argument

Mte90:
Yes please!
If we can use bat as alias for cat as a perfect replacement will be pretty cool :-D

sharkdp:
All of these are supported now:

``` bash
# read from keyboard
bat

# read c++ from keyboard
bat -l c++

# read from stdin/file
bat -l md < README.md

# read from stdin/pipe
bat README.md | bat -l md

# output main.rs, then Cargo.toml, then printer.rs
bat src/main.rs - src/printer.rs < Cargo.toml

# write something to a file quickly
bat > file

# see issue '#4':
curl -s https://raw.githubusercontent.com/sharkdp/bat/master/README.md | bat -l md
```

PR #39, opened 2018-05-02
Add option to specify language

### LINKED ISSUE #19: Add a --language option to explicitly specify the programming language

BrainMaestro:
Closes #19

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -22,7 +22,7 @@ mod terminal;
 use std::collections::HashMap;
 use std::env;
 use std::fs::{self, File};
-use std::io::{self, BufRead, StdoutLock, Write};
+use std::io::{self, BufRead, BufReader, StdoutLock, Write};
 use std::path::Path;
 use std::process;
 
@@ -35,7 +35,7 @@ use directories::ProjectDirs;
 use git2::{DiffOptions, IntoCString, Repository};
 
 use syntect::dumps::{dump_to_file, from_binary, from_reader};
-use syntect::easy::HighlightFile;
+use syntect::easy::HighlightLines;
 use syntect::highlighting::{Theme, ThemeSet};
 use syntect::parsing::SyntaxSet;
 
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -93,6 +94,13 @@ fn print_file<P: AsRef<Path>>(
     line_changes: &Option<LineChanges>,
 ) -> Result<()> {
     let mut highlighter = HighlightFile::new(filename.as_ref(), syntax_set, theme)?;
+    if let Some(lang) = options.language {
+        let syntax = syntax_set.syntaxes().iter()
+            .find(|s| s.name.to_lowercase() == lang.to_lowercase());
```
BrainMaestro: Is there a better way to compare strings by ignoring case? I couldn't find any, and this isn't the best because of the extra allocations

sharkdp: There is [`eq_ignore_ascii_case`](https://doc.rust-lang.org/stable/std/primitive.str.html#method.eq_ignore_ascii_case), which should be good enough for our purposes (we don't need Unicode support). We might need to bump the minimum-required Rust version for this.

src/main.rs
```diff
@@ -93,6 +94,13 @@ fn print_file<P: AsRef<Path>>(
...
+        if syntax.is_some() {
```
sharkdp: Can we use `if let Some(syntax) = syntax {...}` here to avoid the `.unwrap()`?

src/main.rs
```diff
@@ -93,6 +94,13 @@ fn print_file<P: AsRef<Path>>(
...
+            highlighter.highlight_lines = HighlightLines::new(syntax.unwrap(), theme);
```
sharkdp: If we set the language in this way, `HighlightLines` will be constructed twice (once here, once in the `HighlightFile::new` call. I'm not sure if this has any performance implications.

Maybe it would be better to use `HighlightLines` directly (see example in https://docs.rs/syntect/2.0.1/syntect/easy/struct.HighlightLines.html) instead of using `HighlightFile`. This would also allow for more flexibility in the future (e.g. concerning #2)

src/main.rs
BrainMaestro: I don't know how I missed that. Thanks!

src/main.rs
BrainMaestro: Hmm, not sure why I didn't do it that way 😕

PR #43, opened 2018-05-02
Add options to disable

### LINKED ISSUE #5: Add option to disable line numbers

nakulcg:
* line number
* git modification marker
* file name

fix #5

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -46,7 +46,7 @@ lazy_static! {
 }
 
 mod errors {
-    error_chain!{
+    error_chain! {
         foreign_links {
             Io(::std::io::Error);
         }
@@ -57,6 +57,9 @@ use errors::*;
 
 struct Options {
     true_color: bool,
+    file_name: bool,
+    line_number: bool,
+    git_modification_marker: bool,
 }
 
 #[derive(Copy, Clone, Debug)]
@@ -101,15 +104,17 @@ fn print_file<P: AsRef<Path>>(
     let (_, term_width) = term.size();
     let term_width = term_width as usize;
 
-    print_horizontal_line(&mut handle, '┬', term_width)?;
+    if options.file_name {
+        print_horizontal_line(&mut handle, '┬', term_width)?;
 
-    writeln!(
-        handle,
-        "{}{} File {}",
-        " ".repeat(PANEL_WIDTH),
-        Fixed(GRID_COLOR).paint("│"),
-        White.bold().paint(filename.as_ref().to_string_lossy())
-    )?;
+        writeln!(
+            handle,
+            "{}{} File {}",
+            " ".repeat(PANEL_WIDTH),
+            Fixed(GRID_COLOR).paint("│"),
+            White.bold().paint(filename.as_ref().to_string_lossy())
+        )?;
+    }
 
     print_horizontal_line(&mut handle, '┼', term_width)?;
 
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -350,6 +363,24 @@ fn run() -> Result<()> {
...
+        .arg(
+            Arg::with_name("disable-line-number")
+                .short("n")
+                .long("disable-line-number")
+                .help("Disable line number"),
+        )
+        .arg(
+            Arg::with_name("disable-git-modfication-marker")
+                .short("g")
+                .long("disable-git-modfication-marker")
```
sharkdp: typo: modfication => modification. Maybe just `disable-git-markers`?

src/main.rs
```diff
@@ -350,6 +363,24 @@ fn run() -> Result<()> {
...
+        .arg(
+            Arg::with_name("disable-file-name")
+                .short("f")
+                .long("disable-file-name")
+                .help("Disable file name"),
+        )
+        .arg(
+            Arg::with_name("disable-line-number")
+                .short("n")
+                .long("disable-line-number")
```
sharkdp: maybe `disable-line-numbers`?

src/main.rs
```diff
@@ -350,6 +363,24 @@ fn run() -> Result<()> {
                 .multiple(true)
                 .empty_values(false),
         )
+        .arg(
+            Arg::with_name("disable-file-name")
+                .short("f")
+                .long("disable-file-name")
```
sharkdp: maybe `disable-file-names`?

nakulcg:
Thanks for such a quick feedback. Sure I would say 
* We can either retain my additions with the small changes you have suggested
* We can have a single clap arg like as you suggested with below possible values
** plain/minimal - only content, no  horizontal separators above and below, no left sided vertical lines
** numbers - only numbers
** full - numbers, and markers

In addition we can change filename to only show if multiple files

What do you think? I am leaning for the 2nd one

sharkdp:
> We can have a single clap arg like as you suggested with below possible values
> plain/minimal - only content
> numbers - only numbers
> full - numbers, and markers

Sounds great.

> We can change filename to only show if multiple files

I thouhgt about this again.. the header bar with the file name is not super-helpful for a single file right now, but I was planning to add some more information to that line (git status, number of added/deleted/modified lines, last modified by, etc.). We should definitely disable the header for `--style=plain`, but I think I'd like to keep it enabled by default, for now.
 
> In addition we can show the horizontal lines only in case of multiple files

If we keep the header, I'd also like to keep the horizontal lines for now. Again, they should definitely be removed for `--style=plain`. Maybe we could later add some `--grid` option that would allow us to set the style for all grid lines (off, only vertical lines, plain ASCII characters instead of Unicode block-drawing symbols, ...).

Thoughts?

nakulcg:
Sure. Then lets do the style option with
* plain - no line numbers and no modification markers
* line-numbers - only line numbers
* full - line numbers and markers for now, and in future other things

sharkdp:
Thank you!

sharkdp:
Oh, for some reason I thought it would completely disable the horizontal/vertical bars for `--style=plain`.

sharkdp:
> and retained it as we wanted to retain the name anyway

I think "plain" should really just be the file contents, nothing else. Sorry if I confused you with the file-name/header thing.

> and with multiple files this would anyway be an issue.

Why? It would be the same as the normal `cat` behavior. The content of the different files would be concatenated.


Would you agree with these statements? If yes, would you like to address this in another PR? Otherwise, I can also implement this myself.

In any case, thanks!

PR #49, opened 2018-05-03
Add less pager

### LINKED ISSUE #29: Automatic paging
First off, this is nice :).

Have you thought about adding git-style automatic paging to it? If the lines fit on one screen, it would not do paging (like the current behavior). But as soon as it's more than one screen, it could automatically page.

That's one of the things that git got right IMO.

BrainMaestro:
Closes #29 

The huge diff (just whitespace) is because I needed to introduce a new scope because `wait()` borrows mutably

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,8 +55,9 @@ script:
   - cargo build --target $TARGET --verbose
   - cargo test --target $TARGET --verbose
   # Run 'bat' on its own source code and the README
-  - cargo run --target $TARGET -- src/main.rs
-  - cargo run --target $TARGET -- README.md
+  # Piping to 'cat' forces the tty check to fail so that a pager is not used
+  - cargo run --target $TARGET -- src/main.rs | cat
+  - cargo run --target $TARGET -- README.md | cat
 
 before_deploy:
   - bash ci/before_deploy.bash
--- a/src/main.rs
+++ b/src/main.rs
@@ -22,9 +22,9 @@ mod terminal;
 use std::collections::HashMap;
 use std::env;
 use std::fs::{self, File};
-use std::io::{self, BufRead, BufReader, StdoutLock, Write};
+use std::io::{self, BufRead, BufReader, Stdout, StdoutLock, Write};
 use std::path::Path;
-use std::process;
+use std::process::{self, Command, Child, Stdio};
 
 use ansi_term::Colour::{Fixed, Green, Red, White, Yellow};
 use ansi_term::Style;
@@ -65,6 +65,42 @@ struct Options<'a> {
     true_color: bool,
     style: OptionsStyle,
     language: Option<&'a str>,
+    interactive_terminal: bool,
+}
+
+enum OutputType<'a> {
+    Pager(Child),
+    Stdout(StdoutLock<'a>),
+}
+
+impl<'a> OutputType<'a> {
+    fn new_pager() -> Result<Self> {
+        Ok(OutputType::Pager(Command::new("less")
+            .args(&["--quit-if-one-screen", "--RAW-CONTROL-CHARS", "--no-init"])
+            .stdin(Stdio::piped())
+            .spawn()
+            .chain_err(|| "Could not spawn pager")?
+        ))
+    }
+
+    fn new_stdout(stdout: &'a Stdout) -> Self {
+        OutputType::Stdout(stdout.lock())
+    }
+
+    fn stdout(&mut self) -> Result<&mut Write> {
+        Ok(match *self {
+            OutputType::Pager(ref mut command) => command.stdin.as_mut().chain_err(|| "Could not open stdin for pager")?,
+            OutputType::Stdout(ref mut handle) => handle,
+        })
+    }
+}
+
+impl<'a> Drop for OutputType<'a> {
+    fn drop(&mut self) {
+        if let OutputType::Pager(ref mut command) = *self {
+            let _ = command.wait();
+        }
+    }
 }
 
 #[derive(Copy, Clone, Debug)]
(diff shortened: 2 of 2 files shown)


BrainMaestro:
I see the build failed. How do we test this since less doesn't quit automatically, so there'll be no output to stdout

sharkdp:
Thank you very much for working on this!

> How do we test this since less doesn't quit automatically, so there'll be no output to stdout

We should test if the output is an interactive terminal (via `atty::is(Stream::Stdout)`.. this is already used somewhere) and only spawn the pager in this case. It might be worth to save `interactive_terminal: bool` in the `Options` struct. We are going to need this in a few places in the future.

cpp
```diff
@@ -0,0 +1,8 @@
+#include <iostream>
```
sharkdp: can this file be removed?

BrainMaestro: whoa mistake. I just used it to test the language flag

src/main.rs
```diff
@@ -93,55 +93,57 @@ fn print_file<P: AsRef<Path>>(
     line_changes: &Option<LineChanges>,
 ) -> Result<()> {
     let mut highlighter = HighlightFile::new(filename.as_ref(), syntax_set, theme)?;
+    let mut pager = Command::new("less").arg("-FRX").stdin(Stdio::piped()).spawn()?;
+    {
+        let mut handle: &mut Write = pager.stdin.as_mut().expect("could not open stdin for pager");
```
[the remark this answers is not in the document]
BrainMaestro: Oh yeah. not very familiar with `error_chain`

src/main.rs
```diff
@@ -93,55 +93,57 @@ fn print_file<P: AsRef<Path>>(
     line_changes: &Option<LineChanges>,
 ) -> Result<()> {
     let mut highlighter = HighlightFile::new(filename.as_ref(), syntax_set, theme)?;
+    let mut pager = Command::new("less").arg("-FRX").stdin(Stdio::piped()).spawn()?;
```
sharkdp: If there are long option alternatives for these flags (like `-F` => `--quit-if-one-screen`), I would prefer to use those for readability purposes.

BrainMaestro:
I'm not sure about the naming, but this seems like a better structure. let me know what to change

src/main.rs
```diff
@@ -58,6 +58,42 @@ use errors::*;
 struct Options<'a> {
     true_color: bool,
     language: Option<&'a str>,
+    interactive_terminal: bool,
+}
+
+enum Shell<'a> {
```
sharkdp: Very cool!

Maybe call this `OutputTerminal` instead of `Shell`?

src/main.rs
```diff
@@ -58,6 +58,42 @@ use errors::*;
...
+
+    fn stdout(&mut self) -> Result<&mut Write> {
+        Ok(match self {
+            Shell::Interactive(ref mut command) => command.stdin.as_mut().chain_err(|| "Could not open stdin for pager")?,
+            Shell::NonInteractive(ref mut handle) => handle,
+        })
+    }
+}
+
+impl<'a> Drop for Shell<'a> {
```
sharkdp: Nice! :+1:

src/main.rs
```diff
@@ -58,6 +58,42 @@ use errors::*;
...
+    interactive_terminal: bool,
+}
+
+enum Shell<'a> {
+    Interactive(Child),
+    NonInteractive(StdoutLock<'a>),
+}
+
+impl<'a> Shell<'a> {
+    fn new(interactive: bool, stdout: &'a Stdout) -> Result<Self> {
```
sharkdp: I think I'd prefer if this would be split into two methods (`new_*`) instead of passing a boolean argument. The if-else would have to be at the call site.

What do you think?

src/main.rs
sharkdp: Another thought:

Call this `OutputType` and name the variants `Pager` and `Stdout`? The link between interactive=>pager, non-interactive=>stdout happens somewhere else.

src/main.rs
```diff
@@ -65,6 +65,42 @@ struct Options<'a> {
...
+    fn new_pager() -> Result<Self> {
+        Ok(OutputType::Pager(Command::new("less")
+            .args(&["--quit-if-one-screen", "--RAW-CONTROL-CHARS", "--no-init"])
+            .stdin(Stdio::piped())
+            .spawn()
+            .chain_err(|| "Could not spawn pager")?
+        ))
+    }
+
+    fn new_stdout(stdout: &'a Stdout) -> Self {
```
BrainMaestro: I would prefer not to pass in the stdout, but I can't find a way around the lifetime issues

src/main.rs
BrainMaestro: Sure that makes sense

.travis.yml
```diff
@@ -55,8 +55,9 @@ script:
   - cargo build --target $TARGET --verbose
   - cargo test --target $TARGET --verbose
   # Run 'bat' on its own source code and the README
-  - cargo run --target $TARGET -- src/main.rs
-  - cargo run --target $TARGET -- README.md
+  # Piping to 'cat' forces the tty check to fail so that a pager is not used
+  - cargo run --target $TARGET -- src/main.rs | cat
```
BrainMaestro: Do you know another way of doing this so the pager is not used (which causes the build to fail)

BrainMaestro: Do you know another way of doing this so the pager is not used (which causes the build to fail)

sharkdp: We can introduce a command-line flag in the future, but I think this is fine for now. It was just meant as a really basic "smoke" test

sharkdp:
Thank you very much!

PR #61, opened 2018-05-05
Fix git markers

nakulcg:
BTW I can revert the last commit which reformats code if you dont want


src/main.rs
```diff
@@ -231,13 +230,16 @@ fn print_file<P: AsRef<Path>>(
 }
 
 fn get_git_diff(filename: &str) -> Option<LineChanges> {
-    let repo = Repository::open_from_env().ok()?;
+    let repo = Repository::discover(Path::new(&filename)).ok()?;
     let workdir = repo.workdir()?;
-    let current_dir = env::current_dir().ok()?;
-    let filepath = current_dir.join(Path::new(&filename));
+    let absolute_file_path = workdir.join(Path::new(&filename));
```
sharkdp: This will only work if the path `filename` starts at the root of the workdir. I have a fix for this and will push it to you branch. Hope that's okay.

PR #62, opened 2018-05-05
Add color flag

### LINKED ISSUE #26: Default behavior when output is sent to a file
I, for one, prefer the way `exa` handles this. To wit,
battest

With exa, I've aliased `ls` and everything's been fine. I'd consider doing the same for bat if we could get non-colorized output and byte-for-byte copying when the output isn't a terminal ("catting to a file"). Perhaps via a config. If that's not your goal, that's fine too. :+1: 

Cool project, btw.

BrainMaestro:
Closes  #26 

Colors are disabled if the terminal is not interactive unless explicitly
set otherwise

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -66,6 +66,7 @@ struct Options<'a> {
     style: OptionsStyle,
     language: Option<&'a str>,
     interactive_terminal: bool,
+    colored_output: bool,
 }
 
 enum OutputType<'a> {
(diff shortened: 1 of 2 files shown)


BrainMaestro:
We may want to start thinking about splitting the main file into separate modules

BrainMaestro:
After quite some research, it seems like it does not actually print the ANSI codes at all. However, the syntax coloring still produces ANSI codes. I have a fix that I don't like. Let me know what you think

sharkdp:
> I have a fix that I don't like. Let me know what you think

I think this is good enough for now - we can look into other options later on. Thank you very much!

PR #63, opened 2018-05-05
Add style for line-changes only

pitkley:
This PR adds another style to the three existing ones: "line-changes". In comparison to the plain style this style adds the git-diff as the first column, in comparison to the full-style this leaves out the grid and filename.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -57,6 +57,7 @@ use errors::*;
 
 enum OptionsStyle {
     Plain,
+    LineChanges,
     LineNumbers,
     Full,
 }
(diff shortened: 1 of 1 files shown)


sharkdp:
Welcome and thank you very much for your contribution!


If I understand correctly, this shows both line numbers (`line`) and git modifications (`changes`)?

While this request sounds completely reasonable, I'm not quite sure about the way the `--style` option works (even without your PR, see also #43). Maybe we should change this again to work similar to this:

`--style=plain` => `--style=""`
`--style=line-numbers` => `--style="numbers,grid,header"`
`--style=line-changes` => `--style="numbers,changes"`
`--style=full` => `--style="numbers,changes,grid,header"`

We could still provide and alias for some variants (like "full"). What do you think?

sharkdp:
> Should the order of items the user provides be respected? I.e. should numbers,changes be different from changes,numbers?

No, I don't think it should. Line numbers would always be displayed on the left-hand side.

> Personally I'd be fine with predetermining an order and simply using the --style as "enable"-switches for the respective information.

Exactly :+1: 

If you would like to work on this - great! If not, I can also open a new ticket to track this.

pitkley:
@sharkdp thanks for your input, I have implemented the change in #74! :+1:

PR #74, opened 2018-05-06
 Make `--style` parameter more flexible

pitkley:
The parameter now accepts a comma-separated list of strings, where every element defines either a single output component (`changes`, `grid`, `header`, `numbers`) or a predefined style (`full`, `line-numbers`, `plain`).

If available, bat picks the first predefined style in the user-supplied style-list and ignores everything else. If no predefined style was requested, the other parameters that are simple output components will be used.

Examples:

1. `--style changes,full,numbers`
    Will be reduced to only the predefined style `full`.

2. `--style plain,full`
    Will be reduced to only the predefined style `plain`.

3. `--style changes,numbers`
    Will not be reduced, because the list does not contain any predefined styles.

(Note: if `grid` is requested but no other parameters, bat still creates the left-most column with a width of `PANEL_WIDTH`. I didn't want to introduce further logic in this PR that drops or adapts the width of the left column.)

<sup>(This PR supersedes #63.)</sup>


pitkley:
I have updated the PR to work with the new `Printer`. (I have also decreased visibility modifiers where possible to not publish to much of the API we don't want public. I kept most of it on a separate commit, in case this is not wanted.)

From my tests this seems to work correctly, but I'd appreciate if somebody could verify this! :+1:

src/main.rs
```diff
@@ -50,23 +53,82 @@ mod errors {
...
+}
+
+impl FromStr for PredefinedStyle {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self> {
+        match s {
+            "full" => Ok(PredefinedStyle::Full),
+            "plain" => Ok(PredefinedStyle::Plain),
+            _ => Err(ErrorKind::UnknownStyleName(s.to_owned()).into()),
```
BrainMaestro: Should it fail or just default to full? The other options take this more forgiving style

BrainMaestro:
Hey @pitkley, seems good, but the parsing logic seems a little complicated. Something like [this](http://play.rust-lang.org/?gist=cc034b7f1a7bd263ae0a2f891c974df5&version=stable&mode=debug) that uses a `HashSet` could be another way. Let me know what you think.

src/main.rs
pitkley: Defaulting to `full` happens in the clap-configuration: the default value for `--style` is `full`. Also, defaulting here would negate the selection-logic.

src/main.rs
```diff
@@ -417,8 +479,7 @@ fn run() -> Result<()> {
         .arg(
             Arg::with_name("style")
                 .long("style")
-                .possible_values(&["auto", "plain", "line-numbers", "full"])
```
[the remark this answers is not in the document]
pitkley: I have updated the handling of the comma-separated argument and could re-add the possible values! :+1:

src/printer.rs
```diff
@@ -37,26 +38,39 @@ impl<'a> Printer<'a> {
...
+                "{}{} File {}",
+                " ".repeat(PANEL_WIDTH),
+                self.colors.grid.paint("│"),
+                self.colors.filename.paint(filename)
+            )?;
+
+            self.print_horizontal_line('┼')
+        } else {
+            writeln!(self.handle, "File {}", self.colors.filename.paint(filename))
+                .map_err(Into::into)
```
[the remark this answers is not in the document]
pitkley: Since I am returning the result here, no. It would complain about getting a `Result<_, io::Error>` instead of `Result<_, Error>`. I can use `?;` though and simply return `Ok(())` at the end instead.

Let me know if you think that would be easier to understand!

src/printer.rs
```diff
@@ -68,70 +82,93 @@ impl<'a> Printer<'a> {
...
-
-        Ok(())
+        let line = as_terminal_escaped(
+            &regions,
+            self.options.true_color,
+            self.options.colored_output,
+        );
+
+        // Print decorations
+        decorations
```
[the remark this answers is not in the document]
pitkley: After thinking about it a bit more, I have come up with [the following solution](https://github.com/sharkdp/bat/pull/74/commits/a22af299a178993056d54b56a00b013df41ba2c8):

```rust
let line_number_output = self.print_line_number(line_number).unwrap_or("".into());
let git_marker_output = self.print_git_marker(line_number).unwrap_or("".into());
let line_border_output = self.print_line_border().unwrap_or("".into());
let line = as_terminal_escaped(
    &regions,
    self.options.true_color,
    self.options.colored_output,
);

write!(
    self.handle,
    "{}{}{}{}",
    line_number_output, git_marker_output, line_border_output, line
).map_err(Into::into)
```

What do you think about that?

sharkdp:
I'm sorry for not being too active on this pull request, but unfortunately this is quite hard to review as there are many changes that are unrelated to the actual feature. While some of these changes are really good, it would be better to split these into separate pull requests (in the future).

I agree with @pitkley, the parsing logic seems quite complex. Maybe things would also be simpler if we remove `plain` and `full` completely and simply parse a list of components. The default value would simply be a list of all decorations.

pitkley:
I have rebased the PR and performed a few changes:

* I have tried to change less of the surroundings. (The `Cow` change in printer has to stay though if we don't want to needlessly allocate strings required for grid-handling.)
* I have pushed another commit on top. This one combines the two enums `OutputComponent` and `PredefinedStyle` into one and significantly reduces the parsing logic.

    This is at the "expense" of the supplied styles now being additive, rather than one predefined style overruling. But this is probably more in line to what an end-user would expect.

    (For more details, see the [commit message](https://github.com/sharkdp/bat/pull/74/commits/71684ead20d29b7762f733bebe59590f3c436aaa).)

Let me know if this change fits better, or if there are more open points! :+1:

src/main.rs
```diff
@@ -545,18 +598,17 @@ fn run() -> Result<()> {
                 })
                 .unwrap_or_else(|| vec![None]); // read from stdin (None) if no args are given
 
+            let output_components = values_t!(app_matches.values_of("style"), OutputComponent)?
```
BrainMaestro: very nice 👍

src/main.rs
```diff
@@ -545,18 +598,17 @@ fn run() -> Result<()> {
...
+                .into_iter()
+                .map(|style| style.components())
+                .fold(vec![], |mut acc, x| {
```
BrainMaestro: `Vec::new()` would be better since it's empty right?

src/printer.rs
```diff
@@ -31,32 +32,48 @@ impl<'a> Printer<'a> {
...
-            filename.map_or("", |_| "File: "),
-            self.colors.filename.paint(filename.unwrap_or("STDIN"))
-        )?;
-
-        self.print_horizontal_line('┼')
+        if self.options
+            .output_components
+            .contains(&OutputComponent::Grid)
+        {
+            self.print_horizontal_line('┬')?;
```
BrainMaestro: Since this function is only relevant for grid, you could put a check in the function itself to print nothing if grid is not present. it would simplify a lot. what do you think?

src/printer.rs
```diff
@@ -31,32 +32,48 @@ impl<'a> Printer<'a> {
...
+            writeln!(
+                self.handle,
+                "{}{}",
+                filename.map_or("", |_| "File: "),
+                self.colors.filename.paint(filename.unwrap_or("STDIN"))
+            )?;
+
+            self.print_horizontal_line('┼')
+        } else {
+            writeln!(
```
BrainMaestro: This is not quite the same behavior. probably best to avoid the duplication

src/printer.rs
```diff
@@ -72,66 +89,95 @@ impl<'a> Printer<'a> {
...
+            .output_components
+            .contains(&OutputComponent::Grid);
         write!(
             self.handle,
             "{}",
             decorations
                 .into_iter()
-                .filter_map(|dec| dec)
+                .filter_map(|dec| if grid_requested {
+                    Some(dec.unwrap_or(" ".into()))
```
BrainMaestro: Hmm not sure I follow

BrainMaestro:
Nice, this is much easier to follow!
- The `Cow` change, it probably has a negligible effect since we're still building strings after calling `paint`. Only 2-3 places where we have `&str` that can benefit while introducing many `into` calls.
- As an extra suggestion, it would be nice if the checks could look like `self.options.style.grid()` and `self.options.style.headers()`. A simple tuple struct that wraps `Vec<OutputComponent>` would do this. 
- Also, wouldn't HashSet be better since there can be duplicates if the user passes `--style=changes,full,header`. Not that it affects anything

src/main.rs
pitkley: `vec![]` is empty, too. I guess this is just preference, but results in the same code. ([1](https://github.com/rust-lang-nursery/rust-clippy/issues/1542), [2](https://github.com/rust-lang-nursery/fmt-rfcs/issues/70))

src/printer.rs
pitkley: That depends: the way it is implemented here right now, the header can also be shown regardless of if the grid is enabled or not. I suppose this is a decision to make: do we want to display a header if the grid is not requested and if so, how should it be formatted?

src/printer.rs
pitkley: True, thanks for noticing. I'll leave this unchanged for the time being until we have a decision on whether `header` should have an effect without `grid`.

src/printer.rs
pitkley: In a grid we have to at least have a space for a non-existing element to have the right spacing in the left column. Unfortunately I haven't found a cleaner way to express this spacing, without going back to the multiple writes.

pitkley:
> The `Cow` change, it probably has a negligible effect since we're still building strings after calling `paint`. Only 2-3 places where we have `&str` that can benefit while introducing many into calls.

Agreed. I'll go back to `Option<String>`!

> As an extra suggestion, it would be nice if the checks could look like self.options.style.grid() and self.options.style.headers(). A simple tuple struct that wraps Vec<OutputComponent> would do this.

I was thinking about how to make this simpler. I like your suggestion and will give it a try!

> Also, wouldn't HashSet be better since there can be duplicates if the user passes --style=changes,full,header. Not that it affects anything

Good point, I'll update that too.

----

Thanks for the thorough reviews, by the way. I really appreciate them!

pitkley:
Newest updates:

* I have replace `Cow` by `String`. (It doesn't reduce as much noise since we need to convert `&str` to `String`, but it seems better anyway.)
* I have replaced `Vec` by `HashSet` and wrapped the resulting `HashSet<OutputComponent>` into `struct OutputComponents` as suggested by @BrainMaestro. I kept the parameter name `output_components` to be in line with the struct, since renaming the struct to `Style` would have clashed with the `Style` from ansi_term. (If `Style` would be preferred though, I'd suggest going for something like `use ansi_term::Style as AnsiStyle` instead.)

src/printer.rs
BrainMaestro: Sorry if I wasn't clear, I was actually specifically talking about `print_horizontal_line`, but you're right about the header thing. 🤷‍♂️

PR #75, opened 2018-05-06
Add printer

BrainMaestro:
This adds a separate struct that handles printing of the different
tokens.

I think this makes it easier to control the style without making the code too noisy. Just a simple guard check at the top of the relevant function before early exiting. I got it to the same state as master now, but the new proposed style features in #74 would be easier to understand. One caveat of this style is more allocations. Sorry about the huge code dump. Since we didn't discuss this beforehand, it's fine to close if you want to follow a different path.

/cc: @sharkdp @pitkley

### DIFF
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -1,0 +1,145 @@
+use ansi_term::Style;
+use console::Term;
+use errors::*;
+use std::io::Write;
+use syntect::highlighting;
+use terminal::as_terminal_escaped;
+use {Colors, LineChange, LineChanges, Options, OptionsStyle};
+
+const PANEL_WIDTH: usize = 7;
+
+pub struct Printer<'a> {
+    handle: &'a mut Write,
+    colors: Colors,
+    term_width: usize,
+    options: &'a Options<'a>,
+    pub line_changes: Option<LineChanges>,
+}
+
+impl<'a> Printer<'a> {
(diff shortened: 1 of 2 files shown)


sharkdp:
This looks really good - I like the restructuring. This is one of the things I also had in mind concerning splitting things into modules :+1: 

@pitkley I hope you don't mind if I merge this first. Let me know if you want some help in resolving the merge conflicts.

src/printer.rs
```diff
@@ -0,0 +1,145 @@
...
+use terminal::as_terminal_escaped;
+use {Colors, LineChange, LineChanges, Options, OptionsStyle};
+
+const PANEL_WIDTH: usize = 7;
+
+pub struct Printer<'a> {
+    handle: &'a mut Write,
+    colors: Colors,
+    term_width: usize,
+    options: &'a Options<'a>,
```
sharkdp: If some of the options are only needed in the printer, we might also think about inlining them here (instead of passing the whole `Options` struct)? Not sure if that's necessarily better.

BrainMaestro: I thought about that, but we already need 3 of the options in there, and it's likely to grow as well. So, this seemed easier

pitkley:
@sharkdp I'll update my PR, probably this evening or tomorrow! :+1: (Feel free to update my PR if you get to it earlier though!)

PR #79, opened 2018-05-07
Add "--list-languages" command line argument

### LINKED ISSUE #69: Add option to display all supported languages
Maybe:
```
> bat --list-languages
ActionScript			as
AppleScript				applescript
ASP						asa
Batch File				bat, cmd
C#						cs, csx
...
```

See: https://docs.rs/syntect/2.0.1/syntect/parsing/struct.SyntaxSet.html#method.syntaxes

connorkuehl:
Closes #69 

```
$ ./target/debug/bat.exe Cargo.toml --list-languages
ASP                              | asa
HTML (ASP)                       | asp
ActionScript                     | as
AppleScript                      | applescript, script editor
Batch File                       | bat, cmd
NAnt Build File                  | build
C#                               | cs, csx
C++                              | cpp, cc, cp, cxx, c++, C, h, hh, hpp, hxx, h++, inl, ipp
C                                | c, h
--snipped--
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -568,18 +568,23 @@ fn run() -> Result<()> {
             if let Some(_) = app_matches.values_of("list languages") {
                 let languages = assets.syntax_set.syntaxes();
 
-                for lang in languages {
-                    print!("{}\t", lang.name);
+                let longest = match languages.iter()
+                    .map(|s| s.name.len())
+                    .max() {
+                    Some(longlang) => longlang,
+                    None => 32,
+                };
 
+                for lang in languages {
+                    print!("{:width$} | ", lang.name, width = longest);
                     for i in 0..lang.file_extensions.len() {
                         print!("{}", lang.file_extensions[i]);
 
                         if i < lang.file_extensions.len() - 1 {
                             print!(", ");
-                        } else {
-                            println!();
                         }
                     }
+                    println!();
                 }
 
                 return Ok(());
--- a/src/main.rs
+++ b/src/main.rs
@@ -582,14 +582,15 @@ fn run() -> Result<()> {
                     print!("{:width$}{}", lang.name, separator, width = longest);
 
                     // Line-wrapping for the possible file extension overflow.
-                    let desired_width = options.term_width - longest;
+                    let desired_width = options.term_width - longest - separator.len();
                     // Number of characters on this line so far, wrap before `desired_width`
                     let mut num_chars = 0;
 
+                    let comma_separator = ", ";
                     let mut extension = lang.file_extensions.iter().peekable();
                     while let Some(word) = extension.next() {
                         // If we can't fit this word in, then create a line break and align it in.
-                        if word.len() + num_chars >= desired_width {
+                        if word.len() + num_chars + comma_separator.len() >= desired_width {
                             num_chars = 0;
                             print!("\n{:width$}{}", "", separator, width = longest);
                         }
@@ -597,7 +598,7 @@ fn run() -> Result<()> {
                         num_chars += word.len();
                         print!("{}", word);
                         if extension.peek().is_some() {
-                            print!(", ");
+                            print!("{}", comma_separator);
                         }
                     }
                     println!();
--- a/src/main.rs
+++ b/src/main.rs
@@ -569,7 +569,9 @@ fn run() -> Result<()> {
             if app_matches.is_present("list-languages") {
                 let languages = assets.syntax_set.syntaxes();
 
-                let longest = languages.iter()
+                let longest = languages
+                    .iter()
+                    .filter(|s| !s.hidden)
                     .map(|s| s.name.len())
                     .max()
                     .unwrap_or(32); // Fallback width if they have no language definitions.
@@ -590,12 +592,13 @@ fn run() -> Result<()> {
                     let mut extension = lang.file_extensions.iter().peekable();
                     while let Some(word) = extension.next() {
                         // If we can't fit this word in, then create a line break and align it in.
-                        if num_chars + word.len() + comma_separator.len() >= desired_width {
+                        let new_chars = word.len() + comma_separator.len();
+                        if num_chars + new_chars >= desired_width {
                             num_chars = 0;
                             print!("\n{:width$}{}", "", separator, width = longest);
                         }
 
-                        num_chars += word.len() + comma_separator.len();
+                        num_chars += new_chars;
                         print!("{}", word);
                         if extension.peek().is_some() {
                             print!("{}", comma_separator);
--- a/src/main.rs
+++ b/src/main.rs
@@ -565,6 +565,26 @@ fn run() -> Result<()> {
                 )
             })?;
 
+            if let Some(_) = app_matches.values_of("list languages") {
+                let languages = assets.syntax_set.syntaxes();
+
+                for lang in languages {
+                    print!("{}\t", lang.name);
+
+                    for i in 0..lang.file_extensions.len() {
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -437,6 +437,12 @@ fn run() -> Result<()> {
                 .default_value("auto")
                 .help("When to use the pager"),
         )
+        .arg(
+            Arg::with_name("list languages")
+                .long("list-languages")
+                .takes_value(false)
```
sharkdp: I think `false` is the default value, so `.takes_value` can be removed here.

src/main.rs
```diff
@@ -437,6 +437,12 @@ fn run() -> Result<()> {
                 .default_value("auto")
                 .help("When to use the pager"),
         )
+        .arg(
+            Arg::with_name("list languages")
```
sharkdp: I'd prefer if we use list-languages here, just like the `.long(..)` name.

src/main.rs
```diff
@@ -486,6 +492,31 @@ fn run() -> Result<()> {
                 )
             })?;
 
+            if let Some(_) = app_matches.values_of("list languages") {
```
sharkdp: You can use `if app_matches.is_present("list-languages")`.

src/main.rs
```diff
@@ -486,6 +492,31 @@ fn run() -> Result<()> {
...
+                let languages = assets.syntax_set.syntaxes();
+
+                let longest = match languages.iter()
+                    .map(|s| s.name.len())
+                    .max() {
```
sharkdp: you can use `languages.iter().....max().unwrap_or(32)` here instead of the `match` statement.

src/main.rs
```diff
@@ -486,6 +492,31 @@ fn run() -> Result<()> {
...
+                    Some(length) => length,
+                    None => 32, // Fallback width if they have no language definitions.
+                };
+
+                for lang in languages {
+                    print!("{:width$} | ", lang.name, width = longest);
```
sharkdp: Nice! Didn't know this was possible.

src/main.rs
```diff
@@ -486,6 +492,31 @@ fn run() -> Result<()> {
...
+                    for i in 0..lang.file_extensions.len() {
```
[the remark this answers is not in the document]
sharkdp: I think you should be able to use `lang.file_extensions.join(", ")`.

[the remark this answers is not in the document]
connorkuehl: That worked. Thank you!

connorkuehl:
Yes, we should limit the size of the extensions column. I had no idea some languages could have so many extensions lol.

I implemented simple line-wrapping for it. I'm still trying to learn more Rust, so if it is unnecessarily complex I would love some feedback to trim it down. I tried brainstorming a cool and simple iterator solution to this problem but didn't get too far.

Here's what it looks like:

```
--snipped--
reStructuredText                 | rst, rest
Ruby                             | rb, Appfile, Appraisals, Berksfile, Brewfile, capfile, cgi,
                                 | Cheffile, config.ru, Deliverfile, Fastfile, fcgi, Gemfile,
                                 | gemspec, Guardfile, irbrc, jbuilder, podspec, prawn, rabl,
                                 | rake, Rakefile, Rantfile, rbx, rjs, ruby.rail, Scanfile,
                                 | simplecov, Snapfile, thor, Thorfile, Vagrantfile
Cargo Build Results              |
Rust                             | rs
SQL                              | sql, ddl, dml
--snipped--
```

src/main.rs
```diff
@@ -486,6 +491,43 @@ fn run() -> Result<()> {
...
+                    .map(|s| s.name.len())
+                    .max()
+                    .unwrap_or(32); // Fallback width if they have no language definitions.
+
+                let separator = " | ";
+                for lang in languages {
+                    print!("{:width$}{}", lang.name, separator, width = longest);
+
+                    // Line-wrapping for the possible file extension overflow.
+                    let desired_width = 48;
```
sharkdp: Maybe this could be `desired_width = term_width - longest - padding`?

connorkuehl: How should I get the term_width? I looked around and I'm not seeing a function I can call to get that value.

I notice it's set with a literal `90` in `run()`; should I just create a constant and apply it to both places?

sharkdp: The `term_width` is computed/retrieved in `printer.s`. We should probably move that to the `run` function and store it in the `Options` struct.

sharkdp:
We should probably also filter out Languages without any extensions. These are typically syntax definitions which are included by others, e.g. "Regular Expressions (Python)"

sharkdp:
Actually, we should probably just hide the ones that have `hidden == true` (https://docs.rs/syntect/2.0.1/syntect/parsing/syntax_definition/struct.SyntaxDefinition.html)

src/main.rs
```diff
@@ -559,6 +566,46 @@ fn run() -> Result<()> {
...
+
+                let separator = " | ";
+                for lang in languages {
+                    if lang.hidden {
+                        continue;
+                    }
+                    print!("{:width$}{}", lang.name, separator, width = longest);
+
+                    // Line-wrapping for the possible file extension overflow.
+                    let desired_width = options.term_width - longest;
```
sharkdp: Don't we have to subtract `separator.length()` as well?

src/printer.rs
```diff
@@ -18,8 +17,7 @@ pub struct Printer<'a> {
 
 impl<'a> Printer<'a> {
     pub fn new(handle: &'a mut Write, options: &'a Options) -> Self {
-        let (_, term_width) = Term::stdout().size();
-        let term_width = term_width as usize;
+        let term_width = options.term_width;
```
BrainMaestro: Just replace all the usages of `term_width` with `options.term_width` then

src/main.rs
```diff
@@ -559,6 +566,46 @@ fn run() -> Result<()> {
...
+                    // Number of characters on this line so far, wrap before `desired_width`
+                    let mut num_chars = 0;
+
+                    let mut extension = lang.file_extensions.iter().peekable();
+                    while let Some(word) = extension.next() {
+                        // If we can't fit this word in, then create a line break and align it in.
+                        if word.len() + num_chars >= desired_width {
```
sharkdp: Do we have to check for `word.len() + 2 + num_chars >= ..` here due to the ", " part which is possibly appended?

src/printer.rs
```diff
@@ -30,7 +26,7 @@ impl<'a> Printer<'a> {
         Printer {
             handle,
             colors,
-            term_width,
+            term_width: options.term_width,
```
sharkdp: I think `term_width` can be removed from `Printer` if it is already included in `options`.

connorkuehl:
Thanks for putting up with me! I appreciate it, I learned a lot.

src/main.rs
```diff
@@ -559,6 +566,47 @@ fn run() -> Result<()> {
...
+                    let comma_separator = ", ";
+                    let mut extension = lang.file_extensions.iter().peekable();
+                    while let Some(word) = extension.next() {
+                        // If we can't fit this word in, then create a line break and align it in.
+                        if word.len() + num_chars + comma_separator.len() >= desired_width {
+                            num_chars = 0;
+                            print!("\n{:width$}{}", "", separator, width = longest);
+                        }
+
+                        num_chars += word.len();
```
sharkdp: Actually, this also needs to be `num_chars += word.len() + comma_separator.len();`. But I can also change this, if you want.

PR #86, opened 2018-05-09
Split modules

BrainMaestro:
Simple split of diffing and asset related functionality to their own modules. I can do some more cleanup if this is fine


sharkdp:
Thank you very much, fantastic work!

PR #93, opened 2018-05-11
Split more modules

BrainMaestro:


sharkdp:
Again, really nice work. I also like the Option->Config renaming.

PR #95, opened 2018-05-11
add theme option

### LINKED ISSUE #89: Add --theme option
Add a `--theme <name>` command line option to use a different syntax highlighting theme. This should be pretty straightforward as we do already load the whole `ThemeSet` (and only use the one called "Default").

It would also be useful to add a `--list-themes` option in analogy to `--list-languages`.

This can then be used as a basis to provide a few more highlighting themes, see #34.

rleungx:
Closes #89.

### DIFF
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -127,11 +127,12 @@ impl HighlightingAssets {
     }
 
     pub fn get_theme(&self, theme: &str) -> Result<&Theme> {
-        let err = format!("Could not find '{}' theme", theme);
-        Ok(self.theme_set
-            .themes
-            .get(theme)
-            .ok_or_else(|| io::Error::new(io::ErrorKind::Other, err))?)
+        Ok(self.theme_set.themes.get(theme).ok_or_else(|| {
+            io::Error::new(
+                io::ErrorKind::Other,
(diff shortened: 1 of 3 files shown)


src/assets.rs
```diff
@@ -126,11 +126,12 @@ impl HighlightingAssets {
         Ok(())
     }
 
-    pub fn default_theme(&self) -> Result<&Theme> {
+    pub fn get_theme(&self, theme: &str) -> Result<&Theme> {
+        let err = format!("Could not find '{}' theme", theme);
```
sharkdp: I think this should be inlined in the `ok_or_else` call. Otherwise, it will always be constructed.

src/main.rs
```diff
@@ -223,7 +223,10 @@ fn run() -> Result<()> {
             let config = app.config()?;
 
             let assets = HighlightingAssets::new();
-            let theme = assets.default_theme()?;
+            let theme = match config.theme {
```
sharkdp: I think this `match` statement could be simplified to

``` rust
let theme = assets.get_theme(config.theme.unwrap_or("Default"))?;
```

src/assets.rs
rleungx: Agreed.

PR #102, opened 2018-05-12
Added text wrapping. (Fixes #54)

### LINKED ISSUE #54: Text wrapping
Is there any plan for text wrapping? The output right now is *so* polished, lines too long look pretty out-of-place. Definitely understand not wrapping by default, but having a flag for it would be pretty chill I think.

eth-p:
I use bat pretty frequently, so I wanted to give back and make it even better by adding text wrapping (issue #54).

I've never touched Rust before this, and I apologize if the code style is weird or if any of my changes have less-than-ideal performance.

## Changes:
- **Text wrapping.**
   Lines will now automatically wrap at a character boundary.
   Syntax coloring will also wrap.
image

- **Resizing gutter (panel).**
   The gutter (area that shows line numbers and changes) will now automatically resize.
   This cuts back on the amount of wasted space when not showing everything.
image
   It will also disappear if not needed.
   image

### DIFF
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -250,7 +250,7 @@ impl<'a> Printer<'a> {
 
     fn print_horizontal_line(&mut self, grid_char: char) -> Result<()> {
         if self.panel_width == 0 {
-            writeln!(self.handle, "{}", "─".repeat(self.config.term_width))?;
+            writeln!(self.handle, "{}", self.colors.grid.paint("─".repeat(self.config.term_width)))?;
         } else {
             let hline = "─".repeat(self.config.term_width - (self.panel_width + 1));
             let hline = format!("{}{}{}", "─".repeat(self.panel_width), grid_char, hline);
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -142,8 +142,9 @@ impl<'a> Printer<'a> {
             )?;
         } else {
             for &(style, text) in regions.iter() {
-                let mut chars = text.chars().filter(|c| *c != '\n');
-                let mut remaining = text.chars().filter(|c| *c != '\n').count();
+                let text = text.trim_right_matches(|c| c == '\r' || c == '\n');
+                let mut chars = text.chars();
+                let mut remaining = text.chars().count();
 
                 while remaining > 0 {
                     let available = cursor_max - cursor;
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -117,67 +117,59 @@ impl<'a> Printer<'a> {
         let border = if gutter_width > 0 && self.config.output_components.grid() {
             self.gen_border()
         } else {
-            PrintSegment {
-                size: 0,
-                text: "".to_owned(),
-            }
-        };
+            for &(style, text) in regions.iter() {
+                let mut chars = text.chars().filter(|c| *c != '\n');
+                let mut remaining = text.chars().filter(|c| *c != '\n').count();
 
-        cursor_max -= border.size;
-        write!(self.handle, "{} ", border.text)?;
+                while remaining > 0 {
+                    let available = cursor_max - cursor;
 
-        // Line contents.
-        for &(style, text) in regions.iter() {
-            let mut chars = text.chars().filter(|c| *c != '\n');
-            let mut remaining = text.chars().filter(|c| *c != '\n').count();
+                    // It fits.
+                    if remaining <= available {
+                        let text = chars.by_ref().take(remaining).collect::<String>();
+                        cursor += remaining;
 
-            while remaining > 0 {
-                let available = cursor_max - cursor;
+                        write!(
+                            self.handle,
+                            "{}",
+                            as_terminal_escaped(
+                                style,
+                                &*text,
+                                self.config.true_color,
+                                self.config.colored_output
+                            )
+                        )?;
+                        break;
+                    }
 
-                // It fits.
-                if remaining <= available {
-                    let text = chars.by_ref().take(remaining).collect::<String>();
-                    cursor += remaining;
+                    // It wraps.
+                    if self.config.output_wrap == OutputWrap::Character {
+                        let text = chars.by_ref().take(available).collect::<String>();
+                        cursor = 0;
+                        remaining -= available;
 
-                    write!(
-                        self.handle,
-                        "{}",
-                        as_terminal_escaped(
-                            style,
-                            &*text,
-                            self.config.true_color,
-                            self.config.colored_output
-                        )
-                    )?;
-                    break;
-                }
+                        write!(
+                            self.handle,
+                            "{}\n{}{}",
+                            as_terminal_escaped(
+                                style,
+                                &*text,
+                                self.config.true_color,
+                                self.config.colored_output
+                            ),
+                            " ".repeat(gutter_width),
+                            border.text.to_owned()
+                        )?;
 
-                // It wraps.
-                if self.config.output_wrap == OutputWrap::Character {
-                    let text = chars.by_ref().take(available).collect::<String>();
-                    cursor = 0;
-                    remaining -= available;
-
-                    write!(
-                        self.handle,
-                        "{}\n{}{} ",
-                        as_terminal_escaped(
-                            style,
-                            &*text,
-                            self.config.true_color,
-                            self.config.colored_output
-                        ),
-                        " ".repeat(gutter_width),
-                        border.text.to_owned()
-                    )?;
-
-                    continue;
+                        continue;
+                    }
                 }
             }
+
+            write!(self.handle, "\n")?;
         }
 
         // Finished.
-        write!(self.handle, "\n")?;
         Ok(())
     }
 
@@ -230,8 +222,8 @@ impl<'a> Printer<'a> {
     /// Generates the vertical grid border.
     fn gen_border(&self) -> PrintSegment {
         return PrintSegment {
-            text: self.colors.grid.paint("│").to_string(),
-            size: 1,
+            text: self.colors.grid.paint("│ ").to_string(),
+            size: 2,
         };
     }
 
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -135,7 +135,7 @@ impl<'a> Printer<'a> {
                    regions.iter()
                        .map(|&(style, text)| as_terminal_escaped(style, text, true_color, colored_output))
                        .collect::<Vec<_>>()
-                       .join(" ")
+                       .join("")
             )?;
         } else {
             for &(style, text) in regions.iter() {
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -41,7 +41,7 @@ impl<'a> Printer<'a> {
         };
 
         // Generate the panel (gutter) width.
-        let decorations = instance.gen_decorations(0);
+        let decorations = instance.line_decorations(0);
         instance.panel_width = decorations.len() + decorations.iter().fold(0, |a, x| a + x.size);
 
         // Return the instance.
@@ -97,7 +97,7 @@ impl<'a> Printer<'a> {
         let mut cursor_max: usize = self.config.term_width;
 
         // Line decoration.
-        let decorations = self.gen_decorations(line_number);
+        let decorations = self.line_decorations(line_number);
         let gutter_width = decorations.len() + decorations.iter().fold(0, |a, x| a + x.size);
 
         if gutter_width > 0 {
@@ -115,7 +115,7 @@ impl<'a> Printer<'a> {
 
         // Grid border.
         let border = if gutter_width > 0 && self.config.output_components.grid() {
-            self.gen_border()
+            self.line_border()
         } else {
             PrintSegment {
                 size: 0,
@@ -131,11 +131,19 @@ impl<'a> Printer<'a> {
             let true_color = self.config.true_color;
             let colored_output = self.config.colored_output;
 
-            write!(self.handle, "{}",
-                   regions.iter()
-                       .map(|&(style, text)| as_terminal_escaped(style, text, true_color, colored_output))
-                       .collect::<Vec<_>>()
-                       .join("")
+            write!(
+                self.handle,
+                "{}",
+                regions
+                    .iter()
+                    .map(|&(style, text)| as_terminal_escaped(
+                        style,
+                        text,
+                        true_color,
+                        colored_output
+                    ))
+                    .collect::<Vec<_>>()
+                    .join("")
             )?;
         } else {
             for &(style, text) in regions.iter() {
@@ -164,26 +172,24 @@ impl<'a> Printer<'a> {
                     }
 
                     // It wraps.
-                    if self.config.output_wrap == OutputWrap::Character {
-                        let text = chars.by_ref().take(available).collect::<String>();
-                        cursor = 0;
-                        remaining -= available;
+                    let text = chars.by_ref().take(available).collect::<String>();
+                    cursor = 0;
+                    remaining -= available;
 
-                        write!(
-                            self.handle,
-                            "{}\n{}{}",
-                            as_terminal_escaped(
-                                style,
-                                &*text,
-                                self.config.true_color,
-                                self.config.colored_output,
-                            ),
-                            " ".repeat(gutter_width),
-                            border.text.to_owned()
-                        )?;
+                    write!(
+                        self.handle,
+                        "{}\n{}{}",
+                        as_terminal_escaped(
+                            style,
+                            &*text,
+                            self.config.true_color,
+                            self.config.colored_output,
+                        ),
+                        " ".repeat(gutter_width),
+                        border.text.to_owned()
+                    )?;
 
-                        continue;
-                    }
+                    continue;
                 }
             }
 
@@ -194,25 +200,23 @@ impl<'a> Printer<'a> {
         Ok(())
     }
 
-    /// Generates all the line decorations.
-    fn gen_decorations(&self, line_number: usize) -> Vec<PrintSegment> {
+    fn line_decorations(&self, line_number: usize) -> Vec<PrintSegment> {
         let mut decorations = Vec::new();
 
         if self.config.output_components.numbers() {
-            decorations.push(self.gen_deco_line_number(line_number));
+            decorations.push(self.line_number(line_number));
         }
 
         if self.config.output_components.changes() {
-            decorations.push(self.gen_deco_line_changes(line_number));
+            decorations.push(self.line_changes(line_number));
         }
 
         return decorations;
     }
 
-    /// Generates the decoration for displaying the line number.
-    fn gen_deco_line_number(&self, line_number: usize) -> PrintSegment {
+    fn line_number(&self, line_number: usize) -> PrintSegment {
         let plain: String = format!("{:width$}", line_number, width = LINE_NUMBER_WIDTH);
-        let color = self.colors.line_number.paint(plain.to_owned());
+        let color = self.colors.line_number.paint(plain.clone());
 
         return PrintSegment {
             text: color.to_string(),
@@ -220,8 +224,7 @@ impl<'a> Printer<'a> {
         };
     }
 
-    /// Generates the decoration for displaying the git changes.
-    fn gen_deco_line_changes(&self, line_number: usize) -> PrintSegment {
+    fn line_changes(&self, line_number: usize) -> PrintSegment {
         let color = if let Some(ref changes) = self.line_changes {
             match changes.get(&(line_number as u32)) {
                 Some(&LineChange::Added) => self.colors.git_added.paint("+"),
@@ -240,8 +243,7 @@ impl<'a> Printer<'a> {
         };
     }
 
-    /// Generates the vertical grid border.
-    fn gen_border(&self) -> PrintSegment {
+    fn line_border(&self) -> PrintSegment {
         return PrintSegment {
             text: self.colors.grid.paint("│ ").to_string(),
             size: 2,
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -5,14 +5,21 @@ use errors::*;
 use std::io::Write;
 use syntect::highlighting;
 use terminal::as_terminal_escaped;
+use style::OutputWrap;
 use Colors;
 
-const PANEL_WIDTH: usize = 7;
+const LINE_NUMBER_WIDTH: usize = 4;
+
+struct PrintSegment {
+    size: usize,
+    text: String
+}
 
 pub struct Printer<'a> {
     handle: &'a mut Write,
     colors: Colors,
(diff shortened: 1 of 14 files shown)


sharkdp:
This looks super cool - thank you very much for your contribution!

I will only leave a few quick comments for now and add a proper review later when all the checks pass (and the conflicts are resolved).

src/printer.rs
```diff
@@ -24,12 +31,27 @@ impl<'a> Printer<'a> {
...
+        let mut instance = Printer {
             handle,
             colors,
             config,
+            panel_width: 0,
             line_changes: None,
-        }
+        };
+
+        // Generate the panel (gutter) width.
```
sharkdp: Could this be computed first and then simply passed to the `Printer` constructor? This way, we don't need the (mutable) `instance` variable.

src/printer.rs
```diff
@@ -24,12 +31,27 @@ impl<'a> Printer<'a> {
...
+        let decorations = instance.gen_decorations(0);
+        instance.panel_width = decorations.len()
+            + decorations.iter().fold(0, |a, x| a + x.size)
+            + if config.output_components.grid() {
```
sharkdp: 0 is added in both cases here? This can't be right?

src/printer.rs
```diff
@@ -75,86 +101,162 @@ impl<'a> Printer<'a> {
...
+                    ), " ".repeat(gutter_width), border.text.to_owned())?;
+
+                    continue;
+                }
+            }
         }
+
+        // Finished.
+        write!(self.handle, "\n")?;
+        return Ok(());
```
sharkdp: `Ok(())` is equivalent to `return Ok(());` here.

src/printer.rs
```diff
@@ -75,86 +101,162 @@ impl<'a> Printer<'a> {
...
     }
 
-    fn print_git_marker(&self, line_number: usize) -> Option<String> {
+
+
+    #[doc = "
```
sharkdp: You can simply use triple slashes for doc strings:

``` rust
/// Generates all the line decorations
fn gen_decorations(...)
```

src/terminal.rs
```diff
@@ -26,26 +26,46 @@ fn rgb2ansi(r: u8, g: u8, b: u8) -> u8 {
     }
 }
 
+//pub fn as_terminal_escaped(
```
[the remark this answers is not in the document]
eth-p: As I changed the code, I kept comments of the original functions as a reference for how to do certain things. It looks like I forgot to remove that one, sorry.

src/printer.rs
eth-p: I used to count the grid line as part of the gutter's width, but ended up changing it later on to simplify calculations. My bad, it looks like I forgot to remove this leftover code.

src/printer.rs
eth-p: I was thinking that it would be a fair tradeoff, making the code easier to maintain by only having a single function (`gen_decorations`) for determining which components to display and their sizes.

I could change it with a couple of `if`s and consts, if you prefer to avoid the mutable `instance` variable though.

src/printer.rs
```diff
@@ -75,85 +93,173 @@ impl<'a> Printer<'a> {
...
+            write!(self.handle, "{}",
+                   regions.iter()
+                       .map(|&(style, text)| as_terminal_escaped(style, text, true_color, colored_output))
+                       .collect::<Vec<_>>()
+                       .join("")
+            )?;
+        } else {
+            for &(style, text) in regions.iter() {
+                let mut chars = text.chars().filter(|c| *c != '\n');
+                let mut remaining = text.chars().filter(|c| *c != '\n').count();
```
BrainMaestro: You can just clone the chars iterator

src/terminal.rs
```diff
@@ -27,37 +27,36 @@ fn rgb2ansi(r: u8, g: u8, b: u8) -> u8 {
...
+            color.italic()
+        } else {
+            color.normal()
+        }
+    };
 
-    s
+    let mut s: String = String::new();
+    write!(s, "{}", style.paint(text)).unwrap();
+    return s;
```
BrainMaestro: you can remove all of this and just return `style.paint(text).to_string()`. also, the return statement is not necessary. the last expression in a function will be returned

tests/snapshots/output/grid,header.snapshot.txt
```diff
@@ -1,25 +1,25 @@
-───────┬────────────────────────────────────────────────────────────────────────
-       │ File: sample.rs
-───────┼────────────────────────────────────────────────────────────────────────
-       │ struct Rectangle {
```
BrainMaestro: the vertical line is gone?

tests/snapshots/output/grid,numbers.snapshot.txt
```diff
@@ -1,22 +1,22 @@
-   1   │ struct Rectangle {
```
BrainMaestro: 👍

tests/snapshots/output/grid.snapshot.txt
```diff
@@ -1,22 +1,22 @@
-       │ struct Rectangle {
```
BrainMaestro: same as above

src/printer.rs
```diff
@@ -75,85 +93,173 @@ impl<'a> Printer<'a> {
...
+                decorations
+                    .iter()
+                    .map(|seg| seg.text.to_owned())
+                    .collect::<Vec<String>>()
+                    .join(" ")
+            )?;
+        }
+
+        // Grid border.
+        let border = if gutter_width > 0 && self.config.output_components.grid() {
```
BrainMaestro: this can go inside `gen_border` right?

src/printer.rs
```diff
@@ -75,85 +93,173 @@ impl<'a> Printer<'a> {
...
+                                &*text,
+                                self.config.true_color,
+                                self.config.colored_output,
+                            )
+                        )?;
+                        break;
+                    }
+
+                    // It wraps.
+                    if self.config.output_wrap == OutputWrap::Character {
```
BrainMaestro: this is the `else` part, so we already know that this will evaluate to true every time. or am I missing something?

src/printer.rs
```diff
@@ -75,85 +93,173 @@ impl<'a> Printer<'a> {
...
+            write!(self.handle, "\n")?;
+        }
+
+        // Finished.
         Ok(())
     }
 
-    fn print_line_number(&self, line_number: usize) -> Option<String> {
+    /// Generates all the line decorations.
+    fn gen_decorations(&self, line_number: usize) -> Vec<PrintSegment> {
```
BrainMaestro: `line_decorations`?

src/printer.rs
```diff
@@ -75,85 +93,173 @@ impl<'a> Printer<'a> {
...
+        return decorations;
     }
 
-    fn print_line_border(&self) -> Option<String> {
-        if self.config.output_components.grid() {
-            Some(self.colors.grid.paint("│").to_string())
+    /// Generates the decoration for displaying the line number.
+    fn gen_deco_line_number(&self, line_number: usize) -> PrintSegment {
+        let plain: String = format!("{:width$}", line_number, width = LINE_NUMBER_WIDTH);
+        let color = self.colors.line_number.paint(plain.to_owned());
```
BrainMaestro: `plain` is already a string, so `to_owned` is unnecessary

src/printer.rs
```diff
@@ -75,85 +93,173 @@ impl<'a> Printer<'a> {
...
+
+        return PrintSegment {
+            text: color.to_string(),
+            size: plain.len(),
+        };
+    }
+
+    /// Generates the decoration for displaying the git changes.
+    fn gen_deco_line_changes(&self, line_number: usize) -> PrintSegment {
```
BrainMaestro: can it just be `line_changes`? two abbreviations in a method name make it confusing. same for the number one. that way, you can remove the doc comments

src/style.rs
```diff
@@ -13,6 +13,12 @@ pub enum OutputComponent {
     Plain,
 }
 
+#[derive(Debug, Eq, PartialEq, Copy, Clone, Hash)]
+pub enum OutputWrap {
```
BrainMaestro: since there are only two variants, each with no data, maybe this can be represented with a `bool` instead

src/printer.rs
eth-p: I did that [in the initial PR](https://github.com/sharkdp/bat/pull/102/commits/4e4110bf5015009d584ed0ed5d09d19eb5393430#diff-041fe12915c4f607e78f57794ffa89bdR136), but it [failed to build on Rust 1.24.0](https://travis-ci.org/sharkdp/bat/jobs/378014200).

src/printer.rs
eth-p: Yeah, it evaluates to true every time. I left it in there in case different wrapping styles were added in the future, but on second thought, it's unnecessary right now.

src/printer.rs
eth-p: I tried removing `to_owned()`, and ended up with this:

```
Error[E0382]: use of moved value: `plain`
   --> src/printer.rs:217:19
    |
213 |         let color = self.colors.line_number.paint(plain);
    |                                                   ----- value moved here
...
217 |             size: plain.len(),
    |                   ^^^^^ value used here after move
    |
    = note: move occurs because `plain` has type `std::string::String`, which does not implement the `Copy` trait
```

Replacing it with `plain.clone()` works, though. Would that be better to use?

src/printer.rs
eth-p: Good point. I'll rename it.

src/style.rs
eth-p: I used an enum in case anybody wanted to add different wrapping styles (e.g. word wrap) in the future, but I could change it to a bool if that's preferable.

src/printer.rs
eth-p: I did that it there in case different wrapping styles would be added in the future, but looking back at it, it's definitely not necessary right now. I'll remove it.

tests/snapshots/output/grid,header.snapshot.txt
eth-p: That would be part of the resizing gutter change. When there aren't any line decorations, the vertical grid border is removed to save space (since there's nothing to separate).

I could add it back if the old behaviour is preferred, however.

src/printer.rs
BrainMaestro: Huh weird

tests/snapshots/output/grid,header.snapshot.txt
BrainMaestro: I would think that the vertical line is part of the grid and should be there if it's specified, but maybe @sharkdp can clarify

src/printer.rs
BrainMaestro: I think paint can take a `&str`, so you can pass `&plain`. also I didn't notice before, but the variable name `color` is probably not correct

src/style.rs
BrainMaestro: Oh right, I guess you can just go ahead and add that variant then

sharkdp: I think I prefer the enum here, even if there are just two variants for now ("Boolean blindness"). We shouldn't add the third variant if it is not yet used.

tests/snapshots/output/grid,header.snapshot.txt
sharkdp: I like this change. If there is nothing to display in the side-panel, I don't think there is any need for a vertical grid line.

src/printer.rs
```diff
@@ -75,85 +93,175 @@ impl<'a> Printer<'a> {
...
+                            style,
+                            &*text,
+                            self.config.true_color,
+                            self.config.colored_output,
+                        ),
+                        " ".repeat(gutter_width),
+                        border.text.to_owned()
+                    )?;
+
+                    continue;
```
sharkdp: Is this `continue` statement needed?

src/printer.rs
```diff
@@ -75,85 +93,175 @@ impl<'a> Printer<'a> {
...
+                }
+            }
+
+            write!(self.handle, "\n")?;
+        }
+
+        // Finished.
```
sharkdp: I think this can be safely removed :smiley:

sharkdp:
I just performed some simple benchmarks. As probably expected, the text-wrapping has some performance overhead (up to ~30%):

```
▶ hyperfine 'bat --style=full --color=always src/*.rs' \
            'target/release/bat-wrapping --style=full --color=always --wrap=character src/*.rs'
Benchmark #1: bat --style=full --color=always src/*.rs

  Time (mean ± σ):      81.6 ms ±   2.1 ms    [User: 76.4 ms, System: 4.9 ms]
 
  Range (min … max):    78.7 ms …  86.8 ms
 
Benchmark #2: target/release/bat-wrapping --style=full --color=always --wrap=character src/*.rs

  Time (mean ± σ):      97.3 ms ±   2.2 ms    [User: 92.7 ms, System: 4.4 ms]
 
  Range (min … max):    94.0 ms … 101.5 ms
 
Summary

'bat --style=full --color=always src/*.rs' ran
    1.19x faster than 'target/release/bat-wrapping --style=full --color=always --wrap=character src/*.rs'
```

```
▶ hyperfine 'bat --style=full --color=always large.rs' \
            'target/release/bat-wrapping --style=full --color=always --wrap=character large.rs'
Benchmark #1: bat --style=full --color=always large.rs

  Time (mean ± σ):     770.8 ms ±   6.0 ms    [User: 762.6 ms, System: 7.2 ms]
 
  Range (min … max):   762.0 ms … 779.1 ms
 
Benchmark #2: target/release/bat-wrapping --style=full --color=always --wrap=character large.rs

  Time (mean ± σ):      1.016 s ±  0.013 s    [User: 1.006 s, System: 0.008 s]
 
  Range (min … max):    1.001 s …  1.041 s
 
Summary

'bat --style=full --color=always large.rs' ran
    1.32x faster than 'target/release/bat-wrapping --style=full --color=always --wrap=character large.rs'
```

Unfortunately, the performance drops by the same amount even when calling with `--wrap=never`.

src/printer.rs
eth-p: It appears not. Removing it in the next commit.

eth-p:
That would probably be the consequence of the way I rewrote the line decorations. I'll see what I can do to reduce the performance impact.

eth-p:
@sharkdp I split up the decorations into their own file and added caching to them. Here's the results:

```
$ ls -lha | grep larger
-rw-r--r--   1 eth-p  staff   271K 13 May 18:25 larger.rs
```

**Character wrapping**
```
$ bench 'target/release/bat --style full --paging never --wrap character larger.rs'
benchmarking target/release/bat --style full --paging never --wrap character larger.rs
time                 869.2 ms   (724.6 ms .. 1.021 s)
                     0.996 R²   (0.986 R² .. 1.000 R²)
mean                 837.7 ms   (811.1 ms .. 861.4 ms)
std dev              29.52 ms   (24.26 ms .. 32.64 ms)
variance introduced by outliers: 19% (moderately inflated)
```

**No wrapping**
```
$ bench 'target/release/bat --style full --paging never --wrap never larger.rs'
benchmarking target/release/bat --style full --paging never --wrap never larger.rs
time                 791.4 ms   (782.2 ms .. 810.2 ms)
                     1.000 R²   (NaN R² .. 1.000 R²)
mean                 796.2 ms   (791.7 ms .. 802.4 ms)
std dev              6.350 ms   (1.870 ms .. 8.513 ms)
variance introduced by outliers: 19% (moderately inflated)
```

**Bat 0.3**
``` 
$ bench 'bat --style full --paging never larger.rs'
benchmarking bat --style full --paging never larger.rs
time                 812.2 ms   (769.5 ms .. 836.7 ms)
                     1.000 R²   (NaN R² .. 1.000 R²)
mean                 818.7 ms   (813.1 ms .. 823.6 ms)
std dev              5.796 ms   (3.063 ms .. 8.088 ms)
variance introduced by outliers: 19% (moderately inflated)
```

sharkdp:
@eth-p Thank you for looking into the performance-side of this!

Is the caching really the important change here? It seems like it would only help for files with really long lines where the line number would not have to be displayed?



Sorry that the review process has been dragging on for some time now, but there are (still) a few things that I would like to verify/update before I merge this.

To be honest, while I really like the outcome (see screenshots) and the implemented features, I am not quite confident about the changes in the code. Part of the reason might also be that I wasn't really happy with that part of the code (generating the different output elements) to begin with :smile:.

While it's hard to pinpoint any "bad" changes, here are a few things that we might want to discuss:

- If I understand correctly, the `DecorationText` struct basically holds some (formatted) text and its display width (~ length without ANSI codes). While possibly good for performance reasons, I don't really like that we have to pass around this redundant information (and keep it in sync!). An alternative could be to compute the display width (https://docs.rs/console/0.6.1/console/fn.measure_text_width.html)
- The split into the `for_line` and `for_wrap` methods in the `Decoration` trait doesn't feel very natural to me. It leads to quite some code duplication. Another option would be to have a single method ("to_string"?), that receives a `continuation: bool` or `wrapped_line: bool` parameter.
- `LineChangesDecoration` is the only decoration that uses the `printer` argument. Maybe we should pass the `LineChanges` to the `LineChangesDecoration` object on construction and remove the `printer` argument from the `for_line` function.
- Given the points above, we should maybe reconsider if a `trait` is the right abstraction here (the answer can be "yes" :smile:).

Concerning the actual wrapping:
- There is a small problem with windows-style line-breaks (`\r\n`) which can be reproduced by converting some file via `unix2dos` or by simply calling `echo -e "windows\r\nnewline" | bat`. With wrapping enabled, this outputs `^M` characters which is probably due to the wrap-algorithm treating only the `\n` as a line break.
- Instead of implementing character-level wrapping, it might be worth to think about region-based wrapping (regions being defined by syntect throught the highlighting). This might actually be easier than the character-level wrapping, but it's probably not too helpful if regions can be larger than a single line.

eth-p:
@sharkdp 

> Is the caching really the important change here? It seems like it would only help for files with really long lines where the line number would not have to be displayed?

The way I implemented the caching, it should increase the performance whether or not the lines are wrapped. Consider the non-cached versions of the git changes and grid border decorations:

```rust
    fn line_changes(&self, line_number: usize) -> PrintSegment {
        let color = if let Some(ref changes) = self.line_changes {
            match changes.get(&(line_number as u32)) {
                Some(&LineChange::Added) => self.colors.git_added.paint("+"),
                Some(&LineChange::RemovedAbove) => self.colors.git_removed.paint("‾"),
                Some(&LineChange::RemovedBelow) => self.colors.git_removed.paint("_"),
                Some(&LineChange::Modified) => self.colors.git_modified.paint("~"),
                _ => Style::default().paint(" "),
            }
        } else {
            Style::default().paint(" ")
        };

        return PrintSegment {
            text: color.to_string(),
            size: 1,
        };
    }

    fn line_border(&self) -> PrintSegment {
        return PrintSegment {
            text: self.colors.grid.paint("│ ").to_string(),
            size: 2,
        };
    }
```

In both of them, we were calling `Style.paint` on constant strings every line. By generating those strings once and caching the result and cloning it when necessary, we're cutting back on the number of calls to `Style.paint`. While I don't know exactly how expensive a call to `Style.paint` is, it's probably more than `String.clone`. 

> If I understand correctly, the DecorationText struct basically holds some (formatted) text and its display width (~ length without ANSI codes). While possibly good for performance reasons, I don't really like that we have to pass around this redundant information (and keep it in sync!). An alternative could be to compute the display width (https://docs.rs/console/0.6.1/console/fn.measure_text_width.html)

I could definitely change it to use `measure_text_width`, but it's your call.

> The split into the for_line and for_wrap methods in the Decoration trait doesn't feel very natural to me. It leads to quite some code duplication. Another option would be to have a single method ("to_string"?), that receives a continuation: bool or wrapped_line: bool parameter.

I don't think `to_string` would fit, since `Decoration` is more similar to a factory than an object. I could rename it to `generate` and add `continuation: bool` to its parameter list, if that works?

> LineChangesDecoration is the only decoration that uses the printer argument. Maybe we should pass the LineChanges to the LineChangesDecoration object on construction and remove the printer argument from the for_line function.

That was my original plan, but I couldn't figure out a sane way to do it since the `Printer.line_changes` field is modified before each file is printed. I would've preferred to create a PrintContext struct that contained file information like the `LineChanges` object, but I thought that might be over-engineering it a little bit and pushing my understanding of Rust lifetimes.

> Given the points above, we should maybe reconsider if a trait is the right abstraction here (the answer can be "yes" 😄).

I suppose that would depend on whether or not the decorations should be considered as individual modules which are used by the printer, or as a part of the printer itself. I was thinking the former, but I'm likely biased having come from an OOP background where monolithic functions/classes are discouraged. 

I'll leave it up to you to decide whether or not I should move the decorations back into `Printer`, but it might start to look messy if you want me to keep the caching.

> There is a small problem with windows-style line-breaks.

Fixed!

> Instead of implementing character-level wrapping, it might be worth to think about region-based wrapping (regions being defined by syntect throught the highlighting). This might actually be easier than the character-level wrapping, but it's probably not too helpful if regions can be larger than a single line.

That was actually my original idea! For the reason you just mentioned, it ended up not working out too well. I would've had to implement character wrapping on top of region wrapping, and it still would've resulted in funny-looking wrapping like this:

```
1 | This is a single region.                  |
  | [This is an even longer single region, an |
  | d it ended up wasting spacing and was sti |
  | ll broken up.                             |
```

sharkdp:
Thank you very much for the clarifications!

> I could definitely change it to use measure_text_width, but it's your call.

Ok. Let's stay with your solution for now.

> I don't think to_string would fit, since Decoration is more similar to a factory than an object. I could rename it to generate and add continuation: bool to its parameter list, if that works?

`generate` with a `bool` parameter sounds good to me :+1: 

> That was my original plan, but I couldn't figure out a sane way to do it since the Printer.line_changes field is modified before each file is printed. I would've preferred to create a PrintContext struct that contained file information like the LineChanges object, but I thought that might be over-engineering it a little bit and pushing my understanding of Rust lifetimes.

Ok, we can look into this in the future.

> I'll leave it up to you to decide whether or not I should move the decorations back into Printer, but it might start to look messy if you want me to keep the caching.

Ok, let's keep it for now.

> There is a small problem with windows-style line-breaks.
> 
> Fixed!

Great, thank you.

> That was actually my original idea! For the reason you just mentioned, it ended up not working out too well. I would've had to implement character wrapping on top of region wrapping, and it still would've resulted in funny-looking wrapping like this:
> 
> (..)

Yeah, right :pensive:

eth-p:
Merged the `for_line and `for_wrap` methods. I also fixed #117 by disabling the panel when there isn't enough space to show the actual file contents.

sharkdp:
Thank you very much!!

PR #104, opened 2018-05-12
Added fish shell completions.

eth-p:
Added some completions for fish shell.

### DIFF
--- a/completions/bat.fish
+++ b/completions/bat.fish
@@ -1,0 +1,59 @@
+#
+# Completions for the bat command
+# https://github.com/sharkdp/bat
+#
+
+# Helpers
+function __bat_has_subcommand
+	set -l tokens (commandline -poc)
+	if [ (count $tokens) -eq 0 ]
+		return 1
+	end
+
+	if [ -n "$argv[1]" ]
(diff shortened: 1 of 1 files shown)


sharkdp:
Thanks!

There is a way to auto-generate completion files for bash, zsh, fish and PowerShell via `clap` (see for example: https://github.com/sharkdp/fd/pull/66/files, https://github.com/sharkdp/fd/blob/master/build.rs) which I would prefer in order tp prevent maintenance overhead.

What do you think?

eth-p:
That's a pretty cool feature! Automatically generating completion files would definitely be the way to go.

sharkdp:
I hope it's okay if I close this PR. I have opened #115 to track the addition of auto-generated completion files.

sboukortt:
Unfortunately, the generated fish completion does not complete filenames, and so pressing Tab does nothing for them. That’s sufficiently annoying that I simply deleted the completion file on my machine.

sharkdp:
@sboukortt Oh :frowning_face:. Could you please open a new ticket to address this?

sboukortt:
Done (#372), thanks for the suggestion.

PR #105, opened 2018-05-12
Add snapshot tests

BrainMaestro:
Closes #105


tests/snapshots/generate_snapshots.py
```diff
@@ -0,0 +1,49 @@
...
+import itertools
+import subprocess
+import pathlib
+
+def generate_snapshots():
+    single_styles = ["changes", "grid", "header", "numbers"]
+    collective_styles = ["full", "plain"]
+
+    for num in range(len(single_styles)):
+        for grouped in itertools.combinations(single_styles, num + 1):
```
sharkdp: :+1:

tests/snapshots/generate_snapshots.py
```diff
@@ -0,0 +1,49 @@
...
+            generate_snapshot("--style={}".format(",".join(grouped)))
+
+    for style in collective_styles:
+        generate_snapshot("--style={}".format(style))
+
+    generate_snapshot("--color=always")
+    generate_snapshot("--list-languages")
```
sharkdp: I'm not sure if `--list-languages` and `--list-themes` should be added as a snapshot test. We would have to re-generate those files every time we add a new syntax or theme.

I'm also not quite sure about the `--color=always` test because it also depends on the Default highlighting theme.

tests/snapshots/generate_snapshots.py
```diff
@@ -0,0 +1,49 @@
...
+    command = "../../target/debug/bat {} sample.rs > output/{}.snapshot.txt".format(
+        option,
+        option.replace("--", "").replace(" ", "+")
+    )
+    print("generating snapshot for {}".format(option))
+    subprocess.call(command, shell=True)
+
+def prepare_output_dir():
+    output = pathlib.Path("output")
+    if output.exists():
```
sharkdp: There is also [`shutil.rmtree`](https://docs.python.org/3/library/shutil.html#shutil.rmtree), which could be used with `ignore_errors=True`.

tests/snapshots/generate_snapshots.py
```diff
@@ -0,0 +1,49 @@
...
+        for snapshot in output.iterdir():
+            snapshot.unlink()
+        output.rmdir()
+
+    output.mkdir()
+
+def modify_sample_file():
+    print("modifying sample.rs to show changes")
+    subprocess.call("./modify.sh --modify", shell=True)
```
sharkdp: Instead of a separate `modify.sh` script, another option could be to have `sample.rs` and `sample.modified.rs`, which could be copied here (?). This would also allow for a few more modifications, if necessary.

tests/snapshots/generate_snapshots.py
BrainMaestro: Alright, I wasn't sure about those either. I added them after the fact.

tests/snapshots/generate_snapshots.py
BrainMaestro: I'm not sure I follow. I can just modify the file directly in python instead, but will have to implement it the same way in Rust too. But the build failed because of the script, so it's probably for the best

sharkdp: I just thought it might be easier to call `shutil.copyfile("sample.modified.rs", "sample.rs")` instead of calling a shell script that modifies the file via `sed`. From rust, I think it would also be fairly easy to [`copy`](https://doc.rust-lang.org/stable/std/fs/fn.copy.html) the files.

sharkdp: The undo-part would be the reverse copy-operation or a `git checkout sample.rs`.

BrainMaestro: Yeah that makes sense... 👍

sharkdp:
Thanks!

Curious to see how this works out.

PR #113, opened 2018-05-14
sort the language listing

### LINKED ISSUE #111: Sort the language listing
The output of bat --list-languages is difficult to read. Can the output be sorted on the entries in the first column, so that searching for a particular type is easier?

rleungx:
Closes #111.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -226,7 +226,8 @@ fn run() -> Result<()> {
             let theme = assets.get_theme(config.theme.unwrap_or("Default"))?;
 
             if app.matches.is_present("list-languages") {
-                let languages = assets.syntax_set.syntaxes();
+                let mut languages = assets.syntax_set.syntaxes().to_owned();
+                languages.sort_by(|a, b| a.name.to_uppercase().cmp(&b.name.to_uppercase()));
 
                 let longest = languages
                     .iter()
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -226,7 +226,8 @@ fn run() -> Result<()> {
             let theme = assets.get_theme(config.theme.unwrap_or("Default"))?;
 
             if app.matches.is_present("list-languages") {
-                let languages = assets.syntax_set.syntaxes();
+                let mut languages = assets.syntax_set.syntaxes().to_owned();
+                languages.sort_by(|a, b| a.name.to_uppercase().cmp(&b.name.to_uppercase()));
```
sharkdp: Would `sort_by_key` work here?

Something like
``` rust
languages.sort_by_key(|s| s.name.to_uppercase())
```

tests/tester.rs
```diff
@@ -29,10 +29,7 @@ impl BatTester {
 
     pub fn test_snapshot(&self, style: &str) {
         let output = Command::new(&self.exe)
-            .args(&[
-                "tests/snapshots/sample.rs",
-                &format!("--style={}", style),
-            ])
+            .args(&["tests/snapshots/sample.rs", &format!("--style={}", style)])
```
sharkdp: Currently trying to fix this here: #114

src/main.rs
rleungx: Yep. And which one do you prefer? How about `sort_by_cached_key`? See rust-lang/rust#48639.

sharkdp: I don't think we have to worry about performance here, so I would prefer `sort_by_key`, since it is easier to read, in my opinion (`sort_by_cached_key` is also fine, if it's already available in our minimum-required rustc version).

PR #140, opened 2018-05-21
Extract commands to functions

BrainMaestro:
I'm not sure if `features` is the best name for the module. Maybe you have a better idea. I did some minor cleanup for `list_languages`, but it still generates the same diff.


sharkdp:
I think `features` is fine (I only came up with `core`..).

Thanks!

PR #141, opened 2018-05-22
Readme update with the new logo banner.

jraulhernandezi:
I also add the different sizes and formats of the logo right here: [Original Files](https://drive.google.com/open?id=1Psa3lzC5OgAbjCg8E3axyHKwUIbKw9zm)


sharkdp:
Thank you very much!

Would it be possible to get the header-version as an SVG file, too? This way, we could probably even experiment with including the SVG file directly (better filesize, responsibility). I can take care of the README updates myself, if you point me to the SVG file.

sharkdp:
Update: I shortly created the header SVG myself and included it as an SVG file here: #146. @jraulhernandezi What do you think?

jraulhernandezi:
Of course, I just updated the [drive](https://drive.google.com/open?id=1Psa3lzC5OgAbjCg8E3axyHKwUIbKw9zm) with the file .svg file.

sharkdp:
@jraulhernandezi Thank you very much!

Issue #98, opened 2018-05-11
Weird characters showing up


andyperisho:
See pic below. I guess it is the lines that are not being drawn properly.
How can i fix it?

thx

![1526068458](https://user-images.githubusercontent.com/34944098/39944172-cd9a43bc-558f-11e8-8b16-a1b6929492a2.png)

sharkdp:
Oh dear :smile: 

It looks like your terminal can not print the unicode characters `│`, `─`, `┬`, etc. - yes. Which terminal (and which font?) do you use?

I wonder if there is some way to detect this. `bat` could then maybe fall back to ASCII characters in this case.

andyperisho:
i'm using termite.
I think the font I'm using is Hack. I'm on arch linux. how do I check that?

andyperisho:
Nevermind, it had to do with my locale which was not set. Thx for the help!

sharkdp:
Thank you for the feedback.

Berjou:
Hi, I got the same problem with urxvt. However, my locale is set.

```
$ locale
LANG=en_US.utf8
LC_CTYPE="en_US.utf8"
LC_NUMERIC="en_US.utf8"
LC_TIME="en_US.utf8"
LC_COLLATE="en_US.utf8"
LC_MONETARY="en_US.utf8"
LC_MESSAGES="en_US.utf8"
LC_PAPER="en_US.utf8"
LC_NAME="en_US.utf8"
LC_ADDRESS="en_US.utf8"
LC_TELEPHONE="en_US.utf8"
LC_MEASUREMENT="en_US.utf8"
LC_IDENTIFICATION="en_US.utf8"
LC_ALL=
```

```
$ bat --version  
bat 0.3.0
```

Any idea on how to fix this?
Thx.

sharkdp:
Does this work for you?
``` bash
printf "\xE2\x94\x82\n"
```

It shoud print [this unicode character](https://unicode-table.com/en/2502/): `│`.

Berjou:
Yes it does.

sharkdp:
Could you please show a screenshot of
```
echo "\xE2\x94\x82" | bat
```


I cannot reproduce this with `urxvt` (although I did not configure the font).

Berjou:
There is the result with urxvt:
![2018-05-19-063534_1363x667_scrot](https://user-images.githubusercontent.com/10575885/40264977-4603e138-5b2f-11e8-9d4b-9ede8205ffb9.png)

[...]

PR #143, opened 2018-05-22
Fix launching of pager so all control characters are interpreted.

ghuls:
Fix launching of pager so all control characters are interpreted
instead of only ANSI "color" escape sequences.

This fixes issue https://github.com/sharkdp/bat/issues/98


sharkdp:
@ghuls Thank you very much for your investigation and for this fix!

PR #149, opened 2018-05-23
Print themes directory in error message if no themes could be loaded …

ghuls:
…from it.

Before this patch:
    $ bat cache --init
    [bat error]: Could not load themes from '{}'

After:
    $ bat cache --init
    [bat error]: Could not load themes from '/home/user/.config/bat/themes'


sharkdp:
Oops :smile:, thank you!

PR #162, opened 2018-06-01
add line-range

tskinn:
This is my first attempt at actually writing some rust code so bear with me.

I wasn't sure of the best place to put the logic for checking line range. I started with putting it in the print_line function but moved it out to be able to short circuit and stop processing once out of the range.

I didn't add the ability to do multiple ranges yet but it should be fairly simple.

How robust of error checking do we need on validating the line ranges? Like do we want to sort or make sure they are not conflicting or that they are in least to greatest order or anything like that?

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -282,7 +282,7 @@ impl App {
             term_width: Term::stdout().size().1 as usize,
             files,
             theme: self.matches.value_of("theme"),
-            line_range: self.matches.value_of("line-range").map(LineRange::from),
+            line_range: transpose(self.matches.value_of("line-range").map(LineRange::from))?,
         })
     }
 
@@ -351,12 +351,12 @@ pub struct LineRange {
 }
 
 impl LineRange {
-    pub fn from(range_raw: &str) -> LineRange {
-        LineRange::parse_range(range_raw).unwrap_or(LineRange::new())
+    pub fn from(range_raw: &str) -> Result<LineRange> {
+        LineRange::parse_range(range_raw)
     }
 
     pub fn new() -> LineRange {
-        LineRange{
+        LineRange {
             lower: usize::min_value(),
             upper: usize::max_value(),
         }
@@ -365,10 +365,10 @@ impl LineRange {
     pub fn parse_range(range_raw: &str) -> Result<LineRange> {
         let mut new_range = LineRange::new();
 
-        if range_raw.bytes().nth(0).ok_or("No first byte")? == b':' {
+        if range_raw.bytes().nth(0).ok_or("Empty line range")? == b':' {
             new_range.upper = range_raw[1..].parse()?;
             return Ok(new_range);
-        } else if range_raw.bytes().last().ok_or("No last byte")? == b':' {
+        } else if range_raw.bytes().last().ok_or("Empty line range")? == b':' {
             new_range.lower = range_raw[..range_raw.len() - 1].parse()?;
             return Ok(new_range);
         }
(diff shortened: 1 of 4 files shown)


tskinn:
I appreciate the feedback. I will start working on this. Thanks for you patience. Feel free to let me know if you think I'm going in the completely wrong direction with this.

src/app.rs
```diff
@@ -322,10 +336,69 @@ pub struct Config<'a> {
...
         .unwrap_or(false)
 }
+
+pub struct LineRange {
+    pub lower: usize,
+    pub upper: usize,
+}
+
+impl LineRange {
+    pub fn from(range_raw: &str) -> LineRange {
```
sharkdp: I think this should also return a `Result<LineRange>` (by removing the `unwrap_or` call below). This way, we can pass the error messages through to the call-site.

src/app.rs
```diff
@@ -322,10 +336,69 @@ pub struct Config<'a> {
...
+        LineRange{
+            lower: usize::min_value(),
+            upper: usize::max_value(),
+        }
+    }
+
+    pub fn parse_range(range_raw: &str) -> Result<LineRange> {
+        let mut new_range = LineRange::new();
+
+        if range_raw.bytes().nth(0).ok_or("No first byte")? == b':' {
```
sharkdp: Maybe the error message here (and below) should be "Empty line range"?

src/app.rs
```diff
@@ -322,10 +336,69 @@ pub struct Config<'a> {
...
+            new_range.upper = range_raw[1..].parse()?;
+            return Ok(new_range);
+        } else if range_raw.bytes().last().ok_or("No last byte")? == b':' {
+            new_range.lower = range_raw[..range_raw.len() - 1].parse()?;
+            return Ok(new_range);
+        }
+
+        let line_numbers: Vec<&str> = range_raw.split(':').collect();
+        if line_numbers.len() == 2 {
```
sharkdp: I think we should have an `else` branch here which returns a error ("expected single ':' character").

src/app.rs
```diff
@@ -322,10 +336,69 @@ pub struct Config<'a> {
...
+            new_range.lower = line_numbers[0].parse()?;
+            new_range.upper = line_numbers[1].parse()?;
+        }
+        Ok(new_range)                
+    }
+}
+
+#[test]
+fn test_parse_line_range_full() {
+    let range = LineRange::from(Some("40:50")).expect("Shouldn't fail on test!");
```
sharkdp: This would turn into `::from("40:50")`.

src/app.rs
```diff
@@ -322,10 +336,69 @@ pub struct Config<'a> {
...
+    assert_eq!(usize::min_value(), range.lower);
+    assert_eq!(50, range.upper);
+}
+
+#[test]
+fn test_parse_line_range_partial_max() {
+    let range = LineRange::from(Some("40:")).expect("Shouldn't fail on test!");
+    assert_eq!(40, range.lower);
+    assert_eq!(usize::max_value(), range.upper);
+}
```
sharkdp: If we change the signature of `LineRange::from` to return a `Result`, we can also add some tests for negative cases.

src/features.rs
```diff
@@ -85,28 +85,72 @@ fn print_file(
...
+        printer.print_footer()?;
+    }
+    Ok(())
+}
 
-    let mut reader: Box<BufRead> = match filename {
-        None => Box::new(stdin.lock()),
-        Some(filename) => Box::new(BufReader::new(File::open(filename)?)),
-    };
+fn print_file_ranges<'a>(
```
sharkdp: It would be really great if we had just a single function and would avoid this code duplication. I would try to change the `ranges` parameter to accept an `Option<Range>`.

src/main.rs
```diff
@@ -41,6 +41,7 @@ mod errors {
             Clap(::clap::Error);
             Io(::std::io::Error);
             SyntectError(::syntect::LoadingError);
+            ParseIntError(::std::num::ParseIntError);
```
sharkdp: Nice!

src/app.rs
```diff
@@ -269,6 +282,7 @@ impl App {
             term_width: Term::stdout().size().1 as usize,
             files,
             theme: self.matches.value_of("theme"),
+            line_range: self.matches.value_of("line-range").map(LineRange::from),
```
sharkdp: If we change the signature of `LineRange::from`, as suggested below, we could use
``` rust
line_range: tranpose(self.matches.value_of("line-range").map(LineRange::from))
```
such that the parsing-error-messages would bubble up to the main function (and reported to the user).

Here, `transpose` is a little helper function that turns a "Option of Result" into a "Result of Option". It could be implemented like this:
```
fn transpose<T, E>(opt: Option<Result<T, E>>) -> Result<Option<T>, E> {
    opt.map_or(Ok(None), |res| res.map(Some))
}
```

This function is actually an experimental feature which means that we could soon replace this by a call to the standard-library `.transpose()`: https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.transpose

src/app.rs
tskinn: This might be a silly question but how do I go about creating an error? When I try to return
```Err("expected single....")``` I get a mismatched types error: ```expected struct `errors::Error`, found reference```

Do I need to implement a new error type or something like that?

src/app.rs
tskinn: Does this only work if the tests are in a new module called LineRange?

src/app.rs
tskinn: OK I think I see what you are saying now. We want to bubble up any errors that happen when parsing the linerange to bubble up from the config() function.

src/app.rs
sharkdp: You should be able to use `Err("expected single....".into())`.

src/app.rs
sharkdp: Ah, sorry. I just meant to say that the `Some` part will not be needed anymore. I would still use `LineRange::from`.

tskinn: Oh duh. Gotcha

PR #185, opened 2018-07-05
Fix #182

### LINKED ISSUE #182: Escape Sequences get partially stripped
Command to reproduce: `echo -e '\033[1;3mfoo\033[22m bar\033[23m baz' | bat`.
Expected output: _**foo**_ _bar_ baz
Actual output: _**foo**_ bar baz

Note that `bar` should be still italic, but it is not when piped through bat. I cannot reproduce this behaviour with other pagers (less, most) or any other shell utility for that matter.

eth-p:
Tested with:

```shell
printf '\033[33mfoo\033[1m bar\033[22m baz' | cargo run
```

Results (italic = yellow):

cat *foo* **_bar_** *baz*
bat 0.4.1: *foo* **bar** baz
fix-182: *foo* **_bar_** *baz*

### DIFF
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -18,6 +18,7 @@ pub struct Printer<'a> {
     pub config: &'a Config<'a>,
     decorations: Vec<Box<Decoration>>,
     panel_width: usize,
+    pub ansi_prefix_sgr: String,
     pub line_number: usize,
     pub line_changes: Option<LineChanges>,
 }
@@ -67,6 +68,7 @@ impl<'a> Printer<'a> {
             colors,
             config,
             decorations,
+            ansi_prefix_sgr: String::new(),
             line_number: 0,
             line_changes: None,
         }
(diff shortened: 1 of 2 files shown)


src/printer.rs
```diff
@@ -158,12 +159,17 @@ impl<'a> Printer<'a> {
...
             for &(style, region) in regions.iter() {
                 let mut ansi_iterator = AnsiCodeIterator::new(region);
-                let mut ansi_prefix = "";
+                let mut ansi_prefix: String = String::new();
                 for chunk in ansi_iterator {
                     match chunk {
                         // ANSI escape passthrough.
                         (text, true) => {
-                            ansi_prefix = text;
+                            if text.chars().last().unwrap() == 'm' {
```
sharkdp: Can we prevent the usage of `unwrap()` here? Maybe by using `.map_or(false, |c| c == 'm')`?

eth-p:
I used your suggested change, and I added a little extra logic to clear the `ansi_prefix_sgr` variable if it detects a SGR reset code (`ESC[0m`).

sharkdp:
Sorry for the late reply.. I'm happy to merge this once the style changes are fixed (I tried to get an up-to-date rustfmt version to fix it myself, but couldn't get it to work).

Issue #177, opened 2018-06-25
would be nice to change default theme via command line or environment variable versus alias


kaolin:
among other issues, setting the suggested alias breaks subcommands like "bat cache".

sharkdp:
Thank you for the feedback. Definitely agree with this request (see also #173).

Let's introduce `BAT_THEME` and `BAT_STYLE`.

sharkdp:
Released in [bat 0.5](https://github.com/sharkdp/bat/releases/tag/v0.5.0).

PR #187, opened 2018-07-07
Allow specifying the theme via the `BAT_THEME` environment variable

mandx:
The `--theme` command line option stills takes precedence and this change preserves how errors are handled when it's used: If a theme name that doesn't exist is specified using the argument, this error is fatal. However, if a theme that doesn't exist is specified using the environment variable, the error is logged to `stderr` and the "Default" theme is loaded as a fallback.

Code looks a bit too nested, my intention is to keep the changes to a minimum, while preserving the original behavior.

Fixes sharkdp/bat#177

Any feedback is more than welcome!

### DIFF
--- a/src/features.rs
+++ b/src/features.rs
@@ -5,6 +5,7 @@ use diff::get_git_diff;
 use errors::*;
 use output::OutputType;
 use printer::Printer;
+use std::env;
 use std::fs::File;
 use std::io::{self, BufRead, BufReader};
 use syntect::easy::HighlightLines;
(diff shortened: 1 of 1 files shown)


sharkdp:

Please allow for a few more days until I find the time to properly review this.

src/features.rs
```diff
@@ -57,7 +58,24 @@ pub fn list_languages(assets: &HighlightingAssets, term_width: usize) {
 }
 
 pub fn print_files(assets: &HighlightingAssets, config: &Config) -> Result<bool> {
-    let theme = assets.get_theme(config.theme.unwrap_or("Default"))?;
+    let theme = match config.theme {
```
sharkdp: This whole logic should probably be moved to the `App::config` function in `app.rs`. Once `print_files` is called, everything should already be configured properly (and `config` should be ready to use). 

Also, instead of using `match` statements on `Option`, I think we should be able to use combinator functions from [`std::option`](https://doc.rust-lang.org/std/option/enum.Option.html). For example:

```rust
env::var("BAT_THEME").or(self.matches.value_of("theme")).unwrap_or("Default")
```

Concerning error handling, I'd be okay with always using `?` (fatal error).

sharkdp:
Closing in favor of #191

PR #189, opened 2018-07-14
Add spec for using cargo-rpm

cfcosta:
With this, fedora/rhel/centos users can use their default package
manager to install bat, without needing cargo or rust.

### DIFF
--- a/.rpm/bat.spec
+++ b/.rpm/bat.spec
@@ -1,0 +1,33 @@
+%define __spec_install_post %{nil}
+%define __os_install_post %{_dbpath}/brp-compress
+%define debug_package %{nil}
+
+Name: bat
+Summary: A cat(1) clone with wings.
+Version: @@VERSION@@
+Release: 1
+License: MIT or ASL 2.0
+Group: Applications/System
+Source0: %{name}-%{version}.tar.gz
+URL: https://github.com/sharkdp/bat
+
+BuildRoot: %{_tmppath}/%{name}-%{version}-%{release}-root
+
+%description
+%{summary}
+
+%prep
+%setup -q
+
+%install
+rm -rf %{buildroot}
+mkdir -p %{buildroot}
+cp -a * %{buildroot}
+
+%clean
+rm -rf %{buildroot}
+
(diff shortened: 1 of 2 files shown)


sharkdp:
Great! Thank you very much!

Can we move `bat.spec` to a different folder (maybe `assets/packaging` or `ci`) or is `.rpm` required?

Also, it would be fantastic if we could add installation instructions for the mentioned distributions to the README file. I'm not familar with cargo-rpm and RPM- based package managers.

edouardmenayde:
:+1:

stevenpwalsh05:
I was able to compile from source to get this to work on Fedora, but it would have been much simpler if a .rpm existed

moosingin3space:
FYI, the Fedora community is standardizing around [`rust2rpm`](https://pagure.io/fedora-rust/rust2rpm) as our solution to generating RPMs from crates.

moosingin3space:
@sharkdp it means that if you wish `bat` to be packaged in Fedora proper, `rust2rpm` or similar will need to be used to be compatible with Fedora packaging guidelines for Rust.

As a Fedora user and a Rust enthusiast, I think we should figure out a good solution for automatically packaging crates, possibly using COPR, but that's beyond your scope.

sharkdp:
@moosingin3space Thank you very much for your feedback. I'm going to close this for now and leave it up package maintainers to create RPMs/coprs/.. etc.

Hope this is fine for everyone.

sharkdp:
There is actually a copr by @cfcosta, but it's outdated:

https://copr.fedorainfracloud.org/coprs/cfcosta/bat/

moosingin3space:
I just had a look at that copr, and it seems to just distribute binaries in the SRPM, which is a big no-no in Fedora packaging policy.

For the time being, I'm using it from `cargo` and would recommend most users do that. When I get time, I might look into building a copr that's compatible with Fedora policy so `bat` can be officially packaged. It's an awesome tool!

sharkdp:
@moosingin3space Sounds great - Let us know!

PR #191, opened 2018-07-21
Allow specifying the theme via the `BAT_THEME` environment variable

mandx:
The `--theme` command line option stills takes precedence and this
change preserves how errors are handled when it's used: If a theme name
that doesn't exist is specified using the argument, this error is fatal.
However, if a theme that doesn't exist is specified using the environment
variable, the error is logged to `stderr` and the "Default" theme is
loaded as a fallback.

This a refactor/variation of sharkdp/bat#187 based on feedback from @sharkdp.

Fixes sharkdp/bat#177

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -283,7 +283,8 @@ impl App {
             },
             term_width: Term::stdout().size().1 as usize,
             files,
-            theme: self.matches
+            theme: self
+                .matches
                 .value_of("theme")
                 .and_then(|theme_name_arg| Some(String::from(theme_name_arg)))
                 .or_else(|| {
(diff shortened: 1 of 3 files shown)


sharkdp:
Thank you very much for the update.

I hope it's okay that I made a few small modifications. Bat will now always just show a warning when the theme is unknown (either via option or environment variable).

mandx:
This was more of a proof-of-concept of how does the code would look like after your feedback on #187, and I'm glad you like this better!

Of course I don't mind the changes, actually I appreciate them; I am learning a thing or two from your commits.

Thanks!

PR #198, opened 2018-08-17
replace trait object BufRead by generic

tathanhdinh:
Instead of using `Box` we can use generic for `print_file_ranges`. It is just a matter of taste, I don't know which one is better.


tathanhdinh:
Strange, a [built](https://travis-ci.org/sharkdp/bat/jobs/417325218) failed with single message: `An error occurred while generating the build script.`

sharkdp:

I definitely prefer your version. I have restarted the failed build. The `nightly` build is failing due to formatting issues. I can take care of that.

PR #199, opened 2018-08-18
Avoid floating point arithmetic in RGB→8-bit ANSI approximation

mina86:


sharkdp:
Very cool! Thank you very much.

(I ran a quick test to check that this leads to the same result for all RGB colors)

PR #200, opened 2018-08-18
Revert "replace trait object BufRead by generic"

sharkdp:
@tathanhdinh Unfortunatly, I need to revert your commit from #198 because it doesn't quite keep the same behavior. If there is a non-existing file, the following happens (the header is printed on STDOUT):

```
▶ bat non-existing
[bat error]: No such file or directory (os error 2)
───────┬───────────────────────────────────────────────────
       │ File: non-existing
───────┼───────────────────────────────────────────────────
```

while it should just show the error message:

```
▶ bat non-existing
[bat error]: No such file or directory (os error 2)
```

This is due to the reordering of `File::open` and the `print_header` call.

### DIFF
--- a/src/features.rs
+++ b/src/features.rs
@@ -86,34 +86,25 @@ fn print_file(
     printer: &mut Printer,
     filename: Option<&str>,
 ) -> Result<()> {
-    printer.print_header(filename)?;
+    let stdin = io::stdin(); // TODO: this is not always needed
+    {
+        let reader: Box<BufRead> = match filename {
+            None => Box::new(stdin.lock()),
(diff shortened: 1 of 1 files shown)


tathanhdinh:
Hi @sharkdp: Never minds, and thanks a lot for your feedback
IMHO, this is because of `print_header` inconsistency: it shouldn't print anything if the file doesn't exist (the assumed behaviour of `bat`).
What do you think?

sharkdp:
Yes, but I wouldn't like to add the file-exists-logic in the `print_header` function.

Issue #212, opened 2018-08-22
Add '-p' as a shortcut for '--style=plain'


sharkdp:
The "plain" style is sometimes desirable to copy&paste from the terminal without the line numbers or Git modifications. There should be a short option (`-p`) that can be used to set `--style=plain`.

ms2300:
Can I help out and add this feature in?

sharkdp:
Yes, absolutely! Any help would be greatly appreciated.

sharkdp:
Closed via #222

PR #222, opened 2018-08-26
#212 : added -p as an alias for --style=plain

ms2300:
Slight amount of pasta, and can totally refactor if needed

Edit: Tested by diff'ing outputs of --style=plain and -p over multiple files. Should I add formal tests?


ms2300:
You have any idea why the build failed? Something with rust-nightly?

Issue #213, opened 2018-08-22
Show a quick preview of different themes for '--list-themes'


sharkdp:
It would be great if we could show a short snippet of highlighted code for each of the syntax highlighting themes.

PR #223, opened 2018-08-26
#213 Display preview of themes with --list-themes

ms2300:
Outputs every theme with auto style to the terminal one-by-one

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -130,12 +130,11 @@ pub fn list_languages(assets: &HighlightingAssets, term_width: usize) {
 
 pub fn list_themes(assets: &HighlightingAssets, config: &mut Config) {
     let themes = &assets.theme_set.themes;
+    config.files = vec![Some("assets/hello.rs")];
     for (theme, _) in themes.iter() {
         println!("{}", theme);
         config.theme = theme.to_string();
-        config.files = vec![Some("assets/hello.rs")];
-        let controller = Controller::new(&config, &assets);
-        let _res = controller.run();
+        let _controller = Controller::new(&config, &assets).run();
     }
 }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -34,7 +34,7 @@ use std::process;
 
 use ansi_term::Colour::Green;
 
-use app::App;
+use app::{App, Config};
 use assets::{clear_assets, config_dir, HighlightingAssets};
 use controller::Controller;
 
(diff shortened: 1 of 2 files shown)


ms2300:
This is a low-effort way of doing this ticket but it should

src/main.rs
```diff
@@ -128,10 +131,17 @@ pub fn list_languages(assets: &HighlightingAssets, term_width: usize) {
     }
 }
 
-pub fn list_themes(assets: &HighlightingAssets) {
+pub fn list_themes(assets: &HighlightingAssets, config: &mut Config) {
```
sharkdp: I was thinking that `list_themes` could take a non-mutable reference to `Config` and could then  create a mutable copy by calling `.clone()` in the body of the function.

src/main.rs
```diff
@@ -128,10 +131,17 @@ pub fn list_languages(assets: &HighlightingAssets, term_width: usize) {
...
     let themes = &assets.theme_set.themes;
+    let mut style = HashSet::new();
+    style.insert(OutputComponent::Plain);
+    config.files = vec![Some("assets/theme_preview.rs")];
+    config.output_components = OutputComponents(style);
     for (theme, _) in themes.iter() {
-        println!("{}", theme);
+        println!("{}\n", Style::new().bold().paint(theme));
```
sharkdp: It'd be great if we could make this work in Rust 1.26 (build currently fails)

ms2300: I'm using the most recent example from the docs ... if we want it to compile for anything other than stable Rust we may want to unbold the titles or look for another library (which would kinda be overkill)

sharkdp: I think its just that `theme` needs to be casted explicitly ... I can look into this.

src/main.rs
```diff
@@ -128,10 +131,18 @@ pub fn list_languages(assets: &HighlightingAssets, term_width: usize) {
     }
 }
 
-pub fn list_themes(assets: &HighlightingAssets) {
+pub fn list_themes(assets: &HighlightingAssets, cfg: &Config) {
     let themes = &assets.theme_set.themes;
+    let mut config = cfg.clone();
+    let mut style = HashSet::new();
+    style.insert(OutputComponent::Plain);
+    config.files = vec![Some("assets/theme_preview.rs")];
```
sharkdp: I missed that previously, but that's obviously not going to work if we are not in `bat`s source folder.

sharkdp: I think I have a solution for this

ms2300:
So just as a learning experience here, what needed to change in order the theme_preview to work in binary?

sharkdp:
> Thanks for fixing it up and merging by the way

@ms2300 Well, Thank you!

> what needed to change in order the theme_preview to work in binary

The important part is the `include_bytes!(..)` macro in commit 84cd8fe3f135d56e9ed3214c310e51aea8e53d26. This macro reads the contents of the file **at compile time** and includes it as a byte-array. Just as if we had copied the file contents into the source files.

ms2300:
Ahh ok, I'll take a more in depth look at that commit, thanks for the answer

Issue #224, opened 2018-08-27
Theme: Zenburn


colindean:
I tested the canonical [TextMate version](https://github.com/colinta/zenburn/blob/master/zenburn.tmTheme) which I used as the basis for [atom-zenburn](https://atom.io/themes/zenburn) and it works very well.

colindean:
I guess including this is "as easy as" including https://github.com/colinta/zenburn as a submodule off of `assets`?

sharkdp:
Did you see that you can [add your own themes](https://github.com/sharkdp/bat#adding-new-themes) to `bat`?

Do you think *zenburn* should be included by default?

colindean:
Yes, I did see that I could add them and I followed that process to see how well that particular theme version worked. It works well with `bat`.

I think it should be included by default. I believe that zenburn is widely used but I don't have statistics prove that beyond 12,585 downloads of atom-zenburn since I launched it a few years ago and [403,183 code results](https://github.com/search?q=zenburn&type=Code) for this unique word on GitHub.

sharkdp:
Ok. The theme looks quite "orthogonal" to the others, so I'm okay with adding it to the default themes.

> I guess including this is "as easy as" including colinta/zenburn as a submodule off of assets?

Yes, exactly. I will take care of updating the binary theme dump `themes.bin` before releasing the next version.

PR #227, opened 2018-08-28
Adds zenburn theme submodule

colindean:
Work towards #224 but @sharkdp is going to rebuild themes.bin


sharkdp:
Thank you very much!

Issue #210, opened 2018-08-21
Add windows build to releases?


LippyBoy:
Since bat seems to support colors in Windows 10, would it be possible to add windows binaries to the releases page with each version?

sharkdp:
Thank you for the feedback. It is certainly possible.

I quickly tried to build `bat` in a Windows 10 VM and was surprised to see that (at least superficially) everything seems to work :sparkles: 

![image](https://user-images.githubusercontent.com/4209276/44422966-0650eb80-a585-11e8-8793-fe0933bcbc56.png)

I'm a little hesitant because maintaining a working Windows version has lead to a huge amount of work in two other Rust projects of mine [[1]](https://github.com/sharkdp/fd) [[2]](https://github.com/sharkdp/hyperfine).

I'm inclined to say "yes" though, if:
- Someone takes over the work to create an `appveyor.yml` configuration such that we have CI for Windows (Note: the `git2` crate requires CMake, so that needs to be installed in the AppVeyor images). This would also involve some work on getting the unit tests up and running (some of them do fail on Windows).
- We don't have to add huge amounts of `#[cfg(windows)]` parts throughout the code base.
- We only try to support Windows 10 + PowerShell. `cmd.exe` is a pain

be5invis:
> We only try to support Windows 10 + PowerShell. cmd.exe is a pain

I do not think PowerShell is a key part since `bat` is an EXE rather a shell script.
Some notes: box-drawing glyphs under Chinese or Japanese OSes may occupy two console cells rather than one, so be careful.

sharkdp:
> I do not think PowerShell is a key part since bat is an EXE rather a shell script.

I'm certainly not an expert but there are things where we could (or would) run into potential problems:

- ANSI colors do not work in `cmd.exe` and I don't want to switch to a another terminal-color library.
- Argument parsing in `cmd.exe` is weird (https://github.com/sharkdp/fd/issues/299).

keith-hall:
For reference, I can confirm that `bat` is correctly showing syntax highlighting in Windows 7 using PowerShell 6 inside [ConEmu](https://conemu.github.io/), although I had to execute with `--color=always`.

mattpodwysocki:
[...]

PR #231, opened 2018-08-29
Windows version of bat

anykao:
I'd like to execute bat in powershell.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,7 +1,7 @@
 environment:
   global:
     TARGET: x86_64-pc-windows-msvc
-    RUST_VERSION: stable    
+    RUST_VERSION: stable
     CRATE_NAME: bat
     CARGO_HOME: "c:\\cargo"
     RUSTUP_HOME: "c:\\rustup"
@@ -12,7 +12,7 @@ install:
   - set PATH=%PATH%;C:\cargo\bin
   - rustc -Vv
   - cargo -V
-  
+
 build: false
 test_script:
   - cargo test --target %TARGET% --verbose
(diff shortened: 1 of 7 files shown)


anykao:
Note, the auth_token should change, and change email notification as you like.

sharkdp:
I have enabled AppVeyor for this repository, so we should be able to trigger a build with the next commit on your branch.

Can you remind me how to set up the auth_token? Is it the same that we have in `.travis.yml`?

Is there anything else that I need to do?

I guess you have seen #210?

appveyor.yml
```diff
@@ -0,0 +1,41 @@
...
+    RUSTUP_HOME: "c:\\rustup"
+
+install:
+  - appveyor DownloadFile https://win.rustup.rs/ -FileName rustup-init.exe
+  - rustup-init.exe -y --default-host %TARGET%
+  - set PATH=%PATH%;C:\cargo\bin
+  - rustc -Vv
+  - cargo -V
+  
+build: false
```
sharkdp: Why is this set to `false`? It currently doesn't build `bat` or run the tests. If we want automatic deployment for Windows, we definitely need to enable continuous integration.

anykao:
Thank you for reviewing PR.
I have seen #210 , but as you said, there are some tests failed in windows.
image
Should we anyway add `cargo test` to Windows build.

As for auth_token. The same method as `fd` will be suffice, that is 
```
  # Here's how:
  # - Go to 'https://github.com/settings/tokens/new' and generate a Token with only the
  # `public_repo` scope enabled
  # - Then go to 'https://ci.appveyor.com/tools/encrypt' and enter the newly generated token.
  # - Enter the "encrypted value" below
```

anykao:
Add cargo test and disabled test_snapshots on windows.
If there are any other changes should be done, just let me know.

sharkdp:
I have split the style-tests into multiple tests to see which of them fail.

sharkdp:
I found out why the Git modifications were not shown on Windows. Now all unit tests pass :sparkles:

PR #240, opened 2018-08-30
note the LESS enviromental variable

VVelox:


VVelox:
Yeah. This is how to properly pass args to less. This is way more efficient than having a shell script invoke it it different args.

Found it when looking at the man page for less yesterday.

README.md
```diff
@@ -225,6 +225,9 @@ script as a wrapper, for example:
 less --tabs 4 -RF "$@"
 ```
 
+Alternatively if you want to pass options to less, you can also use the enviromental variable 'LESS'.
+So if you wanted to do the item above, you would set set LESS equal to '--tabs 4 -RF "$@"'.
```
sharkdp: * Should the `"$@"` part be removed?
* Word "set" used twice.
* Please use <code>`</code> instead of `'` in both sentences.

sharkdp:
This is not really going to work, as command-line arguments take precedence over the `LESS` environment variable:

From `man less`:
```
       The  environment  variable  is  parsed before the command line, so command line options override the LESS environment
       variable.  If an option appears in the LESS variable, it can be reset to its default value on  the  command  line  by
       beginning the command line option with "-+".
```

VVelox:
@sharkdp Actually it does. The point of it is the suggestion in the README is a terrible solution and exactly how it should not be done.

sharkdp:
you can add new arguments via `LESS`, but you cannot undo the ones that `bat` already sets via the command line.

Issue #245, opened 2018-08-31
Compatibility with POSIX cat


muendelezaji:
First, thanks for this very useful utility. This is not really a bug at this point since it appears to be a design choice (#82), but thought it would be nice to get clarification on this.

`bat` treats files named `cache` (and previously, `init-cache`) as a subcommand and fails when it's the first argument. Granted this only happens when the file appears first AND is in the current directory, so probably not a big deal. But it does mean this just misses being a drop-in replacement for `cat` (esp. in non-interactive scenarios) and potentially breaks [POSIX compliance](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/cat.html) (only `-` is a special filename). I see there's also talk on adding more subcommands on #23 which would depart further from compatibility.

In following on the [project goals](https://github.com/sharkdp/bat#project-goals-and-alternatives), is there any good way around this? Perhaps by using `--cache-X` or other approach

BK1603:
I would like to work on this one if no one else is working on it!

PR #275, opened 2018-09-04
Fixed issue #245

BK1603:
Now if a cache file exists in the current directory, and the user passes
no arguments to the cache command, the cache file would be displayed.
If however the user uses cache command with arguments, the cache command
would be executed as normal regardless of wether the file cache exists
in the current directory or not.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1,5 +1,6 @@
 use std::collections::HashSet;
 use std::env;
+use std::path::Path;
 
 use atty::{self, Stream};
 
@@ -9,7 +10,6 @@ use console::Term;
 
 #[cfg(windows)]
 use ansi_term;
-
 use assets::BAT_THEME_DEFAULT;
 use errors::*;
 use line_range::LineRange;
@@ -102,6 +102,10 @@ impl App {
             AppSettings::ColorNever
         };
 
+        // Check if the current directory contains a file name cache, if it does
+        // do not make the arguements for subcommand 'cache' required.
+        let arg_group_required = !Path::new("./cache").exists();
+
         ClapApp::new(crate_name!())
             .version(crate_version!())
             .global_setting(clap_color_setting)
@@ -125,8 +129,7 @@ impl App {
                          to read from standard input.",
                     ).multiple(true)
                     .empty_values(false),
-                    )
-            .arg(
+            ).arg(
                 Arg::with_name("language")
                     .short("l")
                     .long("language")
@@ -303,7 +306,7 @@ impl App {
                     ).group(
                         ArgGroup::with_name("cache-actions")
                             .args(&["init", "clear", "config-dir"])
-                            .required(true),
+                            .required(arg_group_required),
                     ).arg(
                         Arg::with_name("source")
                             .long("source")
--- a/src/main.rs
+++ b/src/main.rs
@@ -181,8 +181,21 @@ fn run() -> Result<bool> {
 
     match app.matches.subcommand() {
         ("cache", Some(cache_matches)) => {
-            run_cache_subcommand(cache_matches)?;
-            Ok(true)
+            // If there is a file named 'cache' in the current working directory,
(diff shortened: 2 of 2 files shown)


BK1603:
Okay so like I have tried everything I could to pass the travis check, but I just can not pass it :sweat_smile: 

Like, please look at it and let me know what's wrong, cargo fmt says things have been changed but then removes and adds the same line, I am pretty confused.

sharkdp:
> Like, please look at it and let me know what's wrong, cargo fmt says things have been changed but then removes and adds the same line, I am pretty confused.

It's just trailing whitespace that `rustfmt` is complaining about. I fixed it. For the future: you can just use [`cargo fmt`](https://github.com/rust-lang-nursery/rustfmt) and it will just auto-format all the code for you :smile:.

src/main.rs
```diff
@@ -181,8 +181,21 @@ fn run() -> Result<bool> {
...
-            run_cache_subcommand(cache_matches)?;
-            Ok(true)
+            // If there is a file named 'cache' in the current working directory,
+            // arguments for subcommand 'cache' are not mandatory.
+            // If there are non-zero arguments, execute the subcommand cache, else, open the file cache.
+            if !cache_matches.args.is_empty() {
+                run_cache_subcommand(cache_matches)?;
+                Ok(true)
+            } else {
+                let mut config = app.config()?;
```
sharkdp: The code duplication in this `else`-clause is kind of unfortunate (it mirrors most steps that we normally do, if no subcommand is given).

I was wondering whether we could simply take out the `cache` subcommand completely if a `cache` file exists. We wouldn't get the behavior which you implemented here (i.e. that `bat cache --init` still works)

BK1603: Umm, yeah I thought about it, like implementing that in the app.rs file, but, umm, how about we make a function for calling the controller and call it first here and then in the other match clause?

BK1603: Or else we can do what you are suggesting, but yep that is gonna drop the functionality there :sweat_smile:

sharkdp: Ok, sounds good. We should probably move the `HighlightingAssets::new()` part into this new function (and remove the `assets` argument from `list_languages` and `list_themes` and just call `HighlightingAssets::new()` inside those functions).

BK1603: I will be doing what you just said! Should be ready in a few minutes!

src/app.rs
```diff
@@ -102,6 +103,10 @@ impl App {
             AppSettings::ColorNever
         };
 
+        // Check if the current directory contains a file name cache, if it does
+        // do not make the arguements for subcommand 'cache' required.
+        let arg_group_required = !Path::new("./cache").exists();
```
sharkdp: I think this can be just `"cache"`, right?

BK1603:
Sorry for all the mess :sweat_smile: 

Also if there is anything else that isn't like, with the conventions please let me know

And thanks a lot for reviewing the code and telling me about the mistakes :smile:

sharkdp:
Do you want me to call `cargo fmt` again? :smile:

BK1603:
I fixed it :sweat_smile: 
Like when I ran cargo fmt on my laptop, it added a few new lines that travis didn't seem to like :sweat_smile: 
I apologize once again :sweat: 
And thanks a lot once again for reviewing the code :smile:

sharkdp:
> Like when I ran cargo fmt on my laptop, it added a few new lines that travis didn't seem to like 
> I apologize once again

No worries! You need to make sure to have the latest version of rustfmt-nightly. The style changes slightly from time to time.

> And thanks a lot once again for reviewing the code

Well, thank you for your contribution!

BK1603:
IT'S DONE!

sharkdp:
Released in [v0.7.0](https://github.com/sharkdp/bat/releases/tag/v0.7.0).

PR #279, opened 2018-09-05
Add support for JSX in javascript and typescript

tobenna:


sharkdp:
Thank you very much.


How will this work?

Will this be used for every `.js` file? Will it overwrite the default JavaScript syntax?

If we add this, do we still need the TypeScript syntax?

tobenna:
@sharkdp 

> Will this be used for every .js file? Will it overwrite the default JavaScript syntax?

I'm not sure how the language resolution works in syntect.
But since there's already have support for `.js`
we can take [this](https://github.com/sharkdp/bat/pull/279/files#diff-b161992fce76f3192aaf04b511d20ec7R6) out.

So that `.js` files use the default syntax.

> If we add this, do we still need the TypeScript syntax?

We'll still need the other TypeScript syntax this just adds highlighting for `.tsx`

sharkdp:
Thank you for the clarification.

It seems like this is a more modern JavaScript syntax, right? So I guess we can also leave it in.

tobenna:
I tested it locally and it seems to work fine with both of them in there.

PR #290, opened 2018-09-07
Handle `less` in a portable way

Aankhen:
The current logic for special-casing `less` relies on the pager being specified as the literal string or using Unix-style paths. It also ignores the actual pager specified when calling it, so even if you set `PAGER`/`BAT_PAGER` to `/my/path/to/less`, `bat` will simply call `less`.

This PR treats any specified pager as a path and checks only the file stem so that `C:\foo\bar\less.exe`, `/foo/bar/less`, and `less` are all recognized.  It also uses the specified `less`.

I managed to write a test for this by creating a stub `bin/less.rs` that verifies its arguments and the environment. However, this binary was installed along with `bat` when running `cargo install`. [I haven’t yet found a way to test the behaviour without this issue](https://stackoverflow.com/q/52205295/8492116), so for the moment I can only offer the fix without the test (but with the assurance that I’ve tested its behaviour both manually and automatically).

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -1,5 +1,7 @@
 use std::env;
+use std::ffi::OsString;
 use std::io::{self, Write};
+use std::path::PathBuf;
 use std::process::{Child, Command, Stdio};
 
 use app::PagingMode;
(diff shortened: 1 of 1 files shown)


sharkdp:
> for the moment I can only offer the fix without the test (but with the assurance that I’ve tested its behaviour both manually and automatically).

Thank you for thinking about the tests! I'm fine with merging this without tests for now, as I still plan to extend our integration tests anyway.

Aankhen:
Excellent!

sharkdp:
Released in [v0.7.0](https://github.com/sharkdp/bat/releases/tag/v0.7.0).

PR #291, opened 2018-09-07
Disable wrapping when --plain is used. (Fixes #289)

### LINKED ISSUE #289: bat -p should also default --wrap never
When using cat on files containing long string like in `id_rsa.pub`, you are able to easily highlight and select the text and paste it anywhere and its copied verbatim. Using bat, since the default is to wrap output internally and not have the terminal do it, this causes additional line breaks in the copied text. This is extremely misleading when you are doing things like transferring text and using terminals of the same width so you do not see the added line breaks. 

Adding the `--wrap never` along with `-p` is cumbersome and leads to falling back on cat instead. 

Is this option configurable if I was to alter the theme?

eth-p:
This is a fix for #289. As I explained in my reply [here](https://github.com/sharkdp/bat/issues/289#issuecomment-419501495), it didn't make sense to leave wrapping on with `--plain`, since the terminal is more than capable of wrapping the lines itself when the sidebar isn't used.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -415,10 +415,8 @@ impl App {
                         } else {
                             InputFile::Ordinary(filename)
                         }
-                    })
-                    .collect()
-            })
-            .unwrap_or_else(|| vec![InputFile::StdIn])
+                    }).collect()
+            }).unwrap_or_else(|| vec![InputFile::StdIn])
     }
 
     fn output_components(&self) -> Result<OutputComponents> {
--- a/src/app.rs
+++ b/src/app.rs
@@ -190,10 +190,11 @@ impl App {
                     .long("plain")
                     .conflicts_with("style")
                     .conflicts_with("number")
-                    .help("Show plain style (alias for '--style=plain').")
+                    .conflicts_with("wrap")
+                    .help("Show plain style (alias for '--style=plain' and '--wrap=never').")
                     .long_help(
-                        "Only show plain style, no decorations. This is an alias for \
-                         '--style=plain'",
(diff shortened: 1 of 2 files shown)


sharkdp:
@eth-p Thank you very much!

It would be great if we could disable wrapping whenever the style is `plain` (not just of `--plain` is passed on the command line, but also if `--style=plain` is passed).

I also wouldn't care too much about setting `--plain` and `--wrap` as conflicting. The auto-mode of `wrap` would just be `never` whenever `style=="plain"`.

eth-p:
@sharkdp Yeah, that sounds like a better (and more consistent) solution.

I removed that functionality from `--plain` and made it automatic.

src/app.rs
```diff
@@ -190,7 +190,7 @@ impl App {
                     .long("plain")
                     .conflicts_with("style")
                     .conflicts_with("number")
-                    .help("Show plain style (alias for '--style=plain').")
+                    .help("Show plain style (alias for '--style=plain'.")
```
sharkdp: Just a small leftover missing parens here.

sharkdp:
Thank you for the update. This looks great.

src/app.rs
eth-p: Good catch! My bad

sharkdp:
I guess your `rustfmt` version is slightly outdated. I can run `cargo fmt` on this branch, if you want.

eth-p:
If it's easier for you to do it before merging, sure. I'm definitely going to update mine in the meantime, though.

sharkdp:
I install rustfmt via
```
cargo +nightly install rustfmt-nightly -f
```

PR #293, opened 2018-09-08
Do not show header if path is a directory (Fixes #292)

### LINKED ISSUE #292: Do not show header if path is a directory
```
> bat some-directory
[bat error]: Is a directory (os error 21)
───────┬──────────────────────────────────────────────────
       │ File: some-directory
───────┼──────────────────────────────────────────────────

```

When calling `bat` on a directory (more typically, when using `bat *`), we should not print a header for this entry.

Rigbert:
Fixes #292 

Adds a conditional in print_header that checks if the file is in a directory and exits without printing if it is.

### DIFF
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -19,6 +19,8 @@ use errors::*;
 use style::OutputWrap;
 use terminal::{as_terminal_escaped, to_ansi_color};
 
+use std::fs;
+
 pub trait Printer {
     fn print_header(&mut self, handle: &mut Write, file: InputFile) -> Result<()>;
     fn print_footer(&mut self, handle: &mut Write) -> Result<()>;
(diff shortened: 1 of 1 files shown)


sharkdp:

Your approach would work, but I think a better place to handle this would be inside the `Controller::print_file` function. We already have a `match` statement there and call `File::open` to get a handle. We could use this handle to make sure that it's actually a file before we pass it on to `BufReader::new`.

sharkdp:
This would also be a good chance to throw a better error message:
```
[bat error]: 'some-directory' is a directory.
```

instead of the current version:
```
[bat error]: Is a directory (os error 21)
```

sharkdp:
I'm going to close this in favor of e098eb43a28f55942f3185033ebe3e09d86198d7

Issue #260, opened 2018-09-02
Incorrect colours on Windows


Aankhen:
(copied from my comment in #107)

I installed `bat` on a recent Windows 10 Pro x64 build today and tried it in Cmder, the Cmder installation of ConEmu, PowerShell, and the regular Command Prompt. The first two gave me this:

![White-on-grey text everywhere except the filename](https://user-images.githubusercontent.com/978992/44955053-8acb3480-aeca-11e8-8fbd-ab3bd744af1f.png)

PowerShell gave me this (presumably just different default terminal colours; please excuse the crude black censor bar):

![White-on-pink text everywhere except the filename](https://user-images.githubusercontent.com/978992/44955069-d1209380-aeca-11e8-9033-73f1c1aec1d8.png)

Command Prompt gave me this, like the first two:

![White-on-grey text everywhere except the filename](https://user-images.githubusercontent.com/978992/44955084-075e1300-aecb-11e8-9d58-18934c823718.png)

`bat` looks really exciting and I’m eager to use it, so I hope this is just a misconfiguration on my end! FWIW, I get the same results without a `BAT_THEME` and with a `BAT_THEME` of `Monokai Extended Bright`.

sharkdp:
Thank you for the bug report!

This does *not* look like the same error as #252 because you actually get the styling (grid, line numbers, etc.) as well as *some* colors (although incorrect ones).

Something seems wrong with ANSI code handling on your terminals. Can you try to use `--color=never` to see if that produces a correct (colorless) output? Could you also try to switch off the pager (`--paging=never`) to make sure that nothing else is interfering?

Aankhen:
Sure thing. Here’s `--color=never`:

![Correct uncoloured output](https://user-images.githubusercontent.com/978992/45012873-bc610e80-b036-11e8-8163-ca7f4368873b.png)

And here’s `--paging=never`:

![Correct coloured output without paging](https://user-images.githubusercontent.com/978992/45012902-dc90cd80-b036-11e8-930f-c16707533831.png)

Looks like that’s indeed where the issue is, so I did some digging…

The `less` being used was from Git Windows. I also have the one from [UnxUtils](http://unxutils.sourceforge.net/), so I temporarily removed the Git one. Now I get ‘invalid charset name’, because the UnxUtils version of `less` only supports a handful of encodings:

```c
struct charset {
	char *name;
	char *desc;
} charsets[] = {
	{ "ascii",	"8bcccbcc18b95.b"		},
[...]

PR #295, opened 2018-09-08
README.md: Add `Using bat on Windows`

Aankhen:
This PR adds a Windows section to the README, based on #260. (I thought a separate section would be better since these steps are more or less required to use paging and colours.)


README.md
```diff
@@ -235,6 +237,20 @@ script as a wrapper, for example:
...
+
+### Paging
+
+Windows only includes a very limited pager in the form of `more`. You can download a Windows binary for `less` [from its homepage](http://www.greenwoodsoftware.com/less/download.html) or [through Chocolatey](https://chocolatey.org/packages/Less). To use it, place the binary in a directory in your `PATH` or [define an environment variable](#using-a-different-pager).
+
+### Colours
+
+Windows 10 natively supports colours in both `conhost.exe` (Command Prompt) and PowerShell since [v1511](https://en.wikipedia.org/wiki/Windows_10_version_history#Version_1511_(November_Update)), as well as in newer versions of bash. On earlier versions of Windows, you can use [Cmder](http://cmder.net/), which includes [ConEmu](https://conemu.github.io/).
+
+**Note:** The Git and MSYS versions of `less` do not correctly interpret colours on Windows. You can disable `bat`’s coloured output using `--color=never`.
```
sharkdp: Hm.. isn't using `bat` without general color support kind of pointless? I think I would rather advise to use `--paging=never` (or `BAT_PAGER=""`) in this case.

sharkdp:
Thank you very much for writing this up!!

README.md
Aankhen: Sure, that makes sense. I’ll update it.

Aankhen:
Thanks for writing `bat` up! 😉

README.md
aasedb: > I think I would rather advise to use `--paging=never` (or `BAT_PAGER=""`) in this case.

@sharkdp AFAIK empty environment variables are not supported on windows and can neither be set via `set`/`setx` nor via GUI and only by editing the registry

sharkdp: @aasedb Thank you for the hint. Can you please open a new ticket/PR so this can be properly tracked?

Issue #208, opened 2018-08-21
feature request: add BAT_STYLE in the same vein as BAT_THEME


sanga:
I suppose you may already have this in mind but I didn't see a tracking issue for it so I'm adding one just in case.

This would be to avoid silly mistakes such as the one in https://github.com/sharkdp/bat/issues/173

sharkdp:
Yes, thanks for adding this request!

sharkdp:
Closed by @ms2300 via #298

sharkdp:
Released in [v0.7.0](https://github.com/sharkdp/bat/releases/tag/v0.7.0).

PR #298, opened 2018-09-10
#208 Added BAT_STYLE env variable functionality

ms2300:
If an environmental variable of bat style is present then the style will be set accordingly every time while still maintaining ability to change style with --style or a variant of that. #208 

Edit: I don't necessarily like adding more and more else if statements to that area of code and spent the majority of time trying to clean it up but this was seemed like the best strategy to me ... all reviewing is very welcome.


src/app.rs
```diff
@@ -428,14 +430,26 @@ impl App {
...
                 values_t!(matches.values_of("style"), OutputComponent)?
                     .into_iter()
                     .map(|style| style.components(self.interactive_output))
                     .fold(HashSet::new(), |mut acc, components| {
                         acc.extend(components.iter().cloned());
                         acc
                     })
+            } else {
+                let style = env::var("BAT_STYLE").unwrap_or(String::from(BAT_STYLE_DEFAULT));
+                match OutputComponent::from_str(&style) {
```
sharkdp: It would be great if we could remove the code duplication here, yes.

Also, it looks like this will not work for something like `BAT_STYLE="numbers,header"`, right?

sharkdp: I would put this code into the previously existing `else` branch and first build an iterator over the `OutputComponent`s (either from `env` or from the `--style` argument`). Afterwards, we can call `.map(..).fold(...)` etc.

src/app.rs
```diff
@@ -430,26 +430,25 @@ impl App {
...
+                    transpose(env::var("BAT_STYLE").ok().map(|style_str| {
+                        style_str
+                            .split(",")
+                            .map(|x| OutputComponent::from_str(&x))
+                            .collect::<Result<Vec<OutputComponent>>>()
+                    }))?;
+
+                values_t!(matches.values_of("style"), OutputComponent)
+                    .ok()
+                    .or(env_style_components)
```
ms2300: Yeah you know I really should have done that from the beginning... my bad

sharkdp: No worries! Thank you very much for your contribution. I took the liberty to change things myself. Hope that's fine with you!

ms2300: I mean I just try to help out because I like using bat and your change was clearly for the better, it's all good. I learn something new every time as well so thank you!

sharkdp:
So the new behavior is the following:

`BAT_STYLE` can be used to set the style:
```
▶ BAT_STYLE=numbers bat test.cpp
   1 int main() {
   2   std::cout << "Hello World!" << std::endl;
   3   return 1;
   4 }
```

`--style` always overwrites `BAT_STYLE` completely:
```
▶ BAT_STYLE=numbers bat test.cpp --style=header,grid
────────────────────────────────────────────────────────────
File: test.cpp
────────────────────────────────────────────────────────────
int main() {
  std::cout << "Hello World!" << std::endl;
  return 1;
}
────────────────────────────────────────────────────────────
```

Unknown style components in `BAT_STYLE` will be reported as an error:
```
▶ BAT_STYLE=number bat test.cpp                     
[bat error]: Unknown style 'number'
```

ms2300:
Yup that's what I think is best too

Issue #184, opened 2018-07-05
First tab-stopp is off by one to the left, resulting in default unix files not being aligned nicely


monkeydom:
E.g.:

```
dom@dreizehn:~$ cat /etc/hosts
##
# Host Database
#
# localhost is used to configure the loopback interface
# when the system is booting.  Do not change this entry.
##
127.0.0.1	localhost
255.255.255.255	broadcasthost
::1             localhost
dom@dreizehn:~$ bat /etc/hosts
───────┬────────────────────────────────────────────────────────────────────────────────────────────
       │ File: /etc/hosts
───────┼────────────────────────────────────────────────────────────────────────────────────────────
   1   │ ##
   2   │ # Host Database
   3   │ #
   4   │ # localhost is used to configure the loopback interface
   5   │ # when the system is booting.  Do not change this entry.
   6   │ ##
   7   │ 127.0.0.1      localhost
   8   │ 255.255.255.255        broadcasthost
   9   │ ::1             localhost
───────┴────────────────────────────────────────────────────────────────────────────────────────────
```

sharkdp:
Thank you for reporting this. This is related to to #166.

sharkdp:
Fix released in [v0.7.0](https://github.com/sharkdp/bat/releases/tag/v0.7.0).

Issue #166, opened 2018-06-02
Wrapping doesn't work properly with tabs.


eth-p:
When implementing the wrapping, I failed to account for files that use tabs for indentation and/or alignment.

## Image

![image](https://user-images.githubusercontent.com/32112321/40880021-cc6007de-66a1-11e8-8d93-a8fc7d5266c4.png)

## Test File

```markdown
# Leading Spaces
    70 ----------------------------------------------------------- 70
    71 ------------------------------------------------------------ 71
    72 ------------------------------------------------------------- 72
    73 -------------------------------------------------------------- 73
    74 --------------------------------------------------------------- 74
    75 ---------------------------------------------------------------- 75
    76 ----------------------------------------------------------------- 76
    77 ------------------------------------------------------------------ 77
    78 ------------------------------------------------------------------- 78
    79 -------------------------------------------------------------------- 79
    80 --------------------------------------------------------------------- 80
    81 ---------------------------------------------------------------------- 81
    82 ----------------------------------------------------------------------- 82
    83 ------------------------------------------------------------------------ 83
    84 ------------------------------------------------------------------------- 84
    85 -------------------------------------------------------------------------- 85

# Leading Tab (4 or 8 Spaces / Tab)
	70 ----------------------------------------------------------- 70
	71 ------------------------------------------------------------ 71
	72 ------------------------------------------------------------- 72
	73 -------------------------------------------------------------- 73
	74 --------------------------------------------------------------- 74
	75 ---------------------------------------------------------------- 75
	76 ----------------------------------------------------------------- 76
	77 ------------------------------------------------------------------ 77
	78 ------------------------------------------------------------------- 78
	79 -------------------------------------------------------------------- 79
	80 --------------------------------------------------------------------- 80
	81 ---------------------------------------------------------------------- 81
	82 ----------------------------------------------------------------------- 82
	83 ------------------------------------------------------------------------ 83
	84 ------------------------------------------------------------------------- 84
	85 -------------------------------------------------------------------------- 85

# Alignment
1	1
22	2
333	3
4444	4
55555	5
666666	6
7777777	7
88888888	8
```

## Solutions
There are a couple of solutions that I think might solve this issue:

### A. Replace `\t` with 4 or 8 spaces.
**Pros:**
- Easy to implement.
- Configurable for user's choice of either 4 or 8 characters.

**Cons:**
- Output is different from source file (maybe only do it if wrapping and styling is enabled?)
- Breaks tab alignment:
    ```
    1	2
[...]

PR #302, opened 2018-09-11
Tab expansion support.

eth-p:
This pull request adds support for expanding tabs into column-aligned spaces (similar to `expand(1)`). Issues #166 and #184 are fixed when using this feature.

By default, tab width is set to 8 characters. This can be overridden with `--tabs=[n]` or `export BAT_TABS=[n]`. It can also be turned off entirely by setting the value to `0`.

To allow for lossless copy-pasting, when neither the pager nor the decorations are being used, the default tab width will be set to `0` (passthrough).

## Implementation

The function `expand` (`src/preprocessor.rs`) is run for every line inside `InteractivePrinter.print_line`, after the syntax highlighting is processed. ANSI characters are directly passed through while tabs are interpreted and replaced with spaces. The number of spaces depends on the calculated cursor position, and will always try to align the columns.

## Additional Information

Three new tests have been created for this feature:

* tabs_4
* tabs_8
* tabs_passthrough

A couple of small changes were made to the tester scripts and `generate_snapshots.py` was fixed to use the interactive printer for generating snapshots.

## Example

Before:

```
sidebar│         |       |       |         <-- Column Alignment
  24   │ // Tab alignment:
  25   │ /*
  26   │        Indent                     <-- Misaligned Indentation
  27   │        1       2       3       4
  28   │ 1      ?
  29   │ 22     ?
  30   │ 333    ?
  31   │ 4444   ?
  32   │ 55555  ?
  33   │ 666666 ?
  34   │ 7777777        ?                  <-- Misaligned Columns
  35   │ 88888888       ?
  36   │ */
```

After:

```
sidebar│         |       |       |
  20   │ // Tab alignment:
  21   │ /*
  22   │         Indent
  23   │         1       2       3       4
  24   │ 1       ?
  25   │ 22      ?
  26   │ 333     ?
  27   │ 4444    ?
  28   │ 55555   ?
  29   │ 666666  ?
  30   │ 7777777 ?
  31   │ 88888888        ?
  32   │ */
```


sharkdp:
@eth-p Thank you very much for this contribution!!

My only concern is what I mentioned in #166:

> One thing to keep in mind: Using a pre-processing step to expand the tabs could be problematic for languages with a syntax that distinguishes between tabs and spaces (i.e. Makefile).

To be honest, I didn't manage to construct an example that show-cases this issue because the Sublime text syntaxes seem to be rather forgiving, but *in principle* we could get syntax highlighting problems due to the replacement of tabs with spaces. As a stupid example, if we wanted to add a syntax for the [Whitespace programming language](https://en.wikipedia.org/wiki/Whitespace_(programming_language)), this approach would certainly be problematic :smile:

image

sharkdp:
> It might incur a bit of a performance penalty, but I moved the tab expansion step to happen after the syntax is highlighted, so this should no longer be a problem.

Fantastic! Let me know if this is ready for a full review.

eth-p:
@sharkdp Everything's fixed. It should be ready for a review now.

src/app.rs
```diff
@@ -144,195 +147,204 @@ impl App {
...
+                .hidden_short_help(true)
+                .long_help(
+                    "This option exists for POSIX-compliance reasons ('u' is for \
+                         'unbuffered'). The output is always unbuffered - this option \
+                         is simply ignored.",
+                ),
+        ).arg(
+            Arg::with_name("tabs")
+                .long("tabs")
+                .short("t")
```
sharkdp: I would like to reserve the short `-t` for now. Maybe we want to have full compatibility with GNU cat at some point (which uses `-t` for something else).

src/app.rs
```diff
@@ -393,6 +407,18 @@ impl App {
                 || self.matches.value_of("color") == Some("always")
                 || self.matches.value_of("decorations") == Some("always")),
             files,
+            tab_width: self
+                .matches
+                .value_of("tabs")
+                .and_then(|w| w.parse().ok())
+                .or_else(|| env::var("BAT_TABS").ok().and_then(|w| w.parse().ok()))
```
sharkdp: could the `.and_then(|w| w.parse().ok())` part be used only once? Or are we dealing with different string types here?

src/app.rs
```diff
@@ -41,6 +41,9 @@ pub struct Config<'a> {
     /// The character width of the terminal
     pub term_width: usize,
 
+    /// The width of tab characters.
+    pub tab_width: usize,
```
sharkdp: How is "no tab expansion" represented? `tab_width==0`? Could this be `Option<usize>` instead?

tests/snapshots/sample.modified.rs
```diff
@@ -10,7 +10,7 @@ fn main() {
         "The perimeter of the rectangle is {} pixels.",
         perimeter(&rect1)
     );
-    println!(r#"This line contains invalid utf8:  "�"#;
```
sharkdp: What's going on in this changeset?

src/app.rs
eth-p: How about `-x` for expand, since `-e` is already used? Or should I remove the short option entirely?

tests/snapshots/sample.modified.rs
eth-p: My IDE probably did something to that when I re-saved the file after adding the tabs section. I'll attempt to fix that.

src/app.rs
eth-p: It's implemented this way so that an invalid value for --tabs would fall back on the environment variable. Should it be done differently?

src/app.rs
eth-p: It probably could be, but it would probably complicate the argument parsing a little bit.

eth-p:
@sharkdp I could make a unit test for the tab expansion function itself, but I don't think it will do as good of a job as the snapshot tests. The expansion function is pretty simple and straightforward, but the way it's used in the printer isn't. Changes to the printer are the most likely cause of it breaking, and I don't think unit tests could cover that as well as the current snapshot tests.

src/app.rs
sharkdp: I think we can remove the short version. It's not something that people will likely modify all the time.

eth-p: Will do.

src/app.rs
sharkdp: I think we should fail with an error message if the value for `--tabs` can not be parsed (instead of silently falling back on the environment-variable value).

src/app.rs
sharkdp: I would favor the `Option` version but I'm also fine with a short comment for now ("A value of zero indicates that no expansion should be performed").

sharkdp:
> @sharkdp I could make a unit test for the tab expansion function itself, but I don't think it's worth it. The expansion function is pretty simple and straightforward, but the way it's used in the printer isn't. Changes to the printer are the most likely cause of it ever breaking in the future, and I don't think unit tests could cover that as well as the current snapshot tests.

Ok :+1:

tests/snapshots/sample.modified.rs
sharkdp: There are still some changes in this line (which is why the snapshots also show modifications in this line).

The `øˆ€€€` part is just the "interpolated" version of what comes out if you try to decode this as utf-8. You actually need to put the exact same bytes back in place (or tell me if you need any help).

eth-p: I might need a bit of help with that, sorry haha...

sharkdp: Note that it can be pretty irritating when viewing the diffs in GitHub or some editors, as they will show the interpolated `øˆ€€€` string in both cases (if you actually put the `øˆ€€€` string in the file or if you put the invalid utf-8 sequence in the file)

sharkdp: Ok, I'll look into this later today.

src/app.rs
eth-p: Added a comment in d404139

src/app.rs
eth-p: Done in d404139

tests/tests.rs
```diff
@@ -3,33 +3,39 @@ mod tester;
...
+    grid_header:                 "grid,header"                 => [wrap: false, tabs: 8],
+    grid_numbers:                "grid,numbers"                => [wrap: false, tabs: 8],
+    header_numbers:              "header,numbers"              => [wrap: false, tabs: 8],
+    changes_grid_header:         "changes,grid,header"         => [wrap: false, tabs: 8],
+    changes_grid_numbers:        "changes,grid,numbers"        => [wrap: false, tabs: 8],
+    changes_header_numbers:      "changes,header,numbers"      => [wrap: false, tabs: 8],
+    grid_header_numbers:         "grid,header,numbers"         => [wrap: false, tabs: 8],
+    changes_grid_header_numbers: "changes,grid,header,numbers" => [wrap: false, tabs: 8],
+    full:                        "full"                        => [wrap: false, tabs: 8],
+    plain:                       "plain"                       => [wrap: false, tabs: 0],
```
sharkdp: @eth-p I had to change this to `tabs: 0` because `generate_snapshots.py` produced a file that had tabs in them for `--style=plain`. I'm not sure why this was different for you. I tried to add `--paging=never` and `--color=never` to make this more robust.

sharkdp:
@eth-p Thank you very much for all of this!

tests/tests.rs
eth-p: @sharkdp That would be because it doesn't process tabs unless it needs to for alignment reasons (e.g. the pager or decorations are used). On second thought, though, that might be a bit confusing when people inconsistently end up with tabs sometimes.

Should I make a PR to remove that?

sharkdp: I think the current behavior is fine. I just wondered why I got a different version of the file when I ran the `generate_..` script.

PR #307, opened 2018-09-12
Update README (libz-dev may be required)

jdhoek:
On Ubuntu 18.04 I had to install `cmake` as well as `libz-dev` to get bat to compile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,8 +178,8 @@ higher. You can then use `cargo` to build everything:
 cargo install bat
 ```
 
-On macOS, you might have to install `cmake` (`brew install cmake`) in order for
-some dependencies to be built.
+You may have to install `cmake` and the `libz` development package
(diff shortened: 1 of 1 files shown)


sharkdp:

There is a short section about this in the README: https://github.com/sharkdp/bat#error-when-compiling-cannot-find--lz

but apparently this doesn't work on Ubuntu. I think the "From source" section is generally the better place, but it would be great if we could merge these.

sharkdp:
Also, note that this shouldn't be necessary, in principle: https://github.com/alexcrichton/libz-sys/issues/20

jdhoek:
That `zlib` workaround didn't seem to work on Ubuntu 18.04. The build failed for me until I installed `zlib-dev`.

jdhoek:
I've merged the two sections (dropping the latter) and reworded the text.

sharkdp:
Thank you very much!

PR #313, opened 2018-09-18
Disable wrapping when `--style` is plain

### LINKED ISSUE #289: bat -p should also default --wrap never

sharkdp:
closes #289

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -81,6 +81,6 @@ impl OutputComponents {
     }
 
     pub fn plain(&self) -> bool {
-        self.0.is_empty()
+        self.0.iter().all(|c| c == &OutputComponent::Plain)
     }
 }


eth-p:
Oops! Thanks for fixing that, @sharkdp!

PR #315, opened 2018-09-19
Add -X to less wrapper example

### LINKED ISSUE #312: Bat with BAT_PAGER set doesn't work on small files
I set up `bat v0.7.0` to use a small wrapper-script for less:

`~/.local/bin/bat-pager`

```bash
#!/bin/bash

less --tabs 4 -RF "$@"
```

Using it **works** for any file large enough to need a pager, and I can use the scrollwheel on my mouse. So far so good.

But when I bat a short file, I see a brief flash of its contents, and then `bat` just **exits with no output**.

```bash
export BAT_PAGER=bat-pager
bat some.xml
```

Unsetting the `BAT_PAGER` environment variable means everything works as expected — except for the scrolling.

### Environment

Ubuntu `18.04`
*Gnome Terminal with Bash*

jdhoek:
Without `-X` the less wrapper script can fail (seen on Ubuntu 18.04).

Fixes #312.

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,7 +263,7 @@ script as a wrapper, for example:
 ```bash
 #!/bin/bash
 
-less --tabs 4 -RF "$@"
+less --tabs 4 -RXF "$@"
 ```
 
 ## Using `bat` on Windows


sharkdp:

Using `-FRX` doesn't seem to make sense, since this is exactly what `bat` usually passes to `less`?

Does scrolling still work for you if you add `-X`?

jdhoek:
No you're right, `-X` breaks scrolling.

PR #318, opened 2018-09-20
Fix typos

katrinleinweber:
See diff ;-) I'm not sure about the `pre-built` correction, but since `pre…` without `-` appeared more often elsewhere in the file, harmonising the spelling seemed sensible to me.


sharkdp:
Thank you very much!

PR #319, opened 2018-09-21
Use ansi_colours package for better true-colour approximation

mina86:


gjnoonan:
👍 LGTM

sharkdp:
Thank you, that makes sense.

PR #321, opened 2018-09-25
controller.rs: Lock `Stdout` before running.

Aankhen:
`bat` locks `Stdout` when listing themes or languages, but not in regular operation. My understanding is that there’s no real downside to locking it if the output is all coming from one place, and it seems like a good idea when dealing with large files. This PR is an inelegant attempt to implement locking in the normal case too.

### DIFF
--- a/src/controller.rs
+++ b/src/controller.rs
@@ -22,7 +22,14 @@ impl<'b> Controller<'b> {
 
     pub fn run(&self) -> Result<bool> {
         let mut output_type = OutputType::from_mode(self.config.paging_mode);
-        let writer = output_type.handle()?;
+        let mut lock;
+        let writer: &mut Write = match output_type {
+            OutputType::Stdout(ref mut stdout) => {
+                lock = stdout.lock();
+                &mut lock
+            }
+            _ => output_type.handle()?,
+        };
         let mut no_errors: bool = true;
 
         for filename in &self.config.files {


sharkdp:

This could certainly lead to some performance improvements, although I doubt that it will matter much if we have syntax highlighting enabled.

It could be interesting for the loop-through mode, though, when we use `bat` in `cat`-compatibility mode (when *not* writing to an interactive terminal).

It would be great to get some benchmark results before merging this. (We can use [hyperfine](https://github.com/sharkdp/hyperfine) with or without the `--show-output` flag for this).

Implementation-wise, it's unfortunate that we have to `match` on the `OutputType` twice, but I currently don't have a suggestion for a better solution.

Aankhen:
Agreed on all of that. I’m not sure what I would need to do to activate loop-through mode. Since it shouldn’t have a terminal when running under `hyperfine`, does that mean any invocation of it will automatically be in loop-through mode?

sharkdp:
> I’m not sure what I would need to do to activate loop-through mode

"Loop-through mode" is always activated when `bat` writes to anything that is not a interactive terminal (i.e. to a pipe or to a file):
```
bat file.txt | head
bat file.txt | less
bat file.txt > other.txt
bat file.txt > /dev/null
```

> Since it shouldn’t have a terminal when running under `hyperfine`, does that mean any invocation of it will automatically be in loop-through mode?

Exactly. If you run `hyperfine "bat file.txt"`, hyperfine will redirect any output to `/dev/null`, which will put `bat` into loop-through mode.

Aankhen:
Well this is a fine result. I wasn’t seeing any significant difference on small files, as expected, so I tried it on the 1 GB [enwik9](https://cs.fit.edu/~mmahoney/compression/textdata.html). #1 is the latest `master`, #2 is the version with locks, both compiled in release mode:

```
λ hyperfine --warmup 3 ".\bat.orig\target\release\bat.exe -p --color never --paging never E:\Media\enwik9" ".\bat\target\release\bat.exe -p --color never --paging never E:\Media\enwik9"
Benchmark #1: .\bat.orig\target\release\bat.exe -p --color never --paging never E:\Media\enwik9

  Time (mean ± σ):     22.721 s ±  0.626 s    [User: 2.2 ms, System: 12.9 ms]

  Range (min … max):   22.023 s … 23.933 s

Benchmark #2: .\bat\target\release\bat.exe -p --color never --paging never E:\Media\enwik9

  Time (mean ± σ):     23.823 s ±  0.044 s    [User: 1.1 ms, System: 8.2 ms]

  Range (min … max):   23.777 s … 23.895 s

Summary

  '.\bat.orig\target\release\bat.exe -p --color never --paging never E:\Media\enwik9' ran
    1.05x faster than '.\bat\target\release\bat.exe -p --color never --paging never E:\Media\enwik9'
```

So it looks like locking `Stdout` somehow makes it _slower_. I’m glad you suggested the benchmarks. Either I’ve got those wrong or else this is a clear case of premature optimization on my part!

sharkdp:
Thanks!

This is what I get on Linux :smile: 

```
▶ hyperfine --warmup 3 "./bat -p --color never --paging never /home/shark/Downloads/enwik9" "./bat-locked-stdout -p --color never --paging never /home/shark/Downloads/enwik9"
Benchmark #1: ./bat -p --color never --paging never /home/shark/Downloads/enwik9
  Time (mean ± σ):      5.345 s ±  0.027 s    [User: 2.506 s, System: 2.833 s]
  Range (min … max):    5.299 s …  5.386 s
 
Benchmark #2: ./bat-locked-stdout -p --color never --paging never /home/shark/Downloads/enwik9
  Time (mean ± σ):      4.997 s ±  0.022 s    [User: 2.206 s, System: 2.786 s]
  Range (min … max):    4.963 s …  5.032 s
 
Summary
  './bat-locked-stdout -p --color never --paging never /home/shark/Downloads/enwik9' ran
    1.07 ± 0.01 times faster than './bat -p --color never --paging never /home/shark/Downloads/enwik9'
```

Aankhen:
Oh, so it’s just Windows then… or possibly my PC. Do you think it’s worth integrating for that small boost on Linux?

sharkdp:
To be honest, I think I'd rather leave it out for now because I don't really like that we have to `match` in `OutputType` twice (and in this part of the code).

PR #326, opened 2018-09-30
Pipfile as an INI

tiktuk:
`Pipfile` added as INI-type.


sharkdp:
I don't really know `Pipfile`s, but it seems to be TOML instead of INI.

tiktuk:
Sorry, my bad, you're right. I've added it to the TOML bundle instead [here](https://github.com/sharkdp/sublime_toml_highlighting/pull/2). [Pipfiles](https://github.com/pypa/pipfile) are for managing Python packages btw.

Issue #115, opened 2018-05-14
Auto-generate shell-completion files via clap


sharkdp:
Completion files for bash, zsh, fish and PowerShell can be auto-generated via `clap` (see related changes in `fd`: https://github.com/sharkdp/fd/pull/66/files, https://github.com/sharkdp/fd/blob/master/build.rs)

davideGiovannini:
Hi @sharkdp, if this issue is still open I would like to work on it

sharkdp:
@davideGiovannini sounds great!

sharkdp:
Closed via #327

PR #327, opened 2018-10-01
Generate shell completions with clap during build

davideGiovannini:
Auto-generate shell-completion files via clap #115


Cargo.toml
```diff
@@ -43,3 +44,6 @@ features = []
 
 [dev-dependencies]
 tempdir = "0.3"
+
+[build-dependencies]
+clap = "2.32.0"
```
sharkdp: I think we should use  `"2.32"` to be consistent with the dependency above.

build.rs
```diff
@@ -0,0 +1,21 @@
...
+
+use clap::Shell;
+
+include!("src/clap_app.rs");
+
+const BIN_NAME: &str = "bat";
+
+fn main() {
+
+    let outdir = match std::env::var_os("OUT_DIR") {
```
sharkdp: Can we use the same mechanism as in `fd` here:
```rust
let outdir = std::env::var_os("SHELL_COMPLETIONS_DIR").or(std::env::var_os("OUT_DIR"));
```

See https://github.com/sharkdp/fd/issues/121


It would also require us to use `fs::create_dir_all`:
https://github.com/sharkdp/fd/commit/3104729f29f4c1bb3a49b7adc2fa79f355115fd8

ci/before_deploy.bash
```diff
@@ -38,6 +38,11 @@ pack() {
     cp LICENSE-MIT "$tempdir/$package_name"
     cp LICENSE-APACHE "$tempdir/$package_name"
 
+    # various autocomplete
+    cp target/"$TARGET"/release/build/"$PROJECT_NAME"-*/out/"$PROJECT_NAME".bash "$tempdir/$package_name/autocomplete/${PROJECT_NAME}.bash-completion"
```
sharkdp: We need to make sure that the target directory exists.

sharkdp: Also, it would be great to also do this for Windows deployments (in the `before_deploy.ps1` script). I think we don't need a subfolder in case of the Windows archive.

davideGiovannini: I'm not familiar with how it should work on Windows. 

Should it be something like this ?
```ps1
Copy-Item "$SRC_DIR\target\release\build\bat-*/out/_bat.ps1" '.\autocomplete'
```

davideGiovannini:
>You can cargo +nightly install rustfmt-nightly and then run cargo fmt from the repository folder. 

is this the same as running `cargo fmt` installed with `rustup component add rustfmt-preview` ?
It's changing the formatting of files that I did not change 
```
        modified:   src/assets.rs
        modified:   src/diff.rs
        modified:   src/printer.rs
```
> Do wildcards work inside quotes?

Uhm probably not.
I'm removing the quotes

sharkdp:
Hm, I'm not sure, but there might very well be a change in the format between this and rustfmt-nightly (which we currently use to enforce the style guide). This is unfortunate, but apparently there will be a 1.0 release soon which should stabilize the format.

davideGiovannini:
>You can cargo +nightly install rustfmt-nightly and then run cargo fmt from the repository folder. 

is this the same as running `cargo fmt` installed with `rustup component add rustfmt-preview` ?
It's changing the formatting of files that I did not change 
```
        modified:   src/assets.rs
        modified:   src/diff.rs
        modified:   src/printer.rs
```
> Do wildcards work inside quotes?

Uhm probably not.
I'm removing the quotes

sharkdp:
Thank you very much!!

PR #336, opened 2018-10-07
Inputfile wrapper

### LINKED ISSUE #205: Automatic syntax detection/guessing without using filename
Pygments has[ this feature](http://pygments.org/docs/quickstart/#guessing-lexers) where you can `cat <file> | pygmentize` or `pygmentize <file-without-extension>` and the syntax will be guessed according to some patterns in the file. 

It would be great to avoid relying on file extensions only (which are not always present) or integrating with other commands that can output the content and pipe it directly to `bat`.

### LINKED ISSUE #248: Handling of binary files in interactive mode
When `bat` is operating in non-interactive (`cat`-compatibility) mode, binary files are handled fine (see #150).

However, when `bat` is in interactive mode (prints to a terminal), we currently just print all the content in a (lossy) UTF-8 decoded way, which is not helpful at all.

It would be great if `bat` could detect "binary" files (I recently wrote a small crate to do just that: [https://github.com/sharkdp/content_inspector](`content_inspector`)) and print a user-friendly error message / warning.

A potential nice add-on could be to print the first N bytes in `hexdump` style.

sharkdp:
* Add first-line detection for all input types, including pipes.

  So this just works:
  ``` bash
  curl https://sh.rustup.rs | bat
  ```

  This closes #205 
* Do not display binary files in interactive mode.
  
  When attempting to display binary files, we just show this:
  ```
  ▶ bat test.jpg 
  ───────┬─────────────────────────────────────────────────
         │ File: test.jpg   <BINARY>
  ───────┴─────────────────────────────────────────────────
  ```

  This closes #248


zeenix:
Has this regressed in the last (0.10) release? I don't get syntax hilighting for XML output of `virsh dumpxml 7njz4pq0m0|bat`. The first line is `<domain type='kvm'>`.

Issue #310, opened 2018-09-13
env var BAT_OPTS


VladimirAlexiev:
BAT_THEME allows you to set a theme.
I also want to set `--wrap never` permanently, but there's no respective env var.

Why not do like "less", where the `LESS` env var can take any of the command-line options?
I'd suggest to use the name `BAT_OPTS` for this, since in the mind of Windows users `BAT` means "batch files".

(Note: I've added this setting to your `bat()` function as per #263)

eth-p:
@VladimirAlexiev I agree, something like this would be useful. Some options such as `--color` and `--wrap` don't currently have environment variables, and future options would need their own respective variables as well.

That being said, however, I think that it might be getting a bit excessive on environment variables. We already have: 

* `BAT_STYLE`
* `BAT_THEME`
* `BAT_TABS`
* `BAT_PAGER`

Maybe there's a better way to support this kind of functionality? Some sort of configuration file (`toml`, maybe?) that covers all of these different options, and any future options we add. @sharkdp What do you think?

sharkdp:
I agree that the number of environment variables is getting out of hand and I'd rather replace it with one of the suggested options (`BAT_OPTS` or configuration file).

The `BAT_OPTS` variable is kind of appealing to me since it is really easy to implement (I believe, if `clap` supports this somehow :smile:) and very straightforward to use. There is no need for additional documentation of a config format and *N* other environment variables - everything is just configured by command-line options, but they may also be placed in `BAT_OPTS`.

A dedicated configuration file would have the advantage that some things might be easier to configure (for the user) and potentially easier to parse/deserialize (for the developer). I'm thinking about features such as #169 where it might be more convenient to have a dictionary-type structure in a configuration file as opposed to a command line flag which has to be specified multiple times (`--map-extension myini ini --max-extension my-ruby-file rb`).

Note that there is a hybrid variant which [ripgrep is following](https://github.com/BurntSushi/ripgrep/blob/master/GUIDE.md#configuration-file) where command line flags are stored in the configuration file.

llimllib:
(To add some detail to @VladimirAlexiev's comment, in case you got here looking for how to tell `bat` to never wrap, here is one way to do so. Add this alias to your `~/.bashrc` until there is some mechanism to tell `bat` not to page:

```bash
# Add the `--wrap never` arg to all `bat` invocations
bat() {
    command bat --wrap never "${args[@]}"
}
```
)

sharkdp:
[...]

PR #338, opened 2018-10-07
Add simple configuration file

sharkdp:
This allows users to create simple configuration file (`~/.config/bat/config` on Linux) that has the following format:

```bash
--flag1
--flag2
--option1=value1
# lines beginning with '#' are ignored
--option2 value2
# empty lines and trailing whitespace are also ignored

--option3=value3
# Flags and options can also be on a single ine:
--flag4 --flag5 --option4 value4
```

It also adds support for `BAT_OPTS` environment variable with the same format. It takes precedence over the configuration file.

see #310


sharkdp:
There are a few issues with this (apart from the failing tests :smile:)

To add `--theme="Sublime Snazzy"` to the configuration file, users would have to put either
```
--theme
Sublime Snazzy
```

or
```
--theme=Sublime Snazzy
```

into their configuration file. Both are confusing and the more sane alternatives do not work (`--theme="Sublime Snazzy"`) or (--theme "Sublime Snazzy").

sharkdp:
The issue mentioned above is now resolved.

There is one thing missing: `--style` can not be passed in twice, so users will not be able to configure `--style` in their config file and overwrite it on the command-line later.

SimenB:
Would it be possible to set an env variable pointing to the configuration file, similar to `ripgrep`'s `$RIPGREP_CONFIG_PATH`?

sharkdp:
@SimenB Can you please open a new ticket to address this? Thank you

PR #340, opened 2018-10-08
misc: add exec bit for create.sh

sanga:
Anyone who wants to add a syntax to bat will need to do this anyway.


sharkdp:
Thank you

PR #341, opened 2018-10-08
Update dependencies

fornwall:
The driver for this is to to get version 0.3.1 of termios, which supports Android. This means that bat can be built and used out of the box on [Termux](https://termux.com/).


sharkdp:
Awesome, thank you!

sharkdp:
I hadn't heard of Termux. I just tried it out - this is awesome :smile: 

There's even a package for `fd`! :heart:

PR #343, opened 2018-10-09
syntax: Add robot framework support

sanga:
i.e. http://robotframework.org/


sanga:
64bit musl linux build failed, which is curious - is that at all known/transient or might that have been introduced in this PR? The other failure appeared to be apt servers timing out (installing packages in prep for testing).

sharkdp:

I think that are just random failures due to some timeouts. I'll restart the jobs.

PR #344, opened 2018-10-09
Documented nix availability

mgttlinger:
`bat` is available in through the nix package manager. Added that install option to the README.


sharkdp:
Awesome, thank you!

PR #346, opened 2018-10-10
[WIP] add line highlight

tskinn:
There might be a better place to get the background highlight color. Not sure.

### DIFF
--- a/src/clap_app.rs
+++ b/src/clap_app.rs
@@ -146,10 +146,10 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
                 .long("highlight-line")
                 .overrides_with("highlight-line")
                 .takes_value(true)
-                .value_name("n")
+                .value_name("N")
                 .help("Highlight a line.")
                 .long_help(
-                    "Highlight the nth line. The background color is changed to create contrast.",
+                    "Highlight the Nth line. The background color is changed to create contrast.",
                 ),
         )
         .arg(
--- a/src/clap_app.rs
+++ b/src/clap_app.rs
@@ -142,6 +142,17 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
                 ),
         )
         .arg(
+            Arg::with_name("highlight-line")
+                .long("highlight-line")
+                .overrides_with("highlight-line")
+                .takes_value(true)
+                .value_name("n")
+                .help("Highlight a line.")
+                .long_help(
+                    "Highlight the nth line. The background color is changed to create contrast.",
+                ),
+        )
+        .arg(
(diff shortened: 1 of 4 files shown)


src/printer.rs
```diff
@@ -296,6 +301,18 @@ impl<'a> Printer for InteractivePrinter<'a> {
             }
         }
 
+        // Line highlighting
+        let background = match self.config.highlight_line {
```
sharkdp: This whole block could be something like:
``` rust
let background = self.config.highlight_line
    .filter(|line| line == line_number)
    .map(|_| self.background_highlight);
```

Granted, I'm not sure if that is easier to read. Soon, there will be `Option::replace` which could be used instead of `.map(|_| ..)`.

src/clap_app.rs
```diff
@@ -141,6 +141,17 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
                      '--line-range 40:' prints lines 40 to the end of the file",
                 ),
         )
+        .arg(
+            Arg::with_name("highlight-line")
+                .long("highlight-line")
+                .overrides_with("highlight-line")
+                .takes_value(true)
+                .value_name("n")
```
sharkdp: Can we use a captital `N` here to be consistent with the `--line-range` option?

src/clap_app.rs
```diff
@@ -141,6 +141,17 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
...
+                .help("Highlight a line.")
+                .long_help(
+                    "Highlight the nth line. The background color is changed to create contrast.",
```
sharkdp: this would be "the N-th line" then.

src/printer.rs
```diff
@@ -296,6 +301,12 @@ impl<'a> Printer for InteractivePrinter<'a> {
             }
         }
 
+        // Line highlighting
+        let background = self.config.highlight_line
+            .filter(|line| *line == line_number)
+            .map(|_| self.background_highlight)
+            .unwrap();
```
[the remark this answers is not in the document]
sharkdp: You are right. In this case, it should be `and_then` instead of `map`.

sharkdp:
> I agree! Right when I started testing the changes I thought this kind of sucks if it doesn't change more that just the background of the text.

If you have any ideas, feel free to experiment here.

sharkdp:
@tskinn Any updates on this? How do you think we should proceed here?

tskinn:
Sorry, no updates! I haven't looked very deeply into it yet. I've bin a little strapped for time lately but I will try to find some time this week.

sharkdp:
closing this in favor of #453

PR #350, opened 2018-10-11
Fixes #328 - adds info for light backgrounds.

### LINKED ISSUE #328: Default theme broken on default config of Terminal.app (OS X)
Not sure if this is something that you're planning to fix but bat is broken by default on OS X. Not sure what the best fix would be, but my first thought is for the default theme (not sure which one it is since `bat --list-themes` does not tell you the default) should work reasonably with both a black and a white background (although maybe that's too difficult). As far as I can tell the default theme emits some text that is completely white, which makes the default config of bat unusable on a white background.

russtaylor:
Should fix #328 - describes steps for using a theme compatible with light terminal backgrounds.


russtaylor:
I removed the duplicate parts. Thanks for such an awesome tool!

PR #353, opened 2018-10-11
bat- experiment

Andr0meda777:
Not compatible with a number of operating systems


sharkdp:
What is this supposed to achieve!?

Issue #352, opened 2018-10-11
Add a way to pass arguments to the pager via PAGER / BAT_PAGER


sharkdp:
We can use a function to split a (simple) shell command-line into multiple arguments. This would allow use to use just
```
BAT_PAGER="less --tabs 4 -RF"
```
instead of using a wrapper script.

This could be useful: https://crates.io/crates/shell-words

Foxboron:
Taking a stab at this one :)

sharkdp:
Released in [v0.8.0](https://github.com/sharkdp/bat/releases/tag/v0.8.0).

PR #354, opened 2018-10-12
Add arguments from PAGER/BAT_PAGER

Foxboron:
I'm a bit unsure of Rust error handling. Probably also missing a test, but lets just ack the code change first maybe.

Solves #352

Signed-off-by: Morten Linderud <morten@linderud.pw>

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -1,10 +1,10 @@
-extern crate shell_words;
-
 use std::env;
 use std::ffi::OsString;
 use std::io::{self, Write};
 use std::path::PathBuf;
 use std::process::{Child, Command, Stdio};
+
+use shell_words;
 
 use app::PagingMode;
 use errors::*;
--- a/src/output.rs
+++ b/src/output.rs
@@ -33,29 +33,34 @@ impl OutputType {
         let pagerflags = shell_words::split(&pager)
             .chain_err(|| "Could not parse (BAT_)PAGER environment variable.")?;
 
-        let less_path = PathBuf::from(&pagerflags[0]);
-        let is_less = less_path.file_stem() == Some(&OsString::from("less"));
+        match pagerflags.split_first() {
+            Some((pager_name, args)) => {
+                let pager_path = PathBuf::from(pager_name);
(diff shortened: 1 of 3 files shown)


Foxboron:
Fixed up on the `is_less` logic and how it append args. Verified that this gives the desired behavior as described above. Also did a quick swap so the code calling the binary depends on `pagerflags`. Thus we don't pass `/usr/bin/less --flag` as the executable when dealing with flags.

src/output.rs
```diff
@@ -28,23 +30,27 @@ impl OutputType {
             .or_else(|_| env::var("PAGER"))
             .unwrap_or(String::from("less"));
 
-        let less_path = PathBuf::from(&pager);
+        let pagerflags = shell_words::split(&pager).unwrap_or(vec![pager]);
```
sharkdp: `split` will fail if there is a parse error, for example:
```
split("less --option='foo")   # closing quote is missing
```

I don't think that `unwrap_or` would help in such a case?

src/output.rs
```diff
@@ -28,23 +30,27 @@ impl OutputType {
...
+
+        let less_path = PathBuf::from(&pagerflags[0]);
```
sharkdp: What happens if we pass `BAT_PAGER=""`? Does this panic?

Previously, setting `BAT_PAGER=""` would disable the pager. I'd like to keep this behavior.

src/output.rs
```diff
@@ -28,23 +30,27 @@ impl OutputType {
...
+            }
+            p.env("LESSCHARSET", "UTF-8");
             p
         } else {
-            Command::new(pager)
+            Command::new(&less_path)
         };
 
         process
+            .args(&pagerflags[1..])
```
sharkdp: Same here. We should make sure that this does not panic.

sharkdp: Fixed.

src/output.rs
sharkdp: Fixed.

sharkdp:
Thank you for your work!

Issue #351, opened 2018-10-11
Rewrite `--tabs` tests as integration tests


sharkdp:
It would be great to remove all tab-specific things from `tests/tester.rs` and `tests/snapshot_tests.rs` and to re-write these tests as "proper" integration tests in `tests/integration_tests.rs`. The latter are much easier to use and their output is easier to parse in case of failures.

choznerol:
If there's no work in progress, I would like to take this issue!

sharkdp:
As a first step, I would just like to move the `tabs_` snapshot tests to the integration tests file. We can think about migrating the others later on.

sharkdp:
Also, the tab-specific tests should be written in the style of the other integration tests, without the snapshot_test macro.

Thank you for your help!

choznerol:
ok! I opened #356 and we can discuss further steps there :)

PR #356, opened 2018-10-14
Move tabs_* snapshot tests to integration_tests

choznerol:
close https://github.com/sharkdp/bat/issues/351


tests/integration_tests.rs
```diff
@@ -1,14 +1,21 @@
...
 use std::process::Command;
+use tester::BatTester;
 
 fn bat() -> Command {
     let mut cmd = Command::main_binary().unwrap();
     cmd.current_dir("tests/examples");
     cmd
 }
 
+fn tester() -> BatTester {
```
choznerol: This seems a little confusing besides `fn bat()`. Perhaps should I rename it to `fn snapshot_tester()`?

sharkdp:

Sorry if my explanation in #351 was too terse / not clear. My intention was to completely rewrite the tab-related unit tests in the style of the other tests in `integration_tests.rs`. This would involve the creation of a new file in `tests/examples` with a few tab characters (preferably, this file should be rather small and not include as much text as the snapshot test-file). The tests would then use the method of the existing tests in `integration_tests.rs`. For example:
```rust
fn tabs_4() {
    bat()
        .arg("tabs.txt")
        .arg("--tabs=4")
        .assert()
        .success()
        .stdout("...");
}
```

choznerol:
Sure! I was thinking in the wrong direction 😅

1. The arguments `paging`, `color` and `terminal-width` doesn't seem to have an effect on the test result, so I didn't include them here.
2. I kept line 23-35 as the snapshot and didn't [use `\n`](https://github.com/sharkdp/bat/pull/356/files#diff-4baea58d24b03f39f16aeb8f064939abR100). hopefully it makes it more expressive but not verbose :o
3. I realized [the `wrapped` version](https://github.com/sharkdp/bat/blob/master/tests/snapshots/output/tabs_4_wrapped.snapshot.txt) and the [`wrap=naver` version](https://github.com/sharkdp/bat/blob/master/tests/snapshots/output/tabs_4.snapshot.txt) are expected to be identical in the old test. Should we just pick one and test the `--wrap` option separately?

sharkdp:
> * The arguments `paging`, `color` and `terminal-width` doesn't seem to have an effect on the test result, so I didn't include them here. (perhaps I missed something?)

That should be fine.



> * I kept line 23-35 as the snapshot and didn't [use `\n` for newline](https://github.com/sharkdp/bat/blob/e87ebf307695662ef84c6f270d83c5e93464b526/tests/integration_tests.rs#L52). hopefully it makes it more expressive but not verbose :o

Looks good.

> 3\. I realized all [the `wrapped` version](https://github.com/sharkdp/bat/blob/master/tests/snapshots/output/tabs_4_wrapped.snapshot.txt) and the [`wrap=naver` version](https://github.com/sharkdp/bat/blob/master/tests/snapshots/output/tabs_4.snapshot.txt) were expected to be identical in the old test. Should we just pick one of them, and test the `--wrap` option separately?

Ideally, we should also test the combination of `--wrap` and `--tabs`, but we can also do this in a different PR. We would need a long line with tabs for this (or set a very small `--terminal-width`).

tests/integration_tests.rs
```diff
@@ -93,6 +93,144 @@ fn line_range_last_3() {
         .stdout("line 2\nline 3\nline 4\n");
 }
 
+#[test]
+fn tabs_passthrough_wrapped() {
+    bat()
+        .arg("tabs.txt")
+        .arg("--tabs=0")
+        .arg("--wrap=character")
```
sharkdp: I think we can remove `--wrap`?

tests/integration_tests.rs
```diff
@@ -93,6 +93,144 @@ fn line_range_last_3() {
...
+        .arg("--style=plain")
```
sharkdp: I think the current tests also test `--tabs` in combination with other styles, right? I think we should at least test one style like `--style=numbers` here, because it will affect the tab-stops.

tests/integration_tests.rs
```diff
@@ -93,6 +93,144 @@ fn line_range_last_3() {
...
+88888888	?
+");
+}
+
+#[test]
+fn tabs_4_wrapped() {
+    bat()
+        .arg("tabs.txt")
+        .arg("--tabs=4")
+        .arg("--wrap=character")
```
sharkdp: Same here (and for the other tests below)

tests/integration_tests.rs
choznerol: ok! fixed in [f0f0672](https://github.com/sharkdp/bat/pull/356/commits/f0f0672ea922985d14e23e1bb94d892e6213107c)

tests/integration_tests.rs
choznerol: I realized a trailing space would be needed to make [Test 'tabs' with 'numbers'](https://github.com/sharkdp/bat/pull/356/commits/9a42fcb74d1604dddd43466ecd52b6118192d61c) pass. So I just [removed the empty line in snapshot](https://github.com/sharkdp/bat/pull/356/commits/b7620066e107c44849a0d7c8e8544dc69e3e35b6)

sharkdp:
Thank you for the updates!

Can you please run `cargo fmt` to fix the CI error?

```
cargo +nightly install rustfmt-nightly
cargo fmt
```

Issue #23, opened 2018-04-30
Filter to only include hunks?


ErichDonGubler:
First: congrats on the release, and props for creatively putting together the pieces that make this work!

Anyway: It'd be really neat if this could be used as an alternative to the default `pager` in Git -- this is REALLY nice-looking, and I'd love to use this as a pager. Would you be open to a PR or implementing yourself a feature to filter out everything that's not a hunk in a diff? I'm not sure if this should be a special way to process `diff`/`patch` files or just integrated with `libgit2` directly -- the latter is probably easier, since you're already using `libgit2` as a dependency.

sharkdp:
Thank you for the feedback!

Nice idea! I actually was on a similar track here: #12 (see `--context` option).

> Would you be open to a PR 

Definitely! However, maybe it would be worth to first discuss some details here in this ticket?

For example, how would the command line interface (subcommand, option/flag) for that feature look like?

> I'm not sure if this should be a special way to process diff/patch files or just integrated with libgit2 directly -- the latter is probably easier, since you're already using libgit2 as a dependency.

Yes, libgit2 sounds like the better option to me.

ErichDonGubler:
I think that at this point a single flag or subcommand would be appropriate -- the assumption with a `libgit2`-based diff is that you're operating inside of a Git repository. 

## Requirements I'm confident should be part of this

This flag/subcommand (hereafter referred to as `diff`, which is definitely subject to bikeshedding):
* **will** fail the entire commend if not inside of a Git repo
* **should** take a reasonable subset of options that map to common `git diff` invocations; some example invocations might be:
    * `bat diff ` would display all unstaged changes, minus untracked files (a la `git diff`)
    * `bat diff cached ` would display all staged changes (a la `git diff --cached`)
        * Alternative invocations for this use case could look like `bat --diff --cached`

## Unresolved questions

* **Should this be a subcommand or a flag?**
    * Flags would be unambiguous, but a subcommand might look more ergonomic (`git`-esque, in this case).
    * If a subcommand is desired, then maybe the following design would work:
        * `bat print` or some similar subcommand would be the default when only file paths are provided. For instance, given the following file structure: `foo`:
            ```
            $ tree
            .
            ├── diff
            ├── foo.txt
            └── print
            ```
                ...then:
            * `bat foo.txt` would be equivalent to today's functionality, inferred to be `bat [print] foo.txt`.
[...]

PR #368, opened 2018-10-19
Allow for multiple line ranges

sharkdp:
See #23


ErichDonGubler:
Total bikeshed you can ignore: when you first said "multiple line ranges", I originally read it as "*more than one* line range" -- but I know better, since I've been involved with your development of this idea! Maybe "multi-line ranges" is less ambiguous?

Issue #375, opened 2018-10-22
Allow setting environment variable with path to config file


SimenB:
#338 added support for a config file. On mac (at least my machine), it's saved to `$HOME/Library/Preferences/bat/config`. I'd love for there to be an environment variable named e.g. `$BAT_CONFIG_PATH` that allowed me to define a custom path. My use case is that I like to keep config in `$HOME/.config` as I check it into VCS.

This would also match e.g. ripgrep: https://github.com/BurntSushi/ripgrep/blob/fb622666206089cf7092ae116612898ac358c91c/GUIDE.md#configuration-file (although they _only_ support a path as env variable, no defaults)

sharkdp:
Thank you for your feedback. I think it is a good idea to support this.

Another question is if we want to set the default path on Mac OS to `$HOME/.config`. This has been discussed before: #151.

deg4uss3r:
I'd love to help with this is it already being worked?

sharkdp:
@deg4uss3r I don't think so.

deg4uss3r:
Great! @sharkdp do you think just adding in logic to `config.rs` in the `config_file()` function is approperate or do you want to work more in `dirs.rs` for that?

sharkdp:
Closed via #380 by @deg4uss3r.

sharkdp:
This is now available in [bat 0.9](https://github.com/sharkdp/bat/releases/tag/v0.9.0).

kbd:
> Another question is if we want to set the default path on Mac OS to `$HOME/.config`.

I hope this isn't rude to comment on a closed issue, but since the question was asked recently I figured it'd be ok to offer my 2¢ in response. I just updated Homebrew, which pulled down bat 0.9, so I checked the release notes and found that italics went away. I read the background issues to learn the rationale and then went to set up a bat config file, and was surprised to learn that the config file is in `~/Library/...`

I understand the rationale and that the behavior is inherited from an underlying library. FWIW there is no other command line tool I use that stores its config there. As a programmer I expect any command line tool to use the XDG standard location (or less-preferred, `~/.foo`) for its config. I know I'm not alone in source-controlling my dotfiles, and I don't check in anything in `~/Library/...`. So now I need to set a `BAT_CONFIG_PATH`, and at that point I should just `alias bat=bat --italic-text=always` instead.

sharkdp:
I think we should change the default path on Mac OS, see #419

Issue #377, opened 2018-10-24
Add documentation for new configuration file


sharkdp:
See: https://github.com/sharkdp/bat/issues/259#issuecomment-431701670

sharkdp:
Closed via #402 by @deg4uss3r

PR #380, opened 2018-10-25
[WIP] Enviroment var config

deg4uss3r:
Work in progress PR for #375 

Happy to change as you see fit, tested manually with a --style=numbers line, looking for a good way to write an integration test for the config `$VAR`

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -9,7 +9,18 @@ use dirs::PROJECT_DIRS;
 use util::transpose;
 
 pub fn config_file() -> PathBuf {
-    PROJECT_DIRS.config_dir().join("config")
+    match env::var("BAT_CONFIG_PATH") {
+        Ok(env_path) => {
+            let env_path_buf = PathBuf::from(env_path);
+            if env_path_buf.is_file() {
+                return env_path_buf;
+            }
+            else {
+               return PROJECT_DIRS.config_dir().join("config"); 
+            }
+        }
+        Err(_) => PROJECT_DIRS.config_dir().join("config"),
+    }
 }
 
 pub fn get_args_from_config_file() -> Result<Vec<OsString>, shell_words::ParseError> {
--- a/src/config.rs
+++ b/src/config.rs
@@ -9,17 +9,11 @@ use dirs::PROJECT_DIRS;
 use util::transpose;
 
 pub fn config_file() -> PathBuf {
-    match env::var("BAT_CONFIG_PATH") {
-        Ok(env_path) => {
-            let env_path_buf = PathBuf::from(env_path);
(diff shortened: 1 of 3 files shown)


src/config.rs
```diff
@@ -9,7 +9,17 @@ use dirs::PROJECT_DIRS;
 use util::transpose;
 
 pub fn config_file() -> PathBuf {
-    PROJECT_DIRS.config_dir().join("config")
+    match env::var("BAT_CONFIG_PATH") {
```
sharkdp: An alternative way to write this logic would be with the combinators on `Option`:
```rust
env::var("BAT_CONFIG_PATH")
    .ok()
    .map(PathBuf::from)
    .filter(Path::is_file)
    .unwrap_or(PROJECT_DIRS.config_dir().join("config"))
```

sharkdp:
Looks good, thank you very much for your contribution!

sharkdp:
If you want to write an integration test for this (which would be great), you could take a look at the `pager_basic` test in `integration_tests.rs` for an example on how to set an environment variable.

You would probably have to add an exemplary config file somewhere in the `tests` folder and set the `BAT_CONFIG_PATH` variable to that file.

src/config.rs
deg4uss3r: FYI I am working on this looks like below, but seems to be causing some weird issues with my build. I'll keep at it :) 

```rust
env::var("BAT_CONFIG_PATH")
        .ok()
        .map(PathBuf::from)
        .filter(|config_path| config_path.is_file())
        .unwrap_or(PROJECT_DIRS.config_dir().join("config"))
```

sharkdp: Okay, let me know if you need help.

deg4uss3r: Added two integration tests, one to check if it's reading the environment variable path right, and another that enables paging and checks for the paging output like `pager_basic()`

sharkdp:
Thank you very much for the updates. I made some small modifications on top of yours.

deg4uss3r:
Cool, these make sense! Is there anything else you would like me to add? I can type up an addition to the README if you’d like.

sharkdp:
> I can type up an addition to the README if you’d like.

Oh, that would be great! Unfortunately, there is no section about the config file, yet (see #377).

deg4uss3r:
> see #377

Sure! I will open a new PR for that. First commit will be very basic to discuss how it should be formatted and getting all the options included. Is that okay?

sharkdp:
> Sure! I will open a new PR for that. First commit will be very basic to discuss how it should be formatted and getting all the options included. Is that okay?

Sounds great.

Issue #389, opened 2018-10-31
Disable use of italics (by default)


sharkdp:
Disable the use of ANSI-italics by default, but add a new option `--enable-italics` that users can put in their config file, if they want to opt-in.

Rationale:
- A lot of terminals struggle with italics and output weird colors instead of using italic text (see #130, for example).
- Some pagers do not support italics: #384

Ashniu123:
I'll take this.

chaucerbao:
It looks like #96 was the PR that added font styles.

sharkdp:
Closed via #390 by @Ashniu123.

sharkdp:
This is now available in [bat 0.9](https://github.com/sharkdp/bat/releases/tag/v0.9.0).

PR #390, opened 2018-10-31
WIP: Disable use of italics (by default)

Ashniu123:
This is a work in progress for [#389](https://github.com/sharkdp/bat/issues/389).

It is not complete right now but awaiting confirmation that progress is being made in the right direction.

Thank you.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -74,7 +74,7 @@ pub struct Config<'a> {
     /// Command to start the pager
     pub pager: Option<&'a str>,
 
-    pub italics: bool,
+    pub use_italic_text: bool,
 }
 
 fn is_truecolor_terminal() -> bool {
@@ -234,7 +234,7 @@ impl App {
             output_components,
             syntax_mapping,
             pager: self.matches.value_of("pager"),
-            italics: match self.matches.value_of("italic-text") {
+            use_italic_text: match self.matches.value_of("italic-text") {
                 Some("always") => true,
                 _ => false,
             },
--- a/src/app.rs
+++ b/src/app.rs
@@ -73,6 +73,8 @@ pub struct Config<'a> {
 
     /// Command to start the pager
     pub pager: Option<&'a str>,
+
+    pub italics: bool
 }
 
 fn is_truecolor_terminal() -> bool {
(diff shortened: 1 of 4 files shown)


src/app.rs
```diff
@@ -232,6 +234,10 @@ impl App {
             output_components,
             syntax_mapping,
             pager: self.matches.value_of("pager"),
+            italics: match self.matches.value_of("enable-italics") {
```
sharkdp: I like that you suggest `always` and `never`, but in this case, let's maybe use just `--italics=always/never`?

sharkdp: Or maybe just
```
--enable-italics[=yes]  # to enable it, with or without "=yes"
--enable-italics=no     # to switch it off
```

what do you think?

sharkdp:
Looks good, so far!

src/app.rs
Ashniu123: IMHO, to have consistency between flags we should keep `always` and `never`. Because other flags use the `auto, always and never` mostly.

sharkdp: okay, then let's use `--italic-text=always|never`

src/clap_app.rs
```diff
@@ -304,6 +304,17 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
                 .hidden(true)
                 .help("Show path to the configuration file."),
         )
+        .arg(
```
sharkdp: Can we move the whole `arg`ument up, just below `--color`?

src/clap_app.rs
```diff
@@ -304,6 +304,17 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
...
+            Arg::with_name("italic-text")
+                .long("italic-text")
+                .takes_value(true)
+                .multiple(false)
+                .value_name("always|*never*")
+                .possible_values(&["always", "never"])
+                .default_value("never")
+                .hide_default_value(true)
+                .help("Use italics in output (always, *never*)"),
```
sharkdp: Can we also add a `long_help(..)` here:

"Specify when to use ANSI sequences for italic text in the output. Possible values: always, *never*."

sharkdp:
Thank you!

Issue #383, opened 2018-10-27
bat attempts to execute itself when PAGER="bat"


iravid:
I like having other applications use `bat` as a pager, so I `export PAGER="bat"` in my `.zshrc`. However, `bat` will also use that env var and try to execute itself (recursively, endlessly).

To replicate:
```shell
export PAGER="bat"
# This hangs
ls | bat

# In another shell:
ps -fe | grep bat | wc -l
# 177 on my system

unset PAGER
# Works correctly
ls | bat
```

sharkdp:
Nice find :smile:. Thank you for reporting this.

You can easily work around this by setting `PAGER="bat"`, but using `BAT_PAGER` to override this. For example, you can set `BAT_PAGER="less"`. Starting with the next version of `bat` (or current `master`), you can also set `--pager="less"` in `bat`s config file.

I think we should try to fix this by treating `PAGER="bat"` as if `PAGER` had not been set (and similarly for `BAT_PAGER="bat"` or `--pager="bat"`).

iravid:
Cool. I’ll be happy to try fixing this later this week if no one beats me to it :-)

sharkdp:
Closed via #393 by @rodorgas and 7cac07d84559dc6eb02b643f7f640d81a7a4c07e

PR #393, opened 2018-10-31
Avoid endless bat executions with PAGER="bat"

rodorgas:
This PR fix issue #383.

I had to make `pager` a mutable variable, please let me know if there's a better way.

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -37,11 +37,15 @@ impl OutputType {
             .chain_err(|| "Could not parse (BAT_)PAGER environment variable.")?;
 
         match pagerflags.split_first() {
-            Some((initial_pager, args)) => {
-                let pager_name = match initial_pager.as_ref() {
-                    "bat" => "less",
-                    _ => initial_pager,
-                };
+            Some((initial_pager, mut args)) => {
+                let pager_name;
+                if initial_pager == "bat" {
+                    pager_name = "less";
+                    args = &[];
+                }
+                else {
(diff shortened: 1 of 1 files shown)


src/output.rs
```diff
@@ -27,11 +27,16 @@ impl OutputType {
...
     fn try_pager(quit_if_one_screen: bool, pager_from_config: Option<&str>) -> Result<Self> {
         let pager_from_env = env::var("BAT_PAGER").or_else(|_| env::var("PAGER"));
-        let pager = pager_from_config
+
+        let mut pager = pager_from_config
             .map(|p| p.to_string())
             .or(pager_from_env.ok())
             .unwrap_or(String::from("less"));
 
+        if pager == "bat" {
```
sharkdp: I think a better place to check for this would be a few lines below. See how we check for `less` in the `let is_less = ..` line. This also has the benefit of detecting things like `/usr/bin/bat`.

sharkdp:
Thank you for your contribution. I have added an inline comment.

src/output.rs
rodorgas: Done!

rodorgas:
Thanks for the feedback, I've made the changes.

rodorgas:
Additionally, I removed the arguments when the pager is bat, since those arguments won't make sense when we change the pager to less.

sharkdp:
Thank you very much!

Issue #228, opened 2018-08-29
Implement 1:1 compatibility with command line options that cat supports


rahulsundaram:
http://man7.org/linux/man-pages/man1/cat.1.html

In particular, I frequently use -A often but if bat is supposed to be a clone, it really needs to be a drop-in replacement as well.

sharkdp:
Thank you for your feedback.

> if bat is supposed to be a clone, it really needs to be a drop-in replacement as well

As stated in the [Project goals](https://github.com/sharkdp/bat#project-goals-and-alternatives), `bat` tries to be "a drop-in replacement for (POSIX) `cat`" and we are actually pretty close to fulfilling [the POSIX standard](https://github.com/sharkdp/bat/issues/134).

Note that [the POSIX standard](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/cat.html) does not require any command-line arguments except for `-u` (which `bat` supports in the same way as GNU `cat` - by simply ignoring it).

The other command-line options are additional features of GNU `cat` and I have personally never used any of these command-line flags, so I'm not too eager to implement any of them for `bat`. That being said, if there are valid use-cases and a large support for this, I might consider adding some of these (GNU `cat`) flags.

rahulsundaram:
I saw the note on POSIX.  GNU tools have continued to evolve and gain additional features beyond that and -A happens to be one of them.   I will let others voice support for specific options as they see fit

sharkdp:
So I'm curious: what do you do with `cat -A`?

rahulsundaram:
I work on things like puppet templates, yaml files, encrypted hiera etc all of which are sensitive to tab vs space, line ending markers and so forth and can be painful to debug.  Cat -A on these files is a quick way to check if the output looks right especially on a remote system.

sharkdp:
This is actually quite hard to implement correctly due to the interference of syntax highlighting and character replacement. If we switch off syntax highlighting for `-A`, people can also just use `cat`.

I'm closing this, as I don't see this happening in the near future.

sharkdp:
@rahulsundaram I had a change of mind: You can take a look at #395 if you are interested.

rahulsundaram:
Thank you!  Appreciate the heads up as well

sharkdp:
This is now available in [bat 0.9](https://github.com/sharkdp/bat/releases/tag/v0.9.0).

PR #395, opened 2018-11-01
Feature: Highlight non-printable characters

### LINKED ISSUE #381: WISH: Visualize/highlight different whitespace symbols
It is sometimes useful to be able to see the difference between various whitespace symbols, e.g. space (` `), TAB (`\t`), CR (`\r`), LF (`\n`), BACKSPACE (`\b`), ...  This can be particularly helpful when for instance troubleshooting a tab-delimited file (which e.g. uses spaces instead of TABs).

Visualization of whitespace could be done by using different background colors, e.g. (ignoring themes)

* space - default background
* TAB - red
* BACKSPACE - yellow
* CR - magenta
* LF - blue

sharkdp:
Adds a new `-A`/`--show-all` option (in analogy to GNU Linux `cat`s option) that
highlights non-printable characters like space, tab or newline.

This works in two steps:
- **Preprocessing**: replace space by `•`, replace tab by `├──┤`, replace
newline by `␤`, etc.
- **Highlighting**: Use a newly written Sublime syntax to highlight
these special symbols.

#### without `--show-all`
image

#### with `--show-all`
image

Remarks:
- This feature is not technically a drop-in replacement for GNU `cat`s `--show-all` but it has the same purpose.
- This replaces normal syntax highlighting. You can not use syntax highlighting with `--show-all`.
- Tab-width can be controlled via `--tabs N`
- This closes #381 (see also: #228)

### DIFF
--- a/assets/syntaxes/show-nonprintable.sublime-syntax
+++ b/assets/syntaxes/show-nonprintable.sublime-syntax
@@ -11,6 +11,8 @@ contexts:
       scope: support.function.show-nonprintable.space
     - match: "├─*┤"
       scope: constant.character.escape.show-nonprintable.tab
+    - match: "↹"
+      scope: constant.character.escape.show-nonprintable.tab
     - match: "␤"
       scope: keyword.operator.show-nonprintable.newline
     - match: "␍"
--- a/assets/syntaxes/show-nonprintable.sublime-syntax
+++ b/assets/syntaxes/show-nonprintable.sublime-syntax
@@ -1,0 +1,25 @@
+%YAML 1.2
+---
+# http://www.sublimetext.com/docs/3/syntax.html
+name: Highlight non-printables
+file_extensions:
+    - show-nonprintable
+scope: whitespace
+contexts:
+  main:
+    - match: "•"
+      scope: support.function.show-nonprintable.space
+    - match: "├─*┤"
+      scope: constant.character.escape.show-nonprintable.tab
+    - match: "␤"
+      scope: keyword.operator.show-nonprintable.newline
+    - match: "␍"
+      scope: string.show-nonprintable.carriage-return
+    - match: "␀"
+      scope: entity.other.attribute-name.show-nonprintable.null
+    - match: "␇"
+      scope: entity.other.attribute-name.show-nonprintable.bell
+    - match: "␛"
+      scope: entity.other.attribute-name.show-nonprintable.escape
+    - match: "␈"
+      scope: entity.other.attribute-name.show-nonprintable.backspace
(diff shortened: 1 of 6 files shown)


sharkdp:
@HenrikBengtsson Thank you very much for your detailed feedback!

> * `bat --show-all --tabs 1` is rendered identically to `bat --show-all --tabs 2`, i.e. `├┤` (== `├` followed by `┤`).  Could, say, `⇥` (U+21E5, RIGHTWARDS ARROW TO BAR, rightward tab), or `↹` (U+21B9, LEFTWARDS ARROW TO BAR OVER RIGHTWARDS ARROW TO BAR, tab with shift tab) be used for `--tabs 1`? [[en.wikipedia.org/wiki/Tab_key#Unicode](https://en.wikipedia.org/wiki/Tab_key#Unicode)]

Yes, good idea! I was using `↹` initially before I implemented the `├──┤` style. I will use it in the case where `--tabs=1`.

> * I think LF (`␊`) should be used to render the line-feed symbols `\n` (ASCII 0x0A) rather than NL (`␤`) [...] This way the rendering of \n and \r\n matches LF (Unix) and CRLF (Windows), which is how we typically refer to the different new-line alternatives.

Absolutely. I changed it and it will now output either LF or CRLF at the end of a line - much better!

> * which I think should be reserved for `\025` (ASCII 0x15) [[en.wikipedia.org/wiki/Newline](https://en.wikipedia.org/wiki/Newline)]

Hm... `0x15` seems to be something else: Negative Acknowledgement / NAK (https://en.wikipedia.org/wiki/ASCII). I have always thought of "newline" as a synonym for "line feed", but I might be wrong...

> Although not needed to be implemented right away, but in relation to previous comment, plan ahead for adding support for other non-printable ASCII symbols

More non-printable characters should be very easy to add in the future. One entry in `preprocessor.rs` and another in the `show-nonprintable.sublime-syntax`.

HenrikBengtsson:
>> which I think should be reserved for \025 (ASCII 0x15) [<https://en.wikipedia.org/wiki/Newline>]

> Hm... 0x15 seems to be something else: Negative Acknowledgement / NAK (https://en.wikipedia.org/wiki/ASCII). I have always thought of "newline" as a synonym for "line feed", but I might be wrong...

Yeah, you're correct. I was too quick on "NL" in <https://en.wikipedia.org/wiki/Newline>:

"IBM mainframe systems, including z/OS (OS/390) and i5/OS(OS/400) | EBCDIC | NL | 15 | 21 | \025"

That's _not_ part of the ASCII standard.

HenrikBengtsson:
Just tested the updated PR; looks great to me.

PR #402, opened 2018-11-05
[WIP] README Updates for Config File 

deg4uss3r:
Work in progress to update the `README.md` file to include instructions for the config file. Addresses #377

### DIFF
--- a/README.md
+++ b/README.md
@@ -315,21 +315,21 @@ If scrolling still doesn't work for you, you can try to pass the `-S` option in 
 
 ### Configuration file
 
-`bat` can use a config file for options to cut down on command line flags, that you want to save for each run. 
+`bat` can also be customized with a configuration file. The location of the file is dependent on you operating system. To get the default path for your system, call
+```
+bat --config-file
+```
 
-#### Location 
-
-The default location is OS/install location dependent to get the path for your operating system use `bat --config-file` to print out the location of the current config file path.  
-
-You can use the `$BAT_CONFIG_PATH` environment variable to point `bat` to a non-default location of the configuration file like so: 
+Alternatively, you can use the `BAT_CONFIG_PATH` environment variable to point `bat` to a non-default location of the configuration file:
 ```bash
-export BAT_CONFIG_PATH=$HOME/code/setup/bat.conf
+export BAT_CONFIG_PATH="/path/to/bat.conf"
 ```
 
 #### Format
 
-The config file is formatted just as you would pass arguments to the CLI. Here is an example: 
+The configuration file is a simple list of command line arguments. Use `bat --help` to see a full list of possible options and values. In addition, you can add comments by prepending a line with the `#` character.
 
+Example configuration file:
 ```bash
 # Set the theme to "TwoDark"
 --theme="TwoDark"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -313,6 +313,34 @@ will hopefully enable both quit-if-one-screen and mouse-wheel scrolling.
 
 If scrolling still doesn't work for you, you can try to pass the `-S` option in addition.
 
+### Configuration file
+
+`bat` can use a config file for options to cut down on command line flags, that you want to save for each run. 
+
+#### Location 
+
+The default location is operation system/install location dependent (macOS: `Users/<your_user>/Library/Preferences/bat/config`). To get the path of the current location use `bat --config-file` to print out the location of the current config file path.
```
sharkdp: It feels a bit weird that we only mention the path for `macOS`. I would remove the part in parenthesis and just say "To get the path for your operating system, use ...".

README.md
```diff
@@ -313,6 +313,34 @@ will hopefully enable both quit-if-one-screen and mouse-wheel scrolling.
...
 
+### Configuration file
+
+`bat` can use a config file for options to cut down on command line flags, that you want to save for each run. 
+
+#### Location 
+
+The default location is operation system/install location dependent (macOS: `Users/<your_user>/Library/Preferences/bat/config`). To get the path of the current location use `bat --config-file` to print out the location of the current config file path.  
+
+You can change this with the `$BAT_CONFIG_PATH` enviroment variable.
```
sharkdp: (small typo in environment)

Maybe:

You can use the `$BAT_CONFIG_PATH` environment variable to point `bat` to a non-default location of the configuration file.

README.md
```diff
@@ -313,6 +313,34 @@ will hopefully enable both quit-if-one-screen and mouse-wheel scrolling.
...
+
+The default location is operation system/install location dependent (macOS: `Users/<your_user>/Library/Preferences/bat/config`). To get the path of the current location use `bat --config-file` to print out the location of the current config file path.  
+
+You can change this with the `$BAT_CONFIG_PATH` enviroment variable. 
+
+#### Format
+
+The config file is formatted just as you would pass arguments to the CLI. Here is an example: 
+
+```
```
sharkdp: We could use

    ``` bash

here to get some syntax highlighting for the config file.

README.md
```diff
@@ -313,6 +313,34 @@ will hopefully enable both quit-if-one-screen and mouse-wheel scrolling.
...
+
+# Always enable coloring (will send ANSI codes to pipe commands)
+--color=always
+
+# Set the theme to "TwoDark"
+--theme="TwoDark"
+
+# Do not display line numbers or borders
+--decorations=never
+```
```
sharkdp: As for the contents of this configuration file, I would like to make some changes. I don't think that anyone is going to put `--paging=always`, `--color=always` or `--decorations=never` into their configuration file. I certainly wouldn't advise to do so. Instead, I would suggest something like this:

``` bash
# Set the theme to "TwoDark"
--theme="TwoDark"

# Show line numbers, Git modifications and file header (but no grid)
--style="numbers,changes,header"

# Add mouse scrolling support in less (does not work with older
# versions of "less")
--pager="less -FR"

# Use C++ syntax (instead of C) for .h header files
--map-syntax h:cpp

# Use "gitignore" highlighting for ".ignore" files
--map-syntax .ignore:.gitignore
```

deg4uss3r:
Resolved the changes. 

I think we should definitely refer/link to this new section throughout the README. 

I, personally, prefer to have a config over multiple environmental variables makes tracking changes and quick edits a little easier in my opinion :D

Issue #419, opened 2018-11-16
BAT_CONFIG_DIR not being honored on macOS?


mcmillion:
I'm try to get a custom theme set up and I'd like to use `~/.config/bat` as the location for these files. Here's what I'm seeing:

```
➜ echo $BAT_CONFIG_DIR
/Users/mcmillion/.config/bat

➜ bat cache --init
No themes were found in '/Users/mcmillion/Library/Preferences/bat/themes', using the default set
No syntaxes were found in '/Users/mcmillion/Library/Preferences/bat/syntaxes', using the default set.
Writing theme set to /Users/mcmillion/Library/Caches/bat/themes.bin ... okay
Writing syntax set to /Users/mcmillion/Library/Caches/bat/syntaxes.bin ... okay

➜ bat cache --config-dir
/Users/mcmillion/Library/Preferences/bat
```

`BAT_CONFIG_PATH` does seem to be working just fine, however. Am I just misunderstanding how this configuration option is supposed to work? I'd definitely like to keep things under `~/.config` as that's the convention and having to symlink things to `~/Library` on various machines feels wrong.

sharkdp:
Thank you very much for the feedback.

`BAT_CONFIG_DIR` is a variable that I have used in the `README` to make the use of `bat cache --config-dir` clear. It is not actually a configuration option for `bat` (i.e. `bat` will never read this environment variable).

Currently, there is no way to control the directory where user-customized themes and syntaxes are found (Note that - perhaps confusingly - there is a `BAT_CONFIG_PATH` environment variable that can be used to control where `bat`s configuration file is stored).

The discussion about `~/Library` being used on macOS came up a few times in past tickets. I think we should simply change the default for macOS to be `~/.config/bat`. Opinions?

lavifb:
As far as I can tell, `~/Library` is mostly used for regular GUI Mac apps. For command line applications like bat, usually you use `~/.config/bat` or `~/.bat`. I tend to prefer `~/.config/bat` because otherwise your home directory gets very littered with config files.

lavifb:
To reply to the original question, would you consider either changing the variable name in the README or excluding it entirely (using maybe `"$(bat cache --config-dir)/themes"`)? The current name in the example, `BAT_CONFIG_DIR`, is formatted like a configuration option which probably leads to the above confusion.

sharkdp:
Sounds like a good idea. Let's remove it completely and "inline" `$(bat cache --config-dir)`.

[...]

PR #435, opened 2018-12-02
Removed confusing variable from examples in README.md

lavifb:
Solves #419 by removing confusing the variable `BAT_CONFIG_DIR` from examples because it is not a real configuration option for bat.


sharkdp:
Thank you very much!

Issue #441, opened 2018-12-07
Report name of missing file


skuhl:
If you use bat to concatenate a list of files together (i.e., bat file1.txt file2.txt ... > combined.txt), there is an error message when one of the listed files is missing:
[bat error]: No such file or directory (os error 2)

The error message should report the filename of the file that is missing to help the user identify which parameter passed to bat is missing.

If I use cat and a file is missing, the error message is:
/bin/cat: missingfile.txt: No such file or directory

sharkdp:
Thank you for the feedback.

Good suggestion! We should implement this.

sharkdp:
Released in [v0.10.0](https://github.com/sharkdp/bat/releases/tag/v0.10.0).

PR #444, opened 2018-12-10
Report filename in error message

ufuji1984:
This PR fixes issue #441.
( And also adds filename to any other Errors from opening a file. )

I wonder if there is more concise way to do this, so please let me know if there is.
Thanks in advance!

### DIFF
--- a/src/inputfile.rs
+++ b/src/inputfile.rs
@@ -59,7 +59,10 @@ impl<'a> InputFile<'a> {
         match self {
             InputFile::StdIn => Ok(InputFileReader::new(stdin.lock())),
             InputFile::Ordinary(filename) => {
-                let file = File::open(filename)?;
+                let file = match File::open(filename) {
+                    Ok(f) => f,
+                    Err(e) => return Err(format!("{}: {}", filename, e).into()),
+                };
 
                 if file.metadata()?.is_dir() {
                     return Err(format!("'{}' is a directory.", filename).into());


src/inputfile.rs
```diff
@@ -59,7 +59,10 @@ impl<'a> InputFile<'a> {
         match self {
             InputFile::StdIn => Ok(InputFileReader::new(stdin.lock())),
             InputFile::Ordinary(filename) => {
-                let file = File::open(filename)?;
+                let file = match File::open(filename) {
```
sharkdp: I think we could use `.map_err` here. Something like:
```rust
let file = File::open(filename).map_err(format!(...).into())?;
```

ufuji1984: Thanks for your advice!
I tried it, but I couldn't get it built...
Seems out of my hands...

```
62 |                 let file = File::open(filename).map_err(format!("{}: {}", filename, e).into())?;                                                    
   |                                                                                     ^ not found in this scope   
```

```
62 |                 let file = File::open(filename).map_err(|e| format!("{}: {}", filename, e).into())?;
   |                                                 ^^^^^^^ cannot infer type for `F` 
```

```
62 |                 let file = File::open(filename).map_err(|e:Error| format!("{}: {}", filename, e).into())?;  
   |                                                 ^^^^^^^ ------------------- found signature of `fn(errors::Error) -> _`
   |                                                 |                                                           
   |                                                 expected signature of `fn(std::io::Error) -> _`  
```

```
62 |                 let file = File::open(filename).map_err(|e: std::io::Error| -> std::io::Error {format!("{}: {}", filename, e).into()})?;            
   |                                                                                                                               ^^^^ the trait `std::convert::From<std::string::String>` is not implemented for `std::io::Error`
```

sharkdp: Ah, I was wrong. `.into()` is not needed. This should work (I'd like to add the additional quotes around the filename):
``` rust
                let file = File::open(filename).map_err(|e| format!("'{}': {}", filename, e))?;
```

ufuji1984:
Thanks a lot!!!

PR #453, opened 2018-12-16
Line highlight

### LINKED ISSUE #175: highlight individual line
Now that bat supports displaying a range of lines I would like to have bat be able to highlight one line.

My use case has to do with utilizing bat as a preview command in a fuzzy finder like [fzf](https://github.com/junegunn/fzf) or [skim](https://github.com/lotabout/skim).

Here is what we can do currently with bat: fzf
(I'm just using a copy of preview.sh I've put up [here](https://gist.github.com/tskinn/0ed07b1c43968807b24e0e585ae58ce2))

So how can or should this be done? Invert the colors of the line? Does syntec api have something that makes this easy? Or should we just manipulate the grid portion of the display like the line number or insert an arrow in that gutter?

Again I'd be happy to work on it but would like some input before I do.

sharkdp:
Continuing the work of #346 (/CC @tskinn)

* Allow for multiple occurrences of `--highlight-line <N>`
* Highlight the full line
* Proper handling for overlong lines

image

image

closes #175

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -80,8 +80,8 @@ pub struct Config<'a> {
     /// Whether or not to use ANSI italics
     pub use_italic_text: bool,
 
-    /// A line to highlight
-    pub highlight_line: Option<usize>,
+    /// Lines to highlight
+    pub highlight_lines: Vec<usize>,
 }
 
 fn is_truecolor_terminal() -> bool {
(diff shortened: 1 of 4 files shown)


FinallyFinancialFreedom:
perfect!

ReggieMarr:
What happened here with regards to highlighting multiple lines ? Is that subfeature still supported ?

sharkdp:
> s there a reason this wasn't implemented the same was as the line-range argument ? To me that would seem more intuitive.

Sounds like a great idea! Could you please open a new issue for that?

PR #454, opened 2018-12-17
Do not parse arguments in PAGER

### LINKED ISSUE #430: Do *not* parse arguments in PAGER
We recently introduced command-line argument parsing for `PAGER` and `BAT_PAGER` and it's been causing some problems, because some people have set `PAGER="less -X"` (or similar) without including the `-R` argument which is needed in order for `bat` to work with `less` (enables parsing of ANSI sequences).

I think we should *disable* parsing of command line arguments for `PAGER` again but leave it for `BAT_PAGER`. If someone has `PAGER="less -X"` set, we should fall back to the default behavior (use `less` with `-FRX`).

majecty:
This fixes #430

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -26,44 +26,60 @@ impl OutputType {
 
     /// Try to launch the pager. Fall back to stdout in case of errors.
     fn try_pager(quit_if_one_screen: bool, pager_from_config: Option<&str>) -> Result<Self> {
-        let pager_from_env =
-            env::var("BAT_PAGER").or_else(|_| env::var("PAGER").map(add_default_flags_to_less));
+        let mut replace_arguments_to_less = false;
+
+        let pager_from_env = match (env::var("BAT_PAGER"), env::var("PAGER")) {
+            (Ok(bat_pager), _) => Some(bat_pager),
+            (_, Ok(pager)) => {
+                // less needs to be called with the '-R' option in order to properly interpret the
+                // ANSI color sequences printed by bat. If someone has set PAGER="less -F", we
+                // therefore need to overwrite the arguments and add '-R'.
+                //
+                // We only do this for PAGER (as it is not specific to 'bat'), not for BAT_PAGER
+                // or bats '--pager' command line option.
+                replace_arguments_to_less = true;
+                Some(pager)
+            }
+            _ => None,
+        };
 
         let pager = pager_from_config
             .map(|p| p.to_string())
-            .or(pager_from_env.ok())
+            .or(pager_from_env)
             .unwrap_or(String::from("less"));
 
-        let pagerflags = shell_words::split(&pager)
-            .chain_err(|| "Could not parse (BAT_)PAGER environment variable.")?;
+        let pagerflags =
+            shell_words::split(&pager).chain_err(|| "Could not parse pager command.")?;
 
         match pagerflags.split_first() {
-            Some((pager_name, mut args)) => {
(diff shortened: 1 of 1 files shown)


majecty:
@sharkdp I am struggling with fixing the test case `pager_basic`.
Could you give me some hints?

https://github.com/sharkdp/bat/blob/b68c5d857688201db44a188cbd89c93f9841ea7f/tests/integration_tests.rs#L313

majecty:
I tried by creating pager_output.sh file, but the windows test failed. :(

sharkdp:
Thank you very much for looking into this!

Maybe we could do the following (this would also fix the `echo`-based tests, I think):
- Split the `PAGER` environment variable into arguments with `shell_words::split`.
- Check if the (basename of) the first component is `less`
- If yes, call the given version of `less` with our fixed set of arguments (`-FRX`).
- If no, call the pager as specified with all of its arguments.

This way, if someone wants to use `my_special_pager` with a given set of command-line arguments, it can still be done via `PAGER`.

majecty:
Thanks. I'll try it.

src/output.rs
```diff
@@ -26,7 +26,8 @@ impl OutputType {
 
     /// Try to launch the pager. Fall back to stdout in case of errors.
     fn try_pager(quit_if_one_screen: bool, pager_from_config: Option<&str>) -> Result<Self> {
-        let pager_from_env = env::var("BAT_PAGER").or_else(|_| env::var("PAGER"));
+        let pager_from_env =
+            env::var("BAT_PAGER").or_else(|_| env::var("PAGER").map(add_default_flags_to_less));
```
sharkdp: The whole point of this was to have a different handling of `PAGER` and `BAT_PAGER`.  We do not want to modify anything for `BAT_PAGER`.

src/output.rs
```diff
@@ -87,6 +88,20 @@ impl OutputType {
     }
 }
 
+// Since someone could set PAGER to less without arg -R that bat needs,
+// ignore flags for less and change flags to -FRX.
+// If a user wants an env variable with flags unchanged, he/she should use
+// BAT_PAGER.
+fn add_default_flags_to_less(pager: String) -> String {
+    if let Ok(flags) = shell_words::split(&pager) {
+        if flags.first() == Some(&String::from("less")) {
```
sharkdp: This does not work if the first argument is `/usr/bin/less`, for example. See the `is_less` variable example in the same file.

majecty:
When I run the same test in my repository, the test succeeded. I don't know why the test failed. Could you help me?
@sharkdp

majecty:
@sharkdp Thanks! the test passed. Could you review this PR again?

majecty:
@sharkdp Sorry for late.

Then how about this?

```rust
        let pager_from_env = match (env::var("BAT_PAGER"), env::var("PAGER")) {
            (Ok(bat_pager), _) => Some(bat_pager),
            (_, Ok(pager)) => {
                if PathBuf::from(pager.clone()).file_stem() == Some(&OsString::from("less")) {
                    Some(String::from("less"))
                } else {
                    Some(pager)
                }
            }
            _ => None,
        };
```

sharkdp:
> Then how about this?

This wouldn't work if `PAGER="less -F"`, for example, because we need to split off the argument first, before checking the file_stem. I have added an implementation which should hopefully work. It is an adapted version of your latest proposal.

In any case, thank you very much for your work and your patience.

Issue #471, opened 2019-01-18
protobuf format support


netpyoung:
https://developers.google.com/protocol-buffers/docs/proto3

It will be more helpful for support google's protobuf.

caos21:
Added in pull request #474

PR #474, opened 2019-01-19
Protobuf-syntax and ProtobufText syntax

caos21:
Add Protobuf-syntax and ProtobufText syntax as in https://github.com/VcamX/protobuf-syntax-highlighting


caos21:
#471

sharkdp:
Thank you very much. There is no need to copy the `.sublime-syntax` files to the `assets/syntaxes` folder if they are already included in the submodule (it is only needed if we have to manually generate these files from `.tmLanguage` files). Could you please remove these two files again (just add the submodule)?

caos21:
Thanks to you. That's what I thought, but I believe in the first build I made didn't process the new syntax. I will remove it. I am planning to add chapel syntax if you think it is worth.

sharkdp:
Hm.. why did you remove the `syslog.sublime-syntax`? This file is *not* included in the respective submodule, so we are going to need it in `assets/syntaxes`.

caos21:
Yes it is. In the cloned repo. From the original it is not possible to generate the `syslog.sublime-syntax` because some errors, so I cloned and try to correct it. I think the sumodule included it is the cloned repo. Sorry for the misunderstanding, maybe I am missing something.

sharkdp:
> Yes it is. In the cloned repo. From the original it is not possible to generate the `syslog.sublime-syntax` because some errors, so I cloned and try to correct it.

I see. But the submodule is currently still pointing at the [`f9fe104a60280d0ce8e1862f62477a7a9979e92a` ](https://github.com/caos21/syslog-syntax/commit/f9fe104a60280d0ce8e1862f62477a7a9979e92a) commit hash and not at the [`7628d4e0b89d3b4f0dcc014bfb4a3e3433f5133a`](https://github.com/caos21/syslog-syntax/commit/7628d4e0b89d3b4f0dcc014bfb4a3e3433f5133a) commit which you added in addition.

This would have to be updated, too.

caos21:
Thank you, I saw it is already fixed. I am sorry for the issues.

Issue #442, opened 2018-12-07
Change the default configuration directory on macOS to .config/bat


sharkdp:
Change the default path to `bat`s configuration directory to
```
~/.config/bat
```

see #151, #375 and #419

lavifb:
You seem to be using `directories::ProjectDirs` to get the config directory. Do you want just set it in `bat` or do you prefer to try to add this option to the `directories` crate upstream?

sharkdp:
It would be great if that would be supported by `directories`, yes.

However, it seemed to me that there was no intent to change this, see https://github.com/rust-lang-nursery/cli-wg/issues/7#issuecomment-391819505 and https://github.com/rust-lang-nursery/cli-wg/issues/7#issuecomment-391874453
It might be worth asking again, though.

lavifb:
Ok I opened an issue in directories
https://github.com/soc/directories-rs/issues/47#issue-388793645

lavifb:
There does not seem to be much motion on this issue. I don't really want to write a new crate because the various dir crates seem pretty fractured already, so I'll write a quick fix in bat and hopefully this gets fixed upstream at some point.

sharkdp:
Changes by @lavifb released in [v0.10.0](https://github.com/sharkdp/bat/releases/tag/v0.10.0).

PR #491, opened 2019-01-31
Updated bat config dir for MacOs to ~/.config/bat/

lavifb:
Removed directories crate and using dirs_rs
Changed bat config file and cache dir on MacOs:
config: ~/.config/bat/config
cache: ~/.config/bat/cache/


lavifb:
Fix for https://github.com/sharkdp/bat/issues/442

lavifb:
I changed cache_dir on MacOS to ~/.cache/bat/ to agree with Linux

sharkdp:
Thank you very much for working on this!

You are right, it seems like `directories` it not really actively maintained anymore :disappointed:. Still, it would be great if this could be handled in an external crate.

If there is no such crate, I'm fine with a temporary solution like you suggested here. However, it seems to me that this solution here is not quite sufficient, as we probably don't want to hard-code a specific directory, but rather add support for the `$XDG_CONFIG_HOME` and `$XDG_CACHE_HOME` environment variables (and only fall back to the hard-coded paths).

lavifb:
Ah, you're right. I'll modify it to fallback to the XDG env variables first.

sharkdp:
Looks great! I made a few small changes, hope that's fine.

lavifb:
Yeah, of course! I didn't realize you could the `?` operator on `Option`s like you can for `Result`s.

lavifb:
Out of curiosity, why did you have both `cache-dir` and `config-dir` flags for `bat cache`?

sharkdp:
`--config-dir` was there for convenience to look up the configuration directory on any operating system. I use it in the `bat` README to make the documentation OS-independent.

I now added `--cache-dir` for completeness and for debugging purposes. I think it makes sense to let users also look up the caching directory via `bat cache --cache-dir`. Do you see any harm in adding `--cache-dir`?

lavifb:
I agree it makes sense but it may be confusing to someone just reading the man page/help. So if you are trying to add a theme like in the README you use `bat cache --config-dir` and later call `bat cache --init`. With the option to accidentally call `bat cache --cache-dir`, I feel like there is room for a lot of confusion, similar to https://github.com/sharkdp/bat/issues/496. Mostly, since it is a flag for `bat cache` it is hard to tell if `config-dir` is the config for `bat` or `bat cache` or both and whether there is a difference between that and `cache-dir`.

It would possibly make more sense for these to be options for `bat`, so that now it feels like you are changing the config by modifying `bat --config-dir` and then caching the results with `bat cache`. This way it's clearer that `config-dir` and `cache-dir` are both for `bat` and `bat cache` is just there to compile the changes in `config-dir` into `cache-dir`.

lavifb:
Now that I'm looking at it, it also may be clearer to use something like `bat cache --build` instead of `bat cache --init` because `init` feels like something that should only be called once but if I understand this correctly, you should call it anytime you want to add theme/syntax to the config. What do you think?

sharkdp:
@lavifb Excellent points, yes. Completely agreed. I'll try to address this in another PR soon.

lavifb:
Thanks! and thanks for making bat!

Issue #477, opened 2019-01-22
EOF must be sent twice on stdin if no other input is sent


willcassella:
If you run `bat` and then immediately press the EOF sequence (Ctrl-D), bat prints the filename 'STDIN' but continues to read from it. You have to press it again to actually close the file.

sharkdp:
Hm, interesting. Thank you very much for reporting this.

mziter:
I have found and resolved issue locally. However, I am not sure how to exit immeditately as process::exit(0) will exit immediately and not do any destructuring, etc...

Other two options are 1) Wrap new function in a result, which seems idiomatic, but would require a lot of downstream changes I am sure. 2) Wrap in a new ContentType that would basically get passed on and ignored at some future point.

Would like your thoughts...

mziter:
Sorry, just saw this referenced already. :) Will put some comments on pull request.

sharkdp:
closed via #492 by @reidwagner

sharkdp:
Fixed in [v0.10.0](https://github.com/sharkdp/bat/releases/tag/v0.10.0).

reidwagner:
This has been closed - but there was a change in expected behavior, where we no longer get a header and footer for empty input files.

Trying to bring back that behavior while keeping the error logic here seemed to get complex. I think I have a simpler solution all around.

Instead of what's implemented in #492 (that could be reverted), we simply don't call `print_file_ranges()` if our reader.first_line is empty. That appears to give all the desired functionality, + it prints the STDIN header and footer when we press CTRL-D, which seems more consistent in the first place. e.g in `print_file()`:       
```
if !reader.first_line.is_empty() {
    self.print_file_ranges(printer, writer, reader, &self.config.line_ranges)?;
}
```

I _think_ maybe the check that `first_line.is_empty()` could also be removed from `read_line` since the condition above would work for empty files.

sharkdp:
> Instead of what's implemented in #492 (that could be reverted), we simply don't call `print_file_ranges()` if our reader.first_line is empty. That appears to give all the desired functionality, + it prints the STDIN header and footer when we press CTRL-D, which seems more consistent in the first place

Sounds great!

> I _think_ maybe the check that `first_line.is_empty()` could also be removed from `read_line` since the condition above would work for empty files.

[...]

PR #492, opened 2019-02-02
Exit on first EOF from stdin

reidwagner:
Addresses #477.

### DIFF
--- a/src/controller.rs
+++ b/src/controller.rs
@@ -43,6 +43,7 @@ impl<'b> Controller<'b> {
 
         for input_file in &self.config.files {
             match input_file.get_reader(&stdin) {
+                Err(Error(ErrorKind::ImmediateEOF, _)) => (),
                 Err(error) => {
                     handle_error(&error);
                     no_errors = false;
--- a/src/inputfile.rs
+++ b/src/inputfile.rs
@@ -14,9 +14,11 @@ pub struct InputFileReader<'a> {
 }
 
 impl<'a> InputFileReader<'a> {
-    fn new<R: BufRead + 'a>(mut reader: R) -> InputFileReader<'a> {
+    fn new<R: BufRead + 'a>(mut reader: R) -> Result<InputFileReader<'a>> {
         let mut first_line = vec![];
-        reader.read_until(b'\n', &mut first_line).ok();
+        if reader.read_until(b'\n', &mut first_line)? == 0 {
+            return Err(ErrorKind::ImmediateEOF.into());
+        }
 
         let content_type = content_inspector::inspect(&first_line[..]);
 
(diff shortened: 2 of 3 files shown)


mziter:
per docs, [process::exit](https://doc.rust-lang.org/std/process/fn.exit.html): 

> Note that because this function never returns, and that it terminates the process, no destructors on the current stack or any other thread's stack will be run.

Not sure if this is an issue or not?

reidwagner:
Thanks for bringing this up! I was hoping process::exit was sufficient for this simple case, but I can see the argument for a stricter exit sequence as well.

This is also how a BrokenPipe error from get_reader is handled, so there is a precedent.

I'd like to know the maintainer's thoughts as well :).

sharkdp:
Thank you for working on this.

Yes, returning an `Err(..)` with a custom error would definitely be the preferred option here, instead of calling `process::exit` directly.

Using `process::exit` is probably even wrong here if there are multiple input files...

Speaking of wrong behavior: it would be great if we could add integration/regression tests for this (see `tests/integration_tests.rs`)

reidwagner:
@willcassella I added the Result here to handle `read_until()?`. Do you think I should have left it as is, discarding the error?

sharkdp:
> @willcassella I added the Result here to handle `read_until()?`. Do you think I should have left it as is, discarding the error?

I think it should be fine to handle the error with `?` here - I originally thought it would cause problems with broken pipes (when the input file only has one line without line ending), but it returns `Some(..)` in this case.

reidwagner:
Okay, I created a new ErrorKind variant and essentially just ignore it - would any extra actions need to be taken? The behavior is now correct with multiple input files, e.g. `bat foo.txt - foo.txt`.

I haven't figured out how to emulate a CTRL-D for an integration test yet.

src/controller.rs
```diff
@@ -45,7 +45,10 @@ impl<'b> Controller<'b> {
             match input_file.get_reader(&stdin) {
                 Err(error) => {
                     handle_error(&error);
-                    no_errors = false;
+                    match error {
```
sharkdp: Instead of matching on `ErrorKind::ImmediateEOF` here and in `handle_error`, could we also move the `handle_error` call into the `_ => ` branch?

``` rust
match error {
    Error(ErrorKind::ImmediateEOF, _) => (),
    _ => {
        handle_error(&error);
        no_errors = false
    },
}
```

sharkdp:
> I haven't figured out how to emulate a CTRL-D for an integration test yet.

You are right, that's probably very tricky. I'd like to have some tests where we check that `bat empty test == bat test`, `bat test empty == bat test`, `bat empty empty == empty`, etc., but that's not really part of this issue (they would have caught the problem with `process::exit`, though).

reidwagner:
Makes sense. I can write that up.

sharkdp:
Thank you for the update!

willcassella:
> I added the Result here to handle `read_until()?`

Sorry I didn't notice that, I got distracted by how cool `never` is :)

Thanks for fixing this!

PR #497, opened 2019-02-04
add varlink syntax

haraldh:


sharkdp:

For future contributions: please always open a ticket before sending a PR so that potential changes can be discussed first.

Can you please provide a short description to your PR? What is varlink? How popular is it and why should it be added to bat by default?

haraldh:
Sorry, just found out about the possible user customization via `bat cache --init`.

If you still want to merge the PR, I added this commit message:

Varlink is an interface description format and protocol that aims to make services accessible to both humans and machines in the simplest feasible way.

See: https://varlink.org/ and https://github.com/varlink/ for more information

Most prominent varlink user to date is [podman](https://podman.io/), which is a `docker` replacement.
https://github.com/containers/libpod/blob/master/cmd/podman/varlink/io.podman.varlink

sharkdp:
Ok, thank you for the detailed explanation!

PR #498, opened 2019-02-07
Refactor cache subcommand

sharkdp:
* Rename `bat cache --init` to `bat cache --build`
* Move `--config-dir` and `--cache-dir` from 'bat cache' to 'bat' and hide them from the help text.

@lavifb FYI


lavifb:
looks great!

sharkdp:
Thank you for the suggestion!

PR #500, opened 2019-02-08
Add integration tests running bat with empty input files.

reidwagner:
Relates to discussion on PR #492.

### DIFF
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -77,6 +77,58 @@ fn concatenate_stdin() {
 }
 
 #[test]
+fn concatenate_empty_first() {
+    bat()
+        .arg("empty.txt")
+        .arg("test.txt")
+        .assert()
+        .success()
+        .stdout("hello world\n");
+}
+
+#[test]
+fn concatenate_empty_last() {
+    bat()
+        .arg("test.txt")
+        .arg("empty.txt")
+        .assert()
+        .success()
+        .stdout("hello world\n");
+}
+
+#[test]
+fn concatenate_empty_both() {
+    bat()
+        .arg("empty.txt")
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you very much!

One thing I noticed after releasing 0.10 is that there is a change in behavior with your implementation in #492 regarding empty files:

If you call `bat test.txt empty.txt test.txt` on an interactive terminal (where it prints everything with grid, line numbers, header, etc.), you get this:

```
▶ bat test.txt empty.txt test.txt 
───────┬───────────────────────────────────────────
       │ File: test.txt
───────┼───────────────────────────────────────────
   1   │ hello world
───────┴───────────────────────────────────────────
───────┬───────────────────────────────────────────
       │ File: test.txt
───────┼───────────────────────────────────────────
   1   │ hello world
───────┴───────────────────────────────────────────
```

I believe that previous versions of `bat` would have printed the file header. Like this:
```
▶ bat test.txt empty.txt test.txt 
───────┬───────────────────────────────────────────
       │ File: test.txt
───────┼───────────────────────────────────────────
   1   │ hello world
───────┴───────────────────────────────────────────
───────┬───────────────────────────────────────────
       │ File: empty.txt
───────┴───────────────────────────────────────────
───────┬───────────────────────────────────────────
       │ File: test.txt
───────┼───────────────────────────────────────────
   1   │ hello world
───────┴───────────────────────────────────────────
```

It's not really that important, but I think it would be nice to get this behavior back. To check for this in an integration test, we could call something like:
```
bat --style=header --decorations=always empty.txt
```
and make sure that the header is there.

reidwagner:
Ah, I didn't notice that. I'll take a look ASAP. Thanks!

PR #504, opened 2019-02-11
Only print contents of file if not empty.

reidwagner:
This avoids waiting for a second EOF from stdin, and preserves header and footer for empty stdin and files. This is an alternative to #492 (reverted) as a solution to #477.


sharkdp:
Thank you very much!

reidwagner:
Thanks!

PR #518, opened 2019-03-04
Fix syntax on "--config-dir"

issmirnov:
Using the latest bat I got a syntax error. Updating README for others.


sharkdp:
Are you sure that you are using the latest version? What do you get if you call `bat --version`?

issmirnov:
Whoa. I think my ansible script is broken. It should be getting the latest releases, but looks like I'm running `0.4.1`. Oops!!

Issue #372, opened 2018-10-21
Generated fish completion does not complete filenames


sboukortt:
The result is that Tab is only able to complete options, and not the paths to the files that one may want to display.

sharkdp:
Thank you for reporting this.

I think this is related to https://github.com/clap-rs/clap/issues/568

swiesner-dlr:
Personally I scarcely use `bat` options, or at least I use options way less frequently than filenames, so I much prefer to complete filenames instead of options.

Could you disable these completions until the upstream error is fixed?

sharkdp:
@swiesner-dlr I agree. Sounds like a good idea to me. We should enable the shell completions for the next release.

sharkdp:
Shell completions are not included in the new release [bat 0.9](https://github.com/sharkdp/bat/releases/tag/v0.9.0).

ErinCall:
I'm still having this issue—I get tab-completion for options, but not for filenames. I'm using bat 0.9.0 (newly installed via homebrew) and I'm using fish 3.0.0 (recently upgraded from 2.7.1 in case an outdated fish version was the problem). I haven't tried building from source; is that worth an attempt?

sharkdp:
Can you please search your filesystem for any remaining `fd.fish` autocompletion files? 

It seems like the brew formula still ships the completion files: https://github.com/Homebrew/homebrew-core/blob/5df9397fb96b3e74b7aca993afa2b620e0db9d44/Formula/bat.rb

I'm afraid this would have to be reported in the homebrew-core repository.

maraisr:
How long would it take for homebrew to pick up the new changes? Ie, at 1807 25/01/2019 I ran a `brew install bat` and am seeing the `bat.fish` completion files.

sharkdp:
I have now completely disabled the generation of the files, so for the next release package maintainers will be "forced" to remove them from the packages.

alexbepple:
If, somehow, you still have the old completions, you can also just add

    complete -c bat -a '(__fish_complete_path)'

to your config and have it all. ;)

jonas-schievink:
@alexbepple Oooh, thanks for this, very handy!

@sharkdp I'm sure you've thought of this already, but another workaround would be to stop using a subcommand and instead:
* Turn `cache` into normal arguments that use dashes, like `bat --init-cache`, `bat --clear-cache`, ...
* Add another small CLI tool *just* for managing the cache, like `bat-cache --init`, ...

[...]

PR #524, opened 2019-03-11
Minimal fish shell completions

ev-dev:
### Initial Shell Completions
For fish shell only; contains completions for flags but does not include flag sub-options like language lists or subcommands.

From what I understand #482 took place because of bugs with completion autogeneration. I added the 'completions' directory to assets so that perhaps others will follow with ported or manually written bash/zsh completions.


cben:
Tested.  Basically works.
Does not suffer from #372 problem — I can complete file names everywhere.   Including places where strictly I shouldn't, e.g. <code>bat --theme <kbd>TAB</kbd></code> which ideally would complete theme names not files.  But perfect is the enemy of good...

sharkdp:

My only "concern" with this is that it has to be maintained manually. Most contributors will probably not update the completions files when they change command-line options or change help texts (the same happens with the man page). This means that I have to remind people in the PRs or do it myself when I create a new release.

I'm inclined to merge this, but I'm not sure if I will always find the time to keep it up-to-date.

sharkdp:
Oh, before I merge this, we should definitely make sure that the completion file will be added to the release tarballs and Debian packages. See #391 for details.

bat.fish
```diff
@@ -0,0 +1,59 @@
+# Fish Shell Completions (WIP)
+
+# Place or symlink to $XDG_CONFIG_HOME/fish/completions/bat.fish ($XDG_CONFIG_HOME is usually set to ~/.config)
```
sharkdp: I think this can be removed if we properly install this file?

bat.fish
```diff
@@ -0,0 +1,59 @@
...
+complete -c bat -s l -l language -d "Set the language for syntax highlighting"
+
+complete -c bat -l list-languages -d "Display list of supported languages for syntax highlighting"
+
+complete -c bat -s m -l map-syntax -d "<from:to> Map a file extension or file name to an existing syntax"
+
+# TODO: add completion for theme list
```
sharkdp: `bat --list-themes` does produce a newline-separated list of theme names if it is run in non-interactive mode, i.e. if the output is piped somewhere.

ev-dev:
I definitely understand the concern regarding future contributor updates @sharkdp, manually written completions would hopefully be a temporary workaround until some future upgrade of the autogeneration piece. 

Until then, It might be worth adding a note in the readme or a quick CONTRIBUTING.md with some guidelines, I'd add it to this PR but I don't want to impose in that regard.

ev-dev:
I pushed up some minor updates including working --theme autocompletion after that helpful note on piped --list-themes output

I tried getting the completions as part of release packages during testing but I don't have a lot of experience with cargo builds or how to include completions as part of an install for linux/windows (could probably figure out macOS tho if it just involves updating a brew formula or something)

sharkdp:

The deployment should hopefully work just like this.

sharkdp:
Can someone from this thread maybe help with actually shipping the fish completions in the Debian package (#554)?

PR #543, opened 2019-04-13
Add 3 new themes: ansi-light, ansi-dark, base16

### LINKED ISSUE #490: Add a theme that uses terminal colors
I spent a lot of time to customize my terminal named colors, and I wish there was a bat theme that uses them so it doesn’t look bad in my terminal.

The benefit is that it provides a backup theme that works across terminals if no other theme works with the terminal theme.

mk12:
Fixes #490

To summarize that issue: some people change their color theme often, perhaps switching between a light theme and dark theme. In these cases, no single bat theme works well. In addition, some people really want colored output to be congruous with the rest of their terminal programs, and don't care so much about it using a wide range of colors. In both cases, a solution is to use the 8- or 16-ANSI color palette the user has already configured for their terminal. The key difference is the added level of indirection: instead of choosing colors, we choose palette numbers and the user chooses colors.

This PR adds three new themes. They are based on the [base16 textmate theme][1], but modified to encode the 16 palette numbers as RGB colours "#000000" through "#0f0000". This is necessary because tmTheme/[Syntect][2] require concrete RGB colors. It's might seem slightly hacky, but I don't think it's worth generalizing Syntect to other notions of color just so we can pass it the number 10 instead of the string "#0a0000".

Here's how the themes should be used:

- **ansi-light**: This should look decent in any terminal with a light background. It uses only the first 8 colors from the ANSI palette.
- **ansi-dark**: Like ansi-light, but for dark backgrounds. These two themes need to be separate so that we use the right foreground color for ordinary text.
- **base16**: This will look perfect for anyone using a theme designed around the [base16][3] palette. We don't need a light and dark version because base16 themes always have background/foreground in fixed slots.

Note, I just put the new theme files in `assets/themes` for now. It works if I build the cache from that directory. The README seems to say I'm not supposed to check in the new binary files with my PR, however.

[1]: https://github.com/chriskempson/base16-textmate/blob/master/Themes/base16-default-dark.tmTheme
[2]: https://github.com/trishume/syntect
[3]: https://github.com/chriskempson/base16


sharkdp:
Thank you very much, this looks really great! It will take me a few days before I can properly review this, as I'm currently on vacation.

mk12:
Just updated the theme files after realizing the gutter settings were specified incorrectly.

One thing that just occurred to me: since I'm enabling `ColorMode::Base16Encoded` based on the name of the theme, this could cause undesirable behaviour if the user decides to make their own theme called ansi-light/ansi-dark/base16. I think a nicer solution would be to make the encoding a key in the XML file.

trishume:
Author of syntect here, I approve of this method of doing ANSI themes with syntect.

One idea I thought of and posted in the associated issue is to use the alpha channel to select between rendering as ansi and rendering as RGB. It's possible to specify the alpha channel in TextMate themes using longer hex codes, and an alpha of zero (transparent) could be used as a special signifier of an ANSI color.

sharkdp:
@trishume Thank you for the feedback! It's still a "hack", but that sounds like a reasonable option to me. @mk12 what do you think?

mk12:
Yes, I think I'll try doing it with the alpha channel instead.

Advantages:

- Much smaller change to the codebase (just a few lines in `to_ansi_color`).
- Setting is per colour rather than per theme so it's more flexible.
- No terminal that I know of has escapes for transparency, so no one should be using it anyway.
- No issue with people making their own theme named "ansi" or "base16".

mk12:
I've pushed the new, much simpler changes. The old version is in the `old-ansi` branch of my fork if you want to compare.

ariasuni:
Thanks a lot, that’s perfect!

PR #548, opened 2019-04-20
Add syntax highlighting for fish shell

### LINKED ISSUE #484: feature request: support for fish syntax
Would you be amenable to accepting a separate highlighter for fish? Currently bat is using the bash syntax highlighter for fish files - whilst this is close, bash and fish do have different syntaxes. As to usage stats, I'd imagine (but I have no actually evidence of this) that fish is the 3rd most popular shell behind bash and zsh and that it's in pretty wide use. There's a sublime package for it that has about 8K installations. I could make a PR, but I thought I'd ask first

sanga:
fixes #484


sharkdp:
Thank you very much!!

Issue #553, opened 2019-05-02
Potential solution to the TODO in Fish shell completions.


eth-p:
I was just taking a look through the fish completions, and I noticed the TODO at line 19.

https://github.com/sharkdp/bat/blob/e7c5561df736d42734c39712c52abc19ec9aa8d5/assets/completions/bat.fish#L19-L20

After playing around with it, I have a solution that should work on both Linux and MacOS without having to install any extra tools. It's kind of a dirty hack, but it's pure `awk`, `tr`, and `sed` (and optionally `sort` and `uniq`).

```bash
bat --list-languages | awk 'NR==1{dc=0; while(substr($0,dc,2)!="  ") dc++; while(substr($0,dc,1)==" ") dc++}; {print substr($0,dc)}' | tr ',' '\n' | sed 's/^ \+//; s/ \+$//; /^$/d' | sort | uniq
```

## Explanation:

First of all, the language list needs to be parsed. The implementation of `bat --list-languages` wasn't really meant to be read by anything but a human, but with a little hackery, `awk` is perfectly capable of parsing it provided the following two properties are met:

1. The name of the first language is not the longest one in the list (breaks the double space assumption).
2. The name of the first language does not contain two consecutive spaces (same reason as above).

If these properties are met, the following awk script can be used:

```awk
NR == 1 { # Run this for the first column.
    dc = 0                                   # Set the column split counter to zero.
    while (substr($0, dc, 2) != "  ") dc++;  # Skip past the language name column.
    while (substr($0, dc, 1) == " ") dc++;   # Skip past the whitespace that separates the columns.
};

{
    print substr($0, dc) # Prints everything after the `dc` variable (i.e. only the second column).
}
```

After piping `bat --list-languages` to that script, the output will look something like this:

```
as
csv, tsv
applescript, script editor
s, S
adoc, asciidoc, asc
asa
yasm, nasm, asm, inc, mac
awk
bat, cmd
```

From here, we have two options: use the first entry, or just split them all. Since the first entry isn't always the most common extension for the language, it's safer to just use them all. That can be achieved by using `tr` to replace the commas with newlines, and `sed` to clean up the empty lines and garbage (leading/trailing) whitespace.

```bash
... | tr ',' '\n' | sed 's/^ \+//; s/ \+$//; /^$/d'
```

After that, you can optionally sort and de-duplicate any languages using `... | sort | uniq`.

## Options:

If you want to exclude non-extension files, you can modify the `sed` patterns to do the following:

[...]

PR #555, opened 2019-05-02
Add completions for `--language` when using the fish shell.

eth-p:
This a pull request to solve the TODO at line 19 in the fish shell completions.

https://github.com/sharkdp/bat/blob/e7c5561df736d42734c39712c52abc19ec9aa8d5/assets/completions/bat.fish#L19-L20

The basic principle of the solution is the same as what I described in #553, and it has the same caveats, but this is a better implementation that only uses `awk`.

## Features:

image

- Lists all languages from `bat --list-languages`.
- Displays the full name of the language.
- Filters out languages that are probably config files (starts with an upper case letter).

## Caveats:

In order to properly parse the output of `bat --list-languages`, two conditions need to be met with the first language that appears in the list.

1. There is at least two spaces after the language name (i.e. is not the longest name).
2. The language name does not contain two consecutive spaces in it.

### DIFF
--- a/assets/completions/bat.fish
+++ b/assets/completions/bat.fish
@@ -24,7 +24,7 @@ function __bat_autocomplete_languages --description "A helper function used by "
 				sub(/^ +/, "", ext); # Trim leading whitespace.
 				sub(/ +$/, "", ext); # Trim trailing whitespace.
 
-				if (ext != "") {
+				if ((ext != "") && (ext !~ /[A-Z].*/)) {
 					print ext"\t"lang
 				}
 			}
--- a/assets/completions/bat.fish
+++ b/assets/completions/bat.fish
@@ -1,6 +1,38 @@
 # Fish Shell Completions
-
 # Place or symlink to $XDG_CONFIG_HOME/fish/completions/bat.fish ($XDG_CONFIG_HOME is usually set to ~/.config)
+
+# Helper function:
+function __bat_autocomplete_languages --description "A helper function used by "(status filename)
+	bat --list-languages | awk '
+		NR == 1 {
+			dc = 0;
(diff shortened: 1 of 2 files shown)


sharkdp:
Awesome stuff :-)

While I really your `awk` magic here, is there anything that we could do "inside" bat that would help with writing completion code like this? Maybe a (hidden) command-line option or subcommand that would print information like the supported languages or the list of themes in an easily parse-able way?

eth-p:
One option would be to have `bat --list-languages` use the following format without line wrapping when printing to a non-terminal stdout:

```
language <tab> extension extension extension ...
```

This would be a lot easier to parse with awk:

```awk
BEGIN {
    FS = "\t";
}

{
    name = $1;
    split($2, exts, ' ')
    for (i in exts) {
        print exts[i]"\t"name; # For each extension, print a completion with the name of the language.
    }
}
```

sharkdp:
I'd be open to this change. I would probably use
```
language:ext1,ext2,ext3
```
as I am not a big fan of tab characters. I guess that should also be easy to parse.

That being said, I'm also okay with merging this PR as is - as long as we don't have the required change to `--list-languages`.

eth-p:
I added the change to use the `language:ext1,ext2,ext` format for non-interactive printing.

To detect non-interactive printing, I decided to piggyback off of `config.loop_through`. Doing so required less changes, and it's behaviour would be more consistent with how `bat` works normally.

sharkdp:
Thank you very much!

> To detect non-interactive printing, I decided to piggyback off of `config.loop_through`. Doing so required less changes, and it's behaviour would be more consistent with how `bat` works normally.

Yes, it's not perfect ("loop through" doesn't really make sense for `--list-languages`) but I guess it's fine here.

Issue #535, opened 2019-03-30
Document --terminal-width=0 vs --wrap=never


luckydonald:
I tried having it to not wrap.
But `--terminal-width=0` results in a never ending loop of empty lines.
I later learned that `--wrap=never` was there, but maybe `--terminal-width=0` could imply no wrapping.

In any way, I'd recommend including a mention of `--wrap` and `--terminal-width` into the comment of each others at the help output.

```diff
  --wrap <mode>                 
              Specify the text-wrapping mode (*auto*, never, character).
+             See '--terminal-width' to set when to wrap.
  --terminal-width <width>      
              Explicitly set the width of the terminal instead of determining it automatically. If
              prefixed with '+' or '-', the value will be treated as an offset to the actual terminal
-             width.
+             width. This controls when text should wrap, see  '--wrap'.
```

sharkdp:
Thank you for reporting this.

The endless-loop should definitely be fixed.

What would `--terminal-width=0` imply for decorations such as the grid? Does it really make sense to allow for width 0 or should we disallow it and improve the documentation, as you suggested?

luckydonald:
Yeah probably adding references in the `--help` command and making it burp with a "did you mean" would be a cleaner solution.

sharkdp:
Closed via #559 by @eth-p.

luckydonald:
Does it refer to `wrap=never` now?

eth-p:
The documentation still needs to be updated, but it now makes `--terminal-width=0` an explicit error now.

luckydonald:
Can we keep this open please, until we get the documentation done?

PR #559, opened 2019-05-09
Prevent `--terminal-width=0` and invalid forms of `--terminal-width`.

eth-p:
My proposed solution to #535, which involves validating `--terminal-width` to report an error if somebody provides a width of zero.

## Why this approach, and not having it disable wrapping?
User convenience is definitely something to consider, but in this instance I believe having `--terminal-width=0` being analogous to `--wrap=never` would be more of a quirky behaviour than a quality-of-life feature.

1. Having such a feature would dilute the meaning of `--terminal-width`.
2. Using `--terminal-width=0` as `--wrap=never` isn't very clear or intuitive.
3. Using `--terminal-width=0` as a way to disable wrapping results in a conflict when you don't want wrapping, but do want to specify a specific terminal width.
4. It's more explicit (and less characters) to type `--wrap=never`.

### DIFF
--- a/src/clap_app.rs
+++ b/src/clap_app.rs
@@ -331,8 +331,8 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
                 .hidden_short_help(true)
                 .validator(
                     |t| {
-                        let sign = t.chars().next().unwrap();
-                        let num = if (sign == '+') || (sign == '-') {
+                        let is_offset = t.starts_with('+') || t.starts_with('-');
+                        let num = if is_offset {
                             t.chars().skip(1).collect()
                         } else {
                             t
@@ -340,7 +340,7 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
 
                         num.parse::<u32>()
                             .map_err(|_e| "must be an offset or number")
-                            .and_then(|v| if v == 0 && sign != '+' && sign != '-' {
+                            .and_then(|v| if v == 0 && !is_offset {
                                 Err("terminal width cannot be zero".into())
                             } else {
                                 Ok(())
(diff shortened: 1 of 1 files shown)


src/clap_app.rs
```diff
@@ -329,6 +329,24 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
                 .takes_value(true)
                 .value_name("width")
                 .hidden_short_help(true)
+                .validator(
+                    |t| {
+                        let sign = t.chars().next().unwrap();
```
sharkdp: Will this panic if the `t` is an empty string? Or will `clap` treat this case separately?

In any case, I think it would be a bit more clean if we used something like
``` rust
let is_offset = t.starts_with('+') || t.starts_with('-');
```
instead of checking for both `+` and `-` twice in this function.

src/clap_app.rs
```diff
@@ -329,6 +329,24 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
...
+                        let num = if (sign == '+') || (sign == '-') {
```
sharkdp: Instead if constructing `num`, we could also try to `.parse` into a `i32` and then fail if the result is zero and `is_offset` is false.

sharkdp:
Thank you for your contribution and the detailed description. Fully agree with your analysis. I've added two minor inline comments.

src/clap_app.rs
eth-p: Part of the validation I implemented is checking that `z123` isn't considered a valid terminal width. Would `.parse::<i32>()` be able to handle a positive sign character?

If so, that should be able to work.

src/clap_app.rs
eth-p: Done.

sharkdp:
Thank you for the update!

Issue #552, opened 2019-04-29
Add a short form for --paging=never.


rusty-snake:
It's often useful to switch between paging on and paging off without having to typing `--paging=WHEN` every time. A short form (e.g. `-c`) would be useful for this. Alternatively, you could use `-p` as aliase for `--style=plain` and `--paging=never`.

sharkdp:
Fixed via #560 by @eth-p

sharkdp:
Released in [v0.11.0](https://github.com/sharkdp/bat/releases/tag/v0.11.0).

PR #560, opened 2019-05-09
Add `-pp` for plain styling and no pager

eth-p:
A potential solution to #552 without breaking backwards compatibility or `cat` compatibility.

This introduces the `-pp` option, which does the following:

1. `--style=plain` (carried over from `-p`)
2. `--paging=never`


sharkdp:
Love this solution, Thank you!

PR #561, opened 2019-05-09
Improved logic when encountering ANSI/VT100 in character wrap mode

### LINKED ISSUE #551: Box-drawing characters can mess up the output and the terminal
Box-drawing characters can apparently break `bat`'s output, and they might not get displayed correctly. Additionally, it can be side-effecty and leave the terminal emulator / shell into a rather unreadable state, as seen in the screenshot.

Screen Shot 2019-04-25 at 22 20 39

Attached a text file for reproducability. It is a snippet from one of the output files of running [taurus](https://gettaurus.org/), a load testing tool.

[00_garble.txt](https://github.com/sharkdp/bat/files/3119796/00_garble.txt)

This was highly amusing to bump into 😄 but sadly perhaps more of a bug than a feature to keep around.

---
Context:
- MacOS Mojave 10.14.2 (18C54)
- iTerm2 Build 3.2.9
- zsh 5.3 (x86_64-apple-darwin18.0)

eth-p:
Fixes #551.

It turns out that I made a couple mistakes when initially implementing escape passthrough. This improved implementation is still making a few assumptions about the escape sequences (CSI/SGR sequences are safe to concatenate and use as a prefix, and all others are safe to pass through), but it should be a bit more reliable when fed any other sequences.

### DIFF
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -367,7 +367,8 @@ impl<'a> Printer for InteractivePrinter<'a> {
                     match chunk {
                         // ANSI escape passthrough.
                         (text, true) => {
-                            let is_ansi_csi = text.chars().skip(1).nth(0).map_or(false, |c|c == '[');
+                            let is_ansi_csi =
+                                text.chars().skip(1).nth(0).map_or(false, |c| c == '[');
 
                             if is_ansi_csi && text.chars().last().map_or(false, |c| c == 'm') {
                                 // It's an ANSI SGR sequence.
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -367,15 +367,25 @@ impl<'a> Printer for InteractivePrinter<'a> {
                     match chunk {
                         // ANSI escape passthrough.
                         (text, true) => {
-                            if text.chars().last().map_or(false, |c| c == 'm') {
+                            let is_ansi_csi = text.chars().skip(1).nth(0).map_or(false, |c|c == '[');
+
+                            if is_ansi_csi && text.chars().last().map_or(false, |c| c == 'm') {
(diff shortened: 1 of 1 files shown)


src/printer.rs
```diff
@@ -367,15 +367,26 @@ impl<'a> Printer for InteractivePrinter<'a> {
                     match chunk {
                         // ANSI escape passthrough.
                         (text, true) => {
-                            if text.chars().last().map_or(false, |c| c == 'm') {
+                            let is_ansi_csi =
+                                text.chars().skip(1).nth(0).map_or(false, |c| c == '[');
+
+                            if is_ansi_csi && text.chars().last().map_or(false, |c| c == 'm') {
```
sharkdp: Could this be `... && text.ends_with('m')`?

src/printer.rs
```diff
@@ -367,15 +367,26 @@ impl<'a> Printer for InteractivePrinter<'a> {
                     match chunk {
                         // ANSI escape passthrough.
                         (text, true) => {
-                            if text.chars().last().map_or(false, |c| c == 'm') {
+                            let is_ansi_csi =
+                                text.chars().skip(1).nth(0).map_or(false, |c| c == '[');
```
sharkdp: Could this be replaced by `… = text.starts_with("\x1B[")`?

sharkdp:
Thank you very much for looking into this! Seeing #551, I wasn't sure if this could actually be resolved.

src/printer.rs
eth-p: My bad, haha. I'm a bit rusty (pun intended) with the standard library functions.

Issue #569, opened 2019-05-24
Add "..." decoration when using multiple line ranges.


eth-p:
I'm making a shell script that uses `bat` to display the results of `ripgrep` searches, and adding "..." as a decoration between multiple line ranges would be a large quality of life improvement. As you can see in the image below, the lines kind of melt together.

![image](https://user-images.githubusercontent.com/32112321/58355932-be6c3700-7e2a-11e9-8c37-70e1fef9a229.png)

Would it be ok to add a flag that inserts an empty line between the ranges that displays "..." as the line number?

![image](https://user-images.githubusercontent.com/32112321/58356217-a648e780-7e2b-11e9-8e02-422178168d26.png)

sharkdp:
Sounds like a great suggestion, thank you very much!

I am planning to take a detailed look at your PR soon.

bricewge:
@eth-p Will you be willing to share your script with us? I think people at #448 will be interested in it.

eth-p:
@bricewge I haven't tested it on Linux yet (only MacOS), but I definitely can. I made a repo for my script [here](https://github.com/eth-p/bat-extras). As I find other ways to integrate bat with other tools, I'll probably add more scripts there in the future, too.

sharkdp:
closed via #570

sharkdp:
Released in [bat v0.12](https://github.com/sharkdp/bat/releases/tag/v0.12.0).

PR #570, opened 2019-05-25
Add new style component to separate multiple '--line-range's

eth-p:
My solution to my own issue, #569.

## Reasoning

When specifying multiple line ranges with `--line-range=... --line-range=...`, the printed lines tend to blend together and make it difficult to determine the scope of each range.

Example which makes parts of three functions look as though they're a single function:

image

## Implementation

The "snip" separator is implemented as another style component much like header and grid. I chose to include it in the `full` style by default, since doing otherwise would no longer make it "full".

When not using the `full` style, it can be added as `snip`.

## Preview

image

### DIFF
--- a/src/clap_app.rs
+++ b/src/clap_app.rs
@@ -110,7 +110,7 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
                 .validator(|val| {
                     let mut invalid_vals = val.split(',').filter(|style| {
                         !&[
-                            "auto", "full", "plain", "changes", "header", "grid", "numbers",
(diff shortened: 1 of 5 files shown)


sharkdp:
I'm sorry. I completely forgot about this PR. This looks really cool and I would like to merge this, but I didn't have time to look into the changes in detail.

eth-p:
No worries! Take your time, the PR isn't going to disappear anywhere.

sharkdp:
(rebased to fix the merge conflicts)

I finally had a closer look at this. This is awesome :+1:

PR #579, opened 2019-05-31
Print a horizontal header line when 'grid' style without 'header' style

### LINKED ISSUE #571: Output looks unbalanced when using '--style=grid,numbers' without 'header'.
The output looks a bit unbalanced when it's only printing the bottom horizontal line.

image

Would it make more sense to only print that final horizontal line when the `header` component is enabled?

eth-p:
Fixes #571


sharkdp:
Thank you! (sorry for not commenting earlier) Happy to merge this if the tests pass.

eth-p:
It turns out that I forgot to update the snapshots when I added the top grid line.
They should be fixed now.

PR #581, opened 2019-06-07
Add GN syntax.

issmirnov:
GN is a meta-build system that is used extensively at Google.
- https://gn.googlesource.com/gn/

This commit adds https://github.com/eseidelGoogle/gn_sublime with
the appropriate hand-converted GN.sublime-syntax file included.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -141,3 +141,6 @@
 [submodule "assets/syntaxes/ssh-config"]
 	path = assets/syntaxes/ssh-config
 	url = https://github.com/robballou/sublimetext-sshconfig.git
+[submodule "assets/syntaxes/GN"]
+	path = assets/syntaxes/GN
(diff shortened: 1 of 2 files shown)


sharkdp:

Correct me if I'm wrong, but this seems like a rather niche file format. The corresponding sublime package has [600 installs in total](https://packagecontrol.io/packages/GN) (compare with [11K installs](https://packagecontrol.io/packages/SSH%20Config) for the SSH config syntax, which is already rather peculiar). I know, this is probably not the best metric, but it's something I've been using as a rough estimate in the past.

issmirnov:
Totally understood. Since bat offers a way to provide custom syntax files locally, I'm more than happy to keep this in my dotfiles. We can close this one out - if people search for it later they'll have a place to comment.

Thanks for the sound reasoning :)

sharkdp:
>  if people search for it later they'll have a place to comment.

:+1:

PR #582, opened 2019-06-07
Add SSH config syntax

issmirnov:
Tested:
```
bash assets/create.sh
cargo build
target/debug/bat --language ssh_config ~/.ssh/config
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -135,3 +135,6 @@
 [submodule "assets/syntaxes/DotENV"]
 	path = assets/syntaxes/DotENV
 	url = https://github.com/zaynali53/DotENV
+[submodule "assets/syntaxes/ssh-config"]
+	path = assets/syntaxes/ssh-config
+	url = https://github.com/robballou/sublimetext-sshconfig.git


sharkdp:
Thank you very much!

It would be great if we could somehow make this work automatically for `~/.ssh/config`. Obviously, we don't want to enable this for every file named `config`, but maybe there is a good way to add new functionality to `bat` to support this.

issmirnov:
Thanks! Happy to help.

Regarding detection of files, one idea I've been playing with is adding a `first_line_match ` directive ([see here](https://github.com/issmirnov/dotfiles/blob/master/bat/syntaxes/SSH-Config.sublime-syntax#L9)). This opens a few possibilities.

1. Users annotate files manually. It's fine for SSH config files since generally there's only a few of them on each system. First line should be `# -*- mode: ssh-config -*-`
2. The first line match is expanded to guess if it's an SSH conf file. Ie, `^IdentityFile` or `^Host\s(\w*)` match, or some other smart regexes.

If we go that route though, that will likely require a patch to the sublime-syntax file, just like the Java one currently.

Thoughts?

sharkdp:
> Regarding detection of files, one idea I've been playing with is adding a `first_line_match ` directive ([see here](https://github.com/issmirnov/dotfiles/blob/master/bat/syntaxes/SSH-Config.sublime-syntax#L9)).

That's definitely a good first step and something that we could implement before we have a more general solution.

I was thinking it would be even better if we could provide a "out of the box" solution for every user of `bat`. I believe we could achieve something like this if we would add a "Path => Syntax" mapping database somewhere within `bat`. This would work similar to the "File extension => Syntax" and "File name => Syntax" mappings already present in the Sublime Text syntax files.

For the SSH config, we could explicitly check if the (expanded) file path matches `$HOME/.ssh/config`.

> The first line match is expanded to guess if it's an SSH conf file. Ie, `^IdentityFile` or `^Host\s(\w*)` match, or some other smart regexes.

Hmm.. I think that could lead to false positive matches. I'd rather not follow that route.

> [...] will likely require a patch to the sublime-syntax file, just like the Java one currently.

Hm.. I'm not very happy with the hack I did for the Java syntax. I would try to avoid having more patched files like this.

To add the `first_line_match` directive, we could

* Fork the original repository and depend on that in `bat`.
* Think about adding a feature in `bat` which would add such a `# -*- mode: XXX -*-` first-line-match directive automatically for *every* syntax.

What do you think?

issmirnov:
Awesome suggestions!

> add a "Path => Syntax" mapping database

I think this is great. This would work for hosts, ssh config, and likely a whole number of files that are simply called `config` or `hosts`, etc.

> first-line-match directive automatically for every syntax.

If it's not too hard to implement, this would bring `bat` in line with vim/emacs as far as file type identification capabilities. In order to not restart the age old wars, probably best to support both `# -*- mode: XXX -*-` and `# vim:ft=XXX`. Full disclosure, not a emacs user so would be good to double check the syntax.

sharkdp:
I'm going to add open two new feature requests in separate issues. Thank you very much for your feedback!

sharkdp:
Released in [bat v0.12](https://github.com/sharkdp/bat/releases/tag/v0.12.0).

Issue #304, opened 2018-09-11
Moving to the end of 6 Mb file takes forever


mfrw:
I am not really sure if this is an issue or intended behaviour.
It took me forever to move to end of a 6Mb file.

Steps to reproduce:
- `$ seq 1 1000000 > test`
- `$ bat test`
Press `G` to move to end of the file.
This step took a long time. It seems `bat` was going through the file sequentially.

Apologies, if this is not the place to ask this question or this is a known bug/feature.

sharkdp:
Thank you for reporting this. Yes, this is the right place to ask questions like this :+1: 

Let's try to untangle this a little bit:

When you call `bat test`, `bat` runs a pager (presumably `less`) and pipes all its output (the *whole* file contents) to the pager. Moving to the end of the file is slow because `bat` takes some time to output the full file contents.

Next, we can do some benchmarks. I am going to use my [hyperfine](https://github.com/sharkdp/hyperfine) tool to perform the measurements.

First, let's compare `cat` and `bat` when both are printing to `/dev/null`. Note that this is **not the benchmark we are looking for**, because in the interactive use (with or without the pager), we need to print (part of) the output to the terminal. This can cost a significant amount of time, especially if ANSI escape sequences are involved. Also, `bat` will use a faster loop-through-mode if it detects a non-interactive terminal.
```
> hyperfine 'cat test' 'bat test'
Benchmark #1: cat test

  Time (mean ± σ):       2.1 ms ±   0.9 ms    [User: 1.2 ms, System: 1.7 ms]
 
  Range (min … max):     1.2 ms …   6.6 ms
 
  Warning: Command took less than 5 ms to complete. Results might be inaccurate.
 
Benchmark #2: bat test

  Time (mean ± σ):     422.1 ms ±   6.8 ms    [User: 207.3 ms, System: 214.4 ms]
 
  Range (min … max):   415.7 ms … 439.5 ms
 
Summary

  'cat test' ran
  199.43x faster than 'bat test'
```
So even with `bat`s loop-through mode, it is two orders of magnitude slower than `cat`. This is unfortunate, but not too surprising. We have never optimized for speed and `bat` is reading and printing files line-by-line instead of using a larger buffer. Still, this is definitely something we could work on.

Next, we can make `bat` even slower by enabling all components that would be printed if we were printing to an interactive terminal:
```
> hyperfine --warmup 3 'bat --style=full --decorations=always --color=always test'
[...]

PR #594, opened 2019-06-16
Improve SimplePrinter throughput

jonashaag:
bat performance is very slow for `loop_through` compared to native `cat`, which means I have to switch to `/bin/cat` for some tasks.

The reason it's so slow is that bat always reads line-by-line, even if the input is printed as-is (SimplePrinter). This PR addresses that problem by changing abstractions: Reading should be part of the Printers so we can use fast reading for `loop_through` workloads.

Note that the code probably has some weird stuff in it since I'n not very familiar with Rust and its ecosystem; also I didn't change the tests

### DIFF
--- a/src/controller.rs
+++ b/src/controller.rs
@@ -1,11 +1,10 @@
-use std::io::{self, Write};
+use std::io::{self};
 use std::path::Path;
 
 use crate::app::{Config, PagingMode};
 use crate::assets::HighlightingAssets;
 use crate::errors::*;
-use crate::inputfile::{InputFile, InputFileReader};
-use crate::line_range::{LineRanges, RangeCheckResult};
+use crate::inputfile::{InputFile};
 use crate::output::OutputType;
 use crate::printer::{InteractivePrinter, Printer, SimplePrinter};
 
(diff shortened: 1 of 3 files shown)


sharkdp:

I didn't have time to look at the changes in detail, but two comments upfront:

- All of the `--line-range` related unit tests are currently failing.
- Have you seen issue #304 (and the related PR #596, which only appeared just now)?

jonashaag:
For a ~4GiB binary file:

Build | Time required
--- | ---
`bat >/dev/null` without #596 | ~25s
`bat >/dev/null` with #596 | ~15s
`bat >/dev/null` with this patch | ~6s
`bat >/dev/null` with this patch and #596 | ~6s
`cat` | ~3s

Quick and dirty benchmarking (2 runs per build); bat builds are release builds

jonashaag:
Bump this!

sharkdp:
(As stated above,) there are four unit tests which are failing and (without spending a significant amount of time to fully understand the changes in this PR) it's currently hard for me to judge whether or not they can be fixed, given your approach.

I'd be more motivated to look at this in detail if all tests would pass first. But if you are sure that those tests can definitely be fixed, I can also try to take a look now, such that you don't have to spend more time on this before getting some feedback.

jonashaag:
If we can’t fix the unit tests - but rather have to write new one - would that be a show stopper?

sharkdp:
> If we can’t fix the unit tests - but rather have to write new ones - would that be a show stopper?

Unit tests often need to be adapted, but _"can't fix"_ and _"write new ones"_ sounds like a radical change in expected behavior. That's obviously something that would need to be discussed.

jonashaag:
Or a rewrite is necessary because of architectural changes. 

Anyways, at the moment I won’t put any more work into this since it solves my issue (I simply apply the patch to any new version). If anyone feels like continuing from this point please feel free to take over!

jonashaag:
Closing since this now has a shit ton of conflicts

PR #596, opened 2019-06-17
use buffered output by default and implement -u option for unbuffered output

georgmu:
This commit speeds up output by using buffering.

Currently, the downside is that output data is also buffered if there is no input data pending.


sharkdp:

As already discussed in #304, I will not accept changes in `bat`s "streaming" behavior.

For this reason, I'm going to close this PR - but I'm happy to reopen it, in case there should be any new developments / comments.

PR #610, opened 2019-07-21
Small improvements for Fedora instructions

ignatenkobrain:
Anything older than F29 is EOL now, so no need to mention it. Also in F29+ you can use standard `dnf install`.


sharkdp:
Thank you very much!

PR #620, opened 2019-08-02
use explicit dyn with Write to appease compiler

wk:
trait objects without an explicit dyn have been deprecated in 1.37


sharkdp:
Thank you very much!

(the build failure is not related to your PR)

PR #648, opened 2019-09-03
Add Jinja2 syntax

Martin819:
Add sublime-syntax file for a Jinja2

Example:
Screenshot from 2019-09-03 16-24-17

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -153,3 +153,7 @@
 [submodule "assets/syntaxes/Strace"]
 	path = assets/syntaxes/Strace
 	url = https://github.com/djuretic/SublimeStrace
+[submodule "assets/syntaxes/Jinja2"]
+	path = assets/syntaxes/Jinja2
+	url = https://github.com/Martin819/sublime-jinja2
+	branch = master


sharkdp:

Where does this file come from? If you did not create it yourself, we have to add an acknowledgement / LICENSE.

This is one of the reasons why we normally add new syntaxes via submodules, see https://github.com/sharkdp/bat/blob/master/assets/README.md (step 1). This also helps us when updating the syntax in the future.

Martin819:
Hi @sharkdp, the original file is from https://github.com/kudago/jinja2-tmbundle/blob/master/Syntaxes/Jinja2.tmLanguage which is to be installed as a package for a sublime text. I have converted the file to sublime-syntax and made some alterations so it's only based on this file, but not the same file. The need for converting the file is also a reason, why I haven't added it as a submodule.

sharkdp:
Thank you for the explanation.

In this case, please add it as a submodule anyway and shortly document the manual changes in the appropriate section in https://github.com/sharkdp/bat/blob/master/assets/README.md

Martin819:
@sharkdp  I have included forked repo with files already converted to sublime-syntax as a submodule.

sharkdp:
Thank you!

sharkdp:

PR #658, opened 2019-09-09
Add SaltStack SLS syntax

Martin819:
Adding new syntax highlighting for SaltStack states - SLS files.

Example:
Screenshot from 2019-09-09 15-27-47

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -156,4 +156,7 @@
 [submodule "assets/syntaxes/Jinja2"]
 	path = assets/syntaxes/Jinja2
 	url = https://github.com/Martin819/sublime-jinja2
+[submodule "assets/syntaxes/SLS"]
+	path = assets/syntaxes/SLS
+	url = https://github.com/saltstack/sublime-text
 	branch = master


sharkdp:

I've never heard of SaltStack before. It sounds like this syntax would be something very specific. Have you considered [adding this syntax to your local installation](https://github.com/sharkdp/bat#adding-new-syntaxes--language-definitions)?

Martin819:
Hi @sharkdp, I already have it my local installation and it works just fine and that's why I felt it would be great if bat would include it by default.
SaltStack is an open-source configuration management software, very similar to Ansible. It uses SLS files which are in fact conbination of YAML and Jinja2. The problem is that either YAML nor Jinja2 syntax highlighting works for these files. That's why SaltStack itself came with this syntax highlight which wraps aroung Jinja2 highlight and enables it to be used on YAML+Jinja2 files.

sharkdp:
@eth-p Your opinion?

Maybe we should add some policy to make objective decisions on similar PRs that add new syntaxes.

SublimeText package is here:
https://packagecontrol.io/packages/SaltStack-related%20syntax%20highlighting%20and%20snippets

eth-p:
@sharkdp We should definitely make some kind of objective policy for adding new syntaxes. The hard part is going to be figuring out how to gather metrics on the proposed language, though.

While SublimeText is a decent way of figuring out how many people use both Sublime and the language, I feel like it could be missing out on a potentially significant portion of users who don't actually use Sublime.

Some other metrics we could use:

- StackOverflow (survey, tag count, etc.)
- Atom plugins (download count)
- JetBrains language plugins (download count)

Additionally, what do you think would be a good "minimum" number of downloads for a language to be considered popular enough that it should be added?

sharkdp:
> Some other metrics we could use: [...]

That's certainly a good idea!

> Additionally, what do you think would be a good "minimum" number of downloads for a language to be considered popular enough that it should be added?

So far, my threshold was somewhere between 10k and 20k total downloads. But I guess I wasn't strictly following that.

eminence:
Another possibility might be to use the list of languages that github itself will syntax highlight:  https://github.com/github/linguist/blob/master/vendor/README.md

Martin819:
Hi, regarding the questions on the need for this language, there are some results:

- over 2000 questions with tags "salt-stack" or "salt" --> [LINK](https://stackoverflow.com/questions/tagged/salt-stack+or+salt)

- Atom plugin available --> [LINK](https://atom.io/packages/language-salt)

- Not available for Jetbrains, but that's not a surprise since Salt is a cloud management service, you edit the files in 99% cases in CLI and not in the UI based IDEs.

- GitHub syntax is available --> [LINK(from README file from previous post)](https://github.com/saltstack/atom-salt)

sharkdp:
@Martin819 Thank you for the detailed update! Let's merge this for now.

If we have a problem with too many syntax being bundled with `bat` in the future, I guess it's not too bad to remove some of them again - even though it is a regression of some sort.

sharkdp:
(nightly builds are currently broken, but this has nothing to do with this PR)

sharkdp:

PR #664, opened 2019-09-19
Add references to scripts in `bat-extras`

### LINKED ISSUE #655: Add reference to "bat-extras" scripts to the README
https://github.com/eth-p/bat-extras

@eth-p What do you think? Should we add a reference to your script collection to the README?

eth-p:
Fixes #655.

Rather than referring to `bat-extras` directly, I added the supported integrations to the relevant section in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,14 @@ If you happen to use [`fd`](https://github.com/sharkdp/fd), you can use the `-X`
 fd … -X bat
 ```
 
+#### `ripgrep`
+
+With [`batgrep`](https://github.com/eth-p/bat-extras/blob/master/doc/batgrep.md), `bat` can be used as the printer for [`ripgrep`](https://github.com/BurntSushi/ripgrep) search results.
+
+```bash
+batgrep Printer src/
+```
+
 #### `tail -f`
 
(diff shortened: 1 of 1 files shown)


sharkdp:
> Rather than referring to `bat-extras` directly, I added the supported integrations to the relevant section in the README.

Great idea, thank you!

I'll maybe reverse the order of the "manual `man` settings" with `batman`, but the rest looks perfect.

README.md
```diff
@@ -105,6 +105,14 @@ If you happen to use [`fd`](https://github.com/sharkdp/fd), you can use the `-X`
 fd … -X bat
 ```
 
+#### `ripgrep`
+
+With [`batgrep`](https://github.com/eth-p/bat-extras/blob/master/doc/batgrep.md), `bat` can be used as the printer for [`ripgrep`](https://github.com/BurntSushi/ripgrep) search results.
+
+```bash
+batgrep Printer src/
```
sharkdp: This might be slightly confusing, as you mention "printer" above. Maybe `batgrep needle src/` or `batgrep pattern [<path>]`?

eth-p: Hmm... On a second look, that is confusing. The latter would probably be the better example

PR #673, opened 2019-09-27
Parameterized documentation.

### LINKED ISSUE #659: Parameterize bat name, documentation.
In order to make life easier on package maintainers, we should give them the ability to easily rename bat and its documentation when building a package.

- [x] Rename the man page
- [x] Parameterize the executable name in the man page
- [x] Rename the completions
- [x] Parameterize the executable name in the completions
- ~~Rename the executable~~
  This is too hacky with cargo.
  Just use an external shell script.

---

This issue is derived from https://github.com/sharkdp/bat/issues/656

eth-p:
Fixes #659.

Although it's slightly overkill, this uses Liquid for templating.
It's a well-tested templating engine, and it shouldn't break anytime soon.

## Changes
- The manual page has been moved from `doc/bat.1` to `assets/manual/bat.1.in`
- The fish completion file has been moved from `assets/completions/bat.fish` to `assets/completions/bat.fish.in`
- The generated files have been added to the `.gitignore`.
  Keeping them in the repo is redundant, and it will discourage people from copying them directly from GitHub.

## Information
Certain information is now parameterized using environment variables:

- `PROJECT_NAME` -- Set this to change the project name (default: `bat`).
- `PROJECT_EXECUTABLE` -- Set this to change the executable name (in documentation only; default: value of `PROJECT_NAME`).

## Caveats
The executable name can't be changed from an environment variable.
For this reason, I'm also keeping the generated files as `bat.*`.

If any of the files need to be renamed when packaging, it should be done in an external packaging script.


build.rs
```diff
@@ -1,7 +1,69 @@
...
+#[macro_use]
+extern crate lazy_static;
+extern crate liquid;
+
+use std::error::Error;
+use std::fs;
+
+// Read environment variables.
+lazy_static! {
+    pub static ref PROJECT_NAME: &'static str = option_env!("PROJECT_NAME").unwrap_or("bat");
```
sharkdp: Cool! I didn't know `option_env!(…)`.

Issue #661, opened 2019-09-11
create.sh not working


LunarLambda:
I was in the process of making a PR for fixing a tiny issue with the `manpage` syntax

However, create.sh fails catastrophically on my system:

```
Clearing theme set cache ... okay
Clearing syntax set cache ... okay
patching file /home/luna/projects/rust/bat/assets/syntaxes/Packages/Java/JavaDoc.sublime-syntax
Hunk #1 FAILED at 13.
1 out of 1 hunk FAILED -- saving rejects to file /home/luna/projects/rust/bat/assets/syntaxes/Packages/Java/JavaDoc.sublime-syntax.rej
patch: **** Can't reopen file /home/luna/projects/rust/bat/assets/syntaxes/Packages/Java/JavaDoc.sublime-syntax : No such file or directory
patch: **** Can't reopen file /home/luna/projects/rust/bat/assets/syntaxes/Packages/Java/JavaDoc.sublime-syntax : No such file or directory
patch: **** Can't reopen file /home/luna/projects/rust/bat/assets/syntaxes/Packages/Java/JavaDoc.sublime-syntax : No such file or directory
patch: **** Can't reopen file /home/luna/projects/rust/bat/assets/syntaxes/Packages/Java/JavaDoc.sublime-syntax : No such file or directory
patch: **** Can't reopen file /home/luna/projects/rust/bat/assets/syntaxes/Packages/Java/JavaDoc.sublime-syntax : No such file or directory

/* several thousand lines of output omitted */

patch: **** Can't create file /home/luna/projects/rust/bat/assets/syntaxes/Packages/Java/JavaDoc.sublime-syntax.orig : Too many open files
patch: **** Can't create file /home/luna/projects/rust/bat/assets/syntaxes/Packages/Java/JavaDoc.sublime-syntax.orig : Too many open files
patch: **** Can't create file /home/luna/projects/rust/bat/assets/syntaxes/Packages/Java/JavaDoc.sublime-syntax.orig : Too many open files
patch: **** Can't create file /home/luna/projects/rust/bat/assets/syntaxes/Packages/Java/JavaDoc.sublime-syntax.orig : Too many open files
patch: **** Can't create file /home/luna/projects/rust/bat/assets/syntaxes/Packages/Java/JavaDoc.sublime-syntax.orig : Too many open files
patch: **** Can't create file /home/luna/projects/rust/bat/assets/syntaxes/Packages/Java/JavaDoc.sublime-syntax.orig : Too many open files
[...]

PR #674, opened 2019-09-27
Ensure submodules are updated before creating asset cache

eth-p:
This fixes the reason why #661 was opened.

### DIFF
--- a/assets/create.sh
+++ b/assets/create.sh
@@ -3,6 +3,10 @@ set -euo pipefail
 
 ASSET_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
+# Ensure submodules are initialized.
+git submodule init
+git submodule update
+
 # Always remove the local cache to avoid any confusion
 bat cache --clear
 
--- a/assets/create.sh
+++ b/assets/create.sh
@@ -2,10 +2,42 @@
 set -euo pipefail
 
 ASSET_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+REPO_DIR="$ASSET_DIR/.."
 
 # Ensure submodules are initialized.
-git submodule init
-git submodule update
+function update_submodules() {
+	local submodule
+	local submodule_prompt=unspecified
+	local submodule_path
+
+	{
+		while { read -r submodule && read -r submodule_path; } <&3; do
+			if ! [[ -d "${REPO_DIR}/.git/modules/${submodule}" ]] && [[ -d "${REPO_DIR}/${submodule_path}" ]]; then
+				if [[ "$submodule_prompt" = "unspecified" ]]; then
+					echo "One or more submodules were found to be uninitialized."
+					printf "Initialize and update them? [Y/n] "
+					read -r submodule_prompt
+				fi
+
+				case "$submodule_prompt" in
+					y|yes|'') {
+						git -C "$REPO_DIR" submodule update --init "$submodule_path"
+					};;
+					n|no) {
+						return
+					};;
+					*) {
+						echo "Unknown answer. Not updating submodules."
+					};;
+				esac
+			fi
+		done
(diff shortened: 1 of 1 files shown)


assets/create.sh
```diff
@@ -3,6 +3,10 @@ set -euo pipefail
 
 ASSET_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
+# Ensure submodules are initialized.
+git submodule init
+git submodule update
```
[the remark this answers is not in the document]
sharkdp: We need `--recursive` if cloning the bat repository. I think we don't need it when updating the submodules. But it's better to make sure :+1:

assets/create.sh
```diff
@@ -3,6 +3,10 @@ set -euo pipefail
 
 ASSET_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
+# Ensure submodules are initialized.
+git submodule init
```
sharkdp: To make this a little bit more robust, you could use `git -C "$ASSET_DIR" …`. Otherwise, it only works when the CWD is within the `bat` repository (and not within one of the submodules).

eth-p:
Using `git submodule status` turned out to be kind of slow on my machine, so I ended up writing a hacky solution that checks for the existence of submodule data within the .git folder itself. This won't detect submodules that were intentionally deinitialized, but it does a decent job at quickly looking for (and downloading) submodules that were never cloned to begin with.

sharkdp:
Thank you for the update!

I like the interactive prompt, but I would prefer a simpler (and possibly slower) version as compared to the "hacky" one (even though it's not that bad). But I'm also fine with your solution. It's definitely better than not having a check at all. 

It works great on Linux.

assets/create.sh
```diff
@@ -2,6 +2,43 @@
...
+					*) {
+						echo "Unknown answer. Not updating submodules."
+					};;
+				esac
+			fi
+		done
+	} 3< <(git config --file "${REPO_DIR}/.gitmodules" --null --get-regexp path | xargs -0 printf "%s\n" | sed 's/^submodule.//;s/.path$//')
+}
+
+if [ -t 0 ]; then
```
sharkdp: what's this for?

assets/create.sh
```diff
@@ -2,6 +2,43 @@
 set -euo pipefail
 
 ASSET_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+REPO_DIR="$ASSET_DIR/.."
+
+# Ensure submodules are initialized.
+# This might look like black magic, but it works.
```
sharkdp: I think we can remove this line.

assets/create.sh
eth-p: The case statement, the process substitution, or the test for an interactive terminal?

assets/create.sh
eth-p: Oops. I forgot to remove that before committing. Yeah, I'll fix that.

eth-p:
> I like the interactive prompt, but I would prefer a simpler (and possibly slower) version as compared to the "hacky" one (even though it's not that bad). But I'm also fine with your solution. It's definitely better than not having a check at all.

I tried parsing `git submodule status` and 'git submodule foreach', but the former slows down with every initialized submodule, and the latter doesn't work for uninitialized submodules. 

Delving into the source code for [git submodule status](https://github.com/git/git/blob/master/builtin/submodule--helper.c#L775-798), and [submodule.c](https://github.com/git/git/blob/0aae918dd929862d3ce0ea2960897787bb269a3b/submodule.c#L239), it appears to check if a submodule is active based on a config key. That still doesn't help in the case where submodules are initialized and not updated, though.

For the time-being and until I can figure out a better way to interact with submodules, I think this solution is the most user-friendly and efficient way to do things.

assets/create.sh
sharkdp: Ah, nice. I didn't know that.

PR #678, opened 2019-10-04
cmake and libz are not needed to install bat

### LINKED ISSUE #649: cmake no longer build dependency
At least as of bat 0.12.0 (quite possibly sooner) `cmake` is no longer a build dependency of `bat`. The Nix package [updated](https://github.com/NixOS/nixpkgs/pull/67868) its derivation to require `libclang` but to remove `cmake`, and it builds just fine this way. The README should be updated accordingly.

rogeliodh:
Closes #649


sharkdp:
Thank you very much!

Issue #668, opened 2019-09-21
Separate system and user config files


clarfonthey:
I've been trying to export most of my dotfile-type configs to system packages I can install. A lot of programs offer the option for both a system and a user config (e.g. git reads from `/etc/gitconfig`, then `~/.gitconfig`) although currently, bat reads just one file. It'd be nice if we could do something similar for bat as well, loading both `/etc/bat/config` and `~/.config/bat/config`.

sharkdp:
Sounds like a good idea to me! Thank you for the feedback.

eth-p:
Adding a quick note for #659: This should probably be configurable by the package maintainer.

goku321:
I am very new to Rust, like very very new new. Can I try my hands on this if that's okay?

sharkdp:
Sure! Let us know if you need help

goku321:
Hi @sharkdp so, this how I am thinking of approaching this: I would just concatenate config read from `/etc/bat/config` and `~/.config/bat/config` together

sharkdp:
@goku321 Yes, exactly. Maybe we should directly implement it with a list of possible paths to read from, in case there will be more (maybe OS-specific) paths in the future.

etiennemabille:
> Hi @sharkdp so, this how I am thinking of approaching this: I would just concatenate config read from `/etc/bat/config` and `~/.config/bat/config` together

You could use `bat` to concatenate those files, right ? 😉

morgsmccauley:
Is this issue still active? I can see #691 which was closed a while back. I'm keen to pick this up :)

sharkdp:
This is still active, yes.

morgsmccauley:
Should `/etc/bat/config` be static or should we expose an env var to override that also?

sharkdp:
I would try to avoid introducing more environment variables, if possible. What do other CLI applications do? Is `/etc/bat/config` a reasonable place for a system-wide `bat` config?

Teehmon684:
Y'all Crazy 🤣🤣🤣

j0hnmeow:
> Y'all Crazy 🤣🤣🤣

Off-topic.

TurboCartPig:
So out of curiosity, what would the system config paths be on the different platforms? Are there some standards akin to xdg-user-dirs?

Enselic:
> So out of curiosity, what would the system config paths be on the different platforms? Are there any standard conventions?
> 
> I think these might be reasonable:
> 
[...]

PR #691, opened 2019-10-15
Read bat config from multiple config files

goku321:
Summary
- Added logic to read bat config from multiple config files.

Fixes: #668


src/bin/bat/config.rs
```diff
@@ -16,8 +16,22 @@ pub fn config_file() -> PathBuf {
 }
 
 pub fn get_args_from_config_file() -> Result<Vec<OsString>, shell_words::ParseError> {
+    let mut config_file_path = vec![PathBuf::from("/etc/bat/config")];
+    config_file_path.push(default_config_path());
```
sharkdp: so how does that work?

`/etc/bat/config` is always included but `~/.config/bat/config` might be overwritten by `BAT_CONFIG_PATH`? Is that a good idea?

src/bin/bat/config.rs
```diff
@@ -16,8 +16,22 @@ pub fn config_file() -> PathBuf {
...
+
+    let mut config_args = vec![];
+
+    for path in config_file_path.iter() {
+        let args = read_args_from_config_file(path.to_path_buf());
```
sharkdp: I think you could avoid the `.to_path_buf()` if `read_args_from_config_file` simply takes a `&Path`.

src/bin/bat/config.rs
```diff
@@ -16,8 +16,22 @@ pub fn config_file() -> PathBuf {
...
+        config_args.append(&mut args.unwrap());
+    }
```
sharkdp: It's not necessary, but the whole loop could probably be written in functional style with `.map(…)` and `.collect()`. This would also avoid the `mut` on `config_args`.

Something like (not tested and most probably wrong):
``` rust
let config_args = config_file_path
  .iter()
  .map(read_args_from_config_file)
  .transpose()?
  .flatten()
 .collect();
```

src/bin/bat/config.rs
```diff
@@ -16,8 +16,22 @@ pub fn config_file() -> PathBuf {
 }
 
 pub fn get_args_from_config_file() -> Result<Vec<OsString>, shell_words::ParseError> {
+    let mut config_file_path = vec![PathBuf::from("/etc/bat/config")];
```
sharkdp: `config_file_path` => `config_file_paths`?

src/bin/bat/config.rs
goku321: @sharkdp Thanks for the review. I'm thinking of moving the logic for getting path from env var `BAT_CONFIG_PATH` to a separate function (something like `get_config_file_path_from_env` and `default_config_path()` will just return `~/.config/bat/config`.

sharkdp:
Going to close this for now, as there is no further feedback.

PR #679, opened 2019-10-06
Initial library support

DrSensor:
> Fix #423

##### Project Structure
Cargo has a convention where [other executables can be placed in src/bin/*.rs][pkg-layout]. This PR use that convention which move some `src/*.rs` that depend on `app.rs`, `clap_app.rs`, _or_ clap crate into `src/bin/bat`. Summary, the src directory became:

```console
$ tree src
src
├── assets.rs
├── bin
│   └── bat
│       ├── app.rs
│       ├── clap_app.rs
│       ├── config.rs
│       └── main.rs
├── controller.rs
├── decorations.rs
├── diff.rs
├── dirs.rs
├── inputfile.rs
├── lib.rs
├── line_range.rs
├── output.rs
├── preprocessor.rs
├── printer.rs
├── style.rs
├── syntax_mapping.rs
├── terminal.rs
└── util.rs

2 directories, 19 files
```

[pkg-layout]: https://doc.rust-lang.org/cargo/guide/project-layout.html

##### Build and run
```sh
cargo build --bins           # to produce binary executable
cargo run                    # just works fine
cargo doc --no-deps --open   # to build and open generated library docs
```

PR #693, opened 2019-10-20
Library updates

sharkdp:
work in progress... (a follow up on #679)

**To do** before we can release `bat`-as-a-library:

- [ ] Move the `dirs` module to the binary. This involves a restructuring of the `assets` module. For the library, it would first be enough to only provide the cached assets (`HighlightingAssets::new()`).
- [ ] Possibly move the error-handling to the binary (the `handle_error` function). We don't want the library to print `[bat error]: …`. Instead, we could pass a callback to the `Controller`.
- [ ] The public API interface is huge at the moment. We should try to minimize this by making things private. Otherwise, we will have lots of (technically) API-breaking changes in the future.


sharkdp:
@DrSensor I would actually like to make the `printer` and the `output` module private (see reasons given above). This would render your `using_printer` example (here called `advanced`) not possible (for now). I currently don't really see a need for this, but please let me know if you think otherwise.

DrSensor:
As long as I have the flexibility on when to prints header,footer,snip,content and also detect the snip/line_range, then it should be fine. I need that for custom error messages.

PR #696, opened 2019-10-21
Add fstab syntax definitions

flopp:
This commit adds simple syntax definitions for `fstab` files.

### DIFF
--- a/assets/syntaxes/Fstab.sublime-syntax
+++ b/assets/syntaxes/Fstab.sublime-syntax
@@ -1,0 +1,22 @@
+%YAML 1.2
+---
+# http://www.sublimetext.com/docs/3/syntax.html
+name: fstab
+file_extensions:
+  - fstab
+scope: source.fstab
+
+contexts:
+  main:
+    - match: '^\s*\#.*'
+      scope: comment.line.number-sign
+    - match: ^\s*(\S+)\s+(\S+)\s+(\S+)\s+(\S+)\s+(0|1)\s+(0|1|2)\s*$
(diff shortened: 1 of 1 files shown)


sharkdp:
Awesome, thank you very much!

Works great for me:
image


Two questions:
- Since we "understand" which field is which, it would be great if we could highlight the `<type>` and the `<options>` fields in different colors. Maybe even separate colors for the two filepaths/UUIDS. I know it's hard to choose appropriate scopes (http://www.sublimetext.com/docs/3/scope_naming.html), but it would be a great visual aid to parse the file quickly. As a bonus, we could even parse the comma-separated options-string and highlight known keywords.
- I think we could also match on `mtab` (`/etc/mtab`), as it seems to have the same syntax. Not completely sure on that, though.

eth-p:
I agree with @sharkdp. It would be a great visual aid to have different colors for `<type>` and for each field in `<options>`.

eth-p:
@flopp @sharkdp: How about this for colors? I tried to base it off of the scheme in #698.

image

I forked @flopp's branch and edited the syntax file in https://github.com/eth-p/bat/commit/f7ede0d14f01bd71657d436cc785c7aed26327a4 to come up with that. If you like it, you can merge it back into this PR.

sharkdp:
@eth-p looks great.

flopp:
@eth-p good job. I like it!

eth-p:
Great! I'll merge this PR and open a new one for the added colors.

sharkdp:

PR #703, opened 2019-10-25
Add syntax highlighting for /etc/resolv.conf

argentite:
Added simple keyword highlighting for /etc/resolv.conf
Screenshot_2019-10-25_18-45-23


sharkdp:
Thank you very much!

PR #706, opened 2019-10-28
Add more colors to fstab syntax

eth-p:


sharkdp:
Works (and looks) great, thank you.

assets/syntaxes/Fstab.sublime-syntax
```diff
@@ -8,18 +8,126 @@ scope: source.fstab
...
+      set: fstab_pass
+    - match: '\s*[^\s]+\s*'
+      comment: invalid dump field
+      scope: invalid.illegal
+      set: fstab_pass
+    - include: clear
+
+  fstab_pass:
+    - include: comment
+    - match: '\s*[01]\s*'
```
sharkdp: only noticed this after merging, but a `2` also seems to be a valid value here.

eth-p: Whoops. Copy paste got me there. Do you want me to make a PR to fix that, or should one of us just fix and commit directly to master?

sharkdp: No worries :smile:. I think I disabled direct pushes to `master`, so we need a PR.

PR #707, opened 2019-10-29
Add delta to related projects

dandavison:
@sharkdp This PR adds my project  [delta](https://github.com/dandavison/delta) to the list of related projects.

Delta is related to bat in that it honors `BAT_PAGER`, and it uses exactly the same mechanism of including binary syntax and theme definition files (In fact, the delta README instructs users to use bat to create new versions of these files), and I plan to add support for reading custom syntax/theme files from `~/.cache/bat` (https://github.com/dandavison/delta/issues/19). Indeed, delta copies some of your code to do this! I plan to investigate whether I can avoid some of the copying by using bat's new library architecture.

Also, when I first heard of bat and that it had git integration, one thing I wondered was whether bat would syntax-highlight diffs sections according to their language. So it may also make sense to list delta as a related project on the basis that others may have the same thought.

### DIFF
--- a/doc/alternatives.md
+++ b/doc/alternatives.md
@@ -20,5 +20,6 @@ If you think that some entries in this table are outdated or wrong, please open 
 request.
 
 Some other alternatives that are also related, but not yet included in the table:
+- [delta](https://github.com/dandavison/delta)
 - [lesspipe](https://github.com/wofr06/lesspipe)
 - [vimpager](https://github.com/rkitover/vimpager)


sharkdp:
Hi @dandavison 

I've been using `delta` for a few days now and I really like it! I'm happy to add a reference to your project somewhere, but I'm not sure if this is the best place? Is `delta` really an "alternative to `bat`"? I would rather see it as a complement.

sharkdp:
> I plan to investigate whether I can avoid some of the copying by using bat's new library architecture.

I think that would be a good idea, something that both projects will profit from. Note that the `bat`-as-a-library story is not completely finished (see #693).

Also, concerning `delta`s core functionality, note that there is a prominent feature request (#23) here which is somehow *related*.

dandavison:
> Is delta really an "alternative to bat"? I would rather see it as a complement.

Thanks, yes, you're right, it's definitely a complement not an alternative. I wasn't familiar with `vimpager` and `lesspipe` but now I see that they fit into the "alternative" camp, rather than merely "related". So let's close this PR.

> Also, concerning deltas core functionality, note that there is a prominent feature request (#23) here which is somehow related (but definitely not a replacement for delta).

Ah that's interesting, I'd overlooked that. I'll comment in that thread.

sharkdp:
if you like, you could add a new subsection to the "integration with other tools" section: https://github.com/sharkdp/bat#integration-with-other-tools

something like: 

## as a `git` pager

You can use `bat` as a `git` pager, by adding …

Note that syntax highlighting *within* diffs is currently not supported. If you are looking for this, check out [`delta`](...).

PR #712, opened 2019-11-01
Split long long lines into slices.

### LINKED ISSUE #636: files of infinite length cause bat to crash
```
bat /dev/zero
memory allocation of 34359738368 bytes failed
```

treere:
this should fix #636 

Now lines are read using the buffer of BufRead. 
I added an enum to set if the current line is an entire line or a subpart.

(If a line is longer than 1024*1024 the highlight does not work as expected)

### DIFF
--- a/src/inputfile.rs
+++ b/src/inputfile.rs
@@ -1,52 +1,159 @@
 use std::fs::File;
-use std::io::{self, BufRead, BufReader};
+use std::io::{self, BufRead, BufReader, ErrorKind};
 
 use content_inspector::{self, ContentType};
 
 use crate::errors::*;
 
 const THEME_PREVIEW_FILE: &[u8] = include_bytes!("../assets/theme_preview.rs");
+const MAXIMUM_LINE_LENGTH: usize = 1024 * 1024;
+
+/// This function is like the read until present in the standard library
+/// but with a limit in the read buffer size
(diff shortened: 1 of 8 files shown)


Enselic:
Labeling as **needs-work** since there are quite big conflicts that needs to be sorted out at the moment. I took a shot at it, but it was too tricky for a quick rebase.

src/inputfile.rs
```diff
@@ -1,52 +1,159 @@
...
+use std::io::{self, BufRead, BufReader, ErrorKind};
 
 use content_inspector::{self, ContentType};
 
 use crate::errors::*;
 
 const THEME_PREVIEW_FILE: &[u8] = include_bytes!("../assets/theme_preview.rs");
+const MAXIMUM_LINE_LENGTH: usize = 1024 * 1024;
+
+/// This function is like the read until present in the standard library
```
Enselic, line 11: If someone picks this PR up: There is no need to duplicate this function. We can use `take()` instead. See https://github.com/sharkdp/bat/pull/1972/files#r763022008

Enselic:
I'm closing this now to keep our PR inbox clean and easy to get an overview of. I explicitly mentioned in https://github.com/sharkdp/bat/issues/636 that anyone looking to fix this issue should also take a look at this PR.

Issue #592, opened 2019-06-11
Add a dedicated Path => Syntax lookup database


sharkdp:
Sublime text syntax definitions provide two different ways to "detect" specific languages:

* Via the [`file_extensions` field](http://www.sublimetext.com/docs/3/syntax.html#header) in the header of the syntax. This allows us (for example) to choose the "C++" syntax for all files with a `cpp` file extension. It also allows us to choose the "Makefile" syntax for all files called `Makefile`¹.
* Via the [`first_line_match` field](http://www.sublimetext.com/docs/3/syntax.html#header) in the header of the syntax. This allows us (for example) to choose the "Python" syntax for files that start with `#!/usr/bin/env python` - even without a `.py` extension.

It would be great if we could extend our "syntax detection" by allowing bat to choose a specific syntax for certain **file path** patterns. For example, it would be great if we could automatically use the "SSH Config" syntax for the file `$HOME/.ssh/config` (see #582). Or a (hypothetic) "X11 Config" syntax for all `/etc/X11/xorg.conf.d/*.conf` files. Or the "Shell" syntax for `/etc/profile`.

These examples would be rather hard to match with one of the above techniques as (1) the have very generic names and (2) they don't have a specific "first line pattern".

Before someone goes ahead and implements this, we should probably discuss both the general idea and the implementation details.

In any case, I'd be very glad to get some feedback on this idea.

¹ The `file_extensions` lookup does not actually distinguish between file-extension patterns and full-filename patterns. This means that a file called `cpp` will match the "C++" syntax and a file called `foo.Makefile` will match the "Makefile" syntax :frowning_face:

keith-hall:
re the footnote, it may me worth subscribing to this issue in case the situation will improve in future: https://github.com/SublimeTextIssues/Core/issues/1353

sharkdp:
> it may be worth subscribing to this issue in case the situation will improve in future: [SublimeTextIssues/Core#1353](https://github.com/SublimeTextIssues/Core/issues/1353)

Thank you for the reference!

gurgeous:
I love this idea and just today I was dreaming of a nicely highlighted `sshd_config` file... Could we use a regex match against the full file path?

gurgeous:
I might be able to attempt a PR if you give me some direction. I'm an experienced engineer, though this would be my first stab at rust... It's ok to say no. I've been an open source maintainer too :)

sharkdp:
[...]

PR #753, opened 2019-11-19
Highlight systemd units in .ini syntax

zdzichu:


sharkdp:
Thank you for your contribution. Unfortunately, I cannot merge this as is.

Most of these filenames are far too generic (`service`, `path`) to make a specific mapping to the INI syntax.

I think we should wait for #592 to be implemented. I'm not familar with `systemd`, but I guess we could then map certain path patterns to INI.

PR #764, opened 2019-12-06
Update Japanese README file

sorairolake:
The beginning part of untranslated was translated into Japanese text.


sharkdp:
Thank you!

Issue #601, opened 2019-06-28
Keep empty lines empty


jzinn:
It would be nice to be able to use `nl` to number only non-blank lines with this command:
```
bat -p --color=always foo.c | nl -bt
```
However, empty lines in `foo.c` have escape sequences after passing through `bat` so that `nl` adds line numbers to seemingly blank lines.

```
$ bat --version
bat 0.10.0
```

rusty-snake:
@jzinn IMHO it should work with `--color=never` or `--color-auto`.

eth-p:
If I understand correctly, you want to keep the syntax highlighting from `bat` while using `nl` for non-empty line numbering? I can confirm that it doesn't work as expected for that use case.

I'm not sure how @sharkdp would like to handle it, but I have a workaround with `sed` that you could use (probably as an alias) in the meantime:

**Bash:**
```bash
bat -pp --color=always [...] | sed $'s,^\(\x1b\\[[0-9;]*[a-zA-Z]\)*$,,' | nl -bt
```

**Fish:**
```fish
bat -pp --color=always [...] | sed (printf 's,^\(\x1b\\[[0-9;]*[a-zA-Z]\)*$,,') | nl -bt
```

sharkdp:
To be honest, I don't think that this is something that we have to support by all means. If there is an easy solution for this, ok.

jzinn:
It’s marginal for sure, but worth making a issue, I thought, in case it’s an easy change.

This came about trying to highlight example code from Unix Network Programming, which does not number blank lines.

sharkdp:
Closed in #766 by @mbarbar

sharkdp:
This has been fixed in [bat 0.13](https://github.com/sharkdp/bat/releases/tag/v0.13.0).

PR #766, opened 2019-12-07
Leave blank lines blank for terminal output.

mbarbar:
This fixes bug #601. Simply prevents empty strings passed to `as_terminal_escaped` from being escaped.

### DIFF
--- a/src/terminal.rs
+++ b/src/terminal.rs
@@ -27,6 +27,10 @@ pub fn as_terminal_escaped(
     italics: bool,
     background_color: Option<highlighting::Color>,
 ) -> String {
+    if text.is_empty() {
+        return text.to_string();
+    }
+
     let mut style = if !colored {
         Style::default()
     } else {


sharkdp:

That looks like a very simple fix :+1: Are we sure that this wouldn't cause any problems with unclosed ANSI escape sequences?

mbarbar:
From what I could see and understand of the code, no, because text which is passed in to `as_terminal_escaped` is always fully coloured. It will never return an unfinished colouring since it has no notion of syntax or state.

```python
x = """ 
line 1...

line 3.
line 4!
"""
```

In this piece of Python code, even though the multi-line string is a single syntax element, it will open and close the colour for each line.

The disadvantage I suppose with this example is that the third (empty) line which is semantically significant, since it's part of a string, will be uncoloured. I personally don't see that as an issue since `nl` doesn't (can't) give consideration to such a thing when uncoloured anyway.

sharkdp:
Thank you for the clarification.

> From what I could see and understand of the code, no, because text which is passed in to `as_terminal_escaped` is always fully coloured. It will never return an unfinished colouring since it has no notion of syntax or state.

:+1: 

> The disadvantage I suppose with this example is that the third (empty) line which is semantically significant, since it's part of a string, will be uncoloured.

Hm, yes. But printing a colorizing ANSI-escape-sequence and a ANSI-reset sequence directly afterwards is probably not too helpful. The line isn't really colorized. Just the empty string.

PR #784, opened 2019-12-21
Fix '日本文' to '日本語'

PicoSushi:
Hi.

`Japanese` is `日本語` in Japanese. And, `日本文` is `Japanese` in Chinese. This commit fix this wrong expression.


sharkdp:
Thank you.

Issue #758, opened 2019-11-27
Wrapping does not work when piping


desbma:
Line wrapping seems to be disabled when piping to another process, even when forced with `--wrap=character`.

```
$ echo 123456789abcd > /tmp/file
$ bat --style plain --terminal-width 10 --wrap=character /tmp/file
123456789a
bcd
$ bat --style plain --terminal-width 10 --wrap=character /tmp/file | command cat
123456789abcd
```

tzhouhc:
[This line](https://github.com/sharkdp/bat/blob/40e5befdc7c09edde6fe33d4dd79a9a2bf5d0762/src/bin/bat/app.rs#L132) checks and gives up on wrapping if the bat call is not interactive, but if we are providing an absolute width explicitly it should be fine...?

Allowing this would be immensely helpful with fzf preview integration.

sharkdp:
> but if we are providing an absolute width explicitly it should be fine...?

Yes, I think so. Sounds like an easy fix :+1:

fusillicode:
Hi :D,

me and @allevo are trying to tackle this issue and we actually pulled out this preliminary [PR](https://github.com/sharkdp/bat/pull/785).

To solve the issue however it seems that we must extend the `SimplePrinter` to actually consider the wrapping. Does it sounds right? 🤔

If yes how should we proceed? Open a new PR to address the actual `SimplePrinter` changes? 🤔

fusillicode:
Just a quick update, we managed to extend the `SimplePrinter` 😄 

[Here's](https://github.com/fusillicode/bat/pull/1/files) the separated PR.

Let us know how we can proceed 😄

sharkdp:
Maybe I was too fast with merging #785. Wrapping works just fine if we disable interactive mode by using `--color=always`, right?

```
# color and decorations:
bat --terminal-width 50 --wrap=character --color=always src/assets.rs | …

# color only
bat --terminal-width 50 --wrap=character --color=always --plain src/assets.rs | …

# no color
bat --terminal-width 50 --wrap=character --decorations=always src/assets.rs | …

# no color, no decorations
bat --terminal-width 50 --wrap=character --decorations=always --plain src/assets.rs | …
```

desbma:
> Wrapping works just fine if we disable interactive mode by using --color=always or --decorations=always, right?

No, it does not:

    $ echo 1234567890ABCD | bat --plain --color=always --wrap character --terminal-width 10 | command cat
    1234567890ABCD

[...]

PR #785, opened 2019-12-21
Add check for terminal-width in output_wrap config value in not interactive case

fusillicode:


fusillicode:
Thank you for all you work! 🙇 

Let us know what you think about the actual logic implementation 😄 (https://github.com/sharkdp/bat/issues/758#issuecomment-568168618)

Cheers! 🍻

PR #786, opened 2019-12-23
Do not pass '--no-init' on newer less versions

### LINKED ISSUE #749: bat leaves text behind in pager mode
version: bat 0.12.1

Bat leaves the text behind if it is started like
`bat somefile` or `bat somefile --pager='less'`

if i start it like 
`bat somefile --pager='less -r'`
it works fine and leaves nothing behind as less does.

sharkdp:
With this change, we do not pass the `--no-init` option in newer
versions of less (530 or higher).

This fixes #749

### DIFF
--- a/README.md
+++ b/README.md
@@ -435,18 +435,19 @@ export BAT_PAGER="less -RF"
 Instead of using environment variables, you can also use `bat`s [configuration file](https://github.com/sharkdp/bat#configuration-file) to configure the pager (`--pager` option).
 
 **Note**: By default, if the pager is set to `less` (and no command-line options are specified),
-`bat` will pass the following command line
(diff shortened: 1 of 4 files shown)


septatrix:
I think this does not work on current fedora (34) where less reports the following:

```text
less --version
less 581.2 (POSIX regular expressions)
Copyright (C) 1984-2021  Mark Nudelman

```

Should I create a new issue for that?

sharkdp:
Which `bat` version do you use? I think this has been fixed in v0.18.1. If you are using v0.18.1 or newer, please report it as a new bug. Thank you!

septatrix:
Okay, Fedora ships v0.17.1 but based on the release tag and the referenced issues that should resolve the issue. For now I just configured the flags in the config file and maybe I will write the fedora maintainer to update the package if he has some spare time.

PR #790, opened 2019-12-27
Adds Ledger syntax highlighting

colindean:


sharkdp:
Thank you for your contribution. I'd rather not include this by default. It doesn't seem to be a very "popular" syntax file (GitHub and packagecontrol metrics). Note that you can always add custom syntaxes (see README).

colindean:
No worries. I doubt folks would keep their Ledger files publicly visible in either of those systems since those files generally contain private financial information. I don't have good information on the number of Ledger users out there but the intersection of Ledger and bat users is probably small enough that it doesn't make sense to include it in the distribution for everyone.

In case anyone else comes along wanting to add Ledger to bat… I did up a little thing to help me manage any bat syntaxes or themes I'd add: https://github.com/colindean/hejmo/commit/ea7fc1db570386664e30d1be2e0573e459de77b9

sharkdp:
Very cool! @eth-p might be interested in this for `bat-extras`.

eth-p:
That's a pretty clever idea, @colindean. If anybody would be interested in me adding something like that to `bat-extras` (maybe it would be more similar to something `like oh-my-zsh` or `fisher`?), I would be happy to do that.

colindean:
I'd use it! You're certainly welcome to use my script as a starting point. I've basically modeled my interaction with it after [how I use vim-plug with a wrapper script](https://github.com/colindean/hejmo/blob/master/scripts/plug) in combination with the standard [bundles.vim](https://github.com/colindean/hejmo/blob/master/dotfiles/vim/bundles.vim) configuration.

Or perhaps like how [homebrew-bundle](https://github.com/homebrew/homebrew-bundle) with a DSL:

In `~/.config/bat/modules.sh` or something like that:

```shell
syntax "Ledger" "https://github.com/vqv/Ledger3.git"
theme "Zenburn" "https://github.com/exampledoesnotwork/zenburn.git"

# or with some helpers?
github_use_ssh # sets github() to use ssh URLs instead of HTTPS
syntax Ledger $(github vqv/Ledger3)
syntax SomethingElse $(gitlab example/example)
```

Spitballing my 2¢ here.

eth-p:
> Or perhaps like how [homebrew-bundle](https://github.com/homebrew/homebrew-bundle) with a DSL:

I'm actually really liking this idea. Something like what you suggested, but maybe a bit more flexible:

```shell
syntax "vqv/Ledger3"
syntax "vqv/Ledger3@branch"
syntax "vqv/Ledger3#commit"
syntax "vqv/Ledger3" --file "somefile.syntax_set"
syntax "vqv/Ledger3" --patch "./fix.patch"
theme ...
```

Or I could write an actual DSL parser:

```
syntax vqv/Ledger3

syntax vqv/Ledger3
    checkout branch

syntax vqv/Ledger3
    checkout abcdef1

syntax vqv/Ledger3
    file somefile.syntax_set

syntax vqv/Ledger3
    patch ./fix.patch
```

I'm leaning towards the DSL. With a simple DSL, it would be possible to add or remove entries programmatically.

And a couple notes:

1. The language mappings depend on the `. syntax_set` file, so we can't really set them.

2. It would default to GitHub, but we could specify a full git URL or use "gitlab:user/repo" for common websites like GitHub, GitLab, or BitBucket.

3. The file would be `$(dirname "$(bat --config-file)")/[program name]`.

4. What should the script be called? `bat-assets`?

sharkdp:
>     3\. $(dirname "$(bat --config-file)")

there is also `bat --config-dir`.

colindean:
I'd encourage you to go with the DSL as commands rather than parsing.

> What should the script be called?

`bat-assets` is okay, but maybe something indicating modularity, look and feel, or something like that? `bat-mod`? `bat-modules`? `bat-look`? `bat-lookandfeel`? `bat-appearance`? Probably best to optimize for something obvious that will be tab-completed in reality.

eth-p:
> I'd encourage you to go with the DSL as commands rather than parsing.

Is there a specific reason why you think commands would be the better choice? (e.g. parsing isn't flexible enough, harder to maintain, etc.) From my perspective, each of them have their pros and cons:

Parsing
|Pros|Cons|
|:--|:--|
|Can manipulate the file with scripts|No scripting capabilities|
|Can use its own syntax|More difficult to maintain a parser|

Bash
|Pros|Cons|
|:--|:--|
|Relatively easy to implement|Extremely difficult to parse and modify programmatically|
|Can use external commands and functions|Limited by Bash syntax|

> bat-assets is okay, but maybe something indicating modularity, look and feel, or something like that? bat-mod? bat-modules? bat-look? bat-lookandfeel? bat-appearance? Probably best to optimize for something obvious that will be tab-completed in reality.

I like `bat-mod` or `bat-modules`. What do you think, @sharkdp? Would that be too "official" for an unofficial tool?

sharkdp:
I like `bat-modules`, in analogy to git submodules.

> > I'd encourage you to go with the DSL as commands rather than parsing.
> 
> Is there a specific reason why you think commands would be the better choice?

I guess because it seems to be a bit overkill for this task, I would guess. At least that would be my feeling.

What about having a standalone `bat-modules` executable (written in Python?) that is configured via a `modules.yaml` file (or some other format, if you prefer) in `bat --config-dir`?
```yaml
syntaxes:
  - git: vqv/Ledger3
  - git: vqv/Ledger3
     revision: some-feature-branch
  - git: vqv/Ledger3
     revision: abcdef1
  # …
themes:
  - git: some/theme
```

colindean:
It seems like bat-extras is all bash right now. It seems that introducing YAML and Python would also be overkill unless there was some really meaningful reason for it. If you _really_ want to go down that route, this functionality is very similar to that of [Peru](https://github.com/buildinspace/peru). Maybe it is consumable as a library and you can transform the above YAML syntax to its syntax or data structures and leverage its built-in support (well, through deps) for a variety of VCSes.

Given what I've already got in my script, it's about 7/8 of the way to a 1.0 for the functionality. It's maybe an hour's refactor away from where I imagine a 1.0 of bat-modules could be. The big thing would be to extract the more procedural parts into some commands for the DSL.

sharkdp:
> It seems like bat-extras is all bash right now. It seems that introducing YAML and Python would also be overkill unless there was some really meaningful reason for it.

No, agreed. If the bash solution works, that's completely fine. I guess what I wanted to say is: before I would start writing a parser, I'd rather use config file format.

> Given what I've already got in my script, it's about 7/8 of the way to a 1.0 for the functionality. It's maybe an hour's refactor away from where I imagine a 1.0 of bat-modules could be. The big thing would be to extract the more procedural parts into some commands for the DSL.

:+1:

eth-p:
@colindean I had a short break from exams and ended up writing a DSL parser in bash/awk, and I have a minimum viable prototype of `bat-modules.sh` in bat-extras. I plan on refactoring and rewriting a lot of the script, but it works right now.

**~/.config/bat/modules.txt**:
```bash
syntax vqv/Ledger3 
```

**$ bat-modules.sh --update**
```
----- syntax: Ledger3 -----
Cloning...
Updating...
----- bat-modules -----
Done.
Rebuilding cache...
Writing theme set to /Users/Ethan/.cache/bat/themes.bin ... okay
Writing syntax set to /Users/Ethan/.cache/bat/syntaxes.bin ... okay
```

The output formatting needs a lot of work, and the code itself is glued together in a pretty fast and dirty way that I'll clean up when I have another opportunity to work on it.

Issue #811, opened 2020-01-26
Broken layout with unicode


Kogia-sima:
# :bug: Bug Report

## Description

File that contains unicode characters breaks the output layout.

It seems that `bat` regards the all input characters as single-width letter. However, if some line contains multiple-width character, the number of characters and the **true width** of line is different. That makes the `bat` split the line string at wrong position.

## Example

![0126_1036](https://user-images.githubusercontent.com/20660712/73129493-93289000-4028-11ea-9452-3d2fb6ac3f32.jpg)

## Possible Solution

Maybe [unicode-width](https://crates.io/crates/unicode-width) crate could solve this problem.

## Environment

| Software | Version |
| -------- | ------- |
| bat | 0.12.1 |
| qterminal | 0.14.1 |
| OS | Lubuntu 19.10 Eoan Ermine |

sharkdp:
This is a duplicate of #787

sharkdp:
This has been fixed in [bat 0.13](https://github.com/sharkdp/bat/releases/tag/v0.13.0).

PR #815, opened 2020-02-01
Fix wrapping method to support multiple print-width characters

### LINKED ISSUE #787: Wrapping characters spanning multiple print-width
Currently, wrapping will not work as expected if there are characters spanning multiple print-width in a line.

Screenshot from 2019-12-24 15-42-49

I haven't investigated the code, but I think that probably the print-width of a character is always assumed to be `1` in `bat`. In this case it can be fixed by properly calculating the print-width with some crates e.g. `unicode-width`.

Kogia-sima:
Fix #787 and #811


eth-p:
Awesome work, thanks for implementing this!

sharkdp:
@Kogia-sima Thank you very much for your contribution!

Sorry for being pedantic about this, but what is the content of the text that you added as (unit) tests? I'd appreciate if that would be something that is (1) "neutral" in terms of content and (2) uncritical with respect to licensing.

Maybe it would be best to use some very stupid meaningless dummy text, possibly repeating one and the same word (i.e. without reusing any existing text).

Kogia-sima:
Those texts are cited from [here](http://zarniwhoop.uk/files/PDF-languages/HOTO-Write-Multilingual-Text.pdf).

I don't know whether those texts are suitable for MIT/Apache-2.0 license, but it might include affiliated contents. Anyway, I've replaced the input data with the texts which does not reuse any existing sentence.

sharkdp:
@Kogia-sima Thank you for the update and thank you for implementing this!

sharkdp:

Issue #835, opened 2020-02-21
`git log` broken when using bat as pager


tbrodbeck:
I like using `bat` as my pager of choice with syntax highlighting! :). Somewhere across the internet I found the command `export PAGER='bat --paging=always'` to do exactly that.

Now I realized that my `git log` did not work anymore (did not show anything) and I found out that it was because of that setting. Setting `export PAGER=less` and my `git log` is back.

Is there a way how to make `bat`-pager and the `git log` compatible?

FrogBomb:
Do you have more information about your environment and the version of bat you are using?
What you did works fine for me... 
My system is:
```
$ uname -srm
Linux 5.5.4-arch1-1 x86_64
```
I'm using version:
```
$ yay -Q bat
bat 0.12.1-1
```

tbrodbeck:
So my system (macOS Catalina) information is
```
$ uname -srm
Darwin 19.3.0 x86_64
```
And the bat version is:
```
$ brew info bat
bat: stable 0.12.1 (bottled)
[...]
```

eth-p:
Trying this on my system (macOS Mojave) with `git -c core.pager='bat --paging=always --pager="less"' log` (I can't use PAGER because I have a global `core.pager` setting), it works for me.

```bash
$ uname -srm
Darwin 18.7.0 x86_64
```

```bash
$ brew info bat
bat: stable 0.12.1 (bottled)
```

Can you run the following script and paste the output?

```bash
#!/usr/bin/env bash
_bat_config="$(bat --config-file)"
_bat_binary="$(which bat 2>&1)"

printf "\n--- System ---\n"
uname -srm
command -v sw_vers &>/dev/null && sw_vers
command -v lsb_release &>/dev/null && lsb_release -a

printf "\n--- Software ---\n"
printf "less: %s\n" "$(less --version | head -n1)"

printf "\n--- Bat ---\n"
bat --version
printf "\n"

printf "Environment:\n"
env | grep '^BAT\|^PAGER=' && printf "\n"

printf "Config:\n"
[[ -f "$_bat_config" ]] && cat "$(bat --config-file)" && printf "\n"

printf "Wrapper:\n"
file "$_bat_binary" | grep "text executable" &>/dev/null && bat "$_bat_binary" && printf "\n"
```

tbrodbeck:
Sure:
```
$ sh batScript.sh

--- System ---
Darwin 19.3.0 x86_64
ProductName:	Mac OS X
[...]

PR #837, opened 2020-02-23
Add issue templates and a `info.sh` tool for helping collect and report system info.

eth-p:
The shell script is intended to be used to help maintainers and contributors diagnose issues that other people are having with `bat` and/or integration with external tools.

It asks for consent to collect data, and will print the following if everything is allowed:

```
========== system ==========

+ uname -srm
Darwin 18.7.0 x86_64

+ sw_vers # or lab_release on Linux
ProductName:	Mac OS X
ProductVersion:	10.14.6
BuildVersion:	18G95

========== bat ==========

+ bat --version
bat 0.12.1

+ env
BAT_PAGER=
PAGER=/usr/local/bin/less

========== bat_config ==========

+ cat /Users/eth-p/.config/bat/config
--tabs 4

# Syntaxes
#--map-syntax From:to


========== bat_wrapper ==========

+ cat /Users/eth-p/.local/bin/bat
#!/usr/bin/env bash
ARGS=()
case "$(defaults read -globalDomain AppleInterfaceStyle 2>/dev/null || echo "Light")" in
	Light)  ARGS+=(--theme OneHalfLight) ;;
	Dark|*) : ;;
esac
ARGS+=("$@")

exec /usr/local/bin/bat "${ARGS[@]}"


========== tool ==========

+ less --version
less 530 (PCRE regular expressions)
```

The data collected is as follows:

- Kernel name and version.
- Operating system details (distro, release name).
- `bat` version
- `less` version
- `bat` environment variables.
- `bat` config file.
- Any wrapper script ~~(but not function)~~ around `bat`.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -9,7 +9,7 @@ assignees: ''
 
 \-------------------------------------------------------------------------------------------------
 **Hey there, thanks for creating an issue!**  
-In order to replicate your issue, we might need to know a little bit more about the environment
+In order to reproduce your issue, we might need to know a little bit more about the environment
 which you're running `bat` on.
 
 **If you're on Linux or MacOS:**  
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -37,6 +37,7 @@ Please remove this header and continue.
 **How did you install `bat`?**  
 apt-get, homebrew, GitHub release, etc.
 
-**If applicable, can you tell us more about your system?**
-...
+---
+
+[paste the output of `info.sh` here]
 
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -8,28 +8,30 @@ assignees: ''
 ---
 
 \-------------------------------------------------------------------------------------------------
-**IS THIS BUG RELATED TO EXTERNAL TOOLS? (e.g. git, less, etc.)**
-
-If it is, there's a good chance we might need to know a little bit more about the environment
+**Hey there, thanks for creating an issue!**  
+In order to replicate your issue, we might need to know a little bit more about the environment
 which you're running `bat` on.
-
-**If you're on Windows:**  
-Please tell us about your:
-
-- Windows Version
 
 **If you're on Linux or MacOS:**  
 Please run the script at https://github.com/sharkdp/bat/blob/diag-tools/diagnostics/info.sh and
 paste the output at the bottom of the bug report.
 
+**If you're on Windows:**  
+Please tell us about the following at the bottom of the bug report:
+
+- Windows Version (e.g. "Windows 10 1908")
+
 **Once you're done:**  
 Please remove this header and continue.
 \-------------------------------------------------------------------------------------------------
 
-**What do you expect to happen?**  
+**What version of `bat` are you using?**
+[paste the output of `bat --version` here]
+
+**Describe the bug you encountered:**  
 ...
 
-**What actually happens?**  
+**Describe what you expected to happen?**  
 ...
 
 **How did you install `bat`?**  
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,40 @@
+---
+name: Bug Report
+about: Report a bug.
+title: "[Bug] "
+labels: bug
+assignees: ''
+
+---
+
+\-------------------------------------------------------------------------------------------------
+**IS THIS BUG RELATED TO EXTERNAL TOOLS? (e.g. git, less, etc.)**
+
+If it is, there's a good chance we might need to know a little bit more about the environment
+which you're running `bat` on.
+
+**If you're on Windows:**  
+Please tell us about your:
+
+- Windows Version
+
+**If you're on Linux or MacOS:**  
+Please run the script at https://github.com/sharkdp/bat/blob/diag-tools/diagnostics/info.sh and
+paste the output at the bottom of the bug report.
+
+**Once you're done:**  
+Please remove this header and continue.
+\-------------------------------------------------------------------------------------------------
+
+**What do you expect to happen?**  
+...
+
+**What actually happens?**  
+...
+
+**How did you install `bat`?**  
+apt-get, homebrew, GitHub release, etc.
+
+**If applicable, can you tell us more about your system?**
+...
+
--- a/.github/ISSUE_TEMPLATE/syntax_request.md
+++ b/.github/ISSUE_TEMPLATE/syntax_request.md
@@ -15,9 +15,17 @@ Bat supports locally-installed language definitions. See the link below:
 
 https://github.com/sharkdp/bat#adding-new-syntaxes--language-definitions
 
-If you think adding this syntax would help others as well, please remove this header and continue.
+If you think adding this syntax would help others as well, please make sure that it meets our
+guidelines for adding new syntaxes:
+
+ - 10,000 downloads on packagecontrol.io
+
+Once you have confirmed that it meets our guidelines, you can remove this header and continue.
 \-------------------------------------------------------------------------------------------------
 
 
 **Syntax:** syntax/language name
+
+**Guideline Criteria:**
+[packagecontro.io link here]
 
--- a/.github/ISSUE_TEMPLATE/syntax_request.md
+++ b/.github/ISSUE_TEMPLATE/syntax_request.md
@@ -1,0 +1,23 @@
+---
+name: Syntax Request
+about: Request adding a new syntax to bat.
+title: "[Syntax Request] "
+labels: new-syntax
+assignees: ''
+
+---
+
+\-------------------------------------------------------------------------------------------------
+**BEFORE YOU MAKE A REQUEST:**
+
+Are you looking to add a new syntax to use on one of your devices?
+Bat supports locally-installed language definitions. See the link below:
+
+https://github.com/sharkdp/bat#adding-new-syntaxes--language-definitions
+
+If you think adding this syntax would help others as well, please remove this header and continue.
+\-------------------------------------------------------------------------------------------------
+
+
+**Syntax:** syntax/language name
+
--- a/diagnostics/info.sh
+++ b/diagnostics/info.sh
@@ -1,0 +1,157 @@
+#!/usr/bin/env bash
+_modules=('system' 'bat' 'bat_config' 'bat_wrapper' 'tool')
+_modules_consented=()
+
+
+# -----------------------------------------------------------------------------
+# Modules:
+# -----------------------------------------------------------------------------
+
+_bat_:description() {
+	_collects "Version information for 'bat'."
+}
+
+_bat_config_:description() {
+	_collects "The environment variables used by 'bat'."
+	_collects "The 'bat' configuration file."
+}
+
+_bat_wrapper_:description() {
+	_collects "Any wrapper script used by 'bat'."
+}
+
+_system_:description() {
+	_collects "Operating system name."
+	_collects "Operating system version."
+}
+
+_tool_:description() {
+	_collects "Version information for 'less'."
+}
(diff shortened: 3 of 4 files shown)


sharkdp:
This is great, thank you very much!

A few thoughts:
- The new info/debug script looks good and works great. The only (minor) disadvantage I see is that it is harder for users to review before executing (simply because it is longer).
- I quickly thought about shipping that script with `bat` itself. It could be installed in `/usr/share/doc/bat`, but maybe that would make it even harder to access (different paths on different distros). On the other hand, it would maybe increase the "trustworthiness".

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,40 @@
...
+
+**If you're on Linux or MacOS:**  
+Please run the script at https://github.com/sharkdp/bat/blob/diag-tools/diagnostics/info.sh and
+paste the output at the bottom of the bug report.
+
+**Once you're done:**  
+Please remove this header and continue.
+\-------------------------------------------------------------------------------------------------
+
+**What do you expect to happen?**
```
sharkdp: Typically, I see these two points the other way around:
1. Describe your question or bug.
2. What did you expect to happen.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,40 @@
+---
```
sharkdp: Will users be able to select one of the two issue templates? Can they select "none" as well?

Otherwise, what about feature requests and other kind of questions?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,40 @@
...
+\-------------------------------------------------------------------------------------------------
+**IS THIS BUG RELATED TO EXTERNAL TOOLS? (e.g. git, less, etc.)**
+
+If it is, there's a good chance we might need to know a little bit more about the environment
+which you're running `bat` on.
+
+**If you're on Windows:**  
+Please tell us about your:
+
+- Windows Version
```
sharkdp: Maybe add "`bat` version" here as well? Also, can we please move the Windows section below the Linux/MacOS section? I believe it's still a rather small percentage of users.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,40 @@
...
+---
+
+\-------------------------------------------------------------------------------------------------
+**IS THIS BUG RELATED TO EXTERNAL TOOLS? (e.g. git, less, etc.)**
+
+If it is, there's a good chance we might need to know a little bit more about the environment
+which you're running `bat` on.
+
+**If you're on Windows:**  
+Please tell us about your:
```
sharkdp: Maybe: "In the last section of this template, please tell us about …".

.github/ISSUE_TEMPLATE/syntax_request.md
```diff
@@ -0,0 +1,23 @@
...
+Bat supports locally-installed language definitions. See the link below:
+
+https://github.com/sharkdp/bat#adding-new-syntaxes--language-definitions
+
+If you think adding this syntax would help others as well, please remove this header and continue.
+\-------------------------------------------------------------------------------------------------
+
+
+**Syntax:** syntax/language name
+
```
sharkdp: This is great!

We could even add a paragraph that:
- kindly asks the user to check if that syntax is "popular" enough (we recently had a discussion about this somewhere). We could add a link to packagecontrol.io/ and add the 10,000 downloads mark as a guideline.
- has a reference to [this section](https://github.com/sharkdp/bat/tree/master/assets#syntax-highlighting-in-bat) in case somebody wants to open a PR as well.

diagnostics/info.sh
```diff
@@ -0,0 +1,157 @@
...
+
+_ask_module() {
+	_tput clear
+	_tput cup 0 0
+
+cat 1>&2 <<EOF
+--------------------------------------------------------------------------------
+This script will collect information to help assist in issue resolution.
+You will be provided with a description of the data collected, and will be
+offered the opportunity to give consent for each piece of information collected.
```
sharkdp: I really appreciate this. However, it almost sounds a little bit too "dramatic". I wonder if someone could even get the impression that this script will directly upload the information somewhere (and do some magic things to "resolve issues").

I would propose something more straightforward along the lines of:
* this script runs some (harmless) commands to collect information about your system
* It will ask for consent before each command is executed
* In the end, it will output a short report that you can review and then copy into the issue description.

what do you think?

sharkdp:
For further inspiration (not implying that we need to follow this format), ripgreps issue template: https://github.com/BurntSushi/ripgrep/blob/master/ISSUE_TEMPLATE.md

diagnostics/info.sh
eth-p: Sounds good! I'll add it to the next commit

.github/ISSUE_TEMPLATE/bug_report.md
eth-p: I changed the header to say "Please tell us about the following at the bottom of the bug report", which should cover that.

.github/ISSUE_TEMPLATE/bug_report.md
eth-p: I moved it below, and I added the request for `bat --version` to the top above everything else. If people use the `info.sh` script, it will be duplicated in that too, but there shouldn't be any harm in doing so.

.github/ISSUE_TEMPLATE/bug_report.md
eth-p: They should be able to select a "blank" (misc.) issue template.

.github/ISSUE_TEMPLATE/bug_report.md
eth-p: Fixed.

eth-p:
> The new info/debug script looks good and works great. The only (minor) disadvantage I see is that it is harder for users to review before executing (simply because it is longer).

That is an unfortunate consequence of having more complex bash scripts, unfortunately. I was hoping that by asking for consent before running anything, it would remove the need to review it. If it poses to be an issue, we could replace it with the quick script I used in #835?

```bash
#!/usr/bin/env bash
_bat_config="$(bat --config-file)"
_bat_binary="$(which bat 2>&1)"

printf "\n--- System ---\n"
uname -srm
command -v sw_vers &>/dev/null && sw_vers
command -v lsb_release &>/dev/null && lsb_release -a

printf "\n--- Software ---\n"
printf "less: %s\n" "$(less --version | head -n1)"

printf "\n--- Bat ---\n"
bat --version
printf "\n"

printf "Environment:\n"
env | grep '^BAT\|^PAGER=' && printf "\n"

printf "Config:\n"
[[ -f "$_bat_config" ]] && cat "$(bat --config-file)" && printf "\n"

printf "Wrapper:\n"
file "$_bat_binary" | grep "text executable" &>/dev/null && bat "$_bat_binary" && printf "\n"
```

It's significantly easier to read, but it doesn't explain anything, runs everything, and doesn't do the greatest job at formatting the output.

> I quickly thought about shipping that script with bat itself. It could be installed in /usr/share/doc/bat, but maybe that would make it even harder to access (different paths on different distros). On the other hand, it would maybe increase the "trustworthiness".

It would definitely increase the trustworthiness, but it would make it quite a bit more difficult because of the path issue you mentioned. I also don't think it would be used by more than 1% of people who installed bat either.

sharkdp:
> If it poses to be an issue, we could replace it with the quick script I used in #835?

Let's keep it for now. I like it :+1: 

> It would definitely increase the trustworthiness, but it would make it quite a bit more difficult because of the path issue you mentioned. I also don't think it would be used by more than 1% of people who installed bat either.

Okay, let's go with the download first.

sharkdp:
The info.sh script currently prints the following for me:
```
The following data will be collected:
/tmp/info.sh: line 129: _bat_wrapper_function_:description: command not found

The following commands will be run:
"$SHELL" --login -c 'functions bat'
"$SHELL" --login -c 'declare -f bat'
```

sharkdp:
One more thought: it might be useful to format the output of the `info.sh` script as Markdown. In this way, we can be sure that we will properly see all the output. My `bat` config contains `# …` lines, for example. If we do not add a markdown code-section around these (triple-backtick), they would be formatted as markdown headlines.

Instead of printing `=== … ===` headings, we could use markdown-friendly version:
```markdown
system information
------------------

…

bat wrapper script
------------------
```


But that's just a cosmetic thing that we can also do later.

eth-p:
> ```
> /tmp/info.sh: line 129: _bat_wrapper_function_:description: command not found
> ```

Whoops. I missed a colon when writing that function.

> One more thought: it might be useful to format the output of the info.sh script as Markdown.

Absolutely love that idea. I'll get right to it.

eth-p:
New example output:

system
------

**$ uname -srm**
Darwin 18.7.0 x86_64  

**$ sw_vers**
ProductName:	Mac OS X  
ProductVersion:	10.14.6  
BuildVersion:	18G95  

bat
---

**$ bat --version**
bat 0.12.1  

**$ env**
BAT_PAGER=  
PAGER=/usr/local/bin/less  

bat_config
----------

**$ cat /Users/Ethan/.config/bat/config**
```
--tabs 4

# Syntaxes
#--map-syntax From:to

```

bat_wrapper
-----------

**$ cat /Users/Ethan/.local/bin/bat**
```
#!/usr/bin/env bash
ARGS=()
case "$(defaults read -globalDomain AppleInterfaceStyle 2>/dev/null || echo "Light")" in
	Light)  ARGS+=(--theme OneHalfLight) ;;
	Dark|*) : ;;
esac
ARGS+=("$@")

exec /usr/local/bin/bat "${ARGS[@]}"

```

bat_wrapper_function
--------------------

No wrapper function.

tool
----

**$ less --version**
less 530 (PCRE regular expressions)

sharkdp:
Thank you for the updates!

Here is mine (I'm using `zsh`, that's why it outputs *"Unable to determine if a wrapper function is set."*)

system
------

**$ uname -srm**
Linux 5.5.4-arch1-1 x86_64  

**$ lsb_release -a**
LSB Version:	1.4  
Distributor ID:	Arch  
Description:	Arch Linux  
Release:	rolling  
Codename:	n/a  

bat
---

**$ bat --version**
bat 0.12.1  

**$ env**
PAGER=less  

bat_config
----------

**$ cat /home/shark/.config/bat/config**
```
#--style="numbers,changes,header"

# Use italic text on the terminal (not supported on all terminals)
--italic-text=always

# Add mouse scrolling support in less (does not work with older
# versions of "less")
# --pager="less -FR"

# Syntax mappings
--map-syntax .ignore:.gitignore
--map-syntax .fdignore:.gitignore
--map-syntax PKGBUILD:bash
--map-syntax Cargo.lock:toml
--map-syntax h:cpp
--map-syntax ino:cpp


# --theme=DarkNeon
# --theme=Dracula
# --theme="Monokai Extended"
```

bat_wrapper
-----------

No wrapper script.

bat_wrapper_function
--------------------
Unable to determine if a wrapper function is set.

tool
----

**$ less --version**
less 551 (PCRE regular expressions)

eth-p:
Is there any way to check for and print the wrapper in `zsh`? I'm not as familiar with zhs as I am bash, but adding support for more shells would definitely be a good thing.

sharkdp:
This works:
```bash
❯ zsh --login -i -c 'type bat' 2>&1 
bat is a shell function from /home/shark/.zshrc

# modify .zshrc

❯ zsh --login -i -c 'type bat' 2>&1 
bat is a shell function from /home/shark/.zshrc
```

Note that I used `-i` to force the shell to be interactive. Otherwise, it wouldn't parse the `.zshrc`. I wonder if we would need similar arguments for `bash` and `fish` as well?

eth-p:
Awesome! It seems like zsh could be implemented similarly to bash, then.

I did some testing with bash and fish, and both of them worked with my setup. I'm not too sure how bash would handle `.bashrc`, but it did work with functions (but not aliases) in `.bash_profile`.

sharkdp:
I imagine most users would put such a wrapper function in the `.bashrc` file, so we should probably make sure that it would be included in the `info.sh` report.

PR #844, opened 2020-02-28
Simplify issue templates

sharkdp:
Just a few small things that I thought about after looking how the new templates look like on GitHub.

The main change is that I made the "header" section into a HTML comment. This way, users will not have to remove the header part.

@eth-p What do you think?

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -7,37 +7,33 @@ assignees: ''
 
 ---
 
-\-------------------------------------------------------------------------------------------------
-**Hey there, thanks for creating an issue!**  
+<!--
+Hey there, thanks for creating an issue!
+
 In order to reproduce your issue, we might need to know a little bit more about the environment
 which you're running `bat` on.
 
-**If you're on Linux or MacOS:**  
-Please run the script at https://github.com/sharkdp/bat/blob/diag-tools/diagnostics/info.sh and
-paste the output at the bottom of the bug report.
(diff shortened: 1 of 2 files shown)


sharkdp:
Also, I wonder if we should add a "Question" template or rename "Bug report" to "Bug report or question". A lot of opened issues actually fall into this category.

Maybe we could also add a very simple (empty?) template for "Feature request".

eth-p:
> The main change is that I made the "header" section into a HTML comment. This way, users will not have to remove the header part.

I actually forgot that was an option. Not to mention: that's a perfect use for comments, too.

> Also, I wonder if we should add a "Question" template or rename "Bug report" to "Bug report or question". A lot of opened issues actually fall into this category.

True... I think the former would be the best way to tailor the templates. I suspect that no matter which one we choose though, we're going to need to manually change the labels on issues at done point (e.g. a question turns out to be a bug, or a bug turns out bring a question).

> Maybe we could also add a very simple (empty?) template for "Feature request".

After looking at the UI for issues when templates are enabled, I absolutely agree. The blank template option isn't as clearly displayed as the others, and it would make it easier for somebody if they didn't notice it.

sharkdp:
> True... I think the former would be the best way to tailor the templates

Agreed. I added two new templates. 

Thank you very much for starting this. Excited to see how this works out. I'm pretty sure it will save us a lot of work!

eth-p:
> I'm pretty sure it will save us a lot of work!

Hopefully! With the `info.sh` script, it should also help reduce the back-and-forth required to diagnose the cause behind some issues.

PR #852, opened 2020-03-03
Explain how to fix no output issue

sorenbug:
I just ran into this so I figured it might be helpful for others.

I don't know what other details are really needed, so if something needs to be added, just let me know 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -568,6 +568,10 @@ iconv -f ISO-8859-1 -t UTF-8 my-file.php | bat
 Note: you might have to use the `-l`/`--language` option if the syntax can not be auto-detected
 by `bat`.
 
+### No output on Windows
+
+Install the Visual C++ Redistributable: [The latest supported Visual C++ downloads - Microsoft Support](https://support.microsoft.com/en-us/help/2977003/the-latest-supported-visual-c-downloads)
+
 ## Development
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -497,7 +497,7 @@ Example configuration file:
 
 ## Using `bat` on Windows
 
-`bat` mostly works out-of-the-box on Windows, but a few features may need extra configuration.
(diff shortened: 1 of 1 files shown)


sharkdp:

If this is definitely required for the binaries on Windows, maybe we should add the hint in this section: https://github.com/sharkdp/bat#on-windows ?

sorenbug:
Got it, will do

sharkdp:
I was referring to the "installation / on Windows" section, which I think would be more suited

sharkdp:
Sorry for being pedantic here, but: is this required when installing via `scoop`/`choco` as well?

Otherwise, please move it up to the part: "You can download prebuilt binaries from the [Release page](https://github.com/sharkdp/bat/releases) …" and maybe split that sentence into multiple paragraphs or subsections.

sorenbug:
It is at least needed on `choco` as that's how I installed it. I would assume the same for scoop unless there's any additional dependencies in the scoop package that there's not in the choco package.

sharkdp:
@rasmuskriest @meltinglava Can you confirm that this is needed for the Chocolatey and scoop packages, respectively?

meltinglava:
What the scoop maintainers did was suggest `less` as a pager as seen in [bat.json](https://github.com/ScoopInstaller/Main/blob/9059ade74e6a4b5a78e7cdbeb7e45897f093e4c3/bucket/bat.json#L7). I have had no problems with bat once installing less, thow i am not using much windows any more.

rasmuskriest:
I am going to add a dependency for [vcredist140](https://chocolatey.org/packages/vcredist140) to the Chocolatey package so that this manual step will not be necessary in this case (similar to `less`).

sharkdp:
@meltinglava I think this is independent from whether or not `less` is added as a dependency. The question is if users need the "Microsoft Visual C++ Redistributable"  package to be installed. I would guess so, if this is also the case for the manual install and the Chocolatey package.

@rasmuskriest Great! Thank you very much!

PR #855, opened 2020-03-03
Add custom theme/syntax detection to info.sh

eth-p:


diagnostics/info.sh
```diff
@@ -39,6 +41,19 @@ _tool_:description() {
 _bat_:run() {
 	_out bat --version
 	_out env | grep '^BAT_\|^PAGER='
+
+	local cache_dir="$(bat --cache-dir)"
+	if [[ -f "${cache_dir}/syntaxes.bin" ]]; then
+		_print_command "bat" "--list-languages"
+		echo "Found custom syntax set."
+		bat --list-languages
```
sharkdp: Hm, do we really want to include these large lists in the output?

I think the presence of a custom syntax/theme set is an important factor, but the actual content will probably rarely be.

eth-p: That's a fair point. Maybe a diff instead, or just echo the presence of one?

sharkdp: I think the presence alone would be enough for now.

eth-p:
I amended the commit to no longer list the syntax/theme sets, but only the presence of them.

sharkdp:
Thank you!

PR #856, opened 2020-03-03
Add zsh wrapper function detection to info.sh

eth-p:
It should also now support `.bashrc`, but I'm not able to test it on MacOS.
Could you confirm that it works correctly for you, @sharkdp?


sharkdp:
Doesn't work with a `bat` function in `.zshrc`. `zsh` definitely requires the `-i` flag (for interactive shell).

sharkdp:
`bash` also definitely needs the `-i` flag.

Unfortunately, I have another problem with `bash`, because my locale is set to german (`export LANG=de_DE.UTF-8`). In this case, the `grep 'function'` will not work because:
```
❯ bash --login -i -c "type bat"
bat ist eine Funktion.
…
```
I guess this could be fixed by setting `LANG=C` from `info.sh`?

eth-p:
It looks like I forgot to add the `-i` flag in the initial check. I'll fix that with the next commit.

As for the language issue, I'll probably set the export at the top of the script. Most of the script makes the assumption that `LANG=en_US.UTF-8`, and it's safer to override it for the entire script rather than risk forgetting to set it for a particular external command.

eth-p:
All of the issues (`-i` flag, doesn't work with non-English `$LANG`) should be fixed now.

PR #869, opened 2020-03-16
Specify cache dir with BAT_CACHE_DIR

### LINKED ISSUE #829: Specify a directory for cache
Hey :wave:,

I have my `.cache` directory set as `tmpfs` to run all cache related files on RAM. This makes the `.cache` available only per session which is an issue with `bat` as the built custom themes I have are removed.

Is there a way to specify a directory for `bat` to load **theme** and **syntax** from? I've tried `bat cache --source ~/.local/share/bat` but that just creates the cache at that location. I see no option to load the cache from that place.

neuronull:
- if set, BAT_CACHE_DIR is used. otherwise use OS default.

Closes #829

### DIFF
--- a/src/dirs.rs
+++ b/src/dirs.rs
@@ -1,8 +1,6 @@
 use crate::dirs_rs;
-use std::path::{Path, PathBuf};
-
-#[cfg(target_os = "macos")]
 use std::env;
+use std::path::{Path, PathBuf};
 
 /// Wrapper for 'dirs' that treats MacOS more like Linux, by following the XDG specification.
 /// This means that the `XDG_CACHE_HOME` and `XDG_CONFIG_HOME` environment variables are
@@ -14,16 +12,7 @@ pub struct BatProjectDirs {
 
 impl BatProjectDirs {
     fn new() -> Option<BatProjectDirs> {
-        #[cfg(target_os = "macos")]
-        let cache_dir_op = env::var_os("XDG_CACHE_HOME")
-            .map(PathBuf::from)
-            .filter(|p| p.is_absolute())
-            .or_else(|| dirs_rs::home_dir().map(|d| d.join(".cache")));
-
-        #[cfg(not(target_os = "macos"))]
-        let cache_dir_op = dirs_rs::cache_dir();
-
-        let cache_dir = cache_dir_op.map(|d| d.join("bat"))?;
+        let cache_dir = BatProjectDirs::get_cache_dir()?;
 
         #[cfg(target_os = "macos")]
         let config_dir_op = env::var_os("XDG_CONFIG_HOME")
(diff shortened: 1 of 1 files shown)


neuronull:
```
$ env | grep BAT

$ ./target/debug/bat --cache-dir
/home/<user>/.cache/bat

$ export BAT_CACHE_PATH=/tmp/bat_cache

$ ./target/debug/bat --cache-dir
/tmp/bat_cache

$ unset BAT_CACHE_PATH

$ ./target/debug/bat --cache-dir
/home/<user>/.cache/bat
```

src/dirs.rs
```diff
@@ -42,6 +31,27 @@ impl BatProjectDirs {
         })
     }
 
+    pub fn get_cache_dir() -> Option<PathBuf> {
+        // on all OS prefer BAT_CACHE_PATH if set
+        let cache_dir_op = env::var_os("BAT_CACHE_PATH")
+            .map(PathBuf::from)
+            .filter(|p| p.is_absolute());
```
sharkdp: Why this constraint? To help users shooting themselves in the foot? If so, we shouldn't silently fall back to the standard cache dir in this case, but rather print a warning.

If there is no strong reason for doing so, I would just remove this `.filter(…)`. If a user wants to quickly test something, using a relative cache path might even be useful.

src/dirs.rs
```diff
@@ -42,6 +31,27 @@ impl BatProjectDirs {
...
+        if !cache_dir_op.is_none() {
```
sharkdp: you can use:
```rust
if cache_dir_op.is_some() {
  // …
```

src/dirs.rs
neuronull: There wasn't a specific reason for adding that constraint. It was a lazy logic copy from other place in this file doing that. Thinking about it now, it makes much more sense to not require absolute paths here. I'll remove it.

src/dirs.rs
neuronull: Nice! If it wasn't obvious already, I'm still learning Rust.

src/dirs.rs
```diff
@@ -42,6 +31,25 @@ impl BatProjectDirs {
...
+        #[cfg(target_os = "macos")]
+        let cache_dir_op = env::var_os("XDG_CACHE_HOME")
+            .map(PathBuf::from)
+            .filter(|p| p.is_absolute())
+            .or_else(|| dirs_rs::home_dir().map(|d| d.join(".cache")));
+
+        #[cfg(not(target_os = "macos"))]
+        let cache_dir_op = dirs_rs::cache_dir();
+
+        Some(cache_dir_op.map(|d| d.join("bat")))?
```
sharkdp: Couldn't this just be:
```rust
cache_dir_op.map(|d| d.join("bat"))
```

Using `Some(xx)?` is the same as `xx`, right?

neuronull: Good catch. That does appear to be the case. I was trying to maintain the `?` operator (since this part of the function was extracted from the `new() `function) but it appears that removing Some() and ? works the same because its already an Option.

PR #871, opened 2020-03-18
Implement --file-name<name> option

### LINKED ISSUE #654: Option to manually set the name of a file passed in through STDIN.
As is the case with *nix command line tools, they're often going to be on the receiving end of a pipe. Unless we use a hacky workaround (writing though a named pipe or writing to disk), bat is programmed to display "STDIN" in the header when it's being piped to.

With bat gaining in popularity and various wrappers being created to integrate it with other tools (e.g. code formatters), it would be a really nice quality of life improvement for wrapper scripts to be able to set the name displayed in bat's header component.

An option like `--override-file-name` should solve this nicely. With multiple files, the name could be mapped like such:

```
--override-file-name "2:unformatted.rs"
```

This would only override the second file's name if you passed multiple files like `formatted.rs -`.

neuronull:
- can specify filename to be displayed when printing.
- useful for when piping data from STDIN

Closes #654


sharkdp:

Happy to review this once it's possible to apply this to non-STDIN entries as well. It would be great if we could add a test for this. The easiest option is probably to write a new [integration test](https://github.com/sharkdp/bat/blob/master/tests/integration_tests.rs).

neuronull:
> It would be great if we could add a test for this. The easiest option is probably to write a new [integration test](https://github.com/sharkdp/bat/blob/master/tests/integration_tests.rs).

Started looking into this and so far I haven't figured out how to get the execution of the assert_cmd to return the output that we care about for this change (the "File: <filename>"). So far I've only seen it capture the body of the file itself as opposed to the metadata. (when calling `.assert().success().stdout()` )

I'll keep looking into it but thought I'd mention it as I'm new to all this and not sure if I just haven't stumbled across the right thing yet or if I've stumbled across a limitation of the tools used in the integration_tests.

sharkdp:
> Started looking into this and so far I haven't figured out how to get the execution of the assert_cmd to return the output that we care about for this change (the "File: ").

It should work if you pass `--decorations=always` (and set the `--style` appropriately. enable the `header`, but maybe disable some other components)
https://github.com/sharkdp/bat/blob/b1b8addf7e5f7b06b741caf7bdf11c795795ae1f/tests/integration_tests.rs#L135-L144

If needed, you can also use the `--terminal-width` option to set a fixed output width.

neuronull:
> It should work if you pass `--decorations=always` (and set the `--style` appropriately. enable the `header`, but maybe disable some other components)

Cool, yeah this worked great.

neuronull:
Manual testing output:

```
$ ./target/debug/bat foo.json
       File: foo.json
   1   {}

$ ./target/debug/bat foo.json | json_pp | ./target/debug/bat -
       File: STDIN
   1   {}

$ ./target/debug/bat foo.json | json_pp | ./target/debug/bat --file-name=foo.json -
       File: foo.json
   1   {}

$ hexyl foo.binary
┌────────┬─────────────────────────┬─────────────────────────┬────────┬────────┐
│00000000│ 00 00 00 00             ┊                         │0000    ┊        │
└────────┴─────────────────────────┴─────────────────────────┴────────┴────────┘

$ ./target/debug/bat foo.binary | ./target/debug/bat -
       File: STDIN   <BINARY>

$ ./target/debug/bat foo.binary | ./target/debug/bat --file-name=foo.binary -
       File: foo.binary   <BINARY>


$ ./target/debug/bat --file-name="biz.binary" --style=numbers foo.binary
[bat warning]: Binary content from file 'biz.binary' will not be printed to the terminal (but will be present if the output of 'bat' is piped). You can use 'bat -A' to show the binary file contents.


$ ./target/debug/bat --file-name="biz.binary" foo.binary
       File: biz.binary   <BINARY>
```

sharkdp:
Thank you for the updates.

As stated in https://github.com/sharkdp/bat/issues/654#issuecomment-529236134, the idea was to be able to specify `--file-name` multiple times. It doesn't look like the current implementation supports this.

sharkdp:
rebased onto master

neuronull:
> As stated in [#654 (comment)](https://github.com/sharkdp/bat/issues/654#issuecomment-529236134), the idea was to be able to specify `--file-name` multiple times. It doesn't look like the current implementation supports this.

I see now. Started looking into this. If --file-name is specified for one FILE, can it be omitted for another FILE in the same execution? Or is it the case that if --file-name is specified for any FILE, then each FILE has to have a --file-name specified? Seems like former would be harder to handle... Even with the later, I suppose would need to confirm that number of FILES specified matches number of --file-name args received.

If some are optional, would it be best to store these (Config::filenames) as `Vec<Option<&'a str>>` ? 
And if its all or nothing, I'm guessing `Option<Vec<&'a str>>` is a better choice?
Currently struggling with the piece of code in app.rs which will decode the multiple filenames.

Is there a way to say "filenames is the vector of strings or None"
something like: (doesn't compile)

```
 filenames: self
                .matches
               .values_of("file-name")
               .unwrap_or_else(|| None)
```

sharkdp:
> If --file-name is specified for one FILE, can it be omitted for another FILE in the same execution? Or is it the case that if --file-name is specified for any FILE, then each FILE has to have a --file-name specified?

Definitely the latter, I would say. The number of `--file-name` options should always be equal to the number of `FILE`s. If this is not the case, we should print an error.

> And if its all or nothing, I'm guessing `Option<Vec<&'a str>>` is a better choice?

:+1: 

>  Currently struggling with the piece of code in app.rs which will decode the multiple filenames.
> 
> Is there a way to say "filenames is the vector of strings or None"

Yeah, working with these can be tricky. Let's see. `self.matches.values_of("file-name")` already returns `Option<Values<'a>>` and would be `None` if no `--file-name` argument would be present (which is what we want). So we only need to `.map` over that `Option` to handle the `Some(…)` case:
```rust
filenames: self
    .matches
    .values_of("file-name")
    .map(|values| values.collect())
```
Here, the `values.collect()` call turns the `Values` iterator into a `Vec`.

neuronull:
> Yeah, working with these can be tricky. Let's see. `self.matches.values_of("file-name")` already returns `Option<Values<'a>>` and would be `None` if no `--file-name` argument would be present (which is what we want). So we only need to `.map` over that `Option` to handle the `Some(…)` case:
> 
> ```rust
> filenames: self
>     .matches
>     .values_of("file-name")
>     .map(|values| values.collect())
> ```
> 
> Here, the `values.collect()` call turns the `Values` iterator into a `Vec`.

Thank you! I had a feeling that I was over-complicating it.

src/bin/bat/app.rs
```diff
@@ -136,6 +136,12 @@ impl App {
             }
         });
 
+        if self.matches.value_of("file-name").is_some()
+            && self.matches.values_of("file-name").unwrap().len() != files.len()
```
sharkdp: Instead of first checking for `.is_some()` and then using `.unwrap()` later, you can directly match on the `Option`:
``` rust
match self.matches.value_of("file-name") {
    Some(filenames) if filenames.len() != files.len => { … },
    _ => {}
}
```

src/controller.rs
```diff
@@ -45,7 +45,20 @@ impl<'b> Controller<'b> {
...
+                .filenames
+                .as_ref()
+                .unwrap()
+                .into_iter()
+                .map(|name| Some(*name))
+                .collect()
+        };
+
+        for it in self.config.files.iter().zip(filenames) {
+            let (input_file, file_name) = it;
```
sharkdp: You can use:
``` rust
for (input_file, file_name) in self.config.files.iter().zip(filenames) {
```

src/controller.rs
```diff
@@ -45,7 +45,20 @@ impl<'b> Controller<'b> {
...
+            vec![None; self.config.files.len()]
+        } else {
+            self.config
+                .filenames
+                .as_ref()
+                .unwrap()
+                .into_iter()
+                .map(|name| Some(*name))
+                .collect()
+        };
```
sharkdp: Similar here. Using a `match` statement would be more idiomatic and shorter (in general, we try to avoid `.unwrap` calls):
``` rust
let filenames = match self.config.filenames {
    Some(ref filenames) => filenames.into_iter().map(|name| Some(*name)).collect(),
    None => vec![None; self.config.files.len()],
};
```
Seeing that this vector is only being built up to be iterated over directly afterwards, we can also skip the `.collect()` part and just use the iterator directly:
``` rust
let filenames: Box<dyn Iterator<Item = _>> = match self.config.filenames {
    Some(ref filenames) => Box::new(filenames.into_iter().map(|name| Some(*name))),
    None => Box::new(std::iter::repeat(None)),
};
```

src/bin/bat/app.rs
neuronull: nice! The match statement continues to surprise me in its versatility.

src/controller.rs
neuronull: Cool! I need to read up on the Box more. Why use it here? Is it for performance?

sharkdp: `Box` is used for something different here.

We want `filenames` to be an iterator over `&str` (`Iterator<Item = &str>`). However, `Iterator` is a `trait` and not a *concrete type*. It's more like an interface (Java) or an abstract base class (C++, Python, ...). There are many different concrete types that *implement* `Iterator<Item = &str>`. In our case, the concrete types are `std::iter::Map<I, F>` (which is the return type of `.map(…)`) and `std::iter::Repeat<A>` (the return type of `iter::repeat`).

These two types are completely different (i.e. have a different memory layout and size). Consequently, there is no way to store simply say
``` rust
// does not compile:
let filenames = match self.config.filenames {
    Some(ref filenames) => filenames.into_iter().map(|name| Some(*name)),
    None => std::iter::repeat(None),
};
```
because we cannot give a *statically known* type to `filenames`.

The way to solve this problem is to use dynamic memory allocation. Instead of creating the `filenames` object on the stack (where we would need to know the static size), we dynamically allocate space for it on the heap. This is what `Box` does. It's a smart pointer that references a dynamically sized object on the heap.

Using `Box` also allows us to abstract over the concrete type in use. This is where the type annotation comes into play:
``` rust
let filenames: Box<dyn Iterator<Item = _>> = …
```
This basically says: `filenames` is a smart pointer to an object of some type that *implements* `Iterator<…>`. We don't have to specify what `Item` is here, because that can be inferred by the compiler.

If you are familiar with C++, `Box` is like a (smart) pointer to the abstract base class.

sharkdp:
Thank you!

src/controller.rs
neuronull: Whoa thanks for the explanation! I am familiar with C++ and smart pointers so this makes a lot of sense now. Thank you !

idnsunset:
manpage seems not updated? I couldn't find the new `--file-name` option in the manpage.

keith-hall:
You're right, thanks for reporting. Something similar was also mentioned at https://github.com/sharkdp/bat/issues/1176#issuecomment-742388497

Issue #501, opened 2019-02-08
A shell script named `build`


jzinn:
Bat does not highlight a shell script with the file name `build` that contains a shebang line of `#!/bin/bash`.

Bat does highlight other shell scripts that lack a `.sh` extension, like `run`, `setup`, etc.

I suspect bat considers a file named `build` to be an NAnt build file.

Would it be possible to change the precedence so that bat will interpret my `build` script as a shell script?

sharkdp:
> Bat does highlight other shell scripts that lack a `.sh` extension, like `run`, `setup`, etc.
> 
> I suspect bat considers a file named `build` to be an NAnt build file.

Yes, you are right. That's unfortunate.

> Would it be possible to change the precedence so that bat will interpret my `build` script as a shell script?

Yes. You can add this to your config file:
```
--max-syntax build:bash
```

If you just want that behavior temporarily, you can always just use `-l`/`--language`:
```
bat build -lsh
```

jzinn:
Oh snap, that works.  (With `--map-syntax`.)  Thanks.

I thought it would require a code change to get bat to look for shebangs in files without extensions before using the filename to determine the syntax.

PR #877, opened 2020-03-22
glob-based syntax mapping

### LINKED ISSUE #592: Add a dedicated Path => Syntax lookup database

### LINKED ISSUE #685: Files named `build` don't respect shebang lines
Create a file named `hello`:
```ruby
#!/usr/bin/env ruby

puts "hello, world"
```

`bat hello` works just fine, but rename it to `build` and get no syntax highlighting. Or perhaps incorrect highlighting, but in this example I only see white text.

I might tackle this on the weekend if nobody jumps on it. I love the tool btw! <3

sharkdp:
This PR changes the way `--map-syntax` works. The option now works like this:
```
--map-syntax <glob-pattern>:<syntax-name>
```
(this is a breaking change, users will have to adapt their config files)

This PR also adds a builtin database of syntax mappings to `bat` to handle cases such as:

* Glob-based lookups for paths:
  * `/proc/cpuinfo` => `CpuInfo syntax`
  * `**/.ssh/config` => `SSHConfig`
  * `/etc/profile` => `Bash`
  * `*.h` => `C++` (I assume this is a better default than "Objective C")
* Files named `build` are no longer mapped to "NAnt Build File" (which should only match `*.build` files). Instead, we use the first-line detection mechanism on such files.

closes #592, closes #685  
see also:  #501

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -127,6 +127,7 @@ dependencies = [
  "encoding 0.2.33 (registry+https://github.com/rust-lang/crates.io-index)",
  "error-chain 0.12.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "git2 0.13.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "globset 0.4.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "lazy_static 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "liquid 0.19.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "shell-words 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -199,6 +200,14 @@ version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "byte-tools 0.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "bstr"
+version = "0.2.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "memchr 2.3.3 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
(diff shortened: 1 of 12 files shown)


cyqsimon:
Hi sharkdp, I'm working on making this system more extensible in #2747. I noticed that c08d122 you made the match happen in reversed direction (so that mappings inserted later get matched first).

In the spirit of documenting the code behaviour as precisely as possible, I would like to ask you to clarify what your intentions were. My current hypothesis for why this was done is this allows rules inserted by the user at runtime to take precedence; is this correct?

PR #881, opened 2020-03-23
Add --locked flag to cargo commands

chenrui333:
relates to https://github.com/Homebrew/homebrew-core/issues/46025

In homebrew formula, we already enforced the `--locked` flag to achieve reproducible builds https://github.com/Homebrew/homebrew-core/pull/52022

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ matrix:
     #   rust: nightly
     #   # skip the global install step
     #   install:
-    #     - cargo install --debug --force rustfmt-nightly
+    #     - cargo install --locked --debug --force rustfmt-nightly
     #   script: cargo fmt -- --check
 
 sudo: required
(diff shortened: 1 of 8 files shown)


sharkdp:

To be honest, I don't fully understand why we would want that inside the `bat` repository. If the Rust ecosystem decides to change the default for `cargo install`, there's probably a good reason for that (e.g. getting security updates; but see https://github.com/rust-lang/cargo/issues/7169).

You also added the `--locked` flag to `cargo run` and `cargo test` calls. If I understand this correctly, only the behavior of `cargo install` has been changed.

sharkdp:
I'm going to close this for now. Happy to reopen it if there are any new developments.

sharkdp:
I have now added `--locked` to all `cargo install` commands in the README, after #955 happened. Thank you, @chenrui333!

PR #885, opened 2020-03-24
Add option to generate a default config file, fixes #870

### LINKED ISSUE #870: Feature Request: Generate default config file
Branching off from #573 - it would be very useful if bat automatically generated a sane, small, well commented config file for users.

For example:
- `bat --config-file -o $file` writes to `$file` 
- `bat --generate-config-file` writes to the current default config location and prints the path

I assume the implementation would be a small static string or binary blog baked into the binary. Not sure how rust does these things.

jmick414:
I took a shot at implementing the `--generate-config-file` feature request per #870.

This is my 1st foray into Rust so I'm open to feedback/suggestions for improvement.  Also... thoughts on if the default config file content is sufficient or should be modified?

Thanks!

### DIFF
--- a/src/bin/bat/config.rs
+++ b/src/bin/bat/config.rs
@@ -1,6 +1,7 @@
 use std::env;
 use std::ffi::OsString;
 use std::fs;
+use std::io::{self, Write};
 use std::path::PathBuf;
 
 use shell_words;
@@ -13,6 +14,46 @@ pub fn config_file() -> PathBuf {
         .map(PathBuf::from)
         .filter(|config_path| config_path.is_file())
         .unwrap_or_else(|| PROJECT_DIRS.config_dir().join("config"))
+}
+
+pub fn generate_config_file() {
+    let config_file = config_file();
+    if config_file.exists() {
+        println!("A config file already exists at: {}", config_file.to_string_lossy());
+
+        print!("Overwrite? (y/n): ");
+        let _ = io::stdout().flush();
+        let mut decision = String::new();
+        io::stdin().read_line(&mut decision).expect("Failed to read input");
+
+        if !decision.trim().eq_ignore_ascii_case("Y") {
+            return;
+        }
+    } else {
+        let config_dir = config_file.parent().unwrap();
+        if !config_dir.exists() {
+            fs::create_dir(config_dir).expect("Unable to create config directory");
+        }
+    }
+
+    let default_config = "# Specify desired theme (e.g. \"TwoDark\")
+#--theme=\"TwoDark\"
+
+# Enable this to use italic text on the terminal (not supported on all terminals):
+#--italic-text=always
+
+# Uncomment the following line to disable automatic paging:
+#--paging=never
+
+# Use C++ syntax for .ino files
+#--map-syntax \"*.ino:C++\"
+
+# Use \".gitignore\"-style highlighting for \".ignore\" files
+#--map-syntax \".ignore:Git Ignore\"
+";
+
+    fs::write(&config_file, default_config).expect("Error writing config file!");
+    println!("Success! Config file written to {}", config_file.to_string_lossy());
 }
 
 pub fn get_args_from_config_file() -> Result<Vec<OsString>, shell_words::ParseError> {
--- a/src/bin/bat/config.rs
+++ b/src/bin/bat/config.rs
@@ -16,28 +16,31 @@ pub fn config_file() -> PathBuf {
         .unwrap_or_else(|| PROJECT_DIRS.config_dir().join("config"))
 }
 
-pub fn generate_config_file() {
+pub fn generate_config_file() -> bat::errors::Result<()> {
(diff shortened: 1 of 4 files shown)


src/bin/bat/config.rs
```diff
@@ -15,6 +16,46 @@ pub fn config_file() -> PathBuf {
         .unwrap_or_else(|| PROJECT_DIRS.config_dir().join("config"))
 }
 
+pub fn generate_config_file() {
```
sharkdp: In order to properly handle errors, this function should return a `bat::errors::Result<()>`. In `main.rs`, you would then simply call it with
```rust
generate_config_file()?;
```

src/bin/bat/config.rs
```diff
@@ -15,6 +16,46 @@ pub fn config_file() -> PathBuf {
...
+    let config_file = config_file();
+    if config_file.exists() {
+        println!("A config file already exists at: {}", config_file.to_string_lossy());
+
+        print!("Overwrite? (y/n): ");
```
sharkdp: Maybe make it clear that "no" is the default here:
```
Overwrite? (y/N):
```

src/bin/bat/config.rs
```diff
@@ -15,6 +16,46 @@ pub fn config_file() -> PathBuf {
...
+        let _ = io::stdout().flush();
```
sharkdp: If you change this function to return `Result<()>`, you can use
```rust
io::stdout().flush()?;
```
here. Another alternative would be `io::stdout().flush().ok()`.

src/bin/bat/config.rs
```diff
@@ -15,6 +16,46 @@ pub fn config_file() -> PathBuf {
...
+        let mut decision = String::new();
+        io::stdin().read_line(&mut decision).expect("Failed to read input");
```
sharkdp: If this function returns a `Result<…>`, you can simply use `.read_line(&mut decision)?` instead of `.expect(…)` which would kill the process in case of an error.

src/bin/bat/config.rs
```diff
@@ -15,6 +16,46 @@ pub fn config_file() -> PathBuf {
...
+
+        if !decision.trim().eq_ignore_ascii_case("Y") {
+            return;
```
sharkdp: This would need to be a `return Ok(());`. Similarly, you need to use `Ok(())` in the last line of this function.

src/bin/bat/config.rs
```diff
@@ -15,6 +16,46 @@ pub fn config_file() -> PathBuf {
...
+        }
+    } else {
+        let config_dir = config_file.parent().unwrap();
```
sharkdp: `.unwrap()` also kills the process if an error occurs. It's better to use `.parent()?`.

src/bin/bat/config.rs
```diff
@@ -15,6 +16,46 @@ pub fn config_file() -> PathBuf {
...
+        if !config_dir.exists() {
```
sharkdp: Doing a check like this before creating a directory or file is generally discouraged because it can lead to race conditions. Even if that is unlikely to be a problem here, the alternative is much simpler: simply create the directory and ignore the fact that it could already exist.

Now `fs::create_dir` would return an error in case the path already exists, but you can (and should) use `fs::create_dir_all` anyway. `create_dir` would fail if the parent directory would not exist (`~/.config` on Unix systems).

src/bin/bat/config.rs
```diff
@@ -15,6 +16,46 @@ pub fn config_file() -> PathBuf {
...
+            fs::create_dir(config_dir).expect("Unable to create config directory");
+        }
+    }
+
+    let default_config = "# Specify desired theme (e.g. \"TwoDark\")
```
sharkdp: To prevent having to escape all quotes in this default config, you can use a raw string:
``` rust
let default_config = r#"…
…"#;
```

src/bin/bat/config.rs
```diff
@@ -15,6 +16,46 @@ pub fn config_file() -> PathBuf {
...
+#--theme=\"TwoDark\"
```
sharkdp: We could maybe mention `bat --list-themes` here in the comment.

src/bin/bat/config.rs
```diff
@@ -15,6 +16,46 @@ pub fn config_file() -> PathBuf {
...
+#--paging=never
+
+# Use C++ syntax for .ino files
+#--map-syntax \"*.ino:C++\"
+
+# Use \".gitignore\"-style highlighting for \".ignore\" files
+#--map-syntax \".ignore:Git Ignore\"
+";
+
+    fs::write(&config_file, default_config).expect("Error writing config file!");
```
sharkdp: Similar here, please replace `.expect(…)` by a `?`.

jmick414:
Thanks @sharkdp!  I'll be adapting accordingly.

Issue #884, opened 2020-03-23
Will InputFile be able to take strings in the future?


jonathandturner:
I looked into adding `bat` to Nushell last night, and one issue I ran across was that there wasn't a way to send `bat` the string content we have.  Since Nushell handles its own pipelines, we have streams of text that might have been loaded earlier in the pipeline.  Rather than trying to load the original file, it'd be great to be able to display the lines that we have available as string content.

Would this be compatible with the bat's direction, by chance?

eth-p:
We had an old discussion going in #729 for how to rewrite the `InputFile` struct as something more friendly for consumption in a public API. If that ends up being the direction we go down, it should be able to accept arbitrary strings (as an entire file) or a boxed `Read` (for streaming chunks of a file) as an input source.

dtolnay:
+1 for use in https://github.com/dtolnay/cargo-expand. We have the contents in memory so giving bat a String would be sufficient, we wouldn't require a `Read` with streaming chunks.

sharkdp:
I implemented this in #936

Issue #729, opened 2019-11-05
Library API: Input Source


eth-p:
I've been thinking of ideas for the library API that can provide more powerful functionality, while also reducing how much of the bat internals are exposed. Feedback would be appreciated.

---

I think we should rework/refactor input sourcing for `bat`-as-a-library. The current implementation only allows specifying input sources (via `InputFile`) from STDIN, the filesystem, and the embedded preview string.

For an API, we should have something a bit more generalized:

- Accepts any `Read` trait as an input source.
- Accepts any string as an input source.
- Contains custom "properties" (see below)
- Contains metadata (e.g. syntax name, file name, file encoding, etc.)

## Properties:

For extensibility, there should be some way to add extra fields ("properties") to an input source. This would allow features to be implemented cleanly without significantly modifying the internal code. One example would be moving the git change data into a property, and passing the property to the `Decoration` impl instead of the `InteractivePrinter`.

There are a couple ways we could implement this:

**Generics:**

We could make everything generic, and put all the properties in a single struct.
This would be fairly performant since all the hard work is done at compile-time, but it wouldn't work well when mixing external code (e.g. custom decorations) with built-in features.

**HashMap<&'static str, Any>:**

Rather than having to know everything at compile-time, we could use a hashmap of `Any`. This would allow us to split off parts of `bat` into cargo features (e.g. git support), and allows for mixing and matching of various types of properties.

Both the `HashMap` and `Any` will incur a runtime performance hit, but that could easily be mitigated by using a builder that reads and caches the data once per input source:

```rust
trait SomethingBuilder {
    fn create(&mut self, input: &Input) -> Box<dyn Something>;
}

struct MySomethingBuilder {}
impl SomethingBuilder for MySomethingBuilder {
    fn create(&mut self, input: &Input) -> Box<dyn Something> {
        Something::new(
            input.properties()
                .get("thing")
                .and_then(|v| v.downcast_ref::<Thing>())
                .unwrap()
        )
    }
}
```

dandavison:
In case these notes are helpful to the library API design:

[...]

PR #892, opened 2020-03-26
Use --file-name to detect syntax highlighting

### LINKED ISSUE #891: When reading from stdin, bat should use --file-name (if present) to pick the syntax highlighter
Per discussion at https://github.com/sharkdp/bat/issues/237#issuecomment-604443466.

When reading from stdin, bat can't perform highlighting as it doesn't know the filename... unless it is given with --file-name.  (We could even *always* use --file-name (if present) to pick the syntax even if the real filename is available, but that may run into issues if that name is inconsistent with the real name.)

neuronull:
Closes #891

### DIFF
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -184,17 +184,8 @@ impl HighlightingAssets {
             (Some(language), _) => self.syntax_set.find_syntax_by_token(language),
             (None, InputFile::Ordinary(ofile)) => {
                 let path = Path::new(ofile.filename());
-
                 let file_name = path.file_name().and_then(|n| n.to_str()).unwrap_or("");
-                let extension = path.extension().and_then(|x| x.to_str()).unwrap_or("");
-
-                let ext_syntax = self
-                    .syntax_set
-                    .find_syntax_by_extension(&file_name)
-                    .or_else(|| self.syntax_set.find_syntax_by_extension(&extension));
-                let line_syntax = String::from_utf8(reader.first_line.clone())
-                    .ok()
-                    .and_then(|l| self.syntax_set.find_syntax_by_first_line(&l));
+                let line_syntax = self.get_first_line_syntax(reader);
 
                 let absolute_path = path.canonicalize().ok().unwrap_or(path.to_owned());
                 match mapping.get_syntax_for(absolute_path) {
@@ -204,30 +195,38 @@ impl HighlightingAssets {
                         self.syntax_set.find_syntax_by_name(syntax_name)
                     }
                     Some(MappingTarget::MapToUnknown) => line_syntax,
-                    None => ext_syntax.or(line_syntax),
+                    None => self.get_extension_syntax(file_name).or(line_syntax),
                 }
             }
             (None, InputFile::StdIn(None)) => String::from_utf8(reader.first_line.clone())
                 .ok()
                 .and_then(|l| self.syntax_set.find_syntax_by_first_line(&l)),
             (None, InputFile::StdIn(Some(file_name))) => self
-                .syntax_set
-                .find_syntax_by_extension(file_name.to_str().unwrap())
-                .or_else(|| {
-                    self.syntax_set.find_syntax_by_extension(
-                        Path::new(file_name)
-                            .extension()
-                            .and_then(|x| x.to_str())
-                            .unwrap_or(""),
-                    )
-                })
-                .or(String::from_utf8(reader.first_line.clone())
-                    .ok()
-                    .and_then(|l| self.syntax_set.find_syntax_by_first_line(&l))),
+                .get_extension_syntax(file_name.to_str().unwrap())
+                .or(self.get_first_line_syntax(reader)),
             (_, InputFile::ThemePreviewFile) => self.syntax_set.find_syntax_by_name("Rust"),
         };
 
         syntax.unwrap_or_else(|| self.syntax_set.find_syntax_plain_text())
+    }
+
+    fn get_extension_syntax(&self, file_name: &str) -> Option<&SyntaxReference> {
+        self.syntax_set
+            .find_syntax_by_extension(file_name)
+            .or_else(|| {
+                self.syntax_set.find_syntax_by_extension(
+                    Path::new(file_name)
+                        .extension()
+                        .and_then(|x| x.to_str())
+                        .unwrap_or(""),
+                )
+            })
+    }
+
+    fn get_first_line_syntax(&self, reader: &mut InputFileReader) -> Option<&SyntaxReference> {
+        String::from_utf8(reader.first_line.clone())
+            .ok()
+            .and_then(|l| self.syntax_set.find_syntax_by_first_line(&l))
     }
 }
 
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -271,7 +271,7 @@ mod tests {
             let syntax = self.assets.get_syntax(
                 None,
                 input_file,
-                &mut input_file.get_reader(&io::stdin()).unwrap(),
+                &mut input_file.get_reader(io::stdin().lock()).unwrap(),
                 &self.syntax_mapping,
             );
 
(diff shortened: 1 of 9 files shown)


neuronull:
(also tested manually)

src/assets.rs
```diff
@@ -246,17 +261,28 @@ mod tests {
             }
         }
 
-        fn synax_for_file_with_content(&self, file_name: &str, first_line: &str) -> String {
+        fn syntax_for_file_with_content(
+            &self,
+            file_name: &str,
+            first_line: &str,
+            as_stdin: bool,
```
sharkdp: If possible, I'm trying to avoid using Booleans as parameters to functions, as Rust does not have named parameters. At the call site
```
self.syntax_for_file_with_content(file_name, "", false)
```
it's not really clear what this `false` stands for. One alternative is to use `enum` with two entries.

For this particular case, I wonder if the overall implementation would look a bit cleaner if we would attach the (optional) user-provided file name to the `InputFile` enum:
``` rust
#[derive(Debug, Clone, Copy, PartialEq)]
pub enum InputFile<'a> {
    StdIn(Option<&'a OsStr>),
    Ordinary(&'a OsStr),
    ThemePreviewFile,
}
```

This way, we wouldn't need the additional `file_name` parameter in `get_syntax`, for example.

neuronull: Yes I did notice not having named parameters nor default ones, I'm glad you brought it up. I didn't realize enum entries can contain tuples! Would it also make sense to have the optional provided file name stored there for Ordinary files?

`#[derive(Debug, Clone, Copy, PartialEq)]
pub enum InputFile<'a> {
    StdIn(Option<&'a OsStr>),
    Ordinary(&'a OsStr, Option<&'a OsStr>),
    ThemePreviewFile,
}`

I started looking into this and I'm thinking it might make sense to evaluate the file-name(s) option(s) at the same time as the files themselves, essentially returning from this function

`fn files(&self) -> Vec<InputFile>`

, with the vector containing the optional file-names there. What do you think?

sharkdp: I can't see all the implications of that, but it sounds like an idea worth trying.

`enum`s can contain arbitrary data, not just tuples. If we decide to add more fields to the tuples, we should consider using a `struct`. For example:
``` rust
pub struct OrdinaryFile<'a> {
  filename: &'a OsStr,
  user_provided_name: Option<&'a OsStr>,
}

pub enum InputFile<'a> {
    Ordinary(OrdinaryFile),
    // ...
}
```
(names are just an example)


Note that we are discussing the removal of `InputFile` in other tickets, see #884 and #729.

neuronull: > `enum`s can contain arbitrary data, not just tuples. If we decide to add more fields to the tuples, we should consider using a `struct`.

This sounds like a good idea, I will give it a try.

> Note that we are discussing the removal of `InputFile` in other tickets, see #884 and #729.

Good to know. It seems like keeping the user provided file name (or more generally, name of input) close to the InputFile (or whatever it will become as a result of a redesign) makes sense. I'm expecting the changes in this pull request (including potentially making OrdinaryFile struct) , are still safe to do and hopefully will not overly complicate those tickets.

sharkdp:
Rebased and fixed merge conflicts.

src/assets.rs
```diff
@@ -207,9 +207,23 @@ impl HighlightingAssets {
...
+                    self.syntax_set.find_syntax_by_extension(
+                        Path::new(file_name)
+                            .extension()
+                            .and_then(|x| x.to_str())
+                            .unwrap_or(""),
+                    )
+                })
+                .or(String::from_utf8(reader.first_line.clone())
+                    .ok()
+                    .and_then(|l| self.syntax_set.find_syntax_by_first_line(&l))),
```
sharkdp, line 226: I don't like the fact that the whole logic from the `(None, InputFile::Ordinary(ofile))` case above is repeated here. We should extract this into a separate method or try to move both cases into the same match arm.

src/assets.rs
```diff
@@ -333,4 +347,17 @@ mod tests {
...
+        let test = SyntaxDetectionTest::new();
+
+        // from file extension
+        assert_eq!(test.syntax_for_file_with_content("test.cpp", ""), "C++");
+        // from first line (fallback)
+        assert_eq!(
+            test.syntax_for_file_with_content("my_script", "#!/bin/bash"),
+            "Bourne Again Shell (bash)"
+        );
+    }
```
sharkdp, line 362: This does not test the `StdIn` branch?!

src/printer.rs
```diff
@@ -240,20 +230,16 @@ impl<'a> InteractivePrinter<'a> {
...
                 let input = match file {
-                    InputFile::Ordinary(filename) => format!(
-                        "file '{}'",
-                        file_name.unwrap_or(&filename.to_string_lossy())
-                    ),
-                    _ => file_name.unwrap_or("STDIN").to_owned(),
+                    InputFile::Ordinary(ofile) => {
+                        format!("file '{}'", &ofile.filename().to_string_lossy())
+                    }
+                    InputFile::StdIn(Some(name)) => name.to_str().unwrap().to_string(),
```
sharkdp, line 240: Why not `name.to_string_lossy()`?

neuronull: ah, will use that.

src/assets.rs
neuronull: Good catch! That was pure hastiness on my part. I should have taken the changes slower to make sure I didn't lose some code paths.

I am currently working on fixing this, and I've run into an issue that I've been stuck on for a couple days. I have the following:

```rust
          fn syntax_for_stdin_with_content(&self, file_name: &str, first_line: &str) -> String {
              let input_file = InputFile::StdIn(Some(OsStr::new(file_name)));
              let syntax = self.assets.get_syntax(
                  None,
                  &input_file,
                  &mut input_file.get_reader(&io::stdin()).unwrap(),
                  &self.syntax_mapping,
              );
              syntax.name.clone()
          }
```
And both of these test cases hang:

```rust
          // from file extension
          assert_eq!(test.syntax_for_stdin_with_content("test.cpp", "a"), "C++");
          // from first line (fallback)
          assert_eq!(
              test.syntax_for_stdin_with_content("my_script", "#!/bin/bash"),
              "Bourne Again Shell (bash)"
          );
```

at the following line in fn `get_reader` :

`reader.read_until(b'\n', &mut first_line).ok();`

I discerned this from println statement debugging.
And I can't figure out why.
I didn't change this line "`&mut input_file.get_reader(&io::stdin()).unwrap()`" , nor can I see anything that changed in the function get_reader or even the input_file.rs itself.

I saw you pushed some commits too, but I looked at all of those and am not seeing anything which changed. Any ideas? Thanks for your patience with my newbness!

neuronull: Also worth noting that manual tests seem to work fine for the syntax detection from filename extension and from first line using stdin. So it must be something specific to how the test case is setup.

sharkdp: Well, these cases don't hang. They are waiting for input from STDIN :smile:. If you type something on the terminal and press enter (twice), the tests will keep on going and fail or succeed depending on what you input. 

That's obviously not how we want to test this in a unit/integration test.

I have pushed a commit that changes the way that `stdin` is passed into the `get_reader` function. This way, we can just pass in a byte-string as content (instead of reading from *actual* stdin).

neuronull: ahhh, I see now!! Thanks!

src/assets.rs
neuronull: Yep, that makes sense. I chose the helper method approach.

src/assets.rs
```diff
@@ -204,17 +195,39 @@ impl HighlightingAssets {
...
-                    None => ext_syntax.or(line_syntax),
+                    None => self.get_extension_syntax(file_name).or(line_syntax),
                 }
             }
-            (None, InputFile::StdIn) => String::from_utf8(reader.first_line.clone())
+            (None, InputFile::StdIn(None)) => String::from_utf8(reader.first_line.clone())
                 .ok()
                 .and_then(|l| self.syntax_set.find_syntax_by_first_line(&l)),
+            (None, InputFile::StdIn(Some(file_name))) => self
+                .get_extension_syntax(file_name.to_str().unwrap())
```
sharkdp, line 205: Always try to avoid plain `.unwrap()` calls. This will panic if `.to_str()` fails.

Currently working on a fix.

sharkdp:
Thank you for the updates!

PR #898, opened 2020-03-30
Allow using bat-as-a-library with older syntect

### LINKED ISSUE #896: Unbump the syntect dependency
I am interested in using bat in https://github.com/dtolnay/cargo-expand, but I don't want to use syntect 3.2.1 or 3.3.0. Those versions pull in onig 5.0, while syntect 3.2.0 uses onig 4.0 which has a more reliable cross platform build story.

Bat's syntect dependency is currently specified as 3.3.0:

https://github.com/sharkdp/bat/blob/9a050cd87fde4309db5af0179a4773fe97d3d846/Cargo.toml#L37-L38

It looks like this requirement was bumped from 3.2.0 in #628 and #671 for no particular reason. Could I request that the requirement in Cargo.toml be reverted to 3.2.0? Bat works with any of those versions and Cargo's resolver will give your users 3.3.0, but this allows cargo-expand to pin syntect 3.2.0 and still use bat.

Separately, consider disabling these dependabot bumps or find out if there is a way to bump Cargo.lock but not Cargo.toml. Bumping the Cargo.toml is not so bad when you are a binary but needlessly aggressive for libraries and causes unnecessary conflicts like us trying to build with syntect 3.2.0.

---

Longer term, if bat upgrades to syntect 4.0 that would also work for me because that version provides a way to disable the onig dependency. If you are upgrading to 4.0, please consider doing it in a way that allows downstream code to opt out of depending on onig:

```toml
[dependencies]
syntect = { version = "4.0", default-features = false }

[features]
default = ["syntect/default-onig"]
```

dtolnay:
Fixes #896. The "3.2.0" requirement means the same thing as \>=3.2.0 \<4.0.0. This way:

- the bat application will continue using 3.3.0 as per the lockfile;
- bat-as-a-library will pull in the most recent semver compatible version in that range just like before;
- downstream code is free to pin older versions as needed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,7 @@ default-features = false
 features = []
 
 [dependencies.syntect]
-version = "3.3.0"
+version = "3.2.0"
 default-features = false
 features = ["parsing", "yaml-load", "dump-load", "dump-create"]
 


sharkdp:

Happy to merge this, if the question in https://github.com/sharkdp/bat/issues/896#issuecomment-606202411 can clarified.

PR #899, opened 2020-03-30
Strip dependencies of bat-as-a-library

### LINKED ISSUE #895: Remove binary-specific dependencies from bat library
I am considering using bat-as-a-library in https://github.com/dtolnay/cargo-expand but I found that the library pulls in an enormous dependency tree, a lot of which appears to be specific to the bat command line tool. For example `liquid` seems to be related to the bat.1 manpage and fish completions, neither of which should be needed for library mode. `clap` also doesn't seem like it should be necessary in library mode.

I pasted the `liquid` dependency graph below. This by itself is big enough that it really makes me hesitant about adding a library dependency on bat especially when liquid won't even be used at all by the library.

Could you consider adding an on-by-default feature for pulling in the binary-specific dependencies, so that crates that depend on bat as a library can opt out of them? Or alternatively, consider moving the bat library to a different crate name so it doesn't need to share a Cargo.toml with the binary?

```console
    └── liquid v0.19.0
        ├── chrono v0.4.11
        │   ├── num-integer v0.1.42
        │   │   └── num-traits v0.2.11
        │   │       [build-dependencies]
        │   │       └── autocfg v1.0.0
        │   │   [build-dependencies]
        │   │   └── autocfg v1.0.0 (*)
        │   ├── num-traits v0.2.11 (*)
        │   └── time v0.1.42
        │       └── libc v0.2.68 (*)
        ├── deunicode v1.1.0
        ├── doc-comment v0.3.3
        ├── itertools v0.8.2
        │   └── either v1.5.3
        ├── lazy_static v1.4.0 (*)
        ├── liquid-compiler v0.19.0
        │   ├── itertools v0.8.2 (*)
        │   ├── liquid-error v0.19.0
        │   ├── liquid-interpreter v0.19.0
        │   │   ├── anymap v0.12.1
        │   │   ├── itertools v0.8.2 (*)
        │   │   ├── liquid-error v0.19.0 (*)
        │   │   └── liquid-value v0.19.1
        │   │       ├── chrono v0.4.11 (*)
        │   │       ├── itertools v0.8.2 (*)
        │   │       ├── liquid-error v0.19.0 (*)
        │   │       ├── num-traits v0.2.11 (*)
        │   │       └── serde v1.0.105 (*)
        │   ├── liquid-value v0.19.1 (*)
        │   ├── pest v2.1.3
        │   │   └── ucd-trie v0.1.3
        │   └── pest_derive v2.1.0
        │       ├── pest v2.1.3 (*)
        │       └── pest_generator v2.1.3
        │           ├── pest v2.1.3 (*)
        │           ├── pest_meta v2.1.3
        │           │   ├── maplit v1.0.2
        │           │   └── pest v2.1.3 (*)
        │           │   [build-dependencies]
        │           │   └── sha-1 v0.8.2
        │           │       ├── block-buffer v0.7.3
        │           │       │   ├── block-padding v0.1.5
        │           │       │   │   └── byte-tools v0.3.1
        │           │       │   ├── byte-tools v0.3.1 (*)
        │           │       │   ├── byteorder v1.3.4 (*)
        │           │       │   └── generic-array v0.12.3
        │           │       │       └── typenum v1.11.2
        │           │       ├── digest v0.8.1
        │           │       │   └── generic-array v0.12.3 (*)
        │           │       ├── fake-simd v0.1.2
        │           │       └── opaque-debug v0.2.3
        │           ├── proc-macro2 v1.0.9 (*)
        │           ├── quote v1.0.3 (*)
        │           └── syn v1.0.17 (*)
        ├── liquid-derive v0.19.0
        │   ├── liquid-compiler v0.19.0 (*)
        │   ├── liquid-error v0.19.0 (*)
        │   ├── liquid-interpreter v0.19.0 (*)
        │   ├── liquid-value v0.19.1 (*)
        │   ├── proc-macro2 v0.4.30 (*)
        │   ├── proc-quote v0.2.2
        │   │   ├── proc-macro-hack v0.5.14
        │   │   ├── proc-macro2 v0.4.30 (*)
        │   │   ├── proc-quote-impl v0.2.2
        │   │   │   ├── proc-macro-hack v0.5.14 (*)
        │   │   │   ├── proc-macro2 v0.4.30 (*)
        │   │   │   └── quote v0.6.13 (*)
        │   │   ├── quote v0.6.13 (*)
        │   │   └── syn v0.15.44
        │   │       ├── proc-macro2 v0.4.30 (*)
        │   │       ├── quote v0.6.13 (*)
        │   │       └── unicode-xid v0.1.0 (*)
        │   └── syn v0.15.44 (*)
        ├── liquid-error v0.19.0 (*)
        ├── liquid-interpreter v0.19.0 (*)
        ├── liquid-value v0.19.1 (*)
        ├── regex v1.3.6 (*)
        ├── serde v1.0.105 (*)
        ├── unicode-segmentation v1.6.0
        └── url v1.7.2
            ├── idna v0.1.5
            │   ├── matches v0.1.8 (*)
            │   ├── unicode-bidi v0.3.4 (*)
            │   └── unicode-normalization v0.1.12 (*)
            ├── matches v0.1.8 (*)
            └── percent-encoding v1.0.1
```

dtolnay:
Fixes #895. This PR introduces a `features = ["application"]` which is enabled by default and pulls in everything required by bat the application. When depending on bat as a library, downstream Cargo.toml should disable this feature to cut out inapplicable heavy dependencies like liquid.

```toml
[dependencies]
bat = { version = "0.13", default-features = false }
```

I've put some other optional functionality behind features as well: paging and git support, neither of which I would want for cargo-expand.

### DIFF
--- a/ci/script.bash
+++ b/ci/script.bash
@@ -16,3 +16,4 @@ fi
 # Check bat-as-a-library, which has a smaller set of dependencies
 cargo check --target "$TARGET" --verbose --lib --no-default-features
 cargo check --target "$TARGET" --verbose --lib --no-default-features --features git
+cargo check --target "$TARGET" --verbose --lib --no-default-features --features paging
--- a/ci/script.bash
+++ b/ci/script.bash
@@ -17,3 +17,4 @@ fi
 cargo check --target "$TARGET" --verbose --lib --no-default-features
 cargo check --target "$TARGET" --verbose --lib --no-default-features --features git
 cargo check --target "$TARGET" --verbose --lib --no-default-features --features paging
+cargo check --target "$TARGET" --verbose --lib --no-default-features --features git,paging
--- a/ci/script.bash
+++ b/ci/script.bash
@@ -15,3 +15,4 @@ fi
 
 # Check bat-as-a-library, which has a smaller set of dependencies
 cargo check --target "$TARGET" --verbose --lib --no-default-features
+cargo check --target "$TARGET" --verbose --lib --no-default-features --features git
--- a/ci/script.bash
+++ b/ci/script.bash
@@ -12,3 +12,6 @@ if [[ $TARGET != arm-unknown-linux-gnueabihf ]] && [[ $TARGET != aarch64-unknown
     # Run 'bat' on its own source code and the README
     cargo run --target "$TARGET" -- src/bin/bat/main.rs README.md --paging=never
 fi
+
+# Check bat-as-a-library, which has a smaller set of dependencies
+cargo check --target "$TARGET" --verbose --lib --no-default-features
--- a/src/config.rs
+++ b/src/config.rs
@@ -5,14 +5,14 @@ pub use crate::syntax_mapping::{MappingTarget, SyntaxMapping};
 pub use crate::wrap::OutputWrap;
 
 #[derive(Debug, Clone, Copy, PartialEq)]
+#[cfg(feature = "paging")]
 pub enum PagingMode {
-    #[cfg(feature = "paging")]
     Always,
-    #[cfg(feature = "paging")]
     QuitIfOneScreen,
     Never,
 }
 
+#[cfg(feature = "paging")]
 impl Default for PagingMode {
     fn default() -> Self {
         PagingMode::Never
@@ -53,6 +53,7 @@ pub struct Config<'a> {
     pub output_wrap: OutputWrap,
 
     /// Pager or STDOUT
+    #[cfg(feature = "paging")]
     pub paging_mode: PagingMode,
 
     /// Specifies the lines that should be printed
--- a/src/config.rs
+++ b/src/config.rs
@@ -6,7 +6,9 @@ pub use crate::wrap::OutputWrap;
 
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub enum PagingMode {
+    #[cfg(feature = "paging")]
     Always,
+    #[cfg(feature = "paging")]
     QuitIfOneScreen,
     Never,
 }
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -2,57 +2,24 @@ use error_chain::error_chain;
 
 error_chain! {
     foreign_links {
-        Clap(clap::Error);
-        Io(std::io::Error);
-        SyntectError(syntect::LoadingError);
-        ParseIntError(std::num::ParseIntError);
-        GlobParsingError(globset::Error);
+        Clap(::clap::Error) #[cfg(feature = "application")];
+        Io(::std::io::Error);
+        SyntectError(::syntect::LoadingError);
+        ParseIntError(::std::num::ParseIntError);
+        GlobParsingError(::globset::Error);
     }
 }
 
 pub fn default_error_handler(error: &Error) {
     match error {
         Error(ErrorKind::Io(ref io_error), _)
-            if io_error.kind() == std::io::ErrorKind::BrokenPipe =>
+            if io_error.kind() == ::std::io::ErrorKind::BrokenPipe =>
         {
-            std::process::exit(0);
+            ::std::process::exit(0);
         }
         _ => {
             use ansi_term::Colour::Red;
             eprintln!("{}: {}", Red.paint("[bat error]"), error);
         }
     };
-}
-
-// Mock out a type for clap::Error if we aren't pulling in a dependency on clap.
-//
-// This can be removed after migrating away from error_chain to some modern
-// derive-based error library such as thiserror, in favor of:
-//
-//     #[derive(Error)]
-//     pub enum Error {
-//         #[cfg(feature = "application")]
-//         Clap(clap::Error),
-//         ...
-//     }
-//
-#[cfg(not(feature = "application"))]
-mod clap {
-    use std::fmt::{self, Debug, Display};
-
-    pub struct Error(());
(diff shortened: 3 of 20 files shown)


ci/script.bash
```diff
@@ -12,3 +12,8 @@ if [[ $TARGET != arm-unknown-linux-gnueabihf ]] && [[ $TARGET != aarch64-unknown
     # Run 'bat' on its own source code and the README
     cargo run --target "$TARGET" -- src/bin/bat/main.rs README.md --paging=never
 fi
+
+# Check bat-as-a-library, which has a smaller set of dependencies
+cargo check --target "$TARGET" --verbose --lib --no-default-features
+cargo check --target "$TARGET" --verbose --lib --no-default-features --features git
+cargo check --target "$TARGET" --verbose --lib --no-default-features --features paging
```
sharkdp: Does it make sense to also check `--no-default-features --features git,paging` explicitly, even if this is the combination which is also included in the default feature set?

Similarly, shouldn't the unit tests also be executed for all feature set combinations? Or is there a guarantee that they would also run if (1) `cargo check` for that feature set succeeds and (2) `cargo test` succeeds for all features enabled? I would assume that there is no such guarantee?

src/config.rs
```diff
@@ -6,7 +6,9 @@ pub use crate::wrap::OutputWrap;
 
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub enum PagingMode {
+    #[cfg(feature = "paging")]
     Always,
+    #[cfg(feature = "paging")]
     QuitIfOneScreen,
     Never,
 }
```
sharkdp: Is there a reason for leaving the `PagingMode` enum available (with a single variant) in case `paging` is not enabled? I'm assuming there would be too many changes all over the code base if `PagingMode` would not be available?

src/errors.rs
```diff
@@ -2,24 +2,57 @@ use error_chain::error_chain;
...
+            std::process::exit(0);
         }
         _ => {
             use ansi_term::Colour::Red;
             eprintln!("{}: {}", Red.paint("[bat error]"), error);
         }
     };
 }
+
+// Mock out a type for clap::Error if we aren't pulling in a dependency on clap.
```
sharkdp: So I guess it's not possible to put a feature gate on the `Clap(clap::Error);` entry in `foreign_links`? Would it be possible to put one on the full `error_chain!` macro call (which would result in some duplication, but maybe still better than a fake type)?

src/output.rs
```diff
@@ -1,33 +1,39 @@
...
 pub enum OutputType {
+    #[cfg(feature = "paging")]
     Pager(Child),
     Stdout(io::Stdout),
 }
 
 impl OutputType {
     pub fn from_mode(mode: PagingMode, pager: Option<&str>) -> Result<Self> {
-        use self::PagingMode::*;
+        let _ = pager;
```
sharkdp: Silencing a "unused argument" warning? Could we use `#[allow(unused_variables)]` or similar?

sharkdp:
Thank you very much! This looks great :+1:

ci/script.bash
dtolnay: > Does it make sense to also check `--no-default-features --features git,paging` explicitly, even if this is the combination which is also included in the default feature set?

In this case these two features are very orthogonal; no module contains both `cfg(feature = "git")` and `cfg(feature = "paging")` in the same module. But these library-only check builds are very fast once the dependencies have been built by previous builds. `--lib --no-default-features --features git,paging` only takes 0.7 seconds on my machine so we might as well &mdash; I've added it. But in general I haven't found exhaustively checking combinations of orthogonal features to be valuable in my library projects.

> Similarly, shouldn't the unit tests also be executed for all feature set combinations?

In https://github.com/dtolnay/syn we do `check` builds of various common feature combinations but only `test` with all features enabled, and that has been working well. Introducing more `cfg(...)` gunk in the test suite has tended not to be valuable.

src/config.rs
dtolnay: It's a little more invasive in src/controller.rs but not bad, I've made that change. Good call.

src/errors.rs
dtolnay: Ah I tracked down the right syntax, they just do it backwards with the cfg after the variant instead of in front. :thinking: 

```rust
foreign_links {
    Clap(::clap::Error) #[cfg(feature = "application")];
}
```

src/output.rs
dtolnay: Fixed by putting cfg(feature = "paging") on from_mode and exposing OutputType::stdout() for controller.rs to use if building without paging.

ci/script.bash
sharkdp: Thanks!

sharkdp:
Thank you for the updates.

Note that we still pull in `clap` as a dependency via syntect=>onig=>onig_sys=>bindgen (https://github.com/rust-lang/rust-bindgen/blob/f568d095774bd5642fe8404e13268e43bd47327a/Cargo.toml#L50-L51). But this could be resolved by upgrading to `syntect` 4.1.0 and making it possible to use fancy-regex as you suggested in #896.

sharkdp:
> But this could be resolved by upgrading to `syntect` 4.1.0 and making it possible to use fancy-regex as you suggested in #896.

See #900

PR #903, opened 2020-04-02
Factor out writing the output so it can be used in a lib

jchorl:
Hey!

This PR is largely just moving the actual reading/writing logic into a separate function that takes a writer. Going along with the librarizing theme, this allows other libraries to cleanly capture and use the output. I added an example for how this might be used.

I'm not a huge fan of the error handling, but didn't see any better options. It looks like for multiple files, we don't let a single file fail the whole operation. Given that, we can't use standard `Result`s.

I also didn't make with/default error_handler variants. I figure this is a library, people can just pass in the default (it is public, after all). I didn't want to create even more function-nesting.

Let me know what you think -

### DIFF
--- a/src/controller.rs
+++ b/src/controller.rs
@@ -20,16 +20,32 @@ impl<'b> Controller<'b> {
     }
 
     pub fn get_output<'a>(&self) -> Result<Vec<u8>> {
+        let mut writer = Vec::new();
+
         let stdin = io::stdin();
 
-        let mut buf = Vec::new();
+        let filenames: Box<dyn Iterator<Item = _>> = match self.config.filenames {
+            Some(ref filenames) => Box::new(filenames.into_iter().map(|name| Some(*name))),
+            None => Box::new(std::iter::repeat(None)),
+        };
 
-        for input_file in self.config.files.iter() {
+        for (input_file, file_name) in self.config.files.iter().zip(filenames) {
             let mut reader = input_file.get_reader(&stdin)?;
-            while reader.read_line(&mut buf)? {}
+            if self.config.loop_through {
+                let mut printer = SimplePrinter::new();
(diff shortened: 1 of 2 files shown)


src/controller.rs
```diff
@@ -41,8 +41,14 @@ impl<'b> Controller<'b> {
...
         let mut output_type = OutputType::from_mode(paging_mode, self.config.pager)?;
         let writer = output_type.handle()?;
-        let mut no_errors: bool = true;
 
+        let no_errors = self.run_with_writer(writer, handle_error);
+
+        Ok(no_errors)
+    }
+
+    pub fn run_with_writer(&self, mut writer: &mut dyn Write, handle_error: impl Fn(&Error)) -> bool {
```
sharkdp, line 50: `writer` needs to reference a `&mut dyn Write`, but it doesn't have to be mutable itself, right? (which would allow retargeting the borrow).

src/controller.rs
```diff
@@ -41,8 +41,14 @@ impl<'b> Controller<'b> {
...
+        let mut no_errors: bool = true;
```
sharkdp, line 51: I know it was there before, but that `: bool` type annotation is not required.

src/controller.rs
```diff
@@ -41,8 +41,14 @@ impl<'b> Controller<'b> {
 
         let mut output_type = OutputType::from_mode(paging_mode, self.config.pager)?;
         let writer = output_type.handle()?;
-        let mut no_errors: bool = true;
 
+        let no_errors = self.run_with_writer(writer, handle_error);
+
+        Ok(no_errors)
```
sharkdp, line 47: We could simply end the function with:
``` rust
self.run_with_writer(writer, handle_error)
```
(without a trailing semicolon)

PR #911, opened 2020-04-11
Add cat alias detection to info.sh

eth-p:
This should help in diagnosing issues when people have `cat` aliased to `bat`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -69,7 +69,7 @@ name = "atty"
 version = "0.2.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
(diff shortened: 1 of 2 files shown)


sharkdp:
Thank you for looking into this.

This does not detect aliases in `zsh` or `bash`. Only shell functions.

eth-p:
Whoops. I thought I originally accounted for that when I wrote the previous version which this is based on. I'll work on a fix for that, sorry.

eth-p:
Alias detection is now implemented. It tested it on bash, and I believe zsh acts similarly to bash.

sharkdp:
Thank you. I can confirm that it works on `zsh`.

PR #923, opened 2020-04-17
Clarify the Ubuntu/Debian install instructions

dimo414:
A cursory reading of the existing text would lead a reader to think `bat`should be available on Debian, which isn't really true. Most users are not running Sid, so `bat` isn't yet available to them.


sharkdp:
Thank you very much!

Issue #904, opened 2020-04-04
bat in tmux doesn't support mouse scrolling


alexlai97:
`bat` in tmux doesn't support mouse wheel scrolling.
I know `less` can use mouse scrolling by given flag `--mouse`. 
Is `bat` supporting this feature right now ?

sharkdp:
Can you please post the output of the script at https://github.com/sharkdp/bat/blob/master/diagnostics/info.sh in order for us to get more information about your system?

alexlai97:
> Can you please post the output of the script at https://github.com/sharkdp/bat/blob/master/diagnostics/info.sh in order for us to get more information about your system?

system
------

**$ uname -srm**
Linux 5.4.28-2-lts x86_64  

**$ lsb_release -a**
LSB Version:	1.4  
Distributor ID:	Arch  
Description:	Arch Linux  
Release:	rolling  
Codename:	n/a  

bat
---

**$ bat --version**
bat 0.13.0  

**$ env**

bat_config
----------

**$ cat /home/alex/.config/bat/config**
```
#--theme="GitHub"
```

bat_wrapper
-----------

No wrapper script.

bat_wrapper_function
--------------------

No wrapper function.

tool
----

**$ less --version**
less 551 (PCRE regular expressions)

sharkdp:
Thank you for reporting this. I can reproduce it.

You can make this work by adding
``` bash
--pager="less --RAW-CONTROL-CHARS --quit-if-one-screen --mouse"
```
to `bat`s config file (`bat --config-file`).

Interestingly, things work just fine without the `--mouse` option outside of `tmux`.

To be honest, I hadn't heard of the `--mouse` option so far. This is quite new: http://www.greenwoodsoftware.com/less/news.551.html

We have recently added a way to detect the version of `less` within `bat`. I think we should use this to add `--mouse` in versions >= 551.

nicm:
It probably works outside tmux because the terminal in use sends fake Up and Down keys when the application does not support the mouse. tmux doesn't do this (although it is possible to configure it to).

sharkdp:
fixed in #928

sharkdp:
I'm actually going to revert my pull request #928. Adding the `--mouse` option has a huge drawback: you cannot drag-select (and copy) text anymore :frowning_face:.

nicm:
In most terminals you can hold down Shift to bypass application mouse mode.

sharkdp:
[...]

PR #932, opened 2020-04-21
Revert "Pass '--mouse' to less versions >= 551"

sharkdp:
Reverts sharkdp/bat#928

See reason in #904


sharkdp:
I guess we could still do it conditionally (if running from within tmux)

PR #936, opened 2020-04-22
PrettyPrint wrapper, custom 'Read' input, bugfixes, major refactoring

### LINKED ISSUE #729: Library API: Input Source

### LINKED ISSUE #730: Library API: Decorations
I've been thinking of ideas for the library API that can provide more powerful functionality, while also reducing how much of the bat internals are exposed. Feedback would be appreciated.

---

Most of the current implementation for `Decoration` is fine for a public API, but we should remove the parameter that is passing around the internal `InteractivePrinter` struct.

If we choose to implement input source properties as described in #729, this could be easily fixed by using the `HashMap<&'static str, Any>` with builder solution.

I would also suggest that we move the git changes decoration into its own cargo feature (`git-decorations` maybe?), and possibly split off each decoration into its own file.

### LINKED ISSUE #884: Will InputFile be able to take strings in the future?

### LINKED ISSUE #937: "--highlight-line" does not work correctly in combination with "--tabs=0" and "--wrap=never"
```
bat examples/advanced.rs --highlight-line 1:5 --tabs=0 --wrap=never
```
image

I'm already working on a fix.

sharkdp:
This is a major update of *"bat as a library"*. I have implemented a new `PrettyPrinter` struct which is a builder-like wrapper to configure and control the pretty-printing process. I tried to make it similar to [`prettyprint::PrettyPrinter`](https://docs.rs/prettyprint/0.8.1/prettyprint/struct.PrettyPrinter.html) by @mre, but it will not be a drop-in replacement, unfortunately.

**Major new functionality**: the `bat` library can now accept arbitrary `Read`ers as input. This also includes input from a (byte) string.

Here are a few examples of how the library can be used (see `examples` folder for more):

#### `simple.rs`

``` rust
/// A simple program that prints its own source code using the bat library
use bat::PrettyPrinter;

fn main() {
    PrettyPrinter::new().input_file(file!()).print().unwrap();
}
```
image

#### `cat.rs`
``` rust
/// A very simple colorized `cat` clone, using `bat` as a library.
/// See `src/bin/bat` for the full `bat` application.
use bat::PrettyPrinter;

fn main() {
    PrettyPrinter::new()
        .header(true)
        .grid(true)
        .line_numbers(true)
        .input_files(std::env::args_os().skip(1))
        .print()
        .unwrap();
}
```
image

#### `yaml.rs`
``` rust
/// A program that serializes a Rust structure to YAML and pretty-prints the result
use bat::PrettyPrinter;
use serde::Serialize;

#[derive(Serialize)]
struct Person {
    name: String,
    height: f64,
    adult: bool,
    children: Vec<Person>,
}

fn main() {
    let person = Person {
        name: String::from("Anne Mustermann"),
        height: 1.76f64,
        adult: true,
        children: vec![Person {
            name: String::from("Max Mustermann"),
            height: 1.32f64,
            adult: false,
            children: vec![],
        }],
    };

    let bytes = serde_yaml::to_vec(&person).unwrap();
    PrettyPrinter::new()
        .language("yaml")
        .line_numbers(true)
        .input_from_bytes(&bytes)
        .print()
        .unwrap();
}
```
image

#### `advanced.rs`
``` rust
/// A program that prints its own source code using the bat library
use bat::{PagingMode, PrettyPrinter, WrappingMode};

fn main() {
    PrettyPrinter::new()
        .header(true)
        .grid(true)
        .line_numbers(true)
        .use_italics(true)
        // The following line will be highlighted in the output:
        .highlight(line!() as usize)
        .theme("1337")
        .wrapping_mode(WrappingMode::Character)
        .paging_mode(PagingMode::QuitIfOneScreen)
        .input_file(file!())
        .print()
        .unwrap();
}
```
image

closes #884
closes #937 
closes #729
closes #730


sharkdp:
@dandavison Would an API like this work for `delta`?

dandavison:
> @dandavison Would an API like this work for delta?

This looks great! However, regarding delta, the only things it uses from bat are the pager spawning code (`output::OutputType`), the `HighlightingAssets`, and the `to_ansi_color` function (see [comment](https://github.com/sharkdp/bat/issues/729#issuecomment-565125251) for details of delta's usage of bat). I don't _think_ delta could use a pretty printer from bat since what delta has to do is fairly involved (gathering up removed and added lines separately from a diff hunk, running a string alignment algorithm to detect the within-line highlight regions, superimposing foreground color extents from syntect with its own background color extents, etc).

sharkdp:
I see. Than you for the clarification.

PR #940, opened 2020-04-23
bat --diff

### LINKED ISSUE #23: Filter to only include hunks?

sharkdp:
This PR adds a new `--diff` option that can be used to only show lines close to Git changes, i.e. added, removed or modified lines.

The amount of additional context can be controlled with `--diff-context=N`.

Using this like
``` bash
batdiff() {
    git diff --name-only --diff-filter=d | xargs bat --diff
}
```
essentially results in a version of `git diff` with syntax highlighting (but without being able to see the old version).

image

closes #23

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,15 +32,15 @@ matrix:
 
     # Minimum Rust supported channel.
     - os: linux
-      rust: 1.37.0
+      rust: 1.40.0
       env: TARGET=x86_64-unknown-linux-gnu
     - os: linux
-      rust: 1.37.0
+      rust: 1.40.0
       env:
         - TARGET=x86_64-unknown-linux-musl
         - CC_x86_64_unknown_linux_musl=/usr/bin/musl-gcc
     - os: osx
-      rust: 1.37.0
+      rust: 1.40.0
       env: TARGET=x86_64-apple-darwin
 
     # Disable nightly for now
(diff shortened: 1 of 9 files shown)


ErichDonGubler:
YASSS

There is much win here. :)

Mouvedia:
Does it support `strike` for color.diff.old?

sharkdp:
> but without being able to see the old version

no

ErichDonGubler:
> Does it support `strike` for color.diff.old?

No, technically because this doesn't use Git config, but conceptually because...this doesn't display the `old` part of the diff.

eth-p:
This is great! Do you mind if I add a variant of your `batdiff` wrapper to bat-extras?

sharkdp:
> if I add a variant of your `batdiff` wrapper to bat-extras?

of course! :+1:

PR #942, opened 2020-04-24
Fix some clippy warnings

lzutao:
Some expressions should be lazy but it didn't are fixed.

### DIFF
--- a/src/bin/bat/app.rs
+++ b/src/bin/bat/app.rs
@@ -80,6 +80,7 @@ impl App {
         let paging_mode = match self.matches.value_of("paging") {
             Some("always") => PagingMode::Always,
             Some("never") => PagingMode::Never,
+            // FIXME: `_` will always cover in or patterns
             Some("auto") | _ => {
                 if self.matches.occurrences_of("plain") > 1 {
                     // If we have -pp as an option when in auto mode, the pager should be disabled.
@@ -147,6 +148,7 @@ impl App {
                 match self.matches.value_of("wrap") {
                     Some("character") => WrappingMode::Character,
                     Some("never") => WrappingMode::NoWrapping,
+                    // FIXME: `_` will always cover in or patterns
                     Some("auto") | _ => {
                         if style_components.plain() {
                             WrappingMode::NoWrapping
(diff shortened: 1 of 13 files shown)


src/bin/bat/config.rs
```diff
@@ -35,10 +35,10 @@ pub fn generate_config_file() -> bat::error::Result<()> {
         match config_dir {
             Some(path) => fs::create_dir_all(path)?,
             None => {
-                return Ok(Err(format!(
+                return Err(format!(
```
lzutao, line 38: The old code is weird.

src/bin/bat/app.rs
```diff
@@ -80,6 +80,7 @@ impl App {
         let paging_mode = match self.matches.value_of("paging") {
             Some("always") => PagingMode::Always,
             Some("never") => PagingMode::Never,
+            // FIXME: `_` will always cover in or patterns
             Some("auto") | _ => {
```
lzutao, line 84: I don't know what to do about this. I could remove `Some("auto")` but it may change meaning of code.

src/bin/bat/config.rs
sharkdp: Agreed :smile:

sharkdp:

> Does fmt is required?

Yes, running `cargo fmt` would be great.

src/bin/bat/app.rs
sharkdp: It would not change the meaning of the code. Maybe a better alternative would be:
``` rust
  Some("always") => PagingMode::Always,
  Some("never") => PagingMode::Never,
  None | Some("auto") => { … },
  _ => unreachable!(),
```

lzutao: Then `Some("*anything*")` will be reached on unreachable branch. Do you accept that?

sharkdp: Yes. `Some("*anything*")` is `unreachable!()`. The argument parser enforces that:
``` rust
            Arg::with_name("paging")
                // ..
                .possible_values(&["auto", "never", "always"])
```

src/bin/bat/app.rs
```diff
@@ -147,6 +148,7 @@ impl App {
                 match self.matches.value_of("wrap") {
                     Some("character") => WrappingMode::Character,
                     Some("never") => WrappingMode::NoWrapping,
+                    // FIXME: `_` will always cover in or patterns
```
sharkdp, line 151: same reasoning applies here. Please change this in the same way.

sharkdp:
I quickly made the changes myself, as I would like to merge this before making other changes.

sharkdp:
Thank you!

PR #972, opened 2020-05-02
Fix base16, and combine 00 and 0f alpha encodings

mk12:
This changes the base16 theme back from #RRGGBB**0f** to #RRGGBB**00**,
reverting part of #934. That PR used the 0f encoding to produce ANSI
escape sequences 30-37 and 40-47 rather than 38;5 and 48;5 which require
256-color support. Unfortunately, it resulted in base16 using the wrong
colors because ansi_term does not support the bright variants (90-97 and
100-107) so it simply mapped them to the non-bright colors.

This PR makes combines the 00 and 0f alpha encodings into 00, and makes
them use the Color enum for the first 8 colors and Fixed otherwise. This
means the ansi-light and ansi-dark themes will work on terminals without
256-color support, and base16 will render bright colors correctly.

Note: I have not updated themes.bin.

### DIFF
--- a/assets/themes/base16.tmTheme
+++ b/assets/themes/base16.tmTheme
@@ -3,8 +3,8 @@
 <plist version="1.0">
     <dict>
         <!--
-        The colors in this theme are encoded as #RRGGBBAA where RR is an base16
-        color number from 00 to 0f, and AA is the special value 0f to indicate
+        The colors in this theme are encoded as #RRGGBBAA where RR is an ANSI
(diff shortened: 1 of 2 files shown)


sharkdp:
Thank you very much for the fix and the detailed documentation!

> Note: I have not updated themes.bin.

:+1: will do so before preparing the next release.

sharkdp:
Released in [v0.15.1](https://github.com/sharkdp/bat/releases/tag/v0.15.1)

PR #981, opened 2020-05-11
Add padding above headers when no grid

### LINKED ISSUE #968: Extra [vertical] padding around headers when no grid
I like to have `BAT_STYLE=header,numbers,changes`, but when showing multiple files only having a single line header with the file name doesn't visually separate files very well… It would be nice to have the option of a blank line above and/or below the header. Perhaps also an optional blank line instead of `---8<---` with the `snip` option…

pt2121:
closes #968

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 # unreleased
+- Add padding above headers when no grid, see #968 and #981 (@pt2121)
 
 ## Features
 ## Bugfixes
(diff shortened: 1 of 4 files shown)


sharkdp:

I would definitely prefer a version which would not add the blank line before the first file.

pt2121:
Gotcha. I will update my PR. Thanks for the feedback.

sharkdp:
Thank you for the updates. Note that we only want to show the padding if the `header` style component is enabled. We do not want to add padding otherwise.

pt2121:
Thanks again for quick responses. 👍

PR #985, opened 2020-05-12
Unify syntax detection for all variants of InputKind, fix #983

### LINKED ISSUE #983: assets::tests::syntax_detection_invalid_utf8 failed with 0.15.1
I just tried compiling bat 0.15.1 via Nix on macOS and it failed during the tests phase:

```
---- assets::tests::syntax_detection_invalid_utf8 stdout ----
thread 'assets::tests::syntax_detection_invalid_utf8' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 92, kind: Other, message: "Illegal byte sequence" }', src/assets.rs:287:37


failures:
    assets::tests::syntax_detection_invalid_utf8
```

Also of note, the color escapes used by the test runner all stuck a visible `(B` after each colored string. Which is to say, each "ok" was "ok(B" and the "FAILED" was "FAILED(B". Colors printed by rustc itself were fine. I don't know if this is meaningful or just weird though.

eth-p:
This pull request adds a new `InputKind::String` along with `Input::from_string`, and changes a couple things to help fix potential issues and improve the expected syntax detection behavior of `InputKind::CustomReader`/String.

## Additions:
- Adds `InputKind::String` and `Input::from_string` to create files from strings. Currently, this is `pub(crate)` visibility since it was originally added to improve tests. It should be fine to make it public, however.

## Changes:
- Changes the syntax detection behaviour of `InputKind::CustomReader` (and `String`) to work like the syntax detection for a regular file. I feel like `StdIn` should also follow suit, but I didn't want to introduce any significant breaking changes to the CLI in this pull request.

## Fixes:
- Fixes #983 by using `InputKind::String`, so the test files never touch the filesystem. This should also preemptively fix future testing issues related to invalid or unsupported UTF-8 on certain filesystems.


sharkdp:
As you have noticed in the implementation, Rust byte strings implement `Read`. We already use this fact in the high-level API (https://docs.rs/bat/0.15.1/bat/struct.PrettyPrinter.html#method.input_from_bytes):

https://github.com/sharkdp/bat/blob/1a6e8d297ffbfa2e82a20daa9872bac73e9f5005/src/pretty_printer.rs#L89-L92

What do we gain from adding a new enum variant for `OpenedInputKind` and `InputKind`? It seems to duplicate a lot of code.

> Changes the syntax detection behaviour of `InputKind::CustomReader` (and `String`) to work like the syntax detection for a regular file.

What are those changes? What would be examples where the previous behavior would show incorrect results?

> Fixes #983 by using `InputKind::String`, so the test files never touch the filesystem. This should also preemptively fix future testing issues related to invalid or unsupported UTF-8 on certain filesystems.

I see what your motivation is. Removing side effects (OS / file system) to make tests more reliable is something that is always desirable. On the other hand, this change also implies that we are not testing the actual code that will be executed during runtime, right?

eth-p:
> What do we gain from adding a new enum variant for `OpenedInputKind` and `InputKind`? It seems to duplicate a lot of code.

That's a fair point. I was considering eventually migrating the preview theme `InputKind` to use that, but we could just use the `CustomReader` for both.

> What are those changes? What would be examples where the previous behavior would show incorrect results?

image
 
```
// Using the following command line options:
//
// `bat --map-syntax '*.sh:JavaScript' --file-name 'EXAMPLE.sh'`
//
/*
# If we print the file via STDIN (the old CustomReader behavior), then it
# will ignore the syntax mapping and be highlighted as Bash.
# `bat ... < EXAMPLE.sh`
#
# The behavior from passing it in directly as an OrdinaryFile would apply
# the mapping and display this as JavaScript.
# `bat EXAMPLE.sh`
*/
```



> I see what your motivation is. Removing side effects (OS / file system) to make tests more reliable is something that is always desirable. On the other hand, this change also implies that we are not testing the actual code that will be executed during runtime, right?

I did realize that was going to be a problem, and I tried to account for it with the new (in hindsight poorly-named) `syntax_detection_same_for_file_and_string` test. It compares the behavior between the `OrdinaryFile` syntax detection and the `String` detection to make sure that they both detect the same syntax for various test cases. My reasoning behind it was to ensure that even if we avoided touching the filesystem in the other tests, we would still be able to get the same results as if we had used the filesystem.

https://github.com/sharkdp/bat/pull/985/files#diff-ba290b21df65b517c34a37e667a84c1eR395-R415

sharkdp:
> Looking at the original routine, it was subtly different using an `OrdinaryFile` compared to one of the other input types with `--file-name` passed in. It didn't take path canonicalization into consideration (e.g. `cd /etc && bat --input-file="passwd" < passwd` should consider the file it's working on to be `/etc/passwd` instead of `passwd`), and it didn't use any of the user-defined syntax mappings.
> 
> Take this file for example:

Oh, right. That should be fixed. Thanks! For STDIN as well. Opening a new issue/PR would be great, but we can also handle it here, if you like. Adding regression tests according to the examples you designed would be fantastic.

> I did realize that was going to be a problem, and I tried to account for it with the new (in hindsight poorly-named) `syntax_detection_same_for_file_and_string` test. It compares the behavior between the `OrdinaryFile` syntax detection and the `String` detection to make sure that they both detect the same syntax for various test cases.

I see. Hm.. but then we still rely on the file system for the whole test setup to work properly. And we rely on these "same for file and string" tests to be exhaustive and to be maintained in the future.

In principle, I like your idea of moving away from the filesystem because it would also make the tests much faster. Maybe the best way forward would be to further unify the code between the different enum variants in "get_syntax" such that we do not have multiple code paths implementing the same logic. This way, we could maybe move away from filesystem-dependent tests completely (we still have integration tests in another place).

> Should I update the PR to remove the `String` kind

I think that would make the code simpler overall, yes.

eth-p:
A funny thing I found in the 0.15.1 release while working on this PR:

image

I'm going to fix that...

sharkdp:
> A funny thing I found in the 0.15.1 release while working on this PR:

oops :-)

eth-p:
Changes are done. All the `InputKind` variants now roughly follow the same logic for detecting the syntax:

1. If `--list-themes`, use the Rust syntax.
2. If `--language=` is provided, use that.
3. Detect the syntax based on the file name (`--file-name` first, real file path second)
4. Fall back to the detected syntax from reading the first line

I also added the regression test. I encountered a really frustrating issue related to my IDE and unicode normalization, and I'm going to make another issue about it.

src/assets.rs
```diff
@@ -191,42 +191,50 @@ impl HighlightingAssets {
         input: &mut OpenedInput,
         mapping: &SyntaxMapping,
     ) -> &SyntaxReference {
-        let syntax = if let Some(language) = language {
+        let syntax = if match input.kind {
+            OpenedInputKind::ThemePreviewFile => true,
+            _ => false,
+        } {
+            // FIXME: replace the above match statement with matches macro when min Rust >= 1.42.0
```
sharkdp, line 198: :+1: 

In the meantime, could we move this logic to a helper function in the `impl OpenedInputKind` block? This would be much easier to read:

```
let syntax = if input.kind.is_theme_preview_file() {
```

tests/integration_tests.rs
```diff
@@ -670,3 +674,38 @@ fn do_not_panic_regression_tests() {
...
+        .pipe_stdin(Path::new(EXAMPLES_DIR).join(file))
+        .unwrap()
+        .assert()
+        .success()
+        .get_output()
+        .stdout
+        .clone();
+
+    assert_eq!(
+        from_utf8(&output_for_file).unwrap(),
```
sharkdp, line 708: couldn't this be `output_for_file.from_utf8().unwrap()`, without the `std::str::from_utf8` import?

sharkdp: minor: I try to avoid `unwrap` if possible. In test code, it's fine, but I personally prefer `.expect("…")` with a message.

src/assets.rs
```diff
@@ -319,6 +344,22 @@ mod tests {
                 .get_syntax(None, &mut opened_input, &self.syntax_mapping);
             syntax.name.clone()
         }
+
+        fn syntax_is_same_for_inputkinds(&self, file_name: &str, content: &str) -> bool {
+            let as_file = self.syntax_for_real_file_with_content_os(file_name.as_ref(), content);
+            let as_reader = self.syntax_for_file_with_content_os(file_name.as_ref(), content);
+            let consistent = as_file == as_reader;
+            // TODO: Compare StdIn somehow?
```
sharkdp, line 352: maybe we can (later) get rid of `StdIn` and move it to `Input::from_reader`?

tests/integration_tests.rs
```diff
@@ -670,3 +674,38 @@ fn do_not_panic_regression_tests() {
...
+        .arg("--color=always")
+        .arg("--map-syntax=*.js:Markdown")
+        .arg(&format!("--file-name={}", file))
+        .arg("--style=plain")
+        .arg(file)
+        .assert()
+        .success()
+        .get_output()
+        .stdout
+        .clone();
```
sharkdp, line 692: not that it matters too much here, but you can avoid the `.clone` by using:
``` rust
    let cmd_for_file = bat()
        .arg("--color=always")
        .arg("--map-syntax=*.js:Markdown")
        .arg(&format!("--file-name={}", file))
        .arg("--style=plain")
        .arg(file)
        .assert()
        .success();
    let output_for_file = &cmd_for_file.get_output().stdout;
```
similar below.

sharkdp:
Apart from a few minor comments, this looks great - Thank you!

src/assets.rs
eth-p: We will need to make a couple changes to `InputDescription` to allow for custom descriptions, but that would be a good idea. We likely would also get rid of the theme preview file input in the same PR.

tests/integration_tests.rs
eth-p: Great suggestion! Fixed

tests/integration_tests.rs
sharkdp: Oh, right. Sorry for the noise :-)

sharkdp:
Feel free to merge this when the tests have passed :+1:

sharkdp:
Oh, and it would be great if we can update the CHANGELOG as well.

eth-p:
Merged. I didn't include an issue number in the changelog since it didn't have a relevant issue, and only came to light because of this PR.

PR #997, opened 2020-05-13
Allow 'bat' application to built without git feature, remove PrettyPrint git support when feature not enabled.

eth-p:
This pull request changes two things:

1. *(bat-as-a-library; breaking change)*
When the `git` feature is not enabled, it removes `PrettyPrinter::vcs_modification_markers`. This function was previously being exposed when the feature was disabled, despite not having any effect when used.

2. *(bat)*
Adds `cfg` flags to allow the `bat` application to build without requiring the `git` feature. This would allow people to build `bat` without git support if they don't feel they would need or use it.
Caveats: Unfortunately, disabling the feature doesn't change the help text or generated manpage.

### DIFF
--- a/src/bin/bat/app.rs
+++ b/src/bin/bat/app.rs
@@ -198,22 +198,23 @@ impl App {
                     }
                 })
                 .unwrap_or_else(|| String::from(HighlightingAssets::default_theme())),
-            visible_lines: if self.matches.is_present("diff") {
-                VisibleLines::DiffContext(
+            visible_lines: match self.matches.is_present("diff") {
+                #[cfg(feature = "git")]
+                true => VisibleLines::DiffContext(
                     self.matches
                         .value_of("diff-context")
                         .and_then(|t| t.parse().ok())
                         .unwrap_or(2),
-                )
-            } else {
-                VisibleLines::Ranges(
+                ),
+
+                _ => VisibleLines::Ranges(
                     self.matches
                         .values_of("line-range")
                         .map(|vs| vs.map(LineRange::from).collect())
                         .transpose()?
                         .map(LineRanges::from)
                         .unwrap_or_default(),
-                )
+                ),
             },
             style_components,
             syntax_mapping,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@
 ## New syntaxes
 ## New themes
 ## `bat` as a library
+
+- `PrettyPrinter::vcs_modification_markers` is no longer available without the `git` feature, see #997 (@eth-p)
+
 ## Packaging
 
 # v0.15.1
--- a/src/bin/bat/clap_app.rs
+++ b/src/bin/bat/clap_app.rs
@@ -8,7 +8,7 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
         AppSettings::ColorNever
     };
 
-    let app = ClapApp::new(crate_name!())
+    let mut app = ClapApp::new(crate_name!())
         .version(crate_version!())
         .global_setting(clap_color_setting)
         .global_setting(AppSettings::DeriveDisplayOrder)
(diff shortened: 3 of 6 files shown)


sharkdp:
>     2\. Adds `cfg` flags to allow the `bat` application to build without requiring the `git` feature. This would allow people to build `bat` without git support if they don't feel they would need or use it.

Thank you, looks good to me!

>     1. When the `git` feature is not enabled, it removes `PrettyPrinter::vcs_modification_markers`. This function was previously being exposed when the feature was disabled, despite not having any effect when used.

Okay.

>     2\. Caveats: Unfortunately, disabling the feature doesn't change the help text or generated manpage.

I guess that's something we can live with. Help text (`--help`) should be changed, right?


I have two comments:
1. The `match … { true => …, … }` part looks weird and is hard to read, in my opinion. I will add an inline comment
2. Could you please adapt the `CHANGELOG.md` accordingly? I am now trying to always keep this up to date in order to facilitate the release process. I guess we should add a PR template / CONTRIBUTING.md file to explain this to all contributors.

For now, I have set up a "saved reply" along the lines of:

> Could you please add an entry to the "unreleased" section in the `CHANGELOG.md`? Under "Features" or "Other". The format for entries is:
> ```
> - Description what has been changed, see #123 (@user)
> ```
> where `#123` links to the bug ticket and/or this PR and `user` is your username.

src/bin/bat/app.rs
```diff
@@ -198,22 +198,23 @@ impl App {
                     }
                 })
                 .unwrap_or_else(|| String::from(HighlightingAssets::default_theme())),
-            visible_lines: if self.matches.is_present("diff") {
-                VisibleLines::DiffContext(
+            visible_lines: match self.matches.is_present("diff") {
+                #[cfg(feature = "git")]
+                true => VisibleLines::DiffContext(
```
sharkdp, line 203: I see why you chose to use a `match` statement here, but this is a bit difficult to read. I don't know a better solution of the top of my head, though.

I guess `if cfg!(…)` does not work because `DiffContext` is not available if `git` is not set?

eth-p:
> I guess that's something we can live with. Help text (--help) should be changed, right?

Sort of. It removes the related entries, but it doesn't remove any references to them from the descriptions of other arguments (e.g. the description for `--style`). I wasn't sure of how we could do that without writing multiple different help texts and `#[cfg]`ing each of them, and I'm not a fan of doing that because of the maintenance overhead and potential to forget to update one of the different description variants.

> Could you please adapt the CHANGELOG.md accordingly? I am now trying to always keep this up to date in order to facilitate the release process. I guess we should add a PR template / CONTRIBUTING.md file to explain this to all contributors.

For future reference, does this include testing or developer-only (internal, not `bat-as-a-library`) changes?

sharkdp:
> > I guess that's something we can live with. Help text (--help) should be changed, right?
> 
> Sort of. It removes the related entries, but it doesn't remove any references to them from the descriptions of other arguments (e.g. the description for `--style`). I wasn't sure of how we could do that without writing multiple different help texts and `#[cfg]`ing each of them, and I'm not a fan of doing that because of the maintenance overhead and potential to forget to update one of the different description variants.

I agree.

> For future reference, does this include testing or developer-only (internal, not `bat-as-a-library`) changes?

Good question. I would exclude everything that is not of interest to (1) users of `bat` or (2) users of `bat` as a library.

src/bin/bat/app.rs
eth-p: > I guess if cfg!(…) does not work because DiffContext is not available if git is not set?
Pretty much. It's a bit difficult to read, but it was the smallest change to make it work like so:

1. If `--diff` is present with git support, use the diff context.
2. Otherwise, use the `--line-range` option.

eth-p:
Rebased and added a changelog update. I didn't mention building `bat` without support for the git feature since it's not relevant for the majority of `bat` users, and would require them to manually compile the binary.

src/bin/bat/app.rs
sharkdp: Ok

sharkdp:
Released in [bat v0.15.2](https://github.com/sharkdp/bat/releases/tag/v0.15.2).

PR #1002, opened 2020-05-15
Make InputDescription into API, remove special input kind for theme previews.

eth-p:
This pull request refactors the `InputDescription` struct into something consumable in a public API, and removes all special handling for the themes preview file.

## Example

```rust
PrettyPrinter::new()
    .grid(true)
    .header(true)
    .input(
        Input::from_reader(Box::new(BufReader::new("Hello world!\n".as_bytes())))
            .with_description(Some(
                InputDescription::new("String")
                    .with_kind(Some("built-in")),
            )),
    )
    .print()
    .unwrap();
```

image

## Breaking Changes
- `InputKind::ThemePreviewFile` no longer exists.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@
 
 - `PrettyPrinter::vcs_modification_markers` has been marked deprecated when building without the `git` feature, see #997 and #1020 (@eth-p, @sharkdp)
 - Add APIs to provide `Input` descriptions with `InputDescription` (@eth-p)
+- Add function to directly provide `Input`s to `PrettyPrinter` (@eth-p)
 
 ## Packaging
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@
 ## `bat` as a library
 
 - `PrettyPrinter::vcs_modification_markers` has been marked deprecated when building without the `git` feature, see #997 and #1020 (@eth-p, @sharkdp)
+- Add APIs to provide `Input` descriptions with `InputDescription` (@eth-p)
 
 ## Packaging
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@
 ## New syntaxes
 ## New themes
 ## `bat` as a library
+
+- Add APIs to provide `Input` descriptions with `InputDescription` (@eth-p)
+- Add function to directly provide `Input`s to `PrettyPrinter` (@eth-p)
+- <font color="red">`Input::theme_preview_file` is no longer available.</font> (@eth-p)
+
+Changes colored <font color="red">red</font> are breaking changes.
+
 ## Packaging
 
 # v0.15.4
@@ -46,11 +53,6 @@
 ## `bat` as a library
 
 - `PrettyPrinter::vcs_modification_markers` has been marked deprecated when building without the `git` feature, see #997 and #1020 (@eth-p, @sharkdp)
-- Add APIs to provide `Input` descriptions with `InputDescription` (@eth-p)
-- Add function to directly provide `Input`s to `PrettyPrinter` (@eth-p)
-- <font color="red">`Input::theme_preview_file` is no longer available.</font> (@eth-p)
-
-Changes colored <font color="red">red</font> are breaking changes.
 
 ## Packaging
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,7 @@
 
 - Add APIs to provide `Input` descriptions with `InputDescription` (@eth-p)
 - Add function to directly provide `Input`s to `PrettyPrinter` (@eth-p)
-- <font color="red">`Input::theme_preview_file` is no longer available.</font> (@eth-p)
-
-Changes colored <font color="red">red</font> are breaking changes.
+- **Breaking:** `Input::theme_preview_file` is no longer available. (@eth-p)
 
 ## Packaging
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,9 @@
 - `PrettyPrinter::vcs_modification_markers` has been marked deprecated when building without the `git` feature, see #997 and #1020 (@eth-p, @sharkdp)
 - Add APIs to provide `Input` descriptions with `InputDescription` (@eth-p)
 - Add function to directly provide `Input`s to `PrettyPrinter` (@eth-p)
+- <font color="red">`Input::theme_preview_file` is no longer available.</font> (@eth-p)
+
+Changes colored <font color="red">red</font> are breaking changes.
 
 ## Packaging
 
--- a/examples/inputs.rs
+++ b/examples/inputs.rs
@@ -9,9 +9,9 @@ fn main() {
         .line_numbers(true)
         .inputs(vec![
             Input::from_bytes(b"echo 'Hello World!'")
-                .name("embedded.sh")
-                .title("An embedded shell script.")
-                .kind("Embedded"),
+                .name("embedded.sh") // Dummy name provided to detect the syntax.
+                .kind("Embedded")
+                .title("An embedded shell script."),
             Input::from_stdin().title("Standard Input").kind("FD"),
         ])
         .print()
--- a/examples/inputs.rs
+++ b/examples/inputs.rs
@@ -1,0 +1,19 @@
+/// A small demonstration of the Input API.
+/// This prints embedded bytes with a custom header and then reads from STDIN.
+use bat::{Input, PrettyPrinter};
+
+fn main() {
+    PrettyPrinter::new()
+        .header(true)
+        .grid(true)
+        .line_numbers(true)
+        .inputs(vec![
+            Input::from_bytes(b"echo 'Hello World!'")
+                .name("embedded.sh")
+                .title("An embedded shell script.")
+                .kind("Embedded"),
+            Input::from_stdin().title("Standard Input").kind("FD"),
+        ])
+        .print()
+        .unwrap();
+}
--- a/src/bin/bat/app.rs
+++ b/src/bin/bat/app.rs
@@ -257,7 +257,7 @@ impl App {
         let files: Option<Vec<&OsStr>> = self.matches.values_of_os("FILE").map(|vs| vs.collect());
 
         if files.is_none() {
-            let input = Input::stdin().as_file(filenames_or_none.next().unwrap_or(None));
+            let input = Input::stdin_as_file(filenames_or_none.next().unwrap_or(None));
             return Ok(vec![input]);
         }
         let files_or_none: Box<dyn Iterator<Item = _>> = match files {
@@ -269,9 +269,9 @@ impl App {
         for (filepath, provided_name) in files_or_none.zip(filenames_or_none) {
             if let Some(filepath) = filepath {
                 if filepath.to_str().unwrap_or_default() == "-" {
-                    file_input.push(Input::stdin().as_file(provided_name));
+                    file_input.push(Input::stdin_as_file(provided_name));
                 } else {
-                    file_input.push(Input::ordinary_file(filepath).as_file(provided_name));
+                    file_input.push(Input::ordinary_file(filepath).with_name(provided_name));
                 }
             }
         }
--- a/src/bin/bat/app.rs
+++ b/src/bin/bat/app.rs
@@ -257,7 +257,7 @@ impl App {
         let files: Option<Vec<&OsStr>> = self.matches.values_of_os("FILE").map(|vs| vs.collect());
 
         if files.is_none() {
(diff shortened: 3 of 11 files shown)


sharkdp:
I very much like the fact that the theme preview file is removed from the library. I also like the new `InputDescription` structure, but I'm not so sure if we should really make this public and add a new `pub fn input(&mut self, input: Input<'a>) -> &mut Self` on the `PrettyPrinter`. Sure, it adds a lot of flexibility, but it also increases the API surface.

So far, my long-term goal was to remove more and more parts from the "low level" public API (things that are not directly related to the `PrettyPrinter` struct), which would include things like `bat::input::Input`.

With this change, we would basically move the  `Input` struct to the "high level" API, contrary to this goal.

eth-p:
> So far, my long-term goal was to remove more and more parts from the "low level" public API (things that are not directly related to the PrettyPrinter struct), which would include things like bat::input::Input.

I agree that aiming to reduce API surface is a good long term goal, but I believe I can make a good case for why we should directly go against that by moving `bat::input::Input` into the public and making it a stable part of the `bat`-as-a-library API. 

1. While the current `PrettyPrinter` struct serves extremely well as a pretty printer, I don't believe that it has the flexibility by itself to grow along with `bat`-as-a-library and the features that we might add in the future. The input methods that it exposes won't scale well if/when we start adding more metadata to files.
&nbsp;
For example, consider how we currently have two variants of `input_*`: `input_*`, and `input_*_with_name`. If we decide to ever add the ability to manually specify the syntax of an individual input, we would need another two variants for each of those (`...with_syntax`, `...with_name_and_syntax`). And there's still a lot of internal-only metadata that we could make configurable, such as per-file line ranges, diffs, etc.
&nbsp;
Although exposing the `Input` struct does increase the API surface, if we keep the internal details hidden away and only use a builder pattern for it, it provides lots of flexibility without much more of a maintenance burden.

2. One of my goals when working on `bat`-as-a-library is creating a simple, but not overly-restrictive way to use the best parts of `bat` in other projects. Although this is just a small step, having more control over the input files and their metadata opens up a whole new realm of possibilities for other programs that consume it.
&nbsp;
Even the smaller changes like the `InputDescription` API serve a good purpose. Instead of showing `READER file_name` in the header, it can now accurately reflect the true source of the file by showing things such as `FTP file@server:/path` or `CIFS \\NAS\Share\File`.

sharkdp:
> The input methods that it exposes won't scale well if/when we start adding more metadata to files.
>
> For example, consider how we currently have two variants of `input_*`: `input_*`, and `input_*_with_name`. If we decide to ever add the ability to manually specify the syntax of an individual input, we would need another two variants for each of those (`...with_syntax`, `...with_name_and_syntax`)

Agreed. Good point. In this, case, would it make sense to add an additional wrapping layer around the "internal" `Input` struct. Maybe by using an actual builder for `Input`? This way, we could also make it a bit easier to use this part of the API. For example, I would propose that we do not take `Option<T>` as arguments to configure an internal `Option<T>` member. Instead, keep the member optional but provide a "setter" that takes a `T`. This would remove the two `Some(…)`s from your example (I know that there are other examples where taking an `Option<T>` would make things easier, but taking `T` seems more natural to me).

Also, instead of exposing `InputDescription`, we could provide two or three methods in the builder that could be called individually.

Similarly, we could do the `Box` wrapping for the user.

In the end, I would like the example to look more like:
``` rust
PrettyPrinter::new()
    .grid(true)
    .header(true)
    .input(
        Input::from_reader(b"Hello world!\n")
            .kind("HTTP request")
            .title("http://example.com/script.js")
            .language("JavaScript")
    )
    .print()
    .unwrap();
```
(method names are open for discussion)

If we decide to add these methods, should we remove/deprecate the old `input_*` ones?

eth-p:
> Agreed. Good point. In this case, would it make sense to add an additional wrapping layer around the "internal" `Input` struct. Maybe by using an actual builder for `Input`? 

That works. Should we include it in the `pretty_print` module?


> For example, I would propose that we do not take `Option<T>` as arguments to configure an internal `Option<T>` member.

I agree. I would've actually just done that in the original pull request, but I wanted to keep it consistent with the PrettyPrinter interface.

> Also, instead of exposing `InputDescription`, we could provide two or three methods in the builder that could be called individually.

Since we could make a proper builder for it, that would be the more ideal way of handling it I think.
 
> Similarly, we could maybe do the `Box` wrapping for the user.

Good idea.


> In the end, I would like the example to look more like:
> ``` rust
> PrettyPrinter::new()
>     .grid(true)
>     .header(true)
>     .input(
>         Input::from_reader(b"Hello world!\n")
>             .kind("HTTP request")
>             .title("http://example.com/script.js")
>             .language("JavaScript")
>     )
>     .print()
>     .unwrap();
> ```
> (method names are open for discussion)

It looks good, but we should also take `.name` for the file name. The title is purely descriptive, while the name would influence the syntax detection.

We would also need to do the `.language`  function in another pull request. I think ideally, it would be a bit more structured (e.g. `Syntaxes.from_name("JavaScript") -> Result<Syntax>` where Syntax is just a small struct over a String), but that's a discussion for later.

> If we decide to add these methods, should we remove/deprecate the old `input_*` ones? 

I think it would be a good idea to deprecate them in the next release after this is merged, and then remove them one or two releases after?

sharkdp:
> That works. Should we include it in the `pretty_print` module?

Yes, or in a separate file and re-export from `pretty_print`.

> It looks good, but we should also take `.name` for the file name. The title is purely descriptive, while the name would influence the syntax detection.

:+1: 

> We would also need to do the `.language` function in another pull request. I think ideally, it would be a bit more structured (e.g. `Syntaxes.from_name("JavaScript") -> Result<Syntax>` where Syntax is just a small struct over a String), but that's a discussion for later.

Yes, that was only intended as an example.

> I think it would be a good idea to deprecate them in the next release after this is merged, and then remove them one or two releases after?

Okay. Maybe we could even keep the ones without the `_with_name` suffix for now for convenience. We could reimplement them using `.input(…)`.

eth-p:
Changes are done, but I had to make a breaking API change to get the new builder pattern working without creating differing behavior based on the order of functions called:

`bat::input::Input::with_name` no longer sets the kind to "File".

For the preexisting methods, I have them set the kind to "File" to preserve compatibility, but any any usages of the new `Input` API (including files) will not have an implicit kind provided by the constructor. I believe that this would be for the best anyways, since it's more consistent than having defaults that vary based on which `bat::pretty_printer::Input::from_*` was used.

CHANGELOG.md
```diff
@@ -18,6 +18,11 @@
 ## `bat` as a library
 
 - `PrettyPrinter::vcs_modification_markers` is no longer available without the `git` feature, see #997 (@eth-p)
+- Add APIs to provide `Input` descriptions with `InputDescription` (@eth-p)
+- Add function to directly provide `Input`s to `PrettyPrinter` (@eth-p)
+- <font color="red">`Input::theme_preview_file` is no longer available.</font> (@eth-p)
+
+Changes colored <font color="red">red</font> are breaking changes.
```
sharkdp, line 25: :+1: should the first one (vcs_modification_markers) also be marked in red?

examples/inputs.rs
```diff
@@ -0,0 +1,19 @@
...
+fn main() {
+    PrettyPrinter::new()
+        .header(true)
+        .grid(true)
+        .line_numbers(true)
+        .inputs(vec![
+            Input::from_bytes(b"echo 'Hello World!'")
+                .name("embedded.sh")
+                .title("An embedded shell script.")
+                .kind("Embedded"),
```
sharkdp, line 14: minor: maybe we could call `.kind` before `.title` in the same way as the output order?

The `.name(…)` part here is needed for syntax detection? Maybe we could leave it out and add `#!/bin/bash` instead? Or add a comment why the dummy name is needed?

In the future, we could probably also replace it with `.language(…)`.

src/bin/bat/app.rs
```diff
@@ -257,7 +257,7 @@ impl App {
         let files: Option<Vec<&OsStr>> = self.matches.values_of_os("FILE").map(|vs| vs.collect());
 
         if files.is_none() {
-            let input = Input::stdin().with_name(filenames_or_none.next().unwrap_or(None));
+            let input = Input::stdin().as_file(filenames_or_none.next().unwrap_or(None));
```
sharkdp, line 260: I liked the previous name better. Without looking at the definition, it's not clear what `as_file` does and what the argument should be. Maybe `.as_file_with_name(…)` or `.kind("File").with_name(…)`? Or `.as_file().with_name(…)`?

src/bin/bat/main.rs
```diff
@@ -118,11 +120,17 @@ pub fn list_languages(config: &Config) -> Result<()> {
     Ok(())
 }
 
+fn theme_preview_file<'a>() -> Input<'a> {
+    Input::from_reader(Box::new(BufReader::new(THEME_PREVIEW_DATA)))
+        .with_name(Some("theme.rs".as_ref()))
```
sharkdp, line 125: This is now needed for syntax detection?

src/input.rs
```diff
@@ -6,22 +6,82 @@ use content_inspector::{self, ContentType};
...
+    pub fn new(name: impl Into<String>) -> Self {
+        InputDescription {
+            name: name.into(),
+            title: None,
+            kind: None,
+            summary: None,
+        }
+    }
+
+    pub fn set_kind(&mut self, kind: Option<String>) -> () {
```
sharkdp, line 38: `-> ()` is not needed.

src/input.rs
```diff
@@ -6,22 +6,82 @@ use content_inspector::{self, ContentType};
...
+    }
+
+    pub fn kind(&self) -> Option<&String> {
+        self.kind.as_ref()
+    }
+
+    pub fn summary(&self) -> String {
+        self.summary.clone().unwrap_or_else(|| match &self.kind {
+            None => self.name.clone(),
+            Some(kind) => format!("{} '{}'", kind.to_lowercase(), self.name),
```
sharkdp, line 64: Why do we call `.to_lowercase` here? Wouldn't it be more flexible if we leave this out?

src/pretty_printer.rs
```diff
@@ -9,7 +9,7 @@ use crate::{
     config::{Config, VisibleLines},
     controller::Controller,
     error::Result,
-    input::Input,
+    input::Input as BatInput,
```
sharkdp, line 12: why is this needed?

src/input.rs
sharkdp: Oh, because `summary` is only needed in the error message. Maybe we should rewrite that error message a bit and get rid of `.summary`? Is it needed anywhere else?

sharkdp:
With this change, input from STDIN will show "File: STDIN" in the header. It was only "STDIN" before.

examples/inputs.rs
eth-p: Yeah, the name is needed for syntax detection. In the future we definitely should add a `.language` option to replace it, though.

src/bin/bat/main.rs
eth-p: Not particularly. I'll remove it

src/bin/bat/app.rs
eth-p: I'll see if `.as_file()` could work.

src/input.rs
eth-p: It isn't needed elsewhere, but it might be useful to keep around for now in case we do use it.

src/bin/bat/app.rs
eth-p: I ended up just removing `as_file` entirely, since we only ever really use it for pretending a STDIN input is a file. In its place, I added `Input::stdin_as_file(Option)`, which describes the purpose a bit clearer.

In case we ever use reader inputs along with `--file-name` in the `bat` application itself, we can always revert it back to being a separate function, but for now, I think this is most straightforward way to implement it.

It also fixes the `File: STDIN` issue mentioned below.

src/input.rs
eth-p: Fixed

src/pretty_printer.rs
eth-p: Renamed to `input::Input`

CHANGELOG.md
```diff
@@ -6,6 +6,13 @@
 ## New syntaxes
 ## New themes
 ## `bat` as a library
+
+- Add APIs to provide `Input` descriptions with `InputDescription` (@eth-p)
+- Add function to directly provide `Input`s to `PrettyPrinter` (@eth-p)
+- <font color="red">`Input::theme_preview_file` is no longer available.</font> (@eth-p)
+
+Changes colored <font color="red">red</font> are breaking changes.
```
sharkdp, line 14: Unfortunately, GitHub Markdown removes these `<font>` tags. Maybe we just use something like:
* **Breaking**: …

PR #1020, opened 2020-05-24
Make 'vcs_modification_markers' change non-breaking

sharkdp:
I'd like to release a new version of `bat` without breaking changes in the library. @eth-p what do you think?

We could still remove it later when introducing other breaking changes as well.


sharkdp:
Good idea - thanks!

Issue #951, opened 2020-04-26
Improve bat startup speed


sharkdp:
The startup speed of `bat` is currently (v0.15) around 50 ms, which is on (or even past) the edge of being noticeable by humans.

It would be great if this situation could be improved.

The startup speed of `bat` can be measured with [`hyperfine`](github.com/sharkdp/hyperfine):
``` bash
hyperfine \
  --warmup 10 \
  --export-markdown bat_startup.md \
  'bat --no-config --color=always'
```

On my laptop, this results in:

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `bat --no-config --color=always` | 45.3 ± 0.7 | 44.0 | 47.4 | 1.00 |

If we use `perf` to get a profile
``` bash
perf record --call-graph dwarf bat --no-config --color=always < /dev/null
```

![image](https://user-images.githubusercontent.com/4209276/80308095-455d2880-87cd-11ea-8af9-915eb9bb4329.png)

we can see that most of the time is spent deserializing the stored syntaxes and themes via `bat::assets::assets_from_cache_or_binary`.

There are several ideas that come to mind which could improve the situation:

* Somehow parallelize the deserialization process (maybe by splitting up the binaries into smaller chunks. After all, they are made out of ~100 different syntaxes and themes)
* Only load syntaxes and themes "on demand". If syntaxes and themes were serialized separately, we might be able to load them on demand. It's probably not as easy as it sounds, because syntaxes are interlinked (code blocks in markdown files could require the loading of all kinds of syntaxes, for example).
* Try to look into ways to improve the deserialization speed.
* The assets are currently stored in compressed form. I think I have tried storing them in uncompressed form once (and didn't see any performance improvements), but it might be worth re-evaluating that.

We should also validate that we are indeed CPU bound here. The `bat` binary is quite large, due to the fact that all syntaxes and themes are included within the binary. It might take some time to simply load that from disk (or cache).

eth-p:
Adding some profiling data from MacOS Mojave (with a SSD):

Note: The numbers are going to be off (i.e. about 2x worse) due to the added overhead from running a profiler.

**Build:** `RUSTFLAGS=-g cargo build --release`
**Command:** `target/release/bat --no-config --paging=never empty.cs`

**Total:**
![image](https://user-images.githubusercontent.com/32112321/80346226-41f28d00-881f-11ea-9382-5b494cb5b04b.png)

[...]

PR #1023, opened 2020-05-26
Enable GitHub-Actions CI and deployment

### LINKED ISSUE #978: Github actions integration
Is there any interest in using Github actions for this Project? I had a nice experience using them and I would set up one for testing and build automation.

rivy:
Fixes #978

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -1,0 +1,287 @@
+name: CICD
+
+env:
+  PROJECT_NAME: bat
+  PROJECT_DESC: "A `cat` clone with wings"
+  PROJECT_AUTH: "sharkdp"
+  RUST_MIN_SRV: "1.40.0"
+
+on: [push, pull_request]
+
+jobs:
+  min_version:
+    name: MinSRV # Minimum supported rust version
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Install `rust` toolchain (v${{ env.RUST_MIN_SRV }})
+      uses: actions-rs/toolchain@v1
+      with:
+        toolchain: ${{ env.RUST_MIN_SRV }}
+        default: true
+        profile: minimal # minimal component installation (ie, no documentation)
+    - name: Test
+      uses: actions-rs/cargo@v1
+      with:
+        command: test
+
+  build:
+    name: Build
+    runs-on: ${{ matrix.job.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        job:
+          # { os, target, cargo-options, features, use-cross, toolchain }
+          - { os: ubuntu-latest  , target: arm-unknown-linux-gnueabihf , use-cross: use-cross }
+          - { os: ubuntu-18.04   , target: aarch64-unknown-linux-gnu   , use-cross: use-cross }
+          - { os: ubuntu-18.04   , target: i586-unknown-linux-gnu      , use-cross: use-cross }
+          - { os: ubuntu-18.04   , target: i686-unknown-linux-gnu      , use-cross: use-cross }
+          - { os: ubuntu-18.04   , target: i686-unknown-linux-musl     , use-cross: use-cross }
+          - { os: ubuntu-18.04   , target: x86_64-unknown-linux-gnu    , use-cross: use-cross }
+          - { os: ubuntu-18.04   , target: x86_64-unknown-linux-musl   , use-cross: use-cross }
+          - { os: ubuntu-16.04   , target: x86_64-unknown-linux-gnu    , use-cross: use-cross }
+          - { os: macos-latest   , target: x86_64-apple-darwin         }
+          - { os: windows-latest , target: i686-pc-windows-gnu         }
+          - { os: windows-latest , target: i686-pc-windows-msvc        }
+          - { os: windows-latest , target: x86_64-pc-windows-gnu       }  ## note: requires rust >= 1.43.0 to link correctly
+          - { os: windows-latest , target: x86_64-pc-windows-msvc      }
+    steps:
+    - uses: actions/checkout@v1
+    - name: Install any prerequisites
+      shell: bash
+      run: |
+        case ${{ matrix.job.target }} in
+          arm-unknown-linux-gnueabihf) sudo apt-get -y update ; sudo apt-get -y install gcc-arm-linux-gnueabihf ;;
+          aarch64-unknown-linux-gnu) sudo apt-get -y update ; sudo apt-get -y install gcc-aarch64-linux-gnu ;;
+        esac
+    - name: Initialize workflow variables
+      id: vars
+      shell: bash
+      run: |
+        # toolchain
+        TOOLCHAIN="stable" ## default to "stable" toolchain
+        # * specify alternate TOOLCHAIN for *-pc-windows-gnu targets; gnu targets on Windows are broken for the standard *-pc-windows-msvc toolchain (refs: <https://github.com/rust-lang/rust/issues/47048>, <https://github.com/rust-lang/rust/issues/53454>, <https://github.com/rust-lang/cargo/issues/6754>)
+        case ${{ matrix.job.target }} in *-pc-windows-gnu) TOOLCHAIN="stable-${{ matrix.job.target }}" ;; esac;
+        # * use requested TOOLCHAIN if specified
+        if [ -n "${{ matrix.job.toolchain }}" ]; then TOOLCHAIN="${{ matrix.job.toolchain }}" ; fi
+        echo set-output name=TOOLCHAIN::${TOOLCHAIN}
+        echo ::set-output name=TOOLCHAIN::${TOOLCHAIN}
+        # staging directory
+        STAGING='_staging'
+        echo set-output name=STAGING::${STAGING}
+        echo ::set-output name=STAGING::${STAGING}
+        # determine EXE suffix
+        EXE_suffix="" ; case ${{ matrix.job.target }} in *-pc-windows-*) EXE_suffix=".exe" ;; esac;
+        echo set-output name=EXE_suffix::${EXE_suffix}
+        echo ::set-output name=EXE_suffix::${EXE_suffix}
+        # parse commit reference info
+        REF_NAME=${GITHUB_REF#refs/*/}
+        unset REF_BRANCH ; case ${GITHUB_REF} in refs/heads/*) REF_BRANCH=${GITHUB_REF#refs/heads/} ;; esac;
+        unset REF_TAG ; case ${GITHUB_REF} in refs/tags/*) REF_TAG=${GITHUB_REF#refs/tags/} ;; esac;
+        REF_SHAS=${GITHUB_SHA:0:8}
+        echo set-output name=REF_NAME::${REF_NAME}
+        echo set-output name=REF_BRANCH::${REF_BRANCH}
+        echo set-output name=REF_TAG::${REF_TAG}
+        echo set-output name=REF_SHAS::${REF_SHAS}
+        echo ::set-output name=REF_NAME::${REF_NAME}
+        echo ::set-output name=REF_BRANCH::${REF_BRANCH}
+        echo ::set-output name=REF_TAG::${REF_TAG}
+        echo ::set-output name=REF_SHAS::${REF_SHAS}
+        # parse target
+        unset TARGET_ARCH ; case ${{ matrix.job.target }} in arm-unknown-linux-gnueabihf) TARGET_ARCH=arm ;; i686-*) TARGET_ARCH=i686 ;; x86_64-*) TARGET_ARCH=x86_64 ;; esac;
+        echo set-output name=TARGET_ARCH::${TARGET_ARCH}
+        echo ::set-output name=TARGET_ARCH::${TARGET_ARCH}
+        unset TARGET_OS ; case ${{ matrix.job.target }} in *-linux-*) TARGET_OS=linux ;; *-apple-*) TARGET_OS=macos ;; *-windows-*) TARGET_OS=windows ;; esac;
+        echo set-output name=TARGET_OS::${TARGET_OS}
+        echo ::set-output name=TARGET_OS::${TARGET_OS}
+        # package name
+        PKG_suffix=".tar.gz" ; case ${{ matrix.job.target }} in *-pc-windows-*) PKG_suffix=".zip" ;; esac;
+        PKG_BASENAME=${PROJECT_NAME}-${REF_TAG:-$REF_SHAS}-${{ matrix.job.target }}
+        PKG_NAME=${PKG_BASENAME}${PKG_suffix}
+        echo set-output name=PKG_suffix::${PKG_suffix}
+        echo set-output name=PKG_BASENAME::${PKG_BASENAME}
+        echo set-output name=PKG_NAME::${PKG_NAME}
+        echo ::set-output name=PKG_suffix::${PKG_suffix}
+        echo ::set-output name=PKG_BASENAME::${PKG_BASENAME}
+        echo ::set-output name=PKG_NAME::${PKG_NAME}
+        # deployable tag? (ie, leading "vM" or "M"; M == version number)
+        unset DEPLOY ; if [[ $REF_TAG =~ ^[vV]?[0-9].* ]]; then DEPLOY='true' ; fi
+        echo set-output name=DEPLOY::${DEPLOY:-<empty>/false}
+        echo ::set-output name=DEPLOY::${DEPLOY}
+        # DPKG architecture?
+        unset DPKG_ARCH ; case ${{ matrix.job.target }} in i686-*-linux-*) DPKG_ARCH=i686 ;; x86_64-*-linux-*) DPKG_ARCH=amd64 ;; esac;
+        echo set-output name=DPKG_ARCH::${DPKG_ARCH}
+        echo ::set-output name=DPKG_ARCH::${DPKG_ARCH}
+        # DPKG version?
+        unset DPKG_VERSION ; if [[ $REF_TAG =~ ^[vV]?[0-9].* ]]; then DPKG_VERSION=${REF_TAG/#[vV]/} ; fi
+        echo set-output name=DPKG_VERSION::${DPKG_VERSION}
+        echo ::set-output name=DPKG_VERSION::${DPKG_VERSION}
+        # DPKG base name/conflicts?
+        DPKG_BASENAME=${PROJECT_NAME}
+        DPKG_CONFLICTS=${PROJECT_NAME}-musl
+        case ${{ matrix.job.target }} in *-musl) DPKG_BASENAME=${PROJECT_NAME}-musl ; DPKG_CONFLICTS=${PROJECT_NAME} ;; esac;
+        echo set-output name=DPKG_BASENAME::${DPKG_BASENAME}
+        echo set-output name=DPKG_CONFLICTS::${DPKG_CONFLICTS}
+        echo ::set-output name=DPKG_BASENAME::${DPKG_BASENAME}
+        echo ::set-output name=DPKG_CONFLICTS::${DPKG_CONFLICTS}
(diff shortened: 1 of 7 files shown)


rivy:
@sharkdp , I found some time to work on this ...

The CICD file is essentially the same as the one used for `pastel` with an additional build target and the two windows-gnu targets disabled (for linker issues; some explanation and refs are in the commit which disables them). I also added (automatic) 'deb' packaging for two more of the platforms that coincide with your builds.

I don't have a build/pkg for the '..._i386.deb' package. I'm not sure how to build it, and I don't see a target platform on the rust platform list to build toward for that package.

I found two problems (and included solutions):

1. All the text comparison integrations tests were failing on the 'windows' platforms. I thought it might be a line ending issue but couldn't tell from the test output. So, I instrumented the tests to give me more information. Line endings were a problem, so I added some .gitattribute settings to force LF line endings for the test fixture files in examples and snapshots.

As the test time seemed unchanged, I left the instrumentation in the tests (future value?).

2. The pager process execution was failing for `echo` (which is a built-in command for Windows). To execute built-in commands, you have to go execute them via the shell. Also, the direct execution method in-use bypassed usual shell algorithm for finding the executable along the PATH (ie, it doesn't try the various extension in PATHEXT). So, the pager won't be found if it is missing an extension. This is how I worked around that in `env` for `coreutils`.

``` rust
#[cfg(not(windows))]
fn build_command<'a, 'b>(prog_name: &'b str, args: &'a mut Vec<&'b str>) -> (std::borrow::Cow<'b, str>, &'a [&'b str]) {
    (prog_name, &args[..])
}

#[cfg(windows)]
fn build_command<'a, 'b>(prog_name: &'b str, args: &'a mut Vec<&'b str>) -> (std::borrow::Cow<'b, str>, &'a [&'b str]) {
    args.insert(0, prog_name);
    args.insert(0, "/d/c");
    let prog_name = std::env::var("ComSpec")
        .map(std::borrow::Cow::from)
        .unwrap_or_else(|_| std::borrow::Cow::from("cmd"));

    (prog_name, &args[..])
}
```

I created a simpler (maybe slightly hacky 🤷 ) solution for the pager process logic.

``` rust
#[cfg(windows)]
let (pager_path, args) = {
    let p = std::env::var("ComSpec").unwrap_or_else(|_| "cmd".to_string());
    let mut a = args.to_vec();
    a.insert(0, pager_path.to_str().unwrap().to_string());
    a.insert(0, "/d/c".to_string());
    (p, a)
};
```

Of course, with this, and reverting to using `echo`, the EOL issue popped up again. So, I added end of line normalization (via predicates) for the tests expecting `echo` output. The normalization could be used on all tests, but the current hybrid '.gitattributes' and targeted normalization is working.

Ultimately, all [tests are passing](https://github.com/rivy/rust.bat/actions/runs/118193328) and [auto-deployment is working](https://github.com/rivy/rust.bat/releases/tag/v0.15.4.2).

I should be able to muck with this more later this weekend if you'd like any changes.

Cheers!

sharkdp:
@rivy Thank you so much for working on this!

I haven't looked into the details, but a short comment regarding the pager issue: Instead of always using an intermediate `cmd` shell on Windows, couldn't we call it with something like this?
```
bat --pager="cmd /c echo …"
```

.github/workflows/CICD.yml
```diff
@@ -0,0 +1,294 @@
...
+          - { os: ubuntu-18.04   , target: i586-unknown-linux-musl      , use-cross: use-cross }
+          - { os: ubuntu-18.04   , target: i686-unknown-linux-gnu      , use-cross: use-cross }
+          - { os: ubuntu-18.04   , target: i686-unknown-linux-musl     , use-cross: use-cross }
+          - { os: ubuntu-18.04   , target: x86_64-unknown-linux-gnu    , use-cross: use-cross }
+          - { os: ubuntu-18.04   , target: x86_64-unknown-linux-musl   , use-cross: use-cross }
+          - { os: ubuntu-16.04   , target: x86_64-unknown-linux-gnu    , use-cross: use-cross }
+          - { os: macos-latest   , target: x86_64-apple-darwin         }
+          # - { os: windows-latest , target: i686-pc-windows-gnu         }  ## disabled; linker errors
+          - { os: windows-latest , target: i686-pc-windows-msvc        }
+          # - { os: windows-latest , target: x86_64-pc-windows-gnu       }  ## disabled; linker errors ## note: requires rust >= 1.43.0 to link correctly
```
mati865, line 48: 1.43 was released month ago.

rivy: Yes, that's a comment for `coreutils` and a few other projects that I contributed the CICD code too ... and they are all compiling and working. But, here, I see `undefined reference to `_imp____acrt_iob_func'`, for which I don't see a solution yet. I included some references to discussions/issues (some of which you are a part of) about the problem.

- ref: <rust-lang/rust#68887 (comment)>
- ref: <https://users.rust-lang.org/t/linking-with-gcc-failed-exit-code-1/34124>
- ref: <https://www.gitmemory.com/issue/rust-lang/rust/47048/530376978>
- ref: <https://sourceforge.net/p/mingw-w64/mailman/mingw-w64-public/?limit=250&page=7>

I'd be happy to re-enable them here, but v1.43 doesn't fix the noted linker error.
I'll change the comment to be more representative of the situation.

rivy:
> @rivy Thank you so much for working on this!
> 
> I haven't looked into the details, but a short comment regarding the pager issue: Instead of always using an intermediate `cmd` shell on Windows, couldn't we call it with something like this?
> 
> ```
> bat --pager="cmd /c echo …"
> ```

Yes, that's a solution, but I think it's not ideal. The included solution (or a future variation) let's users of `bat` use it in a similar way cross-platform and in a "normal" way for them on the windows platform.

Asking just windows users to specify the pager in an unusual, somewhat arcane, way seems counter-productive to wider use of the tool and creates more "user friction".

(If it's not obvious, I'm a cross-platform developer, using Windows primarily and linux of various stripes quite commonly. And I really like to have all of my tools on both platforms. And, as much as possible, I expect them to work in the same manner, minimizing platform-specific quirks.)

As an example that will bite windows users... `less` when installed on a windows machine is usually saved into a file named `less.exe` somewhere on the PATH, but the user will almost always just use it as just `less`, no extension used (eg, `dir | less`). But `bat --pager=less` won't work. It would have to be `bat --pager=less.exe`. It can be even more complicated with some commands which commonly use a preliminary BAT or CMD file as an alias or preprocessor which calls a later EXE executable (transparent to the user).

As written here, `bat --pager=less` means the same thing and works the same way on all linux/unix, macos, and windows platforms.

So, I've pushed a variation of this solution into `env` for `coreutils` and I'm a fan of it's more common use in other CLIs.

.github/workflows/CICD.yml
mati865: That's because you don't have i686 MinGW installed (GH Actions provide only x86_64), note that x86_64 build passed and failed tests.
It will work if you use MSYS2 which provides both toolchains and was added to GHA recently https://github.com/actions/virtual-environments/issues/30

sharkdp:
> But `bat --pager=less` won't work. It would have to be `bat --pager=less.exe`. It can be even more complicated with some commands which commonly use a preliminary BAT or CMD file as an alias or preprocessor which calls a later EXE executable (transparent to the user).

I absolutely get your points. In principle, I'm all for getting users the best experience. Three counter arguments:
* The `--pager=<cmd>` option is not something that the average user will typically use from the shell. If a user really wants to use a different pager (which one?!), they would put a `--config=my-pager.exe` line in their config file. Typing `.exe` once is not too bad. The large majority of users will never change the pager and never use this option. Which leads us to the next point.
* Spawning an intermediate shell will lead to a performance overhead on every single `bat` call. It may very well be negligible. But it's something I would like to see measured before we activate this. On Linux, spawning an intermediate `bash` shell takes 2.4 ms:

  |:---|---:|---:|---:|---:|
  | `bash -c 'echo test'` | 2.4 ± 0.7 | 1.6 | 6.2 | 1.00 |

  … which would only be a 3% overhead\* (but see #951):
  |:---|---:|---:|---:|---:|
  | `bat --paging=always --pager=less test.txt` | 57.9 ± 2.9 | 53.8 | 66.7 | 1.00 |
  | `bat --paging=always --pager="bash -c less" test.txt` | 59.9 ± 2.3 | 55.8 | 65.3 | 1.03 ± 0.07 |

  but how does the situation look like on Windows with `cmd`?
* Last, I'm not sure what kind of weird side effects we might have by passing the unfiltered `<cmd>` in `--pager=<cmd>` to a shell. But that might just be an uneasy feeling from someone who has seen too many SQL injection bugs.

\* Benchmarks via:
``` bash
hyperfine \
  --warmup 5 \
  'bat --paging=always --pager=less test.txt' \
  'bat --paging=always --pager="bash -c less" test.txt' \
  --export-markdown test.md
```

sharkdp:
> All the text comparison integrations tests were failing on the 'windows' platforms. I thought it might be a line ending issue but couldn't tell from the test output. So, I instrumented the tests to give me more information. Line endings were a problem, so I added some .gitattribute settings to force LF line endings for the test fixture files in examples and snapshots.

These tests seem to be running just fine on AppVeyor: https://ci.appveyor.com/project/sharkdp/bat/builds/33205057 - any idea what is different on the GHA machines?

rivy:
> * The `--pager=<cmd>` option is not something that the average user will typically use from the shell. If a user really wants to use a different pager (which one?!), they would put a `--config=my-pager.exe` line in their config file. Typing `.exe` once is not too bad. The large majority of users will never change the pager and never use this option. Which leads us to the next point.

True, but, reiterating the point, if I, as a user, use a paging program which operates via a BAT entry point, and it "just works" as say `less`. Setting the pager configuration in the configuration file to just `less` will surprisingly fail, without any explanation.

> * Spawning an intermediate shell will lead to a performance overhead on every single `bat` call. It may very well be negligible. But it's something I would like to see measured before we activate this.

Sure, I'd like to see an analysis of that as well.

I do think it's generally irrelevant here as a premature optimization, because when you're using a pager, you're planning to look at the results. And the console display + actual reading of the content should *far* outweigh the time spent starting an extra process. `bat` doesn't start a pager every time, even with redirected output, does it?

> * Last, I'm not sure what kind of weird side effects we might have by passing the unfiltered `<cmd>` in `--pager=<cmd>` to a shell. But that might just be an uneasy feeling from someone who has seen too many SQL injection bugs.

This, I agree with... but "injection bugs" really only occur with permission differentials. On windows, I don't see how the pager would ever have any different permissions than `bat`, so I can't see any way to do harm. And, by not using this kind of implementation, you're exchanging some possible CMD parsing weirdness (and, I agree, it's *wierd* sometimes) for increased user confusion/surprise and more difficult cross-platform testing (you'll have to at least re-write all of the pager tests which are using `echo`, likely with platform-based conditionals).

> * Benchmarks via:
> 
> ```shell
> hyperfine \
>   --warmup 5 \
>   'bat --paging=always --pager=less test.txt' \
>   'bat --paging=always --pager="bash -c less" test.txt' \
>   --export-markdown test.md
> ```

I've not see a way to do this with the CMD shell, as of yet, although I'd live to be able to do timings such as this...

Just my cross-platform centric opinion... I'm happy to help investigate and implement a "best" strategy for you. Let me know what I can do to help.

rivy:
> These tests seem to be running just fine on AppVeyor: https://ci.appveyor.com/project/sharkdp/bat/builds/33205057 - any idea what is different on the GHA machines?

Hmm, well, I'm sure that is a line ending difference. So, my suspicion is that AppVeyor and GHA have different baseline `git` configuration setups. Likely, AppVeyor is using `core.autocrlf=false` (my preferred configuration as well), and GHA is using nothing or, the equivalent, `core.autocrlf=input`, which would checkout text files with CRLF on a windows platform.

I generally use a *.gitattributes* file in the main directory when I'm concerned about this issue. Something like...

``` .gitattributes
# default to LF EOLs for `git`-determined text files
* text=auto eol=lf

# BAT/CMD shell script files require CRLF EOLs
*.[Bb][Aa][Tt] text eol=crlf
*.[Cc][Mm][Dd] text eol=crlf
## *maybe* needed ## *.[Pp][Ss]1 text eol=crlf

# for specific fixtures
*.eol-cr.* text eol=cr
*.eol-lf.* text eol=lf
*.eol-crlf.* text eol=crlf
```

.github/workflows/CICD.yml
rivy: Thanks for the insight!
I'll attempt a new build using the MSYS toolchain.
Any suggestions on a process? I haven't ever tried using that toolchain on GHA.

rivy:
Well, ok then, I didn’t realize you wrote `hyperfine` and that it’s cross-platform. 😃 
I’ll try that ...

rivy:
After testing a few command variations, it appears that there is about a 13.3 ms penalty (or about 12% overhead) for using the `cmd` shell to execute the pager.

``` shell
C:\>env BAT_PAGER="more.com" hyperfine --warmup 5 "bat.exe --paging=always test.txt"
Benchmark #1: bat.exe --paging=always test.txt
  Time (mean ± σ):     108.6 ms ±   2.2 ms    [User: 4.4 ms, System: 6.8 ms]
  Range (min … max):   106.6 ms … 116.7 ms    23 runs
C:\>env BAT_PAGER="cmd.exe /c more.com" hyperfine --warmup 5 "bat.exe --paging=always test.txt"
Benchmark #1: bat.exe --paging=always test.txt
  Time (mean ± σ):     121.9 ms ±   2.0 ms    [User: 2.9 ms, System: 5.4 ms]
  Range (min … max):   118.6 ms … 126.6 ms    21 runs
```

I had to use the `BAT_PAGER` environment variable because, ironically, quoting issues block the use of `hyperfine` to compare the variations directly.

```shell
C:\> hyperfine --warmup 5 "bat.exe --paging=always --pager=more.com test.txt" "bat.exe --paging=always --pager=\"cmd.exe /c more.com\" test.txt"
Benchmark #1: bat.exe --paging=always --pager=more.com test.txt                                                                                 0
  Time (mean ± σ):     109.2 ms ±   2.4 ms    [User: 4.2 ms, System: 9.1 ms]                                                                    0
  Range (min … max):   106.5 ms … 116.0 ms    23 runs

Benchmark #2: bat.exe --paging=always --pager="cmd.exe /c more.com" test.txt
Error: Command terminated with non-zero exit code. Use the '-i'/'--ignore-failure' option if you want to ignore this. Alternatively, use the '--show-output' option to debug what went wrong.
```

mati865:
> Hmm, well, I'm sure that is a line ending difference. So, my suspicion is that AppVeyor and GHA have different baseline git configuration setups. Likely, AppVeyor is using core.autocrlf=false (my preferred configuration as well), and GHA is using nothing or, the equivalent, core.autocrlf=input, which would checkout text files with CRLF on a windows platform.

That's true AppVeyor has sane `false` default but GH Actions actually use dumbest possible choice here which is `true`.
`input` converts CRLF to LF when committing but doesn't convert LF to CRLF on clone (i.e. doesn't break stuff).

sharkdp:
> `bat` doesn't start a pager every time, even with redirected output, does it?

It does not, no. Only if the output goes to a TTY.

> I do think it's generally irrelevant here as a premature optimization, because when you're using a pager, you're planning to look at the results. And the console display + actual reading of the content should _far_ outweigh the time spent starting an extra process. 

> After testing a few command variations, it appears that there is about a 13.3 ms penalty (or about 12% overhead) for using the `cmd` shell to execute the pager.

The actual reading time outweighs the 13 ms by far, sure. However, what I'm more concerned about is latency / startup speed. I really like programs that feel fast. This is why `bat` is written in Rust. The baseline startup time is already quite bad (which is why I opened #951). I really wouldn't like to add another 13 ms for a slight enhancement of a feature that does not seem to be used at all (changing the default pager). `bat` for Windows has been available for two years and I have not heard any reports/complaints about this.

> I generally use a _.gitattributes_ file in the main directory when I'm concerned about this issue.

.github/workflows/CICD.yml
sharkdp: We had MSVC + GNU builds some time ago (in `bat` or `fd`?) and someone suggested that we only ship one of them in order not to confuse users. I'd be okay with only providing the MSVC version.

src/output.rs
```diff
@@ -74,8 +74,18 @@ impl OutputType {
...
+                let (pager_path, args) = {
+                    let p = std::env::var("ComSpec").unwrap_or_else(|_| "cmd".to_string());
+                    let mut a = args.to_vec();
+                    a.insert(0, pager_path.to_str().unwrap().to_string());
+                    a.insert(0, "/d/c".to_string());
+                    (p, a)
+                };
+
+                let mut p = Command::new(&pager_path);
+
```
sharkdp, line 87: As explained in the top level comment, I'd prefer if we could revert this change.

tests/integration_tests.rs
```diff
@@ -24,178 +25,231 @@ fn bat() -> Command {
...
-    bat()
+    let assert = bat()
         .arg("test.txt")
-        .assert()
+        .assert();
+    let stdout = String::from_utf8_lossy(&assert.get_output().stdout);
+    let stderr = String::from_utf8_lossy(&assert.get_output().stderr);
+    println!("stdout={:#?}", stdout);
+    println!("stderr={:#?}", stderr);
+    assert
```
sharkdp, line 35: Could you please revert those changes? The tests already provides a (suboptimal?) debug output in case the tests fail. I'd rather not have these repetitive `println!` statements in here.

rivy:
> The actual reading time outweighs the 13 ms by far, sure. However, what I'm more concerned about is latency / startup speed. I really like programs that feel fast. This is why `bat` is written in Rust. The baseline startup time is already quite bad (which is why I opened #951). I really wouldn't like to add another 13 ms for a slight enhancement of a feature that does not seem to be used at all (changing the default pager). `bat` for Windows has been available for two years and I have not heard any reports/complaints about this.

Alright, I can see I'm not going to prevail here, but, for the record, you officially have one windows user reporting this as a problem (me 👋 😄) and thinks a slightly slower solution is fine. 

I'll drop that code, but it will break testing on windows platforms.

> I didn't really know that CR/LF issues could be handled via a `.gitattributes` file. Thanks! 👍

Alternatively, I just saw another *.gitattributes* variant to solve the same problem...

``` .gitattributes
* -text diff
```

tests/integration_tests.rs
rivy: Just the `println!`'s?

src/output.rs
rivy: Will do...

rivy:
> That's true AppVeyor has sane `false` default but GH Actions actually use dumbest possible choice here which is `true`.

Thanks for the clarification.
And, wow!, that's a poor choice on their part.
🤷‍♂️ ... Maybe, they just want to play the junior developer role, throwing a weird spanner into the works. 😄

tests/integration_tests.rs
sharkdp: Well, no. All of the modifications, except for the `predicate`-changes which are required to make the tests pass.

Is the normalization still required after the `.gitattributes` changes?

rivy: Yes, for the echo's (if they work).

src/output.rs
rivy: Done

rivy:
Reversions added and rebased to master.
Cheers!

sharkdp:
@rivy: Thank you very much!

I'm going to merge this as-is and leave it running in parallel to TravisCI / AppVeyor for now.

One thing that would be really valuable for `bat` would be a special CI run that executes:
```
cargo install --release
bash assets/create.sh
cargo test --release # '--release' because we could prevent another full recompile
```

The `bash assets/create.sh` script is something that I typically run prior to a new release. It uses the installed `bat` to build a binary cache of syntaxes and themes. The subsequent run of `cargo test` will then use these new binary caches.

Quite a lot of tests depend on the builtin set of syntaxes, so this special run would be very valuable in catching bugs/inconsistencies in the syntax additions early.

PR #1025, opened 2020-05-26
Switch from path-absolutize to path_abs

### LINKED ISSUE #1022: [0.15.2] Cannot run `bat` with argument
0.15.2

Problem does not occur with version 0.15.0 which is the last version I tried.

```
$ bat somefile.sh
main' panicked at 'called `Option::unwrap()` on a `None` value', /usr/src/rustc-1.41.0/src/libcore/macros/mod.rs:15:40
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

Displays contents of file.

Cloned repo, checked out v0.15.2, built with cargo.

system
------

**$ uname -srm**
Linux 5.0.0-25-generic x86_64  

**$ lsb_release -a**
No LSB modules are available.  
Distributor ID: LinuxMint  
Description:    Linux Mint 19.1 Tessa  
Release:        19.1  
Codename:       tessa  

bat
---

**$ bat --version**
bat 0.15.2  

```

bat_wrapper
-----------

No wrapper script for 'bat'.

bat_wrapper_function
--------------------

tool
----

**$ less --version**
less 487 (GNU regular expressions)

sharkdp:
closes #1022

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -107,7 +107,7 @@ dependencies = [
  "globset 0.4.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "lazy_static 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "liquid 0.20.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "path-absolutize 1.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "path_abs 0.5.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "semver 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.110 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_yaml 0.8.12 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


ignatenkobrain:
:disappointed:, this makes packaging for Fedora even more time consuming. What about to fix actual problem? :) 

https://github.com/magiclen/path-absolutize/issues/2

sharkdp:
> disappointed, this makes packaging for Fedora even more time consuming

I'm sorry. I don't want to wait for an upstream fix on this. The v0.15.2 release is completely broken. Thank you for reporting these bugs. I was about to do the same after putting out the new `bat` release.

sharkdp:
Is `path_abs` harder to package than `path-absolutize`? I'd be open to switching back at some point if this should really be an issue for Fedora.

ignatenkobrain:
@sharkdp I think latest release of `path-absolutize` should be good (at least its own test suite is being run in CI and works fine now in Fedora).

ignatenkobrain:
@sharkdp yeah, I've built 0.15.2 with new path-absolutize and it works just fine here.

sharkdp:
If it's not too much trouble for you, I'd like to stay with `path_abs` for now. It doesn't use any `unsafe` code.

ignatenkobrain:
It is quite some trouble because it would mean to package 3 new crates that have not been updated for 1-2 years. Which means I need to port them to latest regex / lazy_static and whatnot…

ignatenkobrain:
But I can carry downstream patch that is switching it back in Fedora

decathorpe:
It looks like the situation here has somewhat reversed now - path-absolutize was updated as recently as last year, while the last path_abs release is now over three years old. And as far as I can tell, path-absolutize v3 no longer included unsafe code either.

musicinmybrain:
Revisiting this some time later, the Fedora package is still carrying a patch to use path-absolutize, which has just released v4. Meanwhile, the last path_abs release is now five years old. That doesn’t mean it’s defective, but it does suggest that it isn’t necessarily actively maintained.

Issue #919, opened 2020-04-14
Share sublime syntaxes


Keats:
Bat and [Zola](https://github.com/getzola/zola) both use syntect for highlighting syntaxes and have their own folder for additional syntaxes on top of the official Sublime repo. It would be a good idea to merge so there is one known source of syntaxes that all of syntect users can contribute to.

The idea would be:

- have a third party repo containing the syntax files
- keep the patches that bat has since they solve issues
- pretty much any .sublime-syntax should be accepted as long as it's not a duplicate of an existing one (Zola highlights .srt for example, which is not strictly a programming language but can be useful in various contexts)
- build step is up to the consumer: bat seems to have a built in method for it while Zola has a small Rust script (https://github.com/getzola/zola/blob/master/components/config/examples/generate_sublime.rs) that prints a blurb pasted in the docs.

A bonus point for syntect is to have a Rust CI setup on that repo that tests every syntax for both rust-onig and fancy-regex, which can be used to report potential bugs/issues.

keith-hall:
I've seen that quite often, `tmLanguage` syntax grammars are manually converted to `sublime-syntax`  format for use in syntect. Have you considered adding support for `tmLanguage` files directly in syntect, it seems it could save some hassle? They essentially support a subset of `sublime-syntax` functionality, so the existing data structures can be used, it would just be a case of parsing the plist/xml format...

sharkdp:
@Keats Thank you for your suggestion. As initially discussed in https://github.com/trishume/syntect/issues/287, that sounds like a good idea in principle.

Before we go forward with this, I'd first like to think about the possible downsides as well. I'm only interested in following this route, if it really leads to less maintenance work overall. There are also a few questions/problems that would need to be sorted out.

Pros:
* Synergy effect: all projects immediately benefit from contributions to the shared repo.
* Maintenance work for including new syntaxes can be shared (PR reviews, bug reports, patches, etc.)
* Part of the tooling and testing could be shared. CI would have to be set up just once.

Cons (some of them are specific to `bat`):
[...]

Issue #650, opened 2019-09-04
cargo install failing due to a compilation problem of onig_sys crate


rekka:
`cargo install bat` failing on Ubuntu 18.04 due to <https://github.com/rust-onig/rust-onig/issues/109> (bat version `0.12.1`)

tuxruffian:
Build fails on onig_sys on CentOS 7 as well.

alaaibrahim:
On fix I had for this to install is to modify the Cargo.toml file and fix the syntect version to 3.2.0

```patch
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ default-features = false
 features = []
 
 [dependencies.syntect]
-version = "3.2.1"
+version = "=3.2.0"
 default-features = false
 features = ["parsing", "yaml-load", "dump-load", "dump-create"]
```

I was able to install afterwards

sharkdp:
@alaaibrahim Thank you for your feedback. Unfortunately, pinning the version to 3.2.0 would re-introduce this bug: #626.

l1a:
FYI... fails on Debian 9 as well.

vmedea:
`apt-get install libclang-dev` made it build for me on Debian 10.1 aarch64

i'm reaally surprised,  that a cli tool to view files has a hard dependency on libclang, though

sharkdp:
> i'm reaally surprised, that a cli tool to view files has a hard dependency on libclang, though

That's a build-time-only dependency which (although maybe not optimal) is certainly not too surprising. `bat` has some C-code dependencies and libclang can be used to automatically generate bindings for other languages.

vmedea:
> That's a build-time-only dependency which (although maybe not optimal) is certainly not too surprising.

oh good that it's build-time-only, hadn't noticed that !

sharkdp:
This can be resolved once https://github.com/rust-onig/rust-onig/pull/126 is merged. This would also remove the need for `libclang` to be installed at build time.

trishume:
The `fancy-regex` mode was just merged and released as [v4.0.0](https://github.com/trishume/syntect/pull/270)! You may want to try using it to fix this issue, although it does halve performance so you may only want to use it on platforms where it's needed.

Still hoping that onig's build will eventually improve but for now `fancy-regex` is a good option.

sharkdp:
Thank you very much for the update!

[...]

PR #1034, opened 2020-05-29
Updated AsciiDoc syntax submodule

rxt1077:
I've updated the AsciiDoc syntax to a newer version (from the AsciiDoctor project) which happens to work with fancy-regex. My problem originally was that [nushell](https://github.com/nushell/nushell) was crashing when I tried to open adoc files. I found that they get their syntax from here.

I wholeheartedly agree with the discussion in this issue BTW: https://github.com/sharkdp/bat/issues/919, consolidation might make things easier.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -94,9 +94,6 @@
 [submodule "assets/themes/sublime-snazzy"]
 	path = assets/themes/sublime-snazzy
 	url = https://github.com/greggb/sublime-snazzy
-[submodule "assets/syntaxes/AsciiDoc"]
-	path = assets/syntaxes/02_Extra/AsciiDoc
-	url = https://github.com/SublimeText/AsciiDoc.git
 [submodule "assets/syntaxes/Assembly (ARM)"]
 	path = assets/syntaxes/02_Extra/Assembly (ARM)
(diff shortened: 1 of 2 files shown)


sharkdp:
We do not use `fancy-regex` in `bat`, and I don't think that it's a good idea that `nushell` uses `bat`s binary syntax dump (which was built with `onig` enabled) with the `fancy-regex` feature of `syntect`. They will experience panics for multiple syntaxes. I have documented some of these incompatibilities here: https://github.com/trishume/syntect/issues/287. See also: https://github.com/sharkdp/bat/issues/650#issuecomment-606011484

So for `bat`, there's not really any *need* to update the AsciiDoc syntax. That being said, it looks like the alternative syntax proposed here is more complete? Could you comment on that?

rxt1077:
Thank you for the links and the background. I was unaware of the issues with `nushell`'s usage.

Concerning how this version can benefit `bat`, this syntax is updated and more complete than the previous syntax. It looks like the previous syntax was a bit of a work in progress, whereas this one has been maintained to some extend by a larger project (AsciiDoctor).

sharkdp:
Thank you.

Issue #1024, opened 2020-05-26
Add syntax mappings to 'bat --list-languages' output


sharkdp:
We should think about showing the explicit syntax mappings (the builtin ones and the user provided `--map-syntax=…` ones) to the `bat --list-languages` output, next to the other file extensions. For example:
```
SSH Config                    ssh_config, **/.ssh/config
Bourne Again Shell (bash)     sh, bash, […], /etc/profile, **/bat/config
```

I think we can skip the `MapToUnknown` mappings

alexnovak:
Took a stab at this! Will have a PR up later tonight.

PR #1035, opened 2020-05-30
Issue #1024 -- Gets all configured mappings for --list-languages

alexnovak:
This addresses issue https://github.com/sharkdp/bat/issues/1024 and adds any configurations specified in `--map-syntax` (including those specified in config) as well as all defaults to the `--list-languages` output.

### DIFF
--- a/src/bin/bat/main.rs
+++ b/src/bin/bat/main.rs
@@ -70,7 +70,7 @@ fn get_syntax_mapping_to_paths(
         match mapping {
             (_, MappingTarget::MapToUnknown) => {}
             (matcher, MappingTarget::MapTo(s)) => {
-                let globs = map.entry((*s).into()).or_insert(Vec::new());
+                let globs = map.entry((*s).into()).or_insert_with(Vec::new);
                 globs.push(matcher.glob().glob().into());
             }
         }
--- a/src/bin/bat/main.rs
+++ b/src/bin/bat/main.rs
@@ -63,15 +63,15 @@ fn run_cache_subcommand(matches: &clap::ArgMatches) -> Result<()> {
 }
 
 fn get_syntax_mapping_to_paths(
-    mappings: Vec<(GlobMatcher, MappingTarget)>,
+    mappings: &[(GlobMatcher, MappingTarget)],
 ) -> HashMap<String, Vec<String>> {
-    let mut map = HashMap::new();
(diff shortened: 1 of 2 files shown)


src/bin/bat/main.rs
```diff
@@ -56,15 +59,43 @@ fn run_cache_subcommand(matches: &clap::ArgMatches) -> Result<()> {
     Ok(())
 }
 
+fn get_syntax_mapping_to_paths(
+    mappings: Vec<(GlobMatcher, MappingTarget)>,
+) -> HashMap<String, Vec<String>> {
+    let mut map = HashMap::new();
+    for mapping in mappings {
+        match mapping.1 {
```
sharkdp, line 67: instead of matching on `mapping.1`, you could also match on `mapping` and then capture the tuple:
``` rust
match mapping {
    (_, MappingTarget::MapToUnknown) => {}
    (matcher, MappingTarget::MapTo(s)) => {}
}
```
This way, we don't have to use anonymous names like `.0` and `.1`.

src/bin/bat/main.rs
```diff
@@ -56,15 +59,43 @@ fn run_cache_subcommand(matches: &clap::ArgMatches) -> Result<()> {
...
+        .cloned()
         .collect::<Vec<_>>();
     languages.sort_by_key(|lang| lang.name.to_uppercase());
 
+    let configured_languages =
+        get_syntax_mapping_to_paths(config.syntax_mapping.mappings().clone());
+
+    for lang in languages.iter_mut() {
+        if configured_languages.contains_key(&lang.name) {
+            let additional_paths = configured_languages.get(&lang.name).unwrap();
```
sharkdp, line 93: Instead of using `.contains_key` and then using `.get`, you can simply call `get` and match on the result - avoiding the `.unwrap()` call.

alexnovak:
Thank you so much for your feedback! It's incredibly useful to me (I clearly have yet to earn rank of master). 

I've made your recommended adjustments. Please let me know if there's any other tidying up I could do!

sharkdp:

I made a few more (mostly stylistic) changes in separate commits, in case you are interested.

alexnovak:
Thanks for those modifications! I wasn't aware some of those conventions were options, and the result is much cleaner. 

Thanks for letting me contribute! This has been a great experience.

PR #1047, opened 2020-06-02
Clarify stdin syntax highlighting in the README

### LINKED ISSUE #1028: bat not detecting syntax automatically for some files
[`bat 0.15.3`]

bat does not seem to be detecting and/or performing syntax highlighting by default on some languages. `cat manage.py | bat` does not have any colors, but `cat manage.py | bat -l python` does. 

batcolors

I've tested for the following files:

**Works**: *sh, *.rb,
**Does Not Work**: *.py, *.md, *.org, *.ps1, *.cpp

syntax highlighting on the input from STDIN.

choco install bat

---

My OS is Windows with the following specs:

Edition: Windows 10 Pro
Version: 1909
OS build: 18363.836

alexander-jackson:
Clarify the requirements for automatic syntax highlighting when reading
input from stdin. Fixes #1028.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,9 @@ Display multiple files at once
 > bat src/*.rs
 ```
 
-Read from stdin, determine the syntax automatically
+Read from stdin, determine the syntax automatically (note, this will only work
+if the syntax can be determined from the initial line of the file, usually
+through a shebang such as `#!/bin/sh`)
 
 ```bash
 > curl -s https://sh.rustup.rs | bat
--- a/README.md
+++ b/README.md
@@ -62,9 +62,9 @@ Display multiple files at once
 > bat src/*.rs
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -62,7 +62,9 @@ Display multiple files at once
 > bat src/*.rs
 ```
 
-Read from stdin, determine the syntax automatically
+Read from stdin, determine the syntax automatically (note, this will only work
+if the syntax can be determined from the initial line of the file, usually
```
eth-p, line 66: Small nitpick, but could we use "first line" instead of initial line? That terminology is more ubiquitious, and I think it would be easier for non-native speakers to understand.

alexander-jackson: Yes that makes sense, I've updated that and also clarified a little more about "what" will only work if the syntax can be determined.

sharkdp:
Thanks

Issue #975, opened 2020-05-04
Request: Flag to report missing newline before EOF


lorenzleutgeb:
Hi!

I am a `zsh` user and got used to my shell ["reporting" when there is no newline at the end of the output (a "partial" line)](https://unix.stackexchange.com/questions/167582/):

```
$ /usr/bin/cat id
sha256:197a075f86ebf05daa4509d81e07d6c153443ae0cc5539bf8d82ae1b11ef64a6%
```

(Note the `%`.)

However, that doesn't work with `bat`.

```
$ bat id
   1   sha256:197a075f86ebf05daa4509d81e07d6c153443ae0cc5539bf8d82ae1b11ef64a6
$ bat --plain id
sha256:197a075f86ebf05daa4509d81e07d6c153443ae0cc5539bf8d82ae1b11ef64a6
$ bat --version
bat 0.11.0
```

I'd like to go one step further, and request a flag on `bat` that will print an explicit warning in case the input does not end in a newline character. Such as:

```
$ bat --warn-on-missing-newline id
   1   sha256:197a075f86ebf05daa4509d81e07d6c153443ae0cc5539bf8d82ae1b11ef64a6
   ⚠   No newline at end of file!
```

eth-p:
I like this idea, but I'm not that much of a fan of using an emoji inside the output. There's still a few environments that can't render them, sadly.

We would also need to consider what it looks like with the grid mode. Maybe something like this?

```
  12 | Hello world
   ! | [Missing Newline]
```

lorenzleutgeb:
Sure, that was just what came to mind as an example. We're down to a matter of taste, but I'd prefer `!` (without angle brackets) over `<!>` (with angle brackets). Please go ahead and choose as you like!

eth-p:
Perhaps `[!]` in yellow?

As for the command line options, after a little thought, maybe it should be something similar to how we handle style components:

```
--warning=missing-trailing-newline,...
```

That would be more flexible for if we ever added more warning types (e.g. mixed indentation maybe?).

lorenzleutgeb:
Just FYI, `git diff` says "No newline at end of file"

sharkdp:
Thank you for reporting this.

The fact that `bat -p` prints a trailing newline (even if there is no newline at the end of the file) should be considered a bug, I believe (but we should check older tickets, if there was a reason for this behavior). Note that it works as expected in non-interactive mode, i.e. if the output is piped somewhere else:
```bash
[...]

Issue #1032, opened 2020-05-28
Disable style components with `--style=no-[component]`.


eth-p:
Right now, there isn't a way to disable specific style components from `--style=auto` or `--style=full`. The closest thing possible is manually recreating the "auto" set with a wrapper script, but it's clunky and would require maintenance when new components are added.

I propose we add a way to negate individual style components (i.e. not `plain`, `full`, or `auto`) by prefixing them with `no-` when passed through the `--style` option. If conflicting styles are provided, the rightmost one should be used:

- `--style=no-grid,full` => `--style=full`
- `--style=grid,no-grid,numbers` => `--style=numbers`

eth-p:
A side note to this: I've been experimenting with a way to generalize the `StyleComponent` struct. I came up with a way to do that, but it involves making a number of allocations with `Vec` because you can't have `fn provides(&self) -> &[Self] { return &[self] }`. I could probably work around that by casting `&self` to a raw pointer and using unsafe code to create a slice from it, but I'm not sure if that's very acceptable.

```rs
trait ComponentSet<C: Component> {
    fn defaults(interactive_terminal: bool) -> Vec<C>;
}

trait Component: Sized + Clone {
    fn defaults(interactive_terminal: bool) -> Vec<Self>;

    fn is_negatable(&self) -> bool {
        self.provides().len() == 1
    }

    fn provides(&self) -> Vec<Self> {
        vec![self.clone()]
    }
}

#[derive(Clone, Debug, Eq, PartialEq)]
pub enum ExampleComponent {
    Something,
    SomethingInteractive,
}

impl Component for ExampleComponent {
    fn defaults(interactive_terminal: bool) -> Vec<Self> {
        use ExampleComponent::*;
        if interactive_terminal {
            vec![Something, SomethingInteractive]
        } else {
            vec![Something]
        }
    }
}
```

sharkdp:
:+1: for the feature request.


As for the implementation, I'm not sure I can follow why we would even need a `fn provides(&self) -> &[Self]` function.

In general, note that this is code that is run just once. We definitely don't have to worry about a few allocations. Especially if the alternative is to use `unsafe` code.

eth-p:
> As for the implementation, I'm not sure I can follow why we would even need a fn provides(&self) -> &[Self] function.

This was just my experimenting, but my idea was to use `provides()` like this:

```rs
impl Component for StyleComponent {
    fn provides(&self) -> &[Self] {
[...]

PR #1049, opened 2020-06-03
"No newline at end of file" warning

obfusk:
I created this draft PR (which currently implements a very crude -- and possibly incomplete -- "fix") in the hope that it can serve as a place to identify which code is impacted by and discuss specific ways of solving #975.

### DIFF
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -365,6 +365,8 @@ impl<'a> Printer for InteractivePrinter<'a> {
             }
         };
 
+        let has_eol = line.bytes().next_back() == Some(b'\n');
+
         let regions = {
             let highlighter = match self.highlighter {
                 Some(ref mut highlighter) => highlighter,
@@ -443,8 +445,9 @@ impl<'a> Printer for InteractivePrinter<'a> {
                 }
             }
 
-            if line.bytes().next_back() != Some(b'\n') {
+            if !has_eol {
                 writeln!(handle)?;
+                eprintln!("No newline at end of file");
             }
         } else {
             for &(style, region) in regions.iter() {
@@ -582,6 +585,9 @@ impl<'a> Printer for InteractivePrinter<'a> {
                 )?;
             }
             writeln!(handle)?;
+            if !has_eol {
+                eprintln!("No newline at end of file");
+            }
         }
 
         Ok(())
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -642,7 +642,7 @@ fn header_padding() {
         .arg("single-line.txt")
         .assert()
         .stdout("File: test.txt\nhello world\n\nFile: single-line.txt\nSingle Line\n")
-        .stderr("");
+        .stderr("No newline at end of file\n");
 }
 
 #[cfg(target_os = "linux")]


eth-p:
A couple things should be changed:

- It needs to be configurable with an option (`clap_app.rs`)
    Ideally, this should be added like the style components with `--warning=no-ending-newline`. If you want an easy way to implement this for now though, you can copy parts of the `--wrap` flag to just have a simple enum instead.

- I think the warning should show up after the file contents, instead of above the header.

- You might want to integrate it into the printer a little bit better. You can use `create_fake_panel("<!>")` to create a sidebar that shows "<!>" instead of the line number, and print the warning after that. Coloring the warning sign might require a bit more work, though.

- The solution to figuring out that there isn't a final newline is clever, but it's not immediately obvious that only the last line might not have a `\n` character in it. Is there any chance that you could add a comment that explains that?

obfusk:
Thanks for the feedback!

> 1. It needs to be configurable with an option (`clap_app.rs`)

I absolutely agree. Hence "crude" and "draft" :) I was planning on looking into that when it's clear what the extent of the configurability should be and adding it to the PR.

> 2. You might want to integrate it into the printer a little bit better. You can use `create_fake_panel("/!\")` to create a sidebar that shows `/!\` instead of the line number, and print the warning after that. Coloring the warning sign might require a bit more work, though.

I had not looked into the possibilities for that yet. Thanks for the suggestion. I was also waiting on more feedback on how people want the warning to look before spending significant time on that. I would also assume it doesn't work w/ `--plain`?

> 3. The solution to figuring out that there isn't a final newline is clever, but it's not immediately obvious that only the last line might not have a `\n` character in it. Is there any chance that you could add a comment that explains that?

Good idea. I'll add that to the PR.

I'm also not entirely sure this is the correct place to detect and print the error. It might be better (though more complicated) to detect this at a higher level or pass the "error" up and process it there. Should this be considered a feature/bug of this printer or of `bat` as a whole?

> 4. ~I think the warning should show up after the file contents, instead of above the header.~
> This is a consequence of intertwining `eprintln!` and `write!`, and it actually results in inconsistent behavior (and doesn't work with the pager).
> One possible solution is to use point (2) above and just write it to the writer instead of using `eprintln!`.

I'm not entirely sure why the `eprintln!` shows up before the `write!`. Normally, when outputting to a terminal (and not using a pager), I'd expect `stdout` and `stderr` to be line-buffered and for lines printed to `stderr` and `stdout` to show up in the correct order. I did not yet investigate why this doesn't happen with the output abstraction `bat` uses.

I explicitly chose to output the warning to `stderr` because that way it doesn't show up in the actual output but is still shown to the user. It is common to print errors and warnings to `stderr` instead of to `stdout`. Whether this is the correct choice in this case, I'm not sure. I'd like to hear what others think. At the very least the current PR should be fixed so that the error is printed at the end instead of before the last line.

I see (at least) these options:

1. print an error to `stderr` for each file after writing the complete file, or maybe after writing all files.
2. use an extra line + sidebar to add the warning "in-place" (what to do with `--plain` then?)
3. append a warning symbol to the end of the line (might be hard to understand and might make wrapping more complex though)

Or maybe allow the user to select one of (some of) these options.

eth-p:
> I explicitly chose to output the warning to stderr because that way it doesn't show up in the actual output but is still shown to the user. It is common to print errors and warnings to stderr instead of to stdout. Whether this is the correct choice in this case, I'm not sure. I'd like to hear what others think. At the very least the current PR should be fixed so that the error is printed at the end instead of before the last line.

This is a really complex situation, sadly. While I definitely agree about using `stderr` for its intended purpose, we have a few scenarios to consider:

1. **Printing to the terminal with pager.**
    We can't feed `stderr` into the printer, so all error output has to either be printed before starting the pager so they doesn't interfere with eachother, or mixed into `stdout` so the pager can read it too.

2. **Printing to the terminal without the pager.**
    As long as the inconsistent behavior can be fixed, `stderr` is preferable.

3. **Printing to a pipe.**
    For this case, it should print to `stderr` so the warning isn't mixed into the piped output.

> I had not looked into the possibilities for that yet. Thanks for the suggestion. I was also waiting on more feedback on how people want the warning to look before spending significant time on that. I would also assume it doesn't work w/ --plain?

Correct. If the `numbers` style is missing, it wouldn't be visible. One possible solution would be to take a hybrid approach where it prints `[bat warning]: FILE.EXT is missing an ending newline` when the inline warning wouldn't be possible.

> I'm also not entirely sure this is the correct place to detect and print the error. It might be better (though more complicated) to detect this at a higher level or pass the "error" up and process it there. Should this be considered a feature/bug of this printer or of bat as a whole?

Considering this might be undesired behavior for `bat`-as-a-library, I'm would consider it to be more of a `bat` application feature rather than a printer feature. Passing the warning up would be a good idea, but it would also complicate things if we wanted to print inline warnings. @sharkdp, thoughts?

sharkdp:
I haven't followed the discussion above in details, but my thoughts:
* If we are writing to a pipe, don't do anything about the missing EOL. We want to behave like `cat`.
* I'm not so sure about the `--warn=a,b,c` thing. It would consistent with other `bat` options, but with tickets like #1032 open, I'm not so sure if we should go the same way with this. Especially since we don't have any other options right now. Maybe a simple `--no-missing-eol-warning` would also be fine.
* In `--plain` mode, I wouldn't show any warnings either. `--plain` implies a few other things already. Maybe it should also imply `--no-missing-eol-warning` (or however we decide to call it).
* In non-plain mode, I would print the warning to *STDOUT*. STDERR is for `bat` errors/warnings. If `bat` was not able to perform it's intended action. But here, everything goes according to plan. We print the full file contents. We only want to warn the user of the missing EOL at the EOF.

sharkdp:
Closing for now to clean up the list of open PRs.

PR #1060, opened 2020-06-15
Highlight systemd files and Pacman hooks as .ini files

guidocella:


sharkdp:
Most of these patterns seem far too generic. I wouldn't want to generally map them to the INI syntax. Note that a pattern like `target` matches all files named `*.target`, but also files named `target`.

Where are these files located? Could we use a glob pattern syntax mapping instead? Something like
```
/etc/systemd/*.conf => INI
```

This would need to be added to `src/syntax_mapping.rs`.

guidocella:
I did not realize it matches filenames like target.

systemd files are located in
`/etc/systemd`
`/usr/lib/systemd`
`/usr/local/lib/systemd`
`/lib/systemd` (symlink to `/usr/lib/systemd`)
`/lib64/systemd` (symlink to `/usr/lib/systemd`)
`/run/systemd`
`${XDG_CONFIG_HOME:-$HOME/.config}/systemd/user`
`${XDG_DATA_HOME:-$HOME/.local/share}/systemd/user`

I mapped these patterns specifying the directory:
`**/systemd/**/*.conf`
`**/systemd/**/*.example`
since they would override mappings of other languages (there's only 2 `.example` files in `/usr/lib/systemd/network`, I'm only adding it for completeness).
I specified only the extensions for everything else. This will syntax highlighting a package's systemd files even before installing them to `/usr/lib/systemd` and simplify  the patterns. None of them are used by other mappings anyway and I don't see why you would use them instead of the correct extension for what you're writing, but I'll prefix `**/systemd/**/` to every pattern if you prefer.

I also tried to map all of
`/*/systemd/**`,
`/usr/lib/systemd/**`,
`/usr/local/lib/systemd/**`,
`**/systemd/user/**`
to `ini`
(thus excluding
`/usr/include/systemd`
`/usr/share/systemd`
`/usr/share/doc/systemd`)
but there were many files where the mapping had to be reverted:
`/usr/lib/systemd/catalog/*.catalog` files which translate systemd messages
`/{usr/lib,etc,run}/systemd/ntp-units.d/*.list` files which contain a list of unit names to change the system clock and timezone and to enable/disable network time synchronization
`resolv` files:
`/usr/lib/systemd/resolv.conf`
`/run/systemd/resolve/resolv.conf`
`/run/systemd/resolve/stub-resolv.conf`
executables in `/usr/lib/systemd/{scripts,system-environment-generators,system-shutdown,system-sleep,user-environment-generators}` whose syntax is best detected with their shebang.

Hopefully no other software's files should be this complex to syntax highlight.


I assumed only files with `.hook` extensions are pacman hooks to be highlighted as `ini`, since hook is a generic filename. Otherwise we can specifically highlight only system hooks, `/usr/share/libalpm/hooks/*.hook`, and user hooks, `/etc/pacman.d/hooks/*.hook`, however in `pacman.conf` you can set a different directory from `/etc/pacman.d/hooks` which would break the detection (I would have split this in 2 pull requests if I knew how complex it would turn out).

sharkdp:
Thank you very much for the detailed analysis and the update!

PR #1066, opened 2020-06-20
Enable non-printable characters for redirected output

### LINKED ISSUE #1061: Add ability to enable --show-all output when output is redirected
Hi, I occasionally find myself wanting to capture the (very nice and useful) `--show-all` output. For example, in order to compute a diff on the `--show-all` output itself. Currently, the only way I know how to do this is to print the `--show-all` output to the screen, select it with a mouse, and copy it into a text file.  ~~I understand the reason `bat` must, by default, only enable `-A` if stdout is attached to an interactive terminal.~~ If not by default then perhaps what I'm requesting could be something like

```
bat --show-all=always
```

Thanks for all the work on bat!

gsomix:
Closes #1061

OK, fix looks like straigforward, but I could miss something important.

- [x] Add replacing non-printable characters in `SimplePrinter<'a>` implementation
- [x] Add tests

Thank you!


gsomix:
I read through test suite, but it's doesn't clear where to put test for this fix. Could you help me please?

sharkdp:

`tests/integration_tests.rs` would be a good place, I think. Something like:

``` rust
#[test]
fn show_all_mode() {
    bat()
        .arg("--show-all")
        .arg("test.txt")
        .arg("test.txt")
        .assert()
        .stdout("hello•world␊")
        .stderr("");
}
```

If you like, you could also create a slightly more complicated/complete test file (instead of `test.txt`) for the `--show-all` mode in `tests/examples`.

gsomix:
@sharkdp Thank you!

Almost all checks are green now (Travis is just slow), ready for review!

tests/integration_tests.rs
```diff
@@ -716,3 +716,13 @@ fn do_not_detect_different_syntax_for_stdin_and_files() {
...
     );
 }
+
+#[test]
+fn show_all_mode() {
+    bat()
+        .arg("--show-all")
+        .arg("nonprintable.txt")
+        .assert()
+        .stdout("hello•world␊├──┤␍␀␇␈␛")
```
sharkdp, line 726: I believe this actually demonstrates a bug in the implementation in this PR: there should be an *actual* line-feed character after `␊`, right?

gsomix: Hm, yes, indeed. I'll take a look tomorrow.

gsomix: OK, it might look like a hack, but I could hardcode inserting line-feed right after `␊` character in last position of line. What do you think?

gsomix: I mean something line that:

```rust
let line = replace_nonprintable(line_buffer, self.config.tab_width);
write!(handle, "{}", line)?;
if line.chars().last() == Some('␊') {
    writeln!(handle)?;
}
```

gsomix:
@sharkdp OK, I hope it's fixed now.

gsomix:
@sharkdp Hey, I was wondering if you could find the time to merge it? Thank you!

.gitignore
```diff
@@ -5,3 +5,6 @@
 /assets/completions/bat.fish
 /assets/manual/bat.1
 /assets/metadata.yaml
+
+# IDEA config files
+.idea/
```
sharkdp, line 10: Can we please remove this? I think it rather belongs in a user git ignore file, not into this repo.

tests/integration_tests.rs
sharkdp: Sounds good. I also don't see another option right now.

.gitignore
gsomix: Sure.

sharkdp:
Oh, sorry. One last thing. I'm not yet used to asking contributors about this.

Could you please add an entry to the "unreleased" section in the `CHANGELOG.md`? Maybe under "Bugfixes"? The format for entries is:
```
```

gsomix:
Done.

gsomix:
@sharkdp Thank you for the review! It's my first PR into rust project, and I'm happy now. :)

sharkdp:
:tada:

PR #1070, opened 2020-06-23
Adds GLSL syntax

### LINKED ISSUE #1046: Add GLSL syntax
**Syntax:**
Shaders: `GLSL`/`*.frag`

caioalonso:
Fixes #1046.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -185,3 +185,6 @@
 [submodule "assets/syntaxes/02_Extra/AsciiDoc"]
 	path = assets/syntaxes/02_Extra/AsciiDoc
 	url = https://github.com/asciidoctor/sublimetext-asciidoc.git
+[submodule "assets/syntaxes/02_Extra/GLSL"]
+	path = assets/syntaxes/02_Extra/GLSL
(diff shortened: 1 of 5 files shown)


sharkdp:
Thank you for this contribution.

When I run the tests locally after updating the binary assets (`bash assets/create.sh`), I get a test failure:
```
thread 'no_duplicate_extensions' panicked at 'File extension / pattern "fs" appears twice in the syntax set', tests/no_duplicate_extensions.rs:25:13
```
(this should be added to the CI pipeline)

The `fs` file extension is already used by the F# syntax.

To fix this, we *could* add another "known exception" in `no_duplicate_extensions.rs` and add an appropriate test in `src/assets.rs` (`syntax_detection_well_defined_mapping_for_duplicate_extensions`).

But we should first specify how this collision should be handled

caioalonso:
Thanks for the reply, @sharkdp! Yeah, I was trusting the CI.

I've added the tests and made it default to F# just because it seems a more popular language.

sharkdp:
Thank you very much!

Issue #1075, opened 2020-06-28
Short flag for --paging=never?


LordFlashmeow:
Sometimes, when I'm working with files larger than my screen, I find it more frustrating to with deal with `less` than use iTerm's scrollback, so I can come back to it later. `less` is great for looking at the first n lines of the file, but after that it's a pain.

The auto setting works fine for everything else, the only time I need to use the paging flag is for `--paging=never`, so it feels very long for an effectively "true/false" flag. Would it be possible to add a single letter alias for `--paging=never`? (Maybe `-P`?)

windwp:
You can create a bash alias for it

sharkdp:
Thank you for your request.

Note that you can already use `--plain`/`-p` twice to disable paging (and reduce the style to "plain"):
```
bat -pp …
```

LordFlashmeow:
Hmm. True, but then a lot of the decorations are missing. It looks like it's only a couple of lines to add — one if statement in `config()`, and a couple of lines to add the argument. I can make a PR with my proposed changes, unless you think it's a bad idea (it's your project, after all!)

sharkdp:
Sounds good to me. @eth-p, what do you think?

eth-p:
Short flags are at a bit of a premium, but this seems like a good reason to use one. 👍

sharkdp:
Closed via #1082

sharkdp:
Released in [v0.16.0](https://github.com/sharkdp/bat/releases/tag/v0.16.0).

PR #1082, opened 2020-06-30
Add alias for --paging=never

LordFlashmeow:
From #1075, here's my proposed implementation for adding an alias for `--paging=never`.

I couldn't find a way for clap to create an alias with a value, so I ended up creating a new parameter for it.

I'm not sure if it needs to be added, but I figure I'll share it anyways. I'm not sold on using `-P` as the flag, but it made the most sense to me.


eth-p:
I think the long flag would be better as `--no-paging` with an alias of `--no-pager`. From my experiences, that naming convention is more common than `--disable-x`.

@sharkdp, thoughts?

LordFlashmeow:
I agree on the naming. Currently there is no long flag, but I'm happy to add that

sharkdp:
Thank you very much for your contribution. Code looks good - thank you for adding the tests.

As for the long option, I'm not sure if we really need that. We already have `--paging=never` as an expressive form (for the config file). I would maybe even go as far as removing the new `-P` option from the help text (as a separate entry) and modifying the help text of `--paging` to mention the alias for `--paging=never`. What do you think?

Could you please add an entry to the "unreleased" section in the `CHANGELOG.md`? Under "Features". The format for entries is:
```
```

LordFlashmeow:
Hmm, fair point on the long option, although the `--plain` option had set a precedent for the flag having a long value and distinct help entry. If you want to move away from that, that's fine too.

sharkdp:
> Hmm, fair point on the long option, although the `--plain` option had set a precedent for the flag having a long value and distinct help entry. If you want to move away from that, that's fine too.

You are right. I think I would still rather keep the `--help` text shorter by not introducing a completely new line for the option introduced here.

LordFlashmeow:
I've pulled the long argument and hidden it so it doesn't show up in the help page. I also made a few changes to the pager and paging descriptions in the man page, but I can change it back if I made anything unclear.

Also, as a topic for another issue: adding an environment section to the man page

sharkdp:
Thank you very much for the updates and the high quality PR!

PR #1105, opened 2020-07-09
add webi as cross-platform install method

solderjs:
I've added `bat` to `webi`, which is a cross-platform web-based installer.

See
- https://webinstall.dev/bat
- https://github.com/webinstall/packages/tree/master/bat
- https://github.com/webinstall/packages/blob/master/_webi/template.sh

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,19 @@ The [`prettybat`](https://github.com/eth-p/bat-extras/blob/master/doc/prettybat.
 
 [![Packaging status](https://repology.org/badge/vertical-allrepos/bat.svg)](https://repology.org/project/bat/versions)
 
+### On Linux (using `webi`)
+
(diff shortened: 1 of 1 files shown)


sharkdp:

`webi` looks like an interesting project, but - to be honest - I'd like to wait for it to gain a bit more traction before we advertise this installation method in the `bat` README.

solderjs:
Cool beans. I'll check back in 6 months or so.

In the meantime, if you have any good suggestions for "daily drivers" that you use yourself that we should check out, please let me know.

I found out about `bat` and a number of the other tools on webi from a presentation at the https://utah.rs that I co-run with Eric Gubler, and as I'm working on a Beyond Code Bootcamp course, I'm interested in tools like this that help new devs learn the drop-in replacements like `bat`, `rg`, etc.

PR #1121, opened 2020-08-02
Completely move CI/CI to GitHub Actions

sharkdp:
This PR is intended to fully move our CI/CD from Travis/AppVeyor to GitHub actions.

Change log / to do:
- [x] Disable code coverage comments
- [x] Remove unneeded `*-i586-*` platforms (reduce chance of failure and number of Downloads on the release page)
- [x] Remove Travis, AppVeyor settings and `ci` folder.
- [x] Add (Fish) autocompletion files to Archives and Debian packages
- [x] Add full `copyright` text and Debian `control` file to package file, see: https://github.com/sharkdp/bat/blob/75246fb25b1e68724d636861d24ffa39845c58f2/ci/before_deploy.bash#L60-L177
- [x] Add special job to run
  ```bash
  cargo install --release
  bash assets/create.sh
  cargo test --release # '--release' because we could prevent another full recompile
  ```
  The `bash assets/create.sh` script is something that I typically run prior to a new release. It uses the installed `bat` to build a binary cache of syntaxes and themes. The subsequent run of `cargo test` will then use these new binary caches.
  Quite a lot of tests depend on the builtin set of syntaxes, so this special run would be very valuable in catching bugs/inconsistencies in the syntax additions early.
- [x] Check all feature combinations, see:  https://github.com/sharkdp/bat/blob/75246fb25b1e68724d636861d24ffa39845c58f2/ci/script.bash#L16-L20


sharkdp:
@rivy Just FYI. This is the first time I'm modifying GitHub Actions config files, so I hope I didn't destroy any of your work :smile:

PR #1123, opened 2020-08-03
Ignore 'more' on pager

seeplusplus:
A couple things:
1. There's probably a terser syntax for this. I'd love to see it, I'm still learning Rust.
2. Depending on the desired behavior this may or may not be sufficient. When specifying "more" the fall-over to None means the default "less" will be used. This is in contrast to if you specify, say, "More" where bat will just disable paging. 

Resolves #1063

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -59,6 +59,7 @@ impl OutputType {
 
         let pager = pager_from_config
             .or(pager_from_env)
+            .filter(|p| p != "most")
             .unwrap_or_else(|| String::from("less"));
 
         let pagerflags =
--- a/src/bin/bat/app.rs
+++ b/src/bin/bat/app.rs
@@ -221,7 +221,10 @@ impl App {
             },
             style_components,
             syntax_mapping,
-            pager: self.matches.value_of("pager"),
+            pager: match self.matches.value_of("pager") {
+                Some(s) => if s == "more" { None } else { Some(s) },
+                _ => None,
+            },
             use_italic_text: match self.matches.value_of("italic-text") {
                 Some("always") => true,
                 _ => false,
--- a/src/bin/bat/app.rs
+++ b/src/bin/bat/app.rs
@@ -221,10 +221,7 @@ impl App {
             },
             style_components,
             syntax_mapping,
-            pager: match self.matches.value_of("pager") {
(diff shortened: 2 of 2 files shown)


sharkdp:

Does this really fix #1063 where `more` is only configured via the `PAGER` environment variable? The code you modified only concerns the `--pager` command line option. Check out `OutputType::try_pager` in `src/output.rs`.

sharkdp:
> There's probably a terser syntax for this. I'd love to see it, I'm still learning Rust.

Indeed, manually `match`ing on `Option<T>` is very rarely needed as there are a lot of helper functions that can be used instead. For your use case here, using [`filter`](https://doc.rust-lang.org/std/option/enum.Option.html#method.filter) would be an alternative.

seeplusplus:
> Thank you for your contribution!
> 
> Does this really fix #1063 where `more` is only configured via the `PAGER` environment variable? The code you modified only concerns the `--pager` command line option. Check out `OutputType::try_pager` in `src/output.rs`.

Fixed. Thanks!



> > There's probably a terser syntax for this. I'd love to see it, I'm still learning Rust.
> 
> Indeed, manually `match`ing on `Option<T>` is very rarely needed as there are a lot of helper functions that can be used instead. For your use case here, using [`filter`](https://doc.rust-lang.org/std/option/enum.Option.html#method.filter) would be an alternative:
> 
> ```rust
> self.matches.value_of("pager").filter(|p| p != "most")
> ```

Awesome feedback, thanks for the tip!

sharkdp:
Is this really the behavior we want? Have you tested this?

I think we only want to ignore `more` if it's set as `PAGER`. Not if someone really forces it via the config, via `BAT_PAGER` or via the command line. Also, what if someone uses `PAGER="more -s"`? Shouldn't we shell-split the `PAGER` variable and check the basename only? (see similar code for `less`).

alauq (approved):

src/output.rs
```diff
@@ -59,6 +59,7 @@ impl OutputType {
 
         let pager = pager_from_config
             .or(pager_from_env)
+            .filter(|p| p != "most")
```
alauq, line 62:
```suggestion
            .filter(|p| p != "most")
```

sharkdp:
I'm going to close this for now (see reasons above)

PR #1125, opened 2020-08-03
Clarified which pager is used in docs

mmahmoudian:
For me the default pager is `most` and as the result the text will be pushed to `most` instead of `less`. This change also shows to the user that they can configure the pager according to their preference.


sharkdp:
Thank you

PR #1131, opened 2020-08-06
Add Nginx config file syntax

kjmph:
Add a new submodule for Nginx config file coloring.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -191,3 +191,6 @@
 [submodule "assets/syntaxes/02_Extra/GLSL"]
 	path = assets/syntaxes/02_Extra/GLSL
 	url = https://github.com/euler0/sublime-glsl
+[submodule "assets/syntaxes/02_Extra/Nginx"]
+	path = assets/syntaxes/02_Extra/Nginx
+	url = https://github.com/brandonwamboldt/sublime-nginx.git


sharkdp:

Could you please add an entry to the "unreleased" section in the `CHANGELOG.md`? The format for entries is:
```
```

niklasmohrin:
This syntax matches all `.conf` files, not only `nginx` specific. I don't know how you want to handle this, just wanted to let you know (see #1137).

sharkdp:
> This syntax matches all `.conf` files, not only `nginx` specific.

Oh, now I see. Apache and Nginx clashed on `*.conf`. I thought we had yet another syntax that would match on `*.conf`. In this case, we could also add an explicit `*.conf => MappingTarget::MapToUnknown` mapping in `src/syntax_mapping.rs`. This would make sure that `*.conf` files would not be matched by any of the two syntaxes.

@niklasmohrin Maybe let's keep both syntaxes in your PR and close this one instead. @kjmph: I hope you are fine with this, even though your PR was here first.

niklasmohrin:
I am sorry @kjmph, I didn't see your pull request at first. I can include your handle in the changelog though, if that is okay for you.

kjmph:
It is all good @niklasmohrin; you are running with the patch, I'm happy to see Nginx conf files make it into bat. No need to worry about credit via changelogs, you got this.

Issue #1076, opened 2020-06-29
Improved handling of file-extension conflicts


sharkdp:
With #1070 merged, we now have four known exceptions to the "no duplicate extensions" rule (see `tests/no_duplicate_extensions.rs`). We have a separate test (`syntax_detection_well_defined_mapping_for_duplicate_extensions`) that makes sure that these extensions are handled as expected (mapped to the desired syntax). For example:
``` rust
assert_eq!(test.syntax_for_file("test.sass"), "Sass");
```

However, the duplicate file extensions still appear in the `bat --list-languages` output for both languages:
```
…
Ruby Haml                            haml, sass
…
Sass                                 sass
…
```
In order to prevent this, when printing the list of syntaxes, we could go through each pattern that is displayed and run it through our syntax mapping function to make sure that it's actually mapped to the current syntax. When printing the extensions for `Ruby Haml`, we would check the `sass` pattern (by getting the syntax for a file named `test.sass` AND for a file named `sass`) and find the `Sass` syntax instead. We could then simply skip that pattern.

sharkdp:
Fixed by @Kienyew in #1135.

sharkdp:
Fixed in [v0.16.0](https://github.com/sharkdp/bat/releases/tag/v0.16.0).

PR #1135, opened 2020-08-08
Handle file extension conflicts in --list-languages (#1076)

Kienyew:
Now the each outputted extension should be unique to only one language. Below is the diff output of the old `bat --list-languages` and new `bat --list-languages`.
```diff
13c13
< C:c,h
---
> C:c
49c49
< GLSL:vs,fs,gs,vsh,fsh,gsh,vshader,fshader,gshader,vert,frag,geom,tesc,tese,comp,glsl
---
> GLSL:vs,gs,vsh,fsh,gsh,vshader,fshader,gshader,vert,frag,geom,tesc,tese,comp,glsl
70,71c70,71
< JavaScript:js,htc
< JavaScript (Babel):js,mjs,jsx,babel,es6,cjs
---
> JavaScript:htc
> JavaScript (Babel):js,mjs,jsx,babel,es6,cjs,*.js
92,93c92,93
< Objective-C:m,h
< Objective-C++:mm,M,h
---
> Objective-C:m
> Objective-C++:mm,M
119c119
< Ruby Haml:haml,sass
---
> Ruby Haml:haml
123c123
< Sass:sass
---
> Sass:sass,*.sass

```

But since #1035, some of the output is appearing in the form `Lang ext, *.ext`.
For instance in the C++ entry of `bat --list-languages`, containing both `h` and `*.h`:

```
...
C++                                 cpp, cc, cp, cxx, c++, C, h, hh, hpp, hxx, h++, inl, ipp, *.h
...
```
and likewise 

```
...
Sass                                 sass, *.sass
...
F#                                   fs, fsi, fsx, *.fs
...
JavaScript (Babel)                   js, mjs, jsx, babel, es6, cjs, *.js
...
```

Should this be the desired behavior? @sharkdp

### DIFF
--- a/src/bin/bat/main.rs
+++ b/src/bin/bat/main.rs
@@ -100,11 +100,8 @@ pub fn list_languages(config: &Config) -> Result<()> {
                 true
             } else {
                 let test_file = Path::new("test").with_extension(extension);
-                if let Some(syntax) = assets.syntax_for_file_name(test_file, &config.syntax_mapping) {
-                    syntax.name == lang_name
-                } else {
-                    false
-                }
+                let syntax = assets.syntax_for_file_name(test_file, &config.syntax_mapping).unwrap();
+                syntax.name == lang_name
             }
         });
     }
--- a/src/bin/bat/main.rs
+++ b/src/bin/bat/main.rs
@@ -86,19 +86,6 @@ pub fn list_languages(config: &Config) -> Result<()> {
         .filter(|syntax| !syntax.hidden && !syntax.file_extensions.is_empty())
         .cloned()
         .collect::<Vec<_>>();
-
-    for lang in languages.iter_mut() {
-        let lang_name = lang.name.clone();
-        lang.file_extensions.retain(|extension| {
-            let test_file = Path::new("test").with_extension(extension);
-            match config.syntax_mapping.get_syntax_for(test_file) {
-                Some(MappingTarget::MapTo(primary_lang)) => lang_name == primary_lang,
-                Some(MappingTarget::MapToUnknown) => false,
-                None => true,
-            }
-        });
-    }
-
     languages.sort_by_key(|lang| lang.name.to_uppercase());
 
     let configured_languages = get_syntax_mapping_to_paths(config.syntax_mapping.mappings());
--- a/src/bin/bat/main.rs
+++ b/src/bin/bat/main.rs
@@ -86,6 +86,29 @@ pub fn list_languages(config: &Config) -> Result<()> {
         .filter(|syntax| !syntax.hidden && !syntax.file_extensions.is_empty())
         .cloned()
         .collect::<Vec<_>>();
+
+    // Handling of file-extension conflicts, see issue #1076
+    for lang in languages.iter_mut() {
+        let lang_name = lang.name.clone();
+        lang.file_extensions.retain(|extension| {
+            // The 'extension' variable is not certainly a real extension.
+            // 
+            // Skip if 'extension' starts with '.', likely a hidden file like '.vimrc'
+            // Also skip if the 'extension' contains another real extension, likely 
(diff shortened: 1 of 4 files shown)


src/syntax_mapping.rs
```diff
@@ -24,6 +24,12 @@ impl<'a> SyntaxMapping<'a> {
         let mut mapping = Self::empty();
         mapping.insert("*.h", MappingTarget::MapTo("C++")).unwrap();
         mapping.insert("*.fs", MappingTarget::MapTo("F#")).unwrap();
+        mapping
+            .insert("*.js", MappingTarget::MapTo("JavaScript (Babel)"))
```
sharkdp, line 28: Ideally, we should be able to implement this feature without adding theses new syntax mappings. Instead of just querying the syntax_mapping, we would need to run the full syntax detection on the `test.xx` filename.

src/bin/bat/main.rs
```diff
@@ -86,6 +86,19 @@ pub fn list_languages(config: &Config) -> Result<()> {
...
         .cloned()
         .collect::<Vec<_>>();
+
+    for lang in languages.iter_mut() {
+        let lang_name = lang.name.clone();
+        lang.file_extensions.retain(|extension| {
+            let test_file = Path::new("test").with_extension(extension);
+            match config.syntax_mapping.get_syntax_for(test_file) {
+                Some(MappingTarget::MapTo(primary_lang)) => lang_name == primary_lang,
+                Some(MappingTarget::MapToUnknown) => true,
```
sharkdp, line 96: Shouldn't this be `false`?

sharkdp:

sharkdp:
> But since #1035, some of the output is appearing in the form `Lang ext, *.ext`.
> For instance in the C++ entry of `bat --list-languages`, containing both `h` and `*.h`:
> Should this be the desired behavior?

I think we can leave this for now. The fact that the list contains `h` (and `*.h`) really means that files named `h` would be highlighted with the C++ syntax. This is a limitation of sublime syntax patterns (that we are partially addressing in `bat` through `syntax_mapping.rs`).

src/bin/bat/main.rs
Kienyew: I think you're right, it should be `false`.

src/syntax_mapping.rs
Kienyew: Ops, I was just following the similar commits before and thought it would do 😅, should I revert it and change the implementation?

sharkdp: Yes, please revert this.

By "full syntax detection", I mean: determine the syntax in the same way that `bat` usually does this for a normal file. This involves not just checking the syntax mapping, but also the full database of syntaxes. See `src/assets.rs` => `HighlightingAssets::get_syntax`.

sharkdp (changes requested):

sharkdp:
Thank you very much for the updates. How does the diff for `bat --list-languages` look like now?

Kienyew:
> Thank you very much for the updates. How does the diff for `bat --list-languages` look like now?

This is the new diff for the new `bat --list-languages` by `diff old.txt new.txt` so far.

```diff
13c13
< C:c,h
---
> C:c
49c49
< GLSL:vs,fs,gs,vsh,fsh,gsh,vshader,fshader,gshader,vert,frag,geom,tesc,tese,comp,glsl
---
> GLSL:vs,gs,vsh,fsh,gsh,vshader,fshader,gshader,vert,frag,geom,tesc,tese,comp,glsl
70c70
< JavaScript:js,htc
---
> JavaScript:htc
92,93c92,93
< Objective-C:m,h
< Objective-C++:mm,M,h
---
> Objective-C:m
> Objective-C++:mm,M
119c119
< Ruby Haml:haml,sass
---
> Ruby Haml:haml
```

sharkdp (approved):

sharkdp:
Thank you very much!

PR #1136, opened 2020-08-08
Add zsh completion

Kienyew:


assets/completions/bat.zsh
```diff
@@ -0,0 +1,56 @@
+#compdef bat
+# FIXME: help me with the subcommand `cache`, zsh completion is hard as hell
```
sharkdp, line 2: I have never written any completion scripts. Let's leave this open until someone can help?

sharkdp:

assets/completions/bat.zsh
Kienyew: Maybe I'll dive into the zsh manual again later, sorry to say but is one of the most difficult manual I have read IMO 🤣

sharkdp: I'd be also okay with merging this without support for the `cache` subcommand.

Kienyew: I've added the completion of `cache` subcommand in https://github.com/sharkdp/bat/pull/1136/commits/32905f78dd60a2a181b3792d73ba8e5701736ddb

sharkdp:
Thank you very much for the updates. We should also make sure that the zsh completions is properly included in the tarballs and the Debian packages (see similar code for fish completions).

```
```

Kienyew:
> ... We should also make sure that the zsh completions is properly included in the tarballs and the Debian packages (see similar code for fish completions).

@sharkdp is the change suppose to be done in [.github/workflows/CICD.yml](https://github.com/sharkdp/bat/tree/master/.github/workflows/CICD.yml)? I don't have any experience in packaging and deploying as well as github action. After some searching I think the zsh completion script should be copy to `/usr/share/zsh/site-functions/_bat` when installing.

Kienyew:
I did a few changes. Tell me if I did something wrong.

sharkdp:
Thank you for the updates. As you can see in the failed tests, the packaging stage currently fails with:

> cp: cannot stat 'target/XXX/release/build/bat-*/out/assets/completions/bat.zsh': No such file or directory

Kienyew:
> Thank you for the updates. As you can see in the failed tests, the packaging stage currently fails with:
Thanks for the reply, I hope the things go without problem now.

sharkdp (approved):

sharkdp:
Thank you for the updates!

sharkdp:
I tried copying the generated file to `/usr/share/zsh/site-functions/_bat` and restarted zsh, but I didn't get any auto-completion for `bat`. How did you test this?

Kienyew:
> I tried copying the generated file to `/usr/share/zsh/site-functions/_bat` and restarted zsh, but I didn't get any auto-completion for `bat`. How did you test this?

zsh do the searches in the `fpath` array variable for various functions, including completion functions.

As of my installation of zsh from manjaro's repository, which is derivative of arch linux's, `/usr/share/zsh/site-functions` is included in `fpath` by default. But after I saw your comment, I then try to install zsh in a ubuntu container, and found that  `/usr/share/zsh/site-functions` is not defaultly included in `fpath`, which means that each distributions and versions might have differences in `fpath` initially. 

But what I found:
https://github.com/zsh-users/zsh/blob/04bd9a44a74683ad0d83921bfb3aa0c4d5992c75/INSTALL#L254-L262
(that is really a mess...)

After all, if the release is a deb package, according to [this answer](https://unix.stackexchange.com/a/607810/311205), I believe we should just change the `/usr/share/zsh/site-functions/` to `/usr/share/zsh/vendor-completions`

sharkdp:
> As of my installation of zsh from manjaro's repository, which is derivative of arch linux's, `/usr/share/zsh/site-functions` is included in `fpath` by default.

Hm, weird. I actually tested on Arch Linux. I can't get the completions working by putting `_bat` there. It also doesn't work in `/usr/share/zsh/vendor-completions`.

> After all, if the release is a deb package, according to [this answer](https://unix.stackexchange.com/a/607810/311205), I believe we should just change the `/usr/share/zsh/site-functions/` to `/usr/share/zsh/vendor-completions`

:+1: 

That's what we do for `fd` as well: https://github.com/sharkdp/fd/blob/74c3431a2bf74ce78c26974f3cc07570e1d3baf3/ci/before_deploy.bash#L148

sharkdp:
Just now I had a chance to finally try out the zsh completions myself. They are absolutely fantastic!! Thank you very much for putting so much work into this.

sharkdp:
also... they have been relased in [v0.16.0](https://github.com/sharkdp/bat/releases/tag/v0.16.0).

mjturner:
@Kienyew This is very neat! 

Perhaps I've missed it, but it wasn't obvious to me that the completion file had to be renamed to `_bat` to work. Not so much an issue for those installing distro packages but I'm using the binary on macOS so had to do the renaming myself. Happy to send a PR for an update to the README (perhaps a section under "Customization"?).

Kienyew:
@mjturner 
> Perhaps I've missed it, but it wasn't obvious to me that the completion file had to be renamed to _bat to work. Not so much an issue for those installing distro packages but I'm using the binary on macOS so had to do the renaming myself. 

I agree with you, it really needs some efforts to search up something about this kind of stuff. 

> Happy to send a PR for an update to the README (perhaps a section under "Customization"?).

How about we just tag one of the maintiner and discuss with them?

mjturner:
> How about we just tag one of the maintiner and discuss with them?

Sounds good.

@sharkdp Your thoughts? Would it make sense to add a "Shell customization" section to the README or would a README in `assets/completions` make more sense (the latter being included in `autocomplete/` when a release is built).

sharkdp:
I think we could add such a section to the README, yes :+1: 

But before that, shouldn't we simply change the name of the file to `_bat`?

mjturner:
Thanks for confirming - if no-one beats me to it I'll try and submit a PR later today.

I think the confusing part is that `_bat` doesn't obviously make it ZSH-specific. Perhaps the completion files could go in shell-specific directories rather than just being in `assets/completions`? There could be `assets/completions/{fish,zsh}` - or is that too much clutter as each would only have a single file?

sharkdp:
> I think the confusing part is that `_bat` doesn't obviously make it ZSH-specific. Perhaps the completion files could go in shell-specific directories rather than just being in `assets/completions`? There could be `assets/completions/{fish,zsh}`

sounds good!

PR #1137, opened 2020-08-11
Nginx and Apache config file syntax highlighting

### LINKED ISSUE #1107: nginx conf files
**Syntax:**
Nginx conf files

**Guideline Criteria:**
https://packagecontrol.io/packages/nginx

bat is brilliant. thank you!

niklasmohrin:
Closes #1107

Hey, I picked this issue to get into `bat`. I added a submodule to the [GitHub repo of the linked syntax highlighting file](https://github.com/brandonwamboldt/sublime-nginx). I tried it out with the example from [the nginx wiki](https://www.nginx.com/resources/wiki/start/topics/examples/full/) and worked as expected.

For Apache, I converted the linked `.tmLanguage` file, but the highlighting is not correct yet (closing tag names are not colored). I am still looking into that, but I wanted to open the pull request early.

That said, thanks for creating `bat`, I use it every day!

### DIFF
--- a/assets/README.md
+++ b/assets/README.md
@@ -31,6 +31,7 @@ themes (`bat cache --clear`).
 
 The following files have been manually modified after converting from a `.tmLanguage` file:
 
+* `Apache.sublime_syntax`=> removed `.conf` and `.CONF` file types.
 * `Dart.sublime-syntax` => removed `#regex.dart` include.
 * `INI.sublime-syntax` => added `.hgrc`, `hgrc`, and `desktop` file types.
 * `Org mode.sublime-syntax` => removed `task` file type.
--- a/assets/syntaxes/02_Extra/Apache.sublime-syntax
+++ b/assets/syntaxes/02_Extra/Apache.sublime-syntax
@@ -1,0 +1,223 @@
+%YAML 1.2
+---
+# http://www.sublimetext.com/docs/3/syntax.html
+name: Apache Conf
+file_extensions:
+  - conf
+  - CONF
+  - envvars
+  - htaccess
+  - HTACCESS
+  - htgroups
+  - HTGROUPS
+  - htpasswd
+  - HTPASSWD
+  - .htaccess
+  - .HTACCESS
+  - .htgroups
+  - .HTGROUPS
+  - .htpasswd
+  - .HTPASSWD
+scope: source.apacheconf
+contexts:
+  main:
+    - match: ^(\s)*(#).*$\n?
+      scope: comment.line.hash.ini
+      captures:
+        1: punctuation.definition.comment.apacheconf
+    - match: (<)(Proxy|ProxyMatch|IfVersion|Directory|DirectoryMatch|Files|FilesMatch|IfDefine|IfModule|Limit|LimitExcept|Location|LocationMatch|VirtualHost|Macro|If|Else|ElseIf)(\s(.+?))?(>)
+      captures:
+        1: punctuation.definition.tag.apacheconf
+        2: entity.tag.apacheconf
+        4: string.value.apacheconf
+        5: punctuation.definition.tag.apacheconf
+    - match: (</)(Proxy|ProxyMatch|IfVersion|Directory|DirectoryMatch|Files|FilesMatch|IfDefine|IfModule|Limit|LimitExcept|Location|LocationMatch|VirtualHost|Macro|If|Else|ElseIf)(>)
+      captures:
+        1: punctuation.definition.tag.apacheconf
+        2: entity.tag.apacheconf
+        3: punctuation.definition.tag.apacheconf
+    - match: (?<=(Rewrite(Rule|Cond)))\s+(.+?)\s+(.+?)($|\s)
+      captures:
+        3: string.regexp.apacheconf
+        4: string.replacement.apacheconf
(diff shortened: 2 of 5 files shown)


niklasmohrin:
When running the tests locally, the test `no_duplicate_extensions` fails, but in CI it passes. Is there any particular reason for that?

I found that the nginx syntax highlighting file is also picked for any `*.conf` file, which conflicts with existing syntaxes. I inspected some `.conf` files on my system. For some, the theme worked fairly well, but there are also exceptions like
image

I don't think that there should be an exception like for `.h` files, because this syntax is not supposed to apply to all possible `conf` files. However, the setting is baked into the upstream file, so I will have to make a local copy for this repository (that has to be manually updated, I think?).

Any opinions on which option to pick?

niklasmohrin:
Okay, it turned out, that the generated Apache syntax file was absolutely fine, but in my testing I looked at `.conf` files, which used the `nginx` syntax and therefore were not displayed correctly. I checked again with the `.htaccess` extension and the output looked correct.

I added a submodule and the hand-generated `.sublime-syntax` file. Is there anything I need to setup to have this file generated automatically?

The generated file also specifies `.conf` as an extension, so there is the same problem as with the `nginx` file. How have you handled this problem in the past?

sharkdp:
Thank you for your contribution. Note that we already have a PR for the nginx syntax here: #1131. Could you please modify this PR to only include the Apache config files?

> I added a submodule and the hand-generated `.sublime-syntax` file. Is there anything I need to setup to have this file generated automatically?

I also use manual conversion from Sublime Text. I'm not sure if there is a way to automate this.

> The generated file also specifies .conf as an extension, so there is the same problem as with the nginx file. How have you handled this problem in the past?

We did not have many clashes so far. I think we should:
* Manually remove `.conf` from the converted sublime-syntax file
* Document the manual change in `assets/README.md` => Manual modifications
* Add a new entry to `src/syntax_mapping.rs` which maps `/etc/apache2/**/*.conf` (or something more specific?) to the new Apache syntax. This way, only `*.conf` files in `/etc/apache2` will use the new syntax.


```
```

> That said, thanks for creating `bat`, I use it every day!

Thank you for the feedback!

sharkdp:
> ~When running the tests locally, the test `no_duplicate_extensions` fails, but in CI it passes. Is there any particular reason for that?~

Note that we only have one CI run ("Test with new syntaxes and themes") which actually runs `assets/create.sh` before running the tests. The other jobs will run with the old syntaxes (mainly for saving on overall CI time).

niklasmohrin:
Okay, sounds good! I did not see #1131, I will remove that commit. Still, the `nginx` syntax used here and in #1131 matches all `.conf` files as well, so that should be manually patched out. I will write a comment in the right thread.

As far as I can tell, the Apache config files only live in the `/etc/apache2/` directory, so this restriction should be fine. I think this does not need to be more specific. The only situation I could think of, where this would be a problem, would be serving `.conf` files statically, but this is obviously an edge case.

What do you think?

sharkdp:
> As far as I can tell, the Apache config files only live in the `/etc/apache2/` directory, so this restriction should be fine. I think this does not need to be more specific. The only situation I could think of, where this would be a problem, would be serving `.conf` files statically, but this is obviously an edge case.
> 
> What do you think?

I think we can ignore that edge case, agreed :+1:

niklasmohrin:
Okay, so I did some more research and added some more filenames in `src/syntax_mapping.rs`. Other than that, this should be all on this pull request. I can squash the commits into two or three bigger commits, if you want.

assets/README.md
```diff
@@ -31,6 +31,7 @@ themes (`bat cache --clear`).
 
 The following files have been manually modified after converting from a `.tmLanguage` file:
 
+* `Apache.sublime_syntax`=> removed `.conf` and `.CONF` file types.
```
sharkdp, line 34: Do we still need this if we have the `"*.conf", MappingTarget::MapToUnknown` mapping?

assets/syntaxes/02_Extra/Apache.sublime-syntax
```diff
@@ -0,0 +1,221 @@
+%YAML 1.2
+---
+# http://www.sublimetext.com/docs/3/syntax.html
+name: Apache Conf
+file_extensions:
+  - envvars
+  - htaccess
```
sharkdp, line 7: This already covers files called `htaccess`, `.htaccess` or `test.htaccess`, right? I don't think we need the additional mappings in `syntax_mapping.rs`?

src/syntax_mapping.rs
```diff
@@ -48,6 +48,34 @@ impl<'a> SyntaxMapping<'a> {
...
+            mapping.insert(glob, MappingTarget::MapTo("nginx")).unwrap();
+        }
+
+        for glob in &[
+            "/etc/apache2/**/*.conf",
+            "/etc/apache2/sites-*/**/*",
+            "httpd.conf",
+            "*.htaccess",
+            "*.htgroup",
+            "*.htpasswd",
```
sharkdp, line 72: see above. I don't think these are needed.

assets/README.md
niklasmohrin: Strangely, the test `no_duplicate_extensions` fails when the lines are not removed from the Apache syntax file. Should I add an entry to `KNOWN_EXCEPTIONS` in this test?

sharkdp:
Thank you very much!

Issue #1108, opened 2020-07-16
add a decorations alias


lyndell:
An alias for `--decorations always` so I can pipe the decorations when doing something like `bat FILE | gview -`.  …but without the `^M`'s.

sharkdp:
Sounds good to me. Which flag would you suggest? You also want colors to be enabled, right? (not just grid, line numbers, changes).

lyndell:
`-d` appears available.  Colors could be a problem in `gview` or `text/plain` MIME type,  though it might be nice for someone dumping to monospace HTML.  Maybe use `-h` or `--html` for an HTML dump with colors.  Grid and line numbers would be fine for a pretty-print `cat` or `head` alternative.  `bat FILE | gview -` is the use-case  time **I** would use.

sharkdp:
>  Colors could be a problem in `gview` or `text/plain` MIME type, though it might be nice for someone dumping to monospace HTML. Maybe use `-h` or `--html` for an HTML dump with colors.

How is this related to the feature in question here? We are not planning to add a HTML output feature to `bat`. Note that there are thing like `ansi2html`.

Maybe I don't yet understand your use case. What do you use `… | giew -` for? Is `bat`s paging option not working for you?

lyndell:
`gview`is an excellent file viewer, especially when I want to break out of the Terminal window.  Concatenating the files with `bat` decorations makes the distinct file and line numbers clear.  One example is viewing Kubernetes with `bat pod.yaml deployment.yaml service.yaml | gview -`.  There are times I'd rather use that in stead of `-o` or -O` on `gview`.  It makes the search and navigation easier.

alexanderkarlis:
I was hoping to give this issue a shot. But I noticed that there already is an alias flag for -d (diff). Would it be a good idea to add it as `-D` instead of `-d`? @sharkdp

lyndell:
Sounds good to me.

sharkdp:
If this enables `--color=always` and `--decorations=always`, we could also use `-c`, as the major thing (for me) would be the colorized output, not the decorations.

sharkdp:
Closed via #1141 by @alexanderkarlis

sharkdp:
Released in [v0.16.0](https://github.com/sharkdp/bat/releases/tag/v0.16.0).

PR #1141, opened 2020-08-16
Added new alias for 'decoration=always'

alexanderkarlis:
This PR addresses #1108 .

Hey all, first time making a PR for a rust repo and still pretty new to it.

Think I addressed the feature here for adding an alias for `--decoration=always` with the `-D` arg. The `-D` commands works the same as if `--decoration=always` or if running on a terminal that can support decorations (e.g- `--decoration=auto`)

Let me know. Thanks!


sharkdp:
Thank you very much. The implementation looks good.

My only concern is the letter for the flag. I would actually like to combine `--decorations=always` with `--color=always` into a single flag. I was thinking about `-f` for "force color and decorations", which would still be free (it's also not used in `cat`).

What do you think?

Also, @eth-p: how do you feel about introducing a new short option for this? I tend to use `--color=always` a lot, but my use of `bat` is maybe not the best indicator.

mouse07410:
I think it's a great idea.

eth-p:
I think it would be a good idea. I've found myself wanting to pipe bat output into other programs before, and it's a bit inconvenient to have to pass both by full name.

On top of that, it would also be a nice analogy to the -p flag (albeit doing the exact opposite thing).

alexanderkarlis:
I think that's a good tip. And a seemingly useful alias combo. I will give it another go with both flags.

src/bin/bat/clap_app.rs
```diff
@@ -257,6 +257,15 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
                      an interactive terminal is detected. Possible values: *auto*, never, always.",
                 ),
         )
+        .arg(
+            Arg::with_name("always-decorations")
+                .short("f")
+                .alias("always-decor")
```
sharkdp, line 263: I don't see much value in this. For me, long options are useful for scripts, due to the self-documenting character. Short options are useful when I have to type it in the console. An abbreviated long option would be for what?

src/bin/bat/clap_app.rs
```diff
@@ -257,6 +257,15 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
...
+                .overrides_with("always-decorations")
```
sharkdp, line 264: Should it also "conflict with" `--color=…` and `--decorations=…`?

src/bin/bat/clap_app.rs
```diff
@@ -257,6 +257,15 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
...
+                .hidden(true)
```
sharkdp, line 265: Hiding from short `-h` text is fine, but I think this should appear in the long `--help` text.

src/bin/bat/clap_app.rs
```diff
@@ -257,6 +257,15 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
...
+                .hidden_short_help(true)
+                .help("Alias for '--decorations=always --color=always'")
```
sharkdp, line 267: Can we please use a proper sentence here like for the rest of the command line options? Maybe we could also add:

"This is useful if the output of bat is piped to another program, but you want to keep the colorization/decorations."

src/bin/bat/clap_app.rs
```diff
@@ -257,6 +257,15 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
                      an interactive terminal is detected. Possible values: *auto*, never, always.",
                 ),
         )
+        .arg(
+            Arg::with_name("always-decorations")
```
sharkdp, line 261: With the extended functionality, I would rename this to `--force-colorization` maybe?

sharkdp:
Could you please also update the man page in `assets/manual/bat.1.in`?

```
```

sharkdp:

PR #1195, opened 2020-09-23
Support XDG_CONFIG_HOME git config files

### LINKED ISSUE #1191: Git config inside $XDG_CONFIG_HOME/git is not highlighted
bat 0.15.4

Git configuration files inside `$XDG_CONFIG_HOME/git/{ignore,config,attributes}` are not highlighted

From `man git-config`

```
FILES
       If not set explicitly with --file, there are four files where git config will search for
       configuration options:

       $(prefix)/etc/gitconfig
           System-wide configuration file.

       $XDG_CONFIG_HOME/git/config
           Second user-specific configuration file. If $XDG_CONFIG_HOME is not set or empty,
           $HOME/.config/git/config will be used. Any single-valued variable set in this file will be
           overwritten by whatever is in ~/.gitconfig. It is a good idea not to create this file if you
           sometimes use older versions of Git, as support for this file was added fairly recently.

       ~/.gitconfig
           User-specific configuration file. Also called "global" configuration file.

       $GIT_DIR/config
           Repository specific configuration file.

       $GIT_DIR/config.worktree
           This is optional and is only searched when extensions.worktreeConfig is present in
           $GIT_DIR/config.
```

They should be highlighted

homebrew

---

system
------

**$ uname -srm**
Darwin 19.6.0 x86_64

**$ sw_vers**
ProductName:    Mac OS X
ProductVersion: 10.15.6
BuildVersion:   19G2021

bat
---

**$ bat --version**
bat 0.15.4

**$ env**
BAT_CONFIG_PATH=/Users/ahmed/.config/bat/config
PAGER=less

bat_config
----------

**$ cat /Users/ahmed/.config/bat/config**
```
--theme="TwoDark"
--style="numbers,changes,header"
--italic-text=always
--map-syntax '.*ignore:Git Ignore'
--map-syntax '.gitconfig.local:Git Config'
--map-syntax '.zsh*:Bourne Again Shell (bash)'
--map-syntax '.vimrc.local:VimL'
--map-syntax 'vimrc:VimL'
```

bat_wrapper
-----------

No wrapper script for 'bat'.

bat_wrapper_function
--------------------

**$ /usr/local/bin/zsh --login -i -c type cat**
```
cat is an alias for bat
cat is /usr/local/opt/coreutils/libexec/gnubin/cat
cat is /bin/cat
```

tool
----

**$ less --version**

ahmedelgabri:
@sharkdp this currently what I did & I'm stuck here, not sure why the test
doesn't run.

Closes #1191

### DIFF
--- a/src/syntax_mapping.rs
+++ b/src/syntax_mapping.rs
@@ -114,7 +114,7 @@ impl<'a> SyntaxMapping<'a> {
                         .to_string_lossy(),
                     MappingTarget::MapTo("Git Config"),
                 )
-                .unwrap();
+                .ok();
 
             mapping
                 .insert(
@@ -124,7 +124,7 @@ impl<'a> SyntaxMapping<'a> {
                         .to_string_lossy(),
                     MappingTarget::MapTo("Git Ignore"),
                 )
-                .unwrap();
+                .ok();
 
             mapping
                 .insert(
@@ -134,7 +134,7 @@ impl<'a> SyntaxMapping<'a> {
                         .to_string_lossy(),
                     MappingTarget::MapTo("Git Attributes"),
                 )
-                .unwrap();
+                .ok();
         }
 
         mapping
--- a/src/syntax_mapping.rs
+++ b/src/syntax_mapping.rs
@@ -193,23 +193,6 @@ fn basic() {
 }
 
 #[test]
-fn git_xdg_config_home() {
-    use assert_cmd::Command;
-    let mut cmd = Command::cargo_bin("bat").unwrap();
-    let mut map = SyntaxMapping::builtin();
-
-    cmd.env("XDG_CONFIG_HOME", "/foo/bar");
-
-    map.insert("/foo/bar/git/config", MappingTarget::MapTo("Git Config"))
-        .ok();
-
-    assert_eq!(
-        map.get_syntax_for("/foo/bar/git/config"),
-        Some(MappingTarget::MapTo("Git Config"))
-    );
-}
-
-#[test]
 fn user_can_override_builtin_mappings() {
     let mut map = SyntaxMapping::builtin();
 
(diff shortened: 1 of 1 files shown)


src/syntax_mapping.rs
```diff
@@ -1,3 +1,4 @@
+use assert_cmd::Command;
```
sharkdp, line 1: It's probably easier to use an integration test instead of a unit test (in `tests/`). Otherwise, you need to move this import to the test code.

ahmedelgabri: Thanks, I have moved the import to the test code & now everything is working as expected & tests are passing.

src/syntax_mapping.rs
```diff
@@ -155,6 +192,23 @@ fn basic() {
...
+    cmd.env("XDG_CONFIG_HOME", "/foo/bar");
+
+    map.insert("/foo/bar/git/config", MappingTarget::MapTo("Git Config"))
+        .ok();
+
+    assert_eq!(
+        map.get_syntax_for("/foo/bar/git/config"),
+        Some(MappingTarget::MapTo("Git Config"))
+    );
+}
```
sharkdp, line 210: This does not test the new functionality at all.

1. `cmd.env(…)` will set up a new environment for a command which is yet to be run. But then you never actually run it. The whole `cmd`-related code here has no effect.
2. The test still passes without that code, because you are inserting `/foo/bar/git/config` into the mapping explicitly. It will obviously return the right value in that case.

sharkdp: I removed the unit test. The other code looks good - thank you! I think it's okay to merge this without a test. Writing a proper test for this is nontrivial.

sharkdp:
I made a few changes - hope that's fine for you.

In Rust, we try to prevent `.unwrap()` calls that could actually fail. Here, `XDG_CONFIG_HOME` could have contained an invalid UTF-8 path, which would have lead `.as_str().unwrap()` to fail.

ahmedelgabri:
Thank you for helping & merging the PR I learn quite some stuff while doing this.

PR #1209, opened 2020-10-03
CI: use checkout v2

imba-tjd:
The main difference is that by default v2 uses `--depth=1`.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Git checkout
-      uses: actions/checkout@v1
+      uses: actions/checkout@v2
     - name: Install rust toolchain (v${{ env.MIN_SUPPORTED_RUST_VERSION }})
       uses: actions-rs/toolchain@v1
       with:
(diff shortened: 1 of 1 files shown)


sharkdp:
So `git clone` runs with `--depth=1` then? Don't we control that via
```yaml
    - name: Git checkout
      uses: actions/checkout@v1
      with:
        submodules: true # we need all syntax and theme submodules
```
already?

imba-tjd:
You can see https://github.com/sharkdp/bat/runs/1201014919
and my change https://github.com/sharkdp/bat/runs/1201637712

sharkdp:
Could you please point out the change/advantages to me? I don't want to guess or study everything in detail myself, if there is nothing obviously wrong at the moment.

imba-tjd:
v2 has `--depth=1`:

```
/usr/bin/git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin +7365a2f5eac1c7200265e12c221707aaf654fcf0:refs/remotes/pull/1209/merge
```

v1:

```
git -c http.extraheader="AUTHORIZATION: basic ***" fetch --tags --prune --progress --no-recurse-submodules origin +refs/heads/*:refs/remotes/origin/*
```

v2's message has better layout:

图片

图片

SuperSandro2000:
@sharkdp ``--depth=1`` only loads the latest commit for each file and not the entire history. This is faster and saves disk space and bandwidth.

Issue #1189, opened 2020-09-17
No highlight for Weechat ".conf" files 


Bumbadawg:
Hi,
Is it normal that i have to force INI language for .conf files ?

sharkdp:
I don't think every possible `.conf` file will have INI syntax, so I don't think we should add a mapping like this. If you want, you can add
```
--map-syntax '*.conf:INI'
```
to the bat config file, and it should work.

Bumbadawg:
Thanks that's perfect.

sharkdp:
I think this would be a great example for the README page (in a new section that explains `--map-syntax`), if someone wants to add this.

henil:
@sharkdp I can do that. Is it still pending?

sharkdp:
@henil yes :+1:

sharkdp:
Closing via #1212 by @henil.

PR #1212, opened 2020-10-03
Add new example for --map-syntax

henil:
As mentioned in #1189

### DIFF
--- a/README.md
+++ b/README.md
@@ -490,7 +490,7 @@ alias cat="bat --theme=\$(defaults read -globalDomain AppleInterfaceStyle &> /de
 
 ### Syntax of a file
 
-You can configure `bat` to change the default syntax of the file using `--map-syntax`.
+You can configure `bat` to change the default syntax of the file using `--map-syntax`. Use glob pattern to match exact filename.
 
 For example:
 ```bash
--- a/README.md
+++ b/README.md
@@ -524,12 +524,19 @@ Example configuration file:
 
 # Use C++ syntax for .ino files
 --map-syntax "*.ino:C++"
+```
+
+### Syntax of a file
+
+`bat` can be configured to change the default syntax of the file using `--map-syntax`. For example not all `.conf` file's have `INI` syntax(some do) so it cannot be default. But you can configure it if you want.
+
+For example:
+```bash
+# Use INI syntax for .conf files
+--map-syntax "*.conf:INI"
 
 # Use ".gitignore"-style highlighting for ".ignore" files
 --map-syntax ".ignore:Git Ignore"
-
-# Use INI syntax for .conf files
---map-syntax "*.conf:INI"
 ```
 
 ## Using `bat` on Windows
--- a/README.md
+++ b/README.md
@@ -490,10 +490,10 @@ alias cat="bat --theme=\$(defaults read -globalDomain AppleInterfaceStyle &> /de
 
 ### Syntax of a file
 
-You can change the default syntax highlighting of the file using the `--map-syntax` option.
-It uses glob pattern matching and is matched on full path and filename.
+You can change the default syntax association of the file using the `--map-syntax` option.
+It supports glob pattern matching and is matched on full path and filename.
 
-Use it like this:
+You can use it like this:
 
 ```bash
 bat --map-syntax='PATTERN:SYNTAX' file-name
--- a/README.md
+++ b/README.md
@@ -488,6 +488,19 @@ and the `GitHub` theme when in the _light mode_.
 alias cat="bat --theme=\$(defaults read -globalDomain AppleInterfaceStyle &> /dev/null && echo default || echo GitHub)"
 ```
 
+### Syntax of a file
+
+You can configure `bat` to change the default syntax of the file using `--map-syntax`.
+
+For example:
+```bash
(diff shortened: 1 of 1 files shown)


sharkdp:

Probably it wasn't completely clear from what I wrote in #1118, but my idea was to create a completely new (small) section in the README (in the customization section? https://github.com/sharkdp/bat#customization) to describe the `--max-syntax` feature... with the `*.conf` => INI case as an example.

henil:
Ahh okay. I will make that change and move all the `--map-syntax` examples to that section.

sharkdp:
> and move all the `--map-syntax` examples to that section.

We can show all examples in the new section, but one `--map-syntax` example can stay in the example config file, I think.

henil:
Feel free to suggest better wording in the statement.

sharkdp:
It's in the wrong section now, right? Should be in "Customization".

There are also a few typos/mistakes in the wording. It would be great if you could clean this up.

henil:
Oh sorry didn't see the configuration file was not a subjection.

henil:
Is it better now?

sharkdp:
I'm sorry, but this is not quite what I want. The new section should explain the command-line option `--max-syntax`. You are now showing part of `bat`s config file which hasn't even been introduced at this point in the README (when reading top to bottom).

henil:
> The new section should explain the command-line option `--max-syntax`.

Here you mean `--map-syntax` right?

> but my idea was to create a completely new (small) section in the README (in the customization section?)

As you asked its in the sub-section of `customization` only. Did you mean anything different?

> You are now showing part of `bat`s config file which hasn't even been introduced at this point in the README (when reading top to bottom).

Sorry I am not able to understand what you mean by this.

henil:
@sharkdp pinging you just in case it got buried down in your notification.

henil:
Okay, let me give it a last try. If you are not satisfied with that I will close the PR.

sharkdp:
Sure, thank you!"

henil:
> Not with --map-syntax inside bats config file
I was not aware of `bat` having a config file too. That was the reason for confusion.

sharkdp:
> I was not aware of `bat` having a config file too. That was the reason for confusion. Sorry for that.

no worries, thanks for the update. This looks much better.

There are some minor open points regarding wording/spelling. Maybe someone who actually speaks english could have a look? :smile: (maybe @eth-p if you find the time?)

README.md
```diff
@@ -488,6 +488,28 @@ and the `GitHub` theme when in the _light mode_.
 alias cat="bat --theme=\$(defaults read -globalDomain AppleInterfaceStyle &> /dev/null && echo default || echo GitHub)"
 ```
 
+### Syntax of a file
+
+You can change the default syntax highlighting of the file using the `--map-syntax` option.
```
eth-p, line 493: This would make it more difficult for ESL speakers to understand, but I feel like this can be expressed more accurately as:

> You can change the default syntax association of files by using the `--map-syntax` option.

README.md
```diff
@@ -488,6 +488,28 @@ and the `GitHub` theme when in the _light mode_.
...
+It uses glob pattern matching and is matched on full path and filename.
```
eth-p, line 494: "It supports" instead of "it uses"

This one is more of a pedantic nitpick (and completely optional), but I feel like it reads better. The option does use globs, but saying it supports them is more accurate since it doesn't have the implication of `uses -> relies on` behind it.

README.md
```diff
@@ -488,6 +488,28 @@ and the `GitHub` theme when in the _light mode_.
...
+
+Use it like this:
```
eth-p, line 496: "You can use it like this:"

Small change, but adding "you can" changes the tone to be more welcoming and less instructional. I can't seem to explain why, though 😅

eth-p (commented):
If possible, could you also mention that the `--map-syntax` option is meant to go in the config file and/or move it below the config file section? Without knowing about the config file, someone might get the impression that you have to specify the argument every time (which is just silly)..

Other than those couple of things, it looks great!

henil:
> moving the entire section below the config file section

Do you mean the whole `Customization` section or just the section explaining `--map-syntax`? Because @sharkdp  wanted to have a section containing `--map-syntax` before the `Configuration` section.

README.md
henil: Actually I am an ESL speaker myself. But yeah it is easier to read.

README.md
henil: Not pedantic at all. It sounds better.

README.md
henil: Agreed. Done all the changes in the latest commit.

eth-p:
> > moving the entire section below the config file section
> 
> Do you mean the whole `Customization` section or just the section explaining `--map-syntax`? Because @sharkdp  wanted to have a section containing `--map-syntax` before the `Configuration` section.

Hmm. Alright, then! I trust his reasons for preferring it that way, so it should be fine as is.

eth-p (approved):
Everything looks good! Thanks!

henil:
@sharkdp can you review again?

sharkdp:
@henil Thank you for the updates
@eth-p Thank you for the review

PR #1214, opened 2020-10-03
Mention Homebrew on Linux 🍺🐧

TGotwig:
Was already published on `Homebrew on Linux` by someone else, 
not sure why this isn't already mentioned inside the README 🤔🙂

btw the following link to the `Homebrew on MacOS` formula seems to be wrong 🧐

```
You can install `bat` with [Homebrew](http://braumeister.org/formula/bat):
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,6 +276,14 @@ You can install `bat` with zypper:
 zypper install bat
 ```
 
+### On Linux
+
+You can install `bat` with [Homebrew](https://formulae.brew.sh/formula-linux/bat)::
+
+```bash
+brew install bat
+```
+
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you for your contribution. Instead of naming the new section "On Linux", can we please rename the existing section "On macOS" to "On macOS (or Linux) via Homebrew"? And replace the outdated link with the one you added here?

I don't  think we need two separate sections and "On Linux" is too general for me.

Thanks!

TGotwig:
Done 👍✨

sharkdp (approved):
Thank you!

Issue #1213, opened 2020-10-03
Complete collection of syntax highlighting test files


sharkdp:
*With all the recent news about Hacktoberfest I thought it would be a good idea to point out good beginner issues that would be actually helpful for `bat`. In the past years, I have actually experienced Hacktoberfest as a really great event - both as a contributor as well as a maintainer.*

As of [recently](https://github.com/sharkdp/bat/pull/1126), `bat` has a set of syntax highlighting regression tests (see #1124 for more details). The main idea is that we have a large collection of test files for each and every language that `bat` can highlight. This way, we can make sure that we do not run into issues we had in the past where either (1) syntax highlighting for some language is suddenly not working anymore or (2) `bat` suddenly crashes/panics for some input (due to incompatibilities in the `regex` flavors in `syntect` and Sublime Text).

In order to add a new test file, you can follow these steps (let's take "Ruby" as an example):

1. Make sure that you are running the **latest version of `bat`** (`master` or `bat` 0.16) and that `bat` is available on the path.
2. Find an example Ruby source file or write one yourself. If possible, the file should aim to be "comprehensive" (i.e. include a lot of the possible syntax), but this is not strictly necessary. A simple file is better than none at all. Also, the files shouldn't be gigantic.
3. Save the file in `tests/syntax-tests/source/Ruby` (adapt for your language). The file name could be `test.rb` (adapt extension) but can also be adapted if that is necessary in order for `bat` to highlight it correctly (e.g. `Makefile`).
4. If you have copied the file from somewhere else, please make sure that the file *may* be copied under the respective license and that the license is compatible with `bat`s license. If it requires attribution, please add a `LICENSE.md` in the same folder with a text like this:
    ```
    The `test.rb` file has been added from [enter source here] under the following license: 
    
    [add license text here]
    ```
5. Go to `tests/syntax-tests` and run the `update.sh` Bash script. A new file should be generated in the `highlighted` folder (e.g. `highlighted/Ruby/test.rb`).
6. Use `cat` or `bat --language=txt` to display the content of this file and make sure that the syntax highlighting looks correct.
7. `git add` the new files in the `source` folder as well as the autogenerated files in the `highlighted` folder.
8. Commit and submit a PR! Please reference this issue (#1213).

List of languages / syntaxes:
- [x] **ActionScript**
- [x] Apache Conf
- [x] **AppleScript**
- [x] ARM Assembly
- [x] AsciiDoc (Asciidoctor)
- [x] **ASP**
[...]

PR #1218, opened 2020-10-03
Added Elixir file example

uduDudu:
Added a elixir example file from a discord bot that i made for #1213, i choose it because i think that it covers a good number of the language features that should be highlighted


sharkdp:
Thank you very much!

PR #1220, opened 2020-10-04
Add go highlight test

iamquang95:
This is a part of https://github.com/sharkdp/bat/issues/1213. Simple golang example is added, the file is from https://github.com/golang/example which is under Apache License


sharkdp:
Thank you!

PR #1221, opened 2020-10-04
Update README.md using cargo install --path

iamquang95:
When i follow the `Development` part in readme to set up project, the `cargo install --locked` is deprecated with error
`error: Using `cargo install` to install the binaries for the package in current working directory is no longer supported, use `cargo install --path .` instead. Use `cargo build` if you want to simply build the package.`

My setup:
```
cargo --version                                                                                                                                         
cargo 1.45.1 (f242df6ed 2020-07-22)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -621,11 +621,11 @@ cargo build --bins
 cargo test
 
 # Install (release version)
-cargo install --locked
+cargo install --path . --locked
 
 # Build a bat binary with modified syntaxes and themes
 bash assets/create.sh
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you very much!

We should still keep the `--locked` option though. It would be great if you could adapt accordingly.

iamquang95:
Sorry, i forgot to put `--locked` back. I updated the commit.

sharkdp:
Thank you!

PR #1222, opened 2020-10-04
Add an OCaml file for syntax tests.

dmbaturin:
For the issue #1213 

I wrote that file from scratch, you can use it without attribution. If you want me to add CC-Zero or the Unlicense to make it explicit, I can do that.

It's reasonably comprehensive and in fact has found a highlighting bug already: `val` inside objects isn't highlighted as a keyword. Not sure whether I should open an issue here or in the sublime syntax highlighting file collection repo.


sharkdp:
Thank you very much!



> It's reasonably comprehensive and in fact has found a highlighting bug already: `val` inside objects isn't highlighted as a keyword. Not sure whether I should open an issue here or in the sublime syntax highlighting file collection repo.

It would be great if you could open an issue in the sublime syntax repo! Once the fix is merged upstream, we will include it automatically in `bat` with the next automated submodule update.

PR #1223, opened 2020-10-04
Make tests/syntax-tests/update.sh shell script executable.

dmbaturin:
Issue #1213 suggests running `update.sh`, but that file doesn't have a `+x` bit which makes it a bit harder to execute than it should be.


sharkdp:
Thank you!

PR #1224, opened 2020-10-04
Add Bash syntax tests.

eth-p:
I thought it might be fun to add a Bash syntax test or two.

I added the following syntax tests for Bash:

- `simple.sh` - A small shell script that attempts to demonstrate some commonly-known Bash syntax and features.
- `batgrep.sh` - The entire compiled batgrep shell script.

And I also fixed the `update.sh` shell script to work on Mac OS, which has the `python` executable as Python 2 by default.


sharkdp (approved):
Thank you!

PR #1225, opened 2020-10-04
Add scala code highlight test

iamquang95:
This is a part of #1213. Simple Scala example is added, the file is from https://github.com/typelevel/cats-effect which is under Apache License


sharkdp:
Thank you very much! I have just updated the instructions in #1213 slightly (sorry for that!). It would be great if you could update the PR accordingly (mainly the change in `LICENSE.md`).

iamquang95:
Done, i added the original source of test file.

PR #1226, opened 2020-10-04
Shell script improvements

mdevlamynck:
This PR adds a few improvements to the shell scripts of this project:
* replacing `#!/bin/bash` with `#!/usr/bin/env bash` to improve compatibility with system where `/bin/bash` is absent (for instance NixOS)
* adding the execution flag
* adding `cd $(dirname $0)` to allow running the scripts from another directory (for instance form root `./tests/syntax-tests/regression_test.sh`). I tested every scripts, some of them were already working without.

### DIFF
--- a/tests/benchmarks/comparison.sh
+++ b/tests/benchmarks/comparison.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+cd "$(dirname "${BASH_SOURCE[0]}")"
+
 if ! which hyperfine > /dev/null 2>&1; then
     echo "'hyperfine' does not seem to be installed."
     echo "You can get it here: https://github.com/sharkdp/hyperfine"
--- a/tests/benchmarks/comparison.sh
+++ b/tests/benchmarks/comparison.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 if ! which hyperfine > /dev/null 2>&1; then
     echo "'hyperfine' does not seem to be installed."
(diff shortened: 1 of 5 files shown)


tests/benchmarks/comparison.sh
```diff
@@ -1,4 +1,6 @@
-#!/bin/bash
+#!/usr/bin/env bash
+
+cd $(dirname $0)
```
eth-p, line 3: This will break if the path contains spaces.

It's also better form to be using "${BASH_SOURCE[0]}", which contains the proper path when sourcing the file with `. ` or `source`. Shouldn't matter too much in this case, but it's good to keep in mind.

I would recommend using the following instead:

```bash
cd "$(dirname "${BASH_SOURCE[0]}")"
```

tests/benchmarks/run-benchmarks.sh
```diff
@@ -1,4 +1,6 @@
-#!/bin/bash
+#!/usr/bin/env bash
+
+cd $(dirname $0)
```
eth-p, line 3: Same unquoted expansion issue as the last comment.

tests/syntax-tests/update.sh
```diff
@@ -1,4 +1,6 @@
-#!/bin/bash
+#!/usr/bin/env bash
+
+cd $(dirname $0)
```
eth-p, line 3: Same unquoted expansion issue as the last comment.

eth-p (changes requested):

tests/benchmarks/comparison.sh
mdevlamynck: I didn't know that, thanks :)

eth-p (approved):

sharkdp:
@mdevlamynck Thank you very much for your contribution!

@eth-p Thank you for the review - feel free to merge this.

PR #1227, opened 2020-10-04
Add elm syntax regression test

mdevlamynck:
Adds an elm example as part of the #1213 regression test suite. This is the [triangle](https://elm-lang.org/examples/triangle) example from https://elm-lang.org/examples. The original file comes from https://github.com/elm/elm-lang.org/ where the [license](https://github.com/elm/elm-lang.org/blob/master/LICENSE) can be found.


sharkdp:
The tests are failing. Could it be that you are not running the latest version of `bat`?

mdevlamynck:
Indeed, my system was running an old version of bat. The highlighting is even better now ^^

I'm wondering, maybe the test script shouldn't rely on the system installed bat, but instead on a locally built version?

sharkdp:
> I'm wondering, maybe the test script shouldn't rely on the system installed bat, but instead on a locally built version?

I was wondering that too. We could run `cargo run --release` instead of `bat`, but that would potentially run a build first. It would also add a dependency on `cargo` (which is obviously a build depedency of `bat`, but isn't required for changes like this for now).

In the current version, the script gives control to the user on which version to use. We use this in our CI pipeline because there we want to run `assets/create.sh` first, then install `bat`, and then run the script, in order to test with the very latest syntax highlighting changes.

PR #1228, opened 2020-10-04
Add a ruby source code syntax test

Liberatys:
Issue: #1213 

Add a ruby source for syntax highlighting checks.
Found a bit bad code that I wrote some time ago. Not subject to any lisence


sharkdp:
Thank you very much!

PR #1229, opened 2020-10-04
Add asciidoc highlight test

hertg:
This is part of #1213. I have created the test file myself based on the [AsciiDoctor Syntax Quick Reference](https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#).

I actually found a bug on the syntax highlighting with this one. The first line of the following syntax seems to break the highlighting, that's why it's commented out inside the `test.adoc` file for now, the rest of the syntax highlighting works perfectly. 

```adoc
* [*] checked
* [x] also checked
* [ ] not checked
*     normal list item
```

The first line breaks the markup but is actually considered valid markup based on the [quick reference](https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#).
image

Should I report that on the [syntect](https://github.com/trishume/syntect/) repo, as it's probably an issue over there?


sharkdp:

> I actually found a bug on the syntax highlighting with this one. The first line of the following syntax seems to break the highlighting, that's why it's commented out inside the `test.adoc` file for now, the rest of the syntax highlighting works perfectly.
> 
> ```
> * [*] checked
> * [x] also checked
> * [ ] not checked
> *     normal list item
> ```

Interesting. It seems to be broken on GitHub as well, right?

> Should I report that on the [syntect](https://github.com/trishume/syntect/) repo, as it's probably an issue over there?

If it really is a bug, it should be reported in the AsciiDoc sublime syntax repository (see `.gitmodules` file).

sharkdp:
Merging this as-is for now, thank you!

hertg:
> Interesting. It seems to be broken on GitHub as well, right?

Yeah, i saw that too. But to be fair, Github doesn't really have full-fledged AsciiDoctor markup, so that's pretty much expected i think. :)

> If it really is a bug, it should be reported in the AsciiDoc sublime syntax repository (see `.gitmodules` file).

I just opened an issue for this in the repository you mentioned. But it seems that the repo hasn't been updated in 5 years. There are also open PRs lying around since 3+ years, so I don't expect much to happen. 

It might be worth to consider another library for the asciidoc highlighting, doesn't really pressure though, should I create an issue about that here so it can be resolved in the future?

P.S. I just checked how the highlighting looks in [vim](https://github.com/vim/vim), and it seems that the problem doesn't exist there. But they have implemented the syntax highlighting [completely themselves](https://github.com/vim/vim/blob/master/runtime/syntax/asciidoc.vim) rather than relying on an external dependency.  Their highlighting also lacks some features, but at least it doesn't break anything. :sweat_smile:

sharkdp:
Thanks!

> I just opened an issue for this in the repository you mentioned. But it seems that the repo hasn't been updated in 5 years. There are also open PRs lying around since 3+ years, so I don't expect much to happen.

oh :-/

> It might be worth to consider another library for the asciidoc highlighting, doesn't really pressure though, should I create an issue about that here so it can be resolved in the future?

If there is an alternative project - yes please. Otherwise, it's probably going to stay open for a long time.

Another thing that we can do is to fork the upstream repository, potentially merge some of the open PRs and then change our submodule pointer to the fork. I personally don't want to maintain such a fork for AsciiDoc, but if someone wants to do that, I'd be open to retarget our submodule.

hertg:
> If there is an alternative project - yes please. Otherwise, it's probably going to stay open for a long time.

I created the issue #1290 to make sure it's in the backlog.

> Another thing that we can do is to fork the upstream repository, potentially merge some of the open PRs and then change our submodule pointer to the fork. I personally don't want to maintain such a fork for AsciiDoc, but if someone wants to do that, I'd be open to retarget our submodule.

There actually are already some forks that are more up-to-date, but I don't know if those are just forks meant for personal use or if they are trying to continue the project.

- [AlexCzar/sublimetext-asciidoc](https://github.com/AlexCzar/sublimetext-asciidoc)
  56 commits ahead, last commit: 2020-08-16
- [sambacha/sublimetext-asciidoc](https://github.com/sambacha/sublimetext-asciidoc)
  14 commits ahead, last commit: 2020-05-30

PR #1230, opened 2020-10-04
Add JSON highlight test

hertg:
This is part of #1213. I have created the test file myself and it contains all of the different data types available in JSON.
- String
- Number
- Object
- Array
- Boolean
- *null*


sharkdp:
Thank you! Can we please add a trailing newline in the source file (see #975)

hertg:
Sure, just added it!

Issue #1185, opened 2020-09-10
Open file starting at line number


DanielCardonaRojas:
First of all great work! Loving this CLI 

I checked the man page and did not see an option to open a file starting a line number, otherwise if
this feature does exist please document.

Kienyew:
I think there's a `--line-range` option in bat, which might does what you need 😄
```man
...
       -r, --line-range <N:M>...

              Only print the specified range of lines for each file. For example:

              --line-range 30:40
                     prints lines 30 to 40

              --line-range :40
                     prints lines 1 to 40

              --line-range 40:
                     prints lines 40 to the end of the file
...
```

DanielCardonaRojas:
Kind of does but does not let scroll to lines previous to start of range.

sharkdp:
You can try to use `--pager=…` to pass explicit options to less. Something like:
```
bat --pager="less -FR +10" src/assets_metadata.rs
```

You will need to add an offset of 3 lines for the bat header.

Andriamanitra:
It would be beneficial if bat let line offset arguments pass through to less (or another pager, I think all of the most common ones use similar line offset argument syntax). This would allow users to easily get syntax highlighting from bat when an external program runs a command like `$PAGER +263 /foo/bar/baz`. 

I specifically ran into this issue with Julia's [less function](https://github.com/JuliaLang/julia/blob/master/stdlib/InteractiveUtils/src/editless.jl#L245) -- as it stands you'd need to set $PAGER to quite an impressive bash spell in order to make this work.

sontek:
@sharkdp That doesn't seem to work:

```
rg --vimgrep _files | fzf --delimiter=: --preview='bat --color always -H {2} --pager "less -RF +{2}" {1}'
```

Page never jumps down to the line needed to start at

QuarticCat:
It seems that `--pager="less -FR +10"` cannot correctly handle wrappings. When there are wrapping lines, `less` jumps to a wrong position. This also happens when we want to set line offset in `fzf` because `fzf` doesn't know what is the actual "line". We still need a starting line number flag.

sharkdp:
> When there are wrapping lines, `less` jumps to a wrong position. This also happens when we want to set line offset in `fzf` because `fzf` doesn't know what is the actual "line". We still need a starting line number flag.

[...]

PR #1233, opened 2020-10-04
Add support for opening file starting at a given line number

Lidenburg:
Add support for opening file starting at a given line number.

From #1185 

Some things to note: 
- Only works with less as pager. Does nothing if piped to a file.
- Files that would normally quit because of `--quit-if-one-screen` don't if an offset is provided (because of less behavior)
- If an offset that is greater than the file is long is supplied, an ugly `Cannot seek to line number N  (press RETURN)` is shown (because of less behavior)

### DIFF
--- a/src/bin/bat/app.rs
+++ b/src/bin/bat/app.rs
@@ -238,6 +238,10 @@ impl App {
                 .map(LineRanges::from)
                 .map(|lr| HighlightedLineRanges(lr))
                 .unwrap_or_default(),
+            offset: match self.matches.value_of("offset") {
+                Some(v) => v.parse::<usize>().unwrap_or(0),
+                _ => 0,
+            }
         })
     }
 
--- a/src/bin/bat/app.rs
+++ b/src/bin/bat/app.rs
@@ -223,6 +223,12 @@ impl App {
                         .unwrap_or_default(),
                 ),
             },
+            offset: self.matches.value_of("offset").map_or(0, |v|
+                        v.parse::<usize>().unwrap()
+                        + if style_components.header() {1} else {0} // the header is 1 row
+                        + if style_components.grid() {2} else {0}   // the grid is 2 rows
+                        + 1    // add 1 for this to be an _offset_ and not just what row to show
+            ),
             style_components,
             syntax_mapping,
             pager: self.matches.value_of("pager"),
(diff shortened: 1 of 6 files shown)


sharkdp:

I have a few concerns / questions:

- This would be the first feature that is really `less`-specific. If a user actually runs another pager, the `--offset` argument would be without effect. Even if `less` is the de-facto standard, I'm not sure if we should start introducing `less`-specific features. If we decide do integrate this, the limitation should be documented.
- Does this work in combination with `--line-range`? Say I run `bat --line-range=100:200 --offset 150 …`. 
- The `bat` header is not always visible. Adding a fixed delta of 3 lines is not going to work for other `--style`s. E.g.: `bat --plain --offset 50 …`.

src/bin/bat/app.rs
```diff
@@ -238,6 +238,10 @@ impl App {
                 .map(LineRanges::from)
                 .map(|lr| HighlightedLineRanges(lr))
                 .unwrap_or_default(),
+            offset: match self.matches.value_of("offset") {
+                Some(v) => v.parse::<usize>().unwrap_or(0),
+                _ => 0,
+            }
```
sharkdp, line 244: You might be able to use `map_or` here:
```
offset: self.matches.value_of("offset").map_or(…)
```

Lidenburg: Is something like d3e2bb3346ceaa0c4a3fcb6aab19920f1667f1c1 what you had in mind?

Lidenburg:
1. I have added some extra checks to only show the `--offset` argument if no pager environment variable is set or if it is set to less. It is still possible for it to have no effect if the user were to use `--pager more` in combination with `--offset` however. Not sure what would be a good way to handle this.
2. The `--offset` parameter acts as an offset from the `--line-range` starting offset, I believe this is the most intuitive behavior?
3. I think I have fixed this with commit d3e2bb3346ceaa0c4a3fcb6aab19920f1667f1c1. Unless there are times when the headers and grids can be bigger or smaller.

sharkdp:
>     1. I have added some extra checks to only show the `--offset` argument if no pager environment variable is set or if it is set to less.

Thank you, but I think that's kind of confusing. I'd rather always show that option and add the restriction in the `--help` text (only works with `less` as a pager).

> The `--offset` parameter acts as an offset from the `--line-range` starting offset, I believe this is the most intuitive behavior?

Sounds good.

src/bin/bat/app.rs
```diff
@@ -223,6 +223,12 @@ impl App {
                         .unwrap_or_default(),
                 ),
             },
+            offset: self.matches.value_of("offset").map_or(0, |v|
+                        v.parse::<usize>().unwrap()
+                        + if style_components.header() {1} else {0} // the header is 1 row
```
sharkdp, line 228: I think it would be better if we add a method to the style components that returns their potential header-height in lines. This way, we could use something like:
```
style_components.map(|s| s.height_in_header()).sum()
```
to get the total height.

sharkdp:

PR #1236, opened 2020-10-04
Add an SML syntax test file.

dmbaturin:
For the issue #1213 

StandardML, the ancient greek of typed functional languages. ;)

Highlighting rules look correct. I wrote the file from scratch so there are no licensing concerns.


sharkdp:
Thank you very much!

PR #1237, opened 2020-10-04
Add a protobuf syntax sample.

dmbaturin:
For the issue #1213 

It's a file from my project: https://github.com/vyos/vyconf/blob/master/data/vyconf.proto

I'm its sole author so I can single-handedly give you a permission to use it for any purpose.


sharkdp:
Thank you very much!

PR #1240, opened 2020-10-04
ARM Assembly syntax test file

starsep:
Part of #1213 

I am the author of the code, original can be found here
https://github.com/starsep/asm/blob/master/zad3/grayscale.asm
The only difference is that I removed comments in Polish.


sharkdp:
Thank you very much!

PR #1241, opened 2020-10-04
#1213 Added example.xml for regression tests

loganintech:


loganintech:
#1213

sharkdp:
Thank you very much!

Could you please add a trailing newline to the file? (See #975)

PR #1243, opened 2020-10-04
Add C-Sharp (C#) syntax highlighting test files

GiorgiBeriashvili:
# Description

Refers to the following issue: #1213

The code is written from the scratch and conveys some interesting language features, of which `bat` handled correctly.


sharkdp:

PR #1244, opened 2020-10-04
#1213 Added plaintext.txt for regression test

loganintech:
#1213 

* Added plaintext file for regression tests
* Added README on how the plaintext source file was generated
* Added method for `create_highlighted_versions.py` to use the `--show-all` feature on specifically configured source files.


sharkdp:

Great idea to add a test with `--show-all` as well. If possible, I'd prefer a significantly smaller file that includes some of the important characters (see, for example, `tests/examples/nonprintable.txt` which we use for an integration test).

tests/syntax-tests/create_highlighted_versions.py
```diff
@@ -16,6 +16,16 @@
     "--italic-text=always",
 ]
 
+BINARY_OPTIONS = [
```
sharkdp, line 19: maybe "BAT_OPTIONS"?

tests/syntax-tests/create_highlighted_versions.py
```diff
@@ -16,6 +16,16 @@
...
+    "--show-all",
+    "--no-config",
+    "--style=plain",
+    "--color=always",
+    "--theme='1337'",
+    "--italic-text=always",
+]
+
+binary_folders = ["plaintext.txt"]
```
sharkdp, line 28: To make this more generally useful, what do you think about the following approach:

When reading one of the `source` folders, we check if a `tests/syntax-tests/source/…/bat_options` file exists. If it does, we read the file and expect each line to be an additional command-line option for `bat`.

So for your use case here, you would create `Plaintext/bat_options` with
```
--show-all
```
as the content.

tests/syntax-tests/create_highlighted_versions.py
loganintech: Well, BAT_OPTIONS is what's used in your other, config setup. Have no qualms changing this to whatever you recommend.

tests/syntax-tests/create_highlighted_versions.py
loganintech: Sure, I'll write this up and we can look at it together.

sharkdp:
Thank you, but the input/output files are still huge. And more than 99% of the output will look like
```
\u{6026}␊
```
with a different number, so that's not too helpful. A simple file with a few lines would be more helpful, I believe. Also, we don't want to slow down our tests / CI runs for no apparent benefit.

tests/syntax-tests/create_highlighted_versions.py
```diff
@@ -31,16 +43,24 @@ def create_highlighted_versions(output_basepath):
...
-            source_dirname = path.basename(path.dirname(source))
+            source_dirpath = path.dirname(source)
+            source_dirname = path.basename(source_dirpath)
             source_filename = path.basename(source)
 
-            if source_filename == "LICENSE.md":
+            if source_filename in SKIP_FILENAMES:
                 continue
 
+            options = copy.deepcopy(BAT_OPTIONS)
```
sharkdp, line 53: For simple lists, you can also use `options = BAT_OPTIONS.copy()` without using a separate module.

tests/syntax-tests/create_highlighted_versions.py
```diff
@@ -31,16 +43,24 @@ def create_highlighted_versions(output_basepath):
...
+            files = [f for (_, _, f) in os.walk(source_dirpath)]
+            # If a directory is empty, `files` could possibly be 0-length
+            if len(files) > 0 and "bat_options" in files[0]:
```
sharkdp, line 56: That seems wrong. Why would `bat_options` be the first file to be listed?

eth-p:
Just a nitpick, but is there any chance you can squash or rebase this? Those 45 MB objects are still going to live in the git repository even if they're deleted or replaced, and that feels like a pretty big waste of bandwidth and download time (for slow connections or mobile data).

loganintech:
@sharkdp 
> 
> Thank you, but the input/output files are still huge. And more than 99% of the output will look like
> 
> ```
> \u{6026}␊
> ```
> 
> with a different number, so that's not too helpful. A simple file with a few lines would be more helpful, I believe. Also, we don't want to slow down our tests / CI runs for no apparent benefit.

Noted, I'll significantly reduce the file size.

@eth-p 
> Just a nitpick, but is there any chance you can squash or rebase this before merging? Those 45 MB objects are still going to live in the git repository even if they're deleted or replaced, and that feels like a pretty big waste of bandwidth and download time (for slow connections or mobile data).

Absolutely, I'll take care of that in my next commit.

tests/syntax-tests/create_highlighted_versions.py
loganintech: > I think we can directly check if that file exists. No need to os.walk, right?

Looks like yes. Sorry, python isn't my main language.

loganintech:
@eth-p I believe I squashed the large artifacts out, but if you could double-check for me that'd be greatly appreciated.

eth-p:
That worked, thanks!

tests/syntax-tests/source/Plaintext/README.md
```diff
@@ -0,0 +1,10 @@
+This text file was generated with the following script. Certain invalid characters were removed manually:
```
sharkdp, line 1: I think this is outdated now, right? Can this be removed?

tests/syntax-tests/create_highlighted_versions.py
```diff
@@ -31,16 +42,23 @@ def create_highlighted_versions(output_basepath):
...
+            source_dirpath = path.dirname(source)
+            source_dirname = path.basename(source_dirpath)
             source_filename = path.basename(source)
 
-            if source_filename == "LICENSE.md":
+            if source_filename in SKIP_FILENAMES:
                 continue
 
+            options = BAT_OPTIONS.copy()
+            # If a directory is empty, `files` could possibly be 0-length
```
sharkdp, line 53: outdated comment?

tests/syntax-tests/create_highlighted_versions.py
```diff
@@ -31,16 +42,23 @@ def create_highlighted_versions(output_basepath):
             env.pop("BAT_TABS", None)
             env["COLORTERM"] = "truecolor"  # make sure to output 24bit colors
 
-            bat_output = subprocess.check_output(
-                ["bat"] + BAT_OPTIONS + [source], stderr=subprocess.PIPE, env=env,
-            )
-
-            source_dirname = path.basename(path.dirname(source))
+            source_dirpath = path.dirname(source)
+            source_dirname = path.basename(source_dirpath)
```
sharkdp, line 46: maybe add
``` python
bat_options_path = path.join(source_dirpath, "bat_options")
```
here and then check for existence + pass it in to `get_extra_options` to avoid the duplication?

sharkdp: Another idea would be to return an empty list from `get_extra_options` in case the file is not there (move the existence check to the function). This way, we can just use
``` python
["bat"] + get_extra_options(source) + [source]
```

loganintech:
Sorry for the delay on these changes @sharkdp,

I believe everything is now as was requested.

sharkdp (approved):
Thank you very much!

PR #1245, opened 2020-10-04
Print error message if bat not found on path

Mithil467:
Referring to issue #1213, some users were seeing traceback when bat was not found on path.

### DIFF
--- a/tests/syntax-tests/create_highlighted_versions.py
+++ b/tests/syntax-tests/create_highlighted_versions.py
@@ -64,6 +64,9 @@ def create_highlighted_versions(output_basepath):
                 file=sys.stderr,
             )
             sys.exit(1)
+        except FileNotFoundError:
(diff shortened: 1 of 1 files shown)


tests/syntax-tests/create_highlighted_versions.py
```diff
@@ -64,6 +64,9 @@ def create_highlighted_versions(output_basepath):
                 file=sys.stderr,
             )
             sys.exit(1)
+        except FileNotFoundError:
+            print("'bat' was not found on path.")
```
sharkdp, line 68: Maybe:
```
Error: Could not execute 'bat'. Please make sure that the executable is available on the PATH.
```

sharkdp:

tests/syntax-tests/create_highlighted_versions.py
Mithil467: Thank you. This sounds so much better :)
Updated!

PR #1247, opened 2020-10-04
Add /etc/passwd syntax test file

reima:
Provides a `/etc/passwd` syntax test file for #1213.

I used a stripped down version of `/etc/passwd` on my system. I don't believe the original file has a license.


sharkdp:

PR #1252, opened 2020-10-05
Added a JavaScript test file

DanielVip3:
I made this file which is just a random bunch of (funny) things, but I tried to add everything possible inside it.

Bat highlights it correctly, and in such an enjoyable way.

Refers to #1213.


DanielVip3:
The second test failed (Test with new syntaxes and themes) but I didn't understand what this test was about and it looks like it doesn't depend from my file but from other's files?

sharkdp:

> The second test failed (Test with new syntaxes and themes) but I didn't understand what this test was about and it looks like it doesn't depend from my file but from other's files?

Please make sure that you are running `bat` 0.16.

DanielVip3:
I think I already was
image

DanielVip3:
I noticed #1254 and I was running on Windows, so I updated my local branch and re-executed the script.
Now all those tests should pass.

henil:
Please only include the files you are changing. You have deleted many other files.

DanielVip3:
Okay, gonna close this and open another PR again with a new fork.

PR #1253, opened 2020-10-05
added test.dart

AkshatGadhwal:
added a dart file and tried to cover a vast number of different syntax.
#1213


sharkdp:
I went ahead and updated the highlighted versions for you.

PR #1256, opened 2020-10-05
Add Julia highlight test

henil:
Refer to: #1213 
It has been adapted from [here](https://github.com/mitmath/18S191/blob/master/lecture_notebooks/Basic%20Julia%20syntax.jl) and is subject to no license.

The highlight looks correct. Just a little nitpick `typeof` is a keyword in Julia and it's not being highlighted, But I guess that depends on `syntect`.


sharkdp:
> Just a little nitpick `typeof` is a keyword in Julia and it's not being highlighted

Are you sure? https://docs.julialang.org/en/v1/base/base/#Core.typeof says it's a function. 
Also, https://docs.julialang.org/en/v1/base/base/#Keywords does not list `typeof` as a keyword.

henil:
Oh sorry by "keyword" I mean inbuilt function. Wrong choice of word.

sharkdp:

PR #1257, opened 2020-10-05
Add .gitignore test

henil:
Refer to: #1213 

This is added from [rust-lang/rust's](https://github.com/rust-lang/rust). I preferred that over [bat's](https://github.com/sharkdp/bat/) because it uses more glob patterns and so is better for highlight checking.

I think It does not require to add the License. But if you think it is required I will add it.


sharkdp:
> I think It does not require to add the License. But if you think it is required I will add it.

I'd prefer if we add proper attribution, just to be sure.

henil:
done

sharkdp:
Thank you!

PR #1261, opened 2020-10-05
Add example file for YAML highlighting

audioboxer217:
I tried to add the main types that would be used by YAML
- Strings (single and multi-line)
- Numbers
- Lists (inline and multi-line)
- Boolean (both True/False and Yes/No
- Null

Relates to #1213


sharkdp (approved):
Thank you very much!

PR #1265, opened 2020-10-05
Fix spelling and use PROJECT_EXECUTABLE more

yimmt:
It should be .sublime-syntax instead of .sublime-snytax. I also changed most of "bat" to be "{{PROJECT_EXECUTABLE}}" instead since it seems more consistent. Results looks the same.

### DIFF
--- a/assets/manual/bat.1.in
+++ b/assets/manual/bat.1.in
@@ -90,7 +90,7 @@ values: *auto*, never, always.
 \fB\-f\fR, \fB\-\-force\-colorization\fR
 .IP
 Alias for '--decorations=always --color=always'. This is useful \
-if the output of bat is piped to another program, but you want \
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you very much for your contribution

sharkdp (approved):

eth-p (approved):
Looks great, thanks!

I'm a bit disappointed that I was the one who made almost all of those inconsistencues to begin with, but I guess that's what happens when you write documentation on only 3 hours of sleep 😅

sharkdp:
> I'm a bit disappointed that I was the one who made almost all of those inconsistencies to begin with, but I guess that's what happens when you write documentation on only 3 hours of sleep sweat_smile

That's *not* why I added you as a reviewer :smile:. I wasn't sure if we wanted to keep `bat` in some places (as the project name) even though the executable on that platform would be `batcat`.

PR #1268, opened 2020-10-05
Add option to specify bat target in run-benchmarks.sh

eth-p:
The benchmarking script gets kind of tedious when you have to keep changing the PATH variable. I updated it to accept `--debug`, `--release`, `--system`, and `--bat=path` as a way to select which target to perform benchmarks on.


sharkdp:
This looks great, but I hope that we would never run the benchmarks on the debug version, right? Can we remove the `--debug` part, or do you think that could be useful for something?

sharkdp (commented):
This looks great, but I hope that we would never run the benchmarks on the debug version, right? Can we remove the `--debug` part, or do you think that could be useful for something?

tests/benchmarks/run-benchmarks.sh
```diff
@@ -8,21 +7,47 @@ if ! which hyperfine > /dev/null 2>&1; then
...
+	echo "You can use '--system', '--debug', or '--release'."
+	exit 1
+fi
+
+if ! command -v "$BAT" &>/dev/null; then
+	echo "Could not find the build of bat to benchmark."
+	case "$BAT" in
+		"bat")                        echo "Make you sure to symlink 'batcat' as 'bat'." ;;
+		"../../target/debug/debug")   echo "Make you sure to 'cargo build' first." ;;
+		"../../target/debug/release") echo "Make you sure to 'cargo build --release' first." ;;
```
sharkdp, line 31: those paths seem wrong, right? There is one additional `/debug/`

sharkdp: By the way, for me, those paths would not work. I have a custom
```toml
[build]
target-dir = "/home/shark/.cargo-target"
```
in `~/.cargo/config` (which I can only recommend).

sharkdp:

eth-p:
I seem to have missed all the comments on this... whoops. I'll see if I can work on it right now.

tests/benchmarks/run-benchmarks.sh
eth-p: Fixed in 3ed0081f1fb4b707af026fcf1789909114487975.

sharkdp (approved):
Thanks!

sharkdp:
There is a small merge conflict, but apart from that this looks good.

eth-p:
Sorry for the delay! The conflict should be fixed now.

PR #1270, opened 2020-10-06
Added CSS syntax highlight file and folder

mzegar:
A .css file I wrote a while back for a personal website (MIT license)

https://github.com/mzegar/mzegar.github.io

#1213


eth-p (changes requested):
Could you please run the `tests/syntax-tests/update.sh` script as well? That will generate the comparison files used for regression testing.

I'd also prefer an `@import`, `@font-face`, and media query or two for extra coverage, but that's optional :)

eth-p (approved):
Thanks!

sharkdp (approved):

PR #1271, opened 2020-10-06
Add SCSS syntax tests

eth-p:
As a side note: the existing "Sass" tests were actually SCSS. 

@sharkdp We're going to need to ask for Sass tests again.


sharkdp:
Thanks!

> @sharkdp We're going to need to ask for Sass tests again.

Done

PR #1275, opened 2020-10-06
Add Terraform example file for Syntax Highlighting

audioboxer217:
Simplest actual code I've written for Terraform I could find.  I believe this handles the most common syntax situations as well.  However, I'm happy to add more if anyone can think of anything.

Relates to #1213


sharkdp:

Issue #1262, opened 2020-10-05
Downloads policy


data-man:
As @sharkdp [suggested](https://github.com/sharkdp/bat/pull/948#issuecomment-703782626).

eth-p:
Startup speed is still a large concern, but I'm open to the idea of aggregating other download metrics together to get the 10k minimum.

For example, that would allow the following to meet the criteria:

3000 downloads from JetBrains plugins
7000 downloads from Sublime

It would have to be considered on a case by case basis, however. I wouldn't consider npm to be a valid metric, since something like a templating language could be a dependency of a larger and more popular framework, yet go unused by most projects installing the framework.

sharkdp:
We could also think about other (additional) policies like: the syntax can be added if GitHub also supports it.

> Startup speed is still a large concern

We should probably measure the actual slowdown when adding more and more syntaxes. Shouldn't be too hard with some scripting and the available set of submodules.

rsteube:
@eth-p what's affecting the startup speed so much? Is it the single `syntaxes.bin` where they are all merged in?
And would make sense to just split that in `java.bin`, `python.bin`... and load on demand?

eth-p:
We did some digging in #951, and the majority of the startup time is spent deserializing the syntaxes.

> And would it make sense to just split that in java.bin, python.bin... and load on demand?

It's always possible, but it would likely require a non-trivial amount of refactoring and possibly even a few upstream changes to do it, though. Most of the syntax detection (e.g. by file name, by first content line) is handled by Syntect, which relies on the information provided from the .sublime-syntax files which get serialized into `syntaxes.bin`.

Splitting the `syntaxes.bin` file up would likely require us to build a mapping of that information, manually resolve which asset set it belongs to, and then deserialize that set. To complicate it further, we would probably have to figure this all out before starting any of the printing... which means we need to add a syntax resolution stage to bat's setup (and that might involve opening *all* the files passed as arguments).

Basically, the main thing stopping someone from adding it is the significant amount of time, effort, and regression testing behind it.

rsteube:
Probably not worth it then since the current solution is already quite good enough. I expect that the download limit / current set of embedded syntaxes won't be seen as a problem any more when the process of adding them gets a bit simpler.

data-man:
> We could also think about other (additional) policies like: the syntax can be added if GitHub also supports it.

:+1:

keith-hall:
[...]

PR #1285, opened 2020-10-07
Add Svelte as a syntax

kjmph:
Another syntax definition that is helpful for bat to incorporate.


eth-p:
Before we add built-in syntaxes to bat, it needs to meet a minimum requirement of installs on PackageCloud.

Here are the metrics I found:

- [PackageControl](https://packagecontrol.io/packages/Svelte) - 4k
- [Atom (1)](https://atom.io/packages/ide-svelte) -  6.6k
- [Atom (2)](https://atom.io/packages/language-svelte) - 144
- [JetBrains](https://plugins.jetbrains.com/plugin/12375-svelte) - 37k

While the only metric we go by currently is PackageControl (dicussion: #1262), the number of JetBrains installs easily triples our minimum criteria... @sharkdp, @keith-hall: thoughts?

kjmph:
Thanks, let me know if I can help. Btw, when you say "add built-in syntaxes," should I interpret that there is a way to have syntaxes added via plugins, and not as a submodule?

keith-hall:
Interesting that the syntax is most popular with JetBeans :) nice detective work, @eth-p. I say we can allow it.

> Btw, when you say "add built-in syntaxes," should I interpret that there is a way to have syntaxes added via plugins, and not as a submodule?

@kjmph [It is possible to customize bat to use syntax definitions not included by default in the installation.](https://github.com/sharkdp/bat#adding-new-syntaxes--language-definitions) Here we are just discussing whether we want Svelte included in the installation by default. Having a plugin system to provide more syntaxes is something we are actively discussing, but doesn't exist yet.

kjmph:
Ahh, I see now. I'll change my workflow and start using the cache build more effectively. Thanks for that pointer.

sharkdp:
> the number of JetBrains installs easily triples our minimum criteria...

Ok, but I don't think these numbers are comparable. Maybe JetBrains auto-downloads the syntax package every month for update checks (just as an example). Or maybe there are much more or much less JetBrains users compared to Sublime users.

On the other hand, I think there are definitely some syntaxes that would not make sense for Sublime Text users to install.. but would still be very valuable for `bat`. A stupid example would be the `/proc/cpuinfo` syntax. Another example might be a syntax for a language that comes with its own de-facto IDE. Something like `Mathematica` comes to mind.

Anyhow.. I understand we want to merge this. @kjmph Thank you very much for your contribution!

Could you please add an entry to the "unreleased" section in `CHANGELOG.md`? The format for entries is:
```
```

kjmph:
My apologies on the delay. I believe the latest commit takes care of the Changelog request.

sharkdp:
@kjmph Thank you for the update. I'm happy to merge this if we can also add a syntax regression test along the lines of #1213.

We should document this as a requirement for future syntax additions.

kjmph:
I believe this latest commit takes care of the regression syntax test.

sharkdp (approved):
Thank you!

sharkdp:
Added in [bat v0.17](https://github.com/sharkdp/bat/releases/tag/v0.17.0).

PR #1287, opened 2020-10-07
Add `-S` flag to less when `--wrap=never` (closes #1255)

### LINKED ISSUE #1255: Pass -S option to less if wrapping is explictly disabled

    bat 0.15.4

I have a file with long lines (longer than 100 characters). If I call `bat` like this, lines are wrapping:

    bat --wrap=never Vagrantfile

I expect the lines not to wrap.

On Ubuntu, I used the DEB package provided in the GitHub releases. On macOS, I used Homebrew.

The following system information is for Ubuntu, but I see the same behavior on macOS Catalina as well.

---

system
------

**$ uname -srm**
Linux 5.4.0-47-generic x86_64  

**$ lsb_release -a**
No LSB modules are available.  
Distributor ID:	Ubuntu  
Description:	Ubuntu 20.04.1 LTS  
Release:	20.04  
Codename:	focal  

bat
---

**$ bat --version**
bat 0.15.4  

**$ env**

bat_config
----------

bat_wrapper
-----------

No wrapper script for 'bat'.

bat_wrapper_function
--------------------

**$ /usr/bin/fish --login -i -c functions cat**
```
# Defined in - @ line 1
function cat --wraps=bat --description 'alias cat=bat'
bat  $argv;
end
```

tool
----

**$ less --version**
less 551 (GNU regular expressions)

gahag:
Prevent less from wrapping lines by setting the proper flag when `--wrap=never`.
If the user set a custom value for `--pager`, no additional flag is set.

The implementation is done, I just have to finish testing with the many less versions.
I'm considering building a docker image with the relevant less versions to help testing.

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -10,6 +10,15 @@ use crate::paging::PagingMode;
 #[cfg(feature = "paging")]
 use crate::wrapping::WrappingMode;
 
+
+#[cfg(feature = "paging")]
+#[derive(Debug)]
+enum SingleScreenAction {
+    Quit,
+    Nothing,
+}
+
+
 #[derive(Debug)]
 pub enum OutputType {
     #[cfg(feature = "paging")]
@@ -21,10 +30,9 @@ impl OutputType {
     #[cfg(feature = "paging")]
     pub fn from_mode(paging_mode: PagingMode, wrapping_mode: WrappingMode, pager: Option<&str>) -> Result<Self> {
         use self::PagingMode::*;
-        use self::WrappingMode::*;
         Ok(match paging_mode {
-            Always => OutputType::try_pager(false, wrapping_mode == Character, pager)?,
-            QuitIfOneScreen => OutputType::try_pager(true, wrapping_mode == Character, pager)?,
+            Always => OutputType::try_pager(SingleScreenAction::Nothing, wrapping_mode, pager)?,
+            QuitIfOneScreen => OutputType::try_pager(SingleScreenAction::Quit, wrapping_mode, pager)?,
             _ => OutputType::stdout(),
         })
     }
@@ -32,8 +40,8 @@ impl OutputType {
     /// Try to launch the pager. Fall back to stdout in case of errors.
     #[cfg(feature = "paging")]
     fn try_pager(
-        quit_if_one_screen: bool,
-        line_wrap: bool,
+        single_screen_action: SingleScreenAction,
+        wrapping_mode: WrappingMode,
         pager_from_config: Option<&str>
     ) -> Result<Self> {
         use std::env;
@@ -85,12 +93,12 @@ impl OutputType {
                     let mut p = Command::new(&pager_path);
                     if args.is_empty() || replace_arguments_to_less {
                         p.arg("--RAW-CONTROL-CHARS");
-                        if quit_if_one_screen {
+                        if let SingleScreenAction::Quit = single_screen_action {
                             p.arg("--quit-if-one-screen");
                         }
 
-                        if !line_wrap {
-                            p.arg("-S");
+                        if let WrappingMode::NoWrapping = wrapping_mode {
+                            p.arg("--chop-long-lines");
                         }
 
                         // Passing '--no-init' fixes a bug with '--quit-if-one-screen' in older
--- a/src/output.rs
+++ b/src/output.rs
@@ -12,7 +12,7 @@ use crate::wrapping::WrappingMode;
 
 
 #[cfg(feature = "paging")]
-#[derive(Debug)]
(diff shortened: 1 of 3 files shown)


gahag:
I have built a debian-based docker image with the following less versions: 563, 551, 530, 487. I have tested my implementation with all of them, and it seems to work. There's just one minor detail: with less 487, the wrapped lines don't have an indicator:

less 487: 
image

less 530+:
image

I can also supply the dockerfile if you people are interested.

src/output.rs
```diff
@@ -82,6 +89,10 @@ impl OutputType {
                             p.arg("--quit-if-one-screen");
                         }
 
+                        if !line_wrap {
+                            p.arg("-S");
```
sharkdp, line 93: Minor thing, but can we please use the long option `--chop-long-lines`.

src/output.rs
```diff
@@ -17,18 +19,23 @@ pub enum OutputType {
...
             _ => OutputType::stdout(),
         })
     }
 
     /// Try to launch the pager. Fall back to stdout in case of errors.
     #[cfg(feature = "paging")]
-    fn try_pager(quit_if_one_screen: bool, pager_from_config: Option<&str>) -> Result<Self> {
+    fn try_pager(
+        quit_if_one_screen: bool,
+        line_wrap: bool,
```
sharkdp, line 36: As we don't have named parameters in Rust, I really like to avoid `bool` parameters. One is bad enough already (*"What does `OutputType::try_pager(false, pager)` do?"*), but two are definitely not ideal (I know that one was here already). It would be great if we could introduce two `enum`s instead. For example:
``` rust
enum ActionForShortFiles {
  QuitPager,
  Nothing,
}
```
or maybe just use the `less` wording with something like
``` rust
enum ChopLongLines {
  Yes,
  No,
}
```

This is a small point overall, but I think it would be great to improve this.

src/output.rs
```diff
@@ -17,18 +19,23 @@ pub enum OutputType {
...
     #[cfg(feature = "paging")]
-    pub fn from_mode(mode: PagingMode, pager: Option<&str>) -> Result<Self> {
+    pub fn from_mode(paging_mode: PagingMode, wrapping_mode: WrappingMode, pager: Option<&str>) -> Result<Self> {
         use self::PagingMode::*;
-        Ok(match mode {
-            Always => OutputType::try_pager(false, pager)?,
-            QuitIfOneScreen => OutputType::try_pager(true, pager)?,
+        use self::WrappingMode::*;
+        Ok(match paging_mode {
+            Always => OutputType::try_pager(false, wrapping_mode == Character, pager)?,
```
sharkdp, line 26: Let's try to not repeat ourselves here, even if its a small piece of code: I would extract `wrapping_mode == Character` into a separate variable (also: see the proposed `enum` change below).

sharkdp (changes requested):
Thank you for your contribution! This looks good - I added a few comments.

sharkdp:
> I have built a debian-based docker image with the following less versions: 563, 551, 530, 487.

nice :smile: I'd be interested to see the `Dockerfile`, but I don't think we need to add it in the repo.

src/output.rs
gahag: Sure, seems better.

src/output.rs
gahag: I agree, I'll do it right away.

gahag:
There it is. I've implemented the suggested details. About the dockerfile, I've created a gist: https://gist.github.com/gahag/19bd5b88f006a2c85ca3ee634e5c8590

src/output.rs
```diff
@@ -78,10 +93,14 @@ impl OutputType {
                     let mut p = Command::new(&pager_path);
                     if args.is_empty() || replace_arguments_to_less {
                         p.arg("--RAW-CONTROL-CHARS");
-                        if quit_if_one_screen {
+                        if let SingleScreenAction::Quit = single_screen_action {
```
sharkdp, line 96: Using `single_screen_action == SingleScreenAction::Quit` would be a bit clearer here.. but maybe that's just me. It would require deriving `PartialEq`.

Similar below.

gahag: Initially I used PartialEq , but then switched to the `if let` in order to skip yet another derive which slows down compile time. Nevertheless, I've update the code with the suggestion.

sharkdp (approved):
The code changes look great, thank you very much!

Could you please add an entry to the "unreleased" section in `CHANGELOG.md`? The format for entries is:
```
```

gahag:
I had repeated the tests before commiting, everything seems to work fine :D. I've updated the changelog as requested.

sharkdp:
Fantastic - thank you!

PR #1288, opened 2020-10-07
test: x86-64 Assembly NASM syntax test

starsep:
Part of #1213 

Similarly as in #1240 I'm the author of the code and I only removed comments.
https://github.com/starsep/asm/blob/master/zad1/enlight.asm


sharkdp:
Thank you very much!

PR #1279, opened 2020-10-06
Add Batch syntax test

R1tschY:
Created a Batch syntax test from an existing Batch script from GitHub.

Relates to #1213

PR #1289, opened 2020-10-07
Add Diff syntax highlight test

SuperTux88:
I used the diff of commit 82e7786e747b1fcfac1f963ae6415a22ec9caae1 to generate the test file (thought that fits here :smiley: and also doesn't need an extra license file that way).

Relates to #1213

(also offtopic, when running `update.sh` it also changed the line endings in `highlighted/Batch/build.bat` from #1279, but I didn't include that in this PR, but tell me if I should add it in a separate commit or create a new PR for that)


sharkdp:
> I used the diff of commit [82e7786](https://github.com/sharkdp/bat/commit/82e7786e747b1fcfac1f963ae6415a22ec9caae1) to generate the test file (thought that fits here smiley and also doesn't need an extra license file that way).

:+1: 

> (also offtopic, when running `update.sh` it also changed the line endings in `highlighted/Batch/build.bat` from #1279, but I didn't include that in this PR, but tell me if I should add it in a separate commit or create a new PR for that)

Hm, strange. A separate PR would be great. Maybe make sure that the original file has Unix line endings.

PR #1295, opened 2020-10-08
Added test file for Erlang.

jackyhui96:
added bat_erlang.erl file for syntax highlighting.
issue #1213


sharkdp:
Thank you very much!

PR #1296, opened 2020-10-09
Set line endings for batch files to LF for syntax-tests

SuperTux88:
I tried to fix the outgoing change in `highlighted/Batch/build.bat` after running `update.sh`, and as you mentioned in https://github.com/sharkdp/bat/pull/1289#issuecomment-705490210, the source file has CRLF line endings too.

But it looks like this is a git thing, git always commits .bat files with CRLF (because it looks like they don't run correctly with LF line endings, and while we don't want to run these syntax-test files, git doesn't know that). And whatever I tried to make git ignore this and just commit these files with only LF, it ignored all of my tries and just converts the file back to CRLF (`warning: LF will be replaced by CRLF in tests/syntax-tests/highlighted/Batch/build.bat`).

So I started comparing the generated files on a binary level and saw that the difference is only the last newline. The commited file has CRLF as every newline, but after running the `update.sh` the last newline was only a LF. So I changed the `create_highlighted_versions.py` script to replace that last newline with a CRLF which is what git would do too. That way the script always already generates the file git already knows, so it doesn't show an outgoing change after running `update.sh`.

This maybe isn't the best solution, but as said, everything else I tried didn't work. If somebody has a better solution, I'm open for feedback. The alternative would be to just ignore this, because when trying to commit the outgoing change it just disappears, but I still think it's confusing because it's showing an outgoing change I wouldn't expect when running `update.sh`. :shrug: 

Relates to #1213

### DIFF
--- a/tests/.gitattributes
+++ b/tests/.gitattributes
@@ -4,10 +4,6 @@ snapshots/** text=auto eol=lf
 syntax-tests/source/** text=auto eol=lf
 syntax-tests/highlighted/** text=auto eol=lf
 
-# BAT/CMD files always need CRLF EOLs
-*.[Bb][Aa][Tt] text eol=crlf
-*.[Cc][Mm][Dd] text eol=crlf
-
 examples/* linguist-vendored
 snapshots/* linguist-vendored
 benchmarks/* linguist-vendored


sharkdp:
> But it looks like this is a git thing, git always commits .bat files with CRLF (because it looks like they don't run correctly with LF line endings, and while we don't want to run these syntax-test files, git doesn't know that).

Ah, sorry. Maybe I should have mentioned that. Take a look at the `tests/.gitattributes` file. It specifies `LF` line endings for the `syntax-tests` files.

And there is a (conflicting) entry for `bat` files. Maybe that messes everything up.

SuperTux88:
Ah, I tried `.gitattributes`, but didn't notice that there is already a file in `tests/.gitattributes` which then overrides my file in the repo-root again. That explains why nothing worked, thanks for the hint.

It looks like when I create another `.gitattributes` file directly in the `syntax-tests` folder for `*.bat` and `*.cmd` it overwrites the config in the parent tests folder again, so it has the wanted effect. So all syntax-tests should now have LF line endings.

sharkdp:
> Ah, I tried `.gitattributes`, but didn't notice that there is already a file in `tests/.gitattributes` which then overrides my file in the repo-root again. That explains why nothing worked, thanks for the hint.

Ahh.. sorry you had to go through this :smile:. Thank you for investigating!

I think we can actually remove those `bat`/`cmd` lines from the existing `.gitattributes`. I think they were just added because someone used that as his default `.gitattributes` file. There are no other `bat`/`cmd` files, as far as I can tell.

SuperTux88:
> I think we can actually remove those `bat`/`cmd` lines from the existing `.gitattributes`. I think they were just added because someone used that as his default `.gitattributes` file. There are no other `bat`/`cmd` files, as far as I can tell.

Ah, OK, that works for me too, I updated the PR. :+1:

sharkdp (approved):
Thank you!

PR #1297, opened 2020-10-09
Add mapping for Cake (.cake) files

patriksvensson:
This PR adds syntax mappings for [Cake (C# Make)](https://github.com/cake-build/cake) build script definitions.

I couldn't find a CONTRIBUTING.md so not sure if you accept pull requests for things like this, but it would be super convenient to have support for this out of the box.

### DIFF
--- a/src/syntax_mapping.rs
+++ b/src/syntax_mapping.rs
@@ -24,6 +24,7 @@ impl<'a> SyntaxMapping<'a> {
         let mut mapping = Self::empty();
         mapping.insert("*.h", MappingTarget::MapTo("C++")).unwrap();
         mapping.insert("*.fs", MappingTarget::MapTo("F#")).unwrap();
+        mapping.insert("*.cake", MappingTarget::MapTo("C#")).unwrap();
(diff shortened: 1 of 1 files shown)


keith-hall:
Thanks for your contribution. As Cake is a C# DSL, this could also be achieved by adding `.cake` to the [C# syntax definition's `file_extensions` directly](https://github.com/sublimehq/Packages/blob/759d6eed9b4beed87e602a23303a121c3a6c2fb3/C%23/C%23.sublime-syntax#L9), which would have the added advantage of allowing other tools which use this `.sublime-syntax` file to also map Cake files correctly by default, rather than just correcting the handling in `bat`.

(my understanding is that the explicit mapping in the `syntax_mapping.rs` file is mainly to prevent the wrong mapping when there are conflicts, and for scenarios the `. sublime-syntax` files don't support, like mappings for specific paths.)

patriksvensson:
@keith-hall Thank you for your explanation, that makes perfect sense. 
I'm closing this PR then.

PR #1301, opened 2020-10-10
Added BibTeX syntax tests

venky18:
Added BibTeX syntax tests

Relates to #1213.


sharkdp:

PR #1302, opened 2020-10-10
Add HTML syntax test sample

zjr:
Hi there! Tried to catch every basic feature of HTML including oddness like skipping closing tags. Added a script tag w/ JS inside but didn't take the JS very far aside from assuring basic highlighting. Happy to amend that if necessary.

Related to #1213


sharkdp:

PR #1303, opened 2020-10-10
Add Clojure syntax test

felipe-fg:
Added an example with Clojure code.

Relates to #1213


sharkdp (approved):

PR #1308, opened 2020-10-11
added php test file

okezieuc:
#1213  I added a sample test.php file that covers plenty of php syntax.


eth-p (changes requested):
The PHP looks good, but could you amend the commit to not update the Vue submodule? I believe it got caught by `git add -A`, and that would be a change for a different pull request.

okezieuc:
I have reverted the commit and prevented vue submodules from being updated.

eth-p:
Awesome, thanks! I'll merge it as soon as the checks pass.

eth-p (approved):

PR #1309, opened 2020-10-12
Add Perl test file

Saul-BT:
Related to the issue #1213,
Advice: It not covers all possible syntax (but the most used).

All this syntax is from Perl 5.


sharkdp:

PR #1310, opened 2020-10-12
added new syntax highlighting files

okezieuc:
#1213 I added a sample etc/hosts file and a sample requirements.txt file. I didn't know there was already a sample YAML file because it wasn't ticked in the issue so I added some extra syntax that I had the one I made to the previous example.yaml file.


sharkdp:

PR #1312, opened 2020-10-13
Add apache httpd.conf example

bofh69:
Issue: #1213
Adds an Apache2 httpd.conf file


sharkdp (approved):

Issue #1148, opened 2020-08-21
`man`: Optional parts of arguments are not hightlighting


SuperSandro2000:
**What version of `bat` are you using?**
bat 0.15.4

**Describe the bug you encountered:**
Optional parts of arguments in manpages are not highlighted.

I used the following commands:

```
export MANPAGER=sh -c 'col -bx | bat -l man -p'
git push --help
```

![image](https://user-images.githubusercontent.com/7258858/90928264-d8624400-e3f6-11ea-8e4c-d94aacfdde00.png)
![image](https://user-images.githubusercontent.com/7258858/90933897-9e963b00-e400-11ea-9384-ba4ee8de502b.png)

**Describe what you expected to happen?**
``--[no-]signed`` should be green and ``(true|false|if-asked)`` should be orange.

**How did you install `bat`?**
apt-get via GitHub release

sharkdp:
Thank you for reporting this. The [`man` syntax definition](https://github.com/sharkdp/bat/blob/master/assets/syntaxes/02_Extra/Manpage.sublime-syntax) has been manually written for `bat` and is not extensively tested. Contributions are very welcome!

SuperSandro2000:
I started gathering some examples and I don't think this is easily match able by regex:

```
--no-recurse-submodules, --recurse-submodules=check|on-demand|only|no
--[no-]thin
--repo=<repository>
--[no-]signed, --signed=(true|false|if-asked)
[--[no-]signed|--signed=(true|false|if-asked)]
[--all | --mirror | --tags] [--follow-tags] [--atomic] [-n | --dry-run] [--receive-pack=<git-receive-pack>]
```

Especially this line is hard to match:

``
[--[no-]signed|--signed=(true|false|if-asked)]
``

It translates to either arg A or B with optionals parts.

This is my attempt but I assume the regex is broken for other examples right now: ``(?:[^a-zA-Z0-9_-]|^|\s)(--?[A-Za-z0-9\[][A-Za-z0-9-\]]*)(?:(=)?("?)([A-Za-z0-9<>()|=-]+)("?))?``.

sharkdp:
Thank you for looking into this. I'm sure we will not catch all cases, but we could try to solve some of the more often used ones.

The existing regex should probably be "cleaned up" first. Sublime syntax files allow for multiline regexes with comments.

SuperSandro2000:
[...]

PR #1315, opened 2020-10-14
Improve Manpage syntax

keith-hall:
Fixes https://github.com/sharkdp/bat/issues/1148
(and also relates to #1213 )

How `git push --help` looks after these changes when using bat as the pager:

image


sharkdp:
Thank you very much! The new man page for `git push` looks great :+1: 

Unfortunately, this seems to break C syntax highlighting for man pages like `select(2)` which look like this, currently:

image

I hope it's fine that I push a `select(2)` syntax regression test that should fail on this branch. We don't have to keep this exact style (meaning: it's fine if we have to push a new `highlighted/Manpage/select-2.man` version) but it would be great if the C syntax highlighting still works.

sharkdp:
It seems like I can not push to your fork, but you can cherry-pick e1af838805e53b4f56ea944b51ce5fc532821ee3 if you want.

keith-hall:
Well spotted, thanks! It is definitely useful to have these regression tests :+1: I have cherry picked your commit and pushed some fixes. It uses some "heuristics" to determine if it is C code or not - namely presence of a block comment or `#include` directive - if you know of any man pages where this won't work, please let me know!

(about not being able to push to my fork: sorry about that. I like to keep forks separate from my own repos, so I created a GitHub org for them, but unfortunately GitHub since stopped allowing maintainers push access - it doesn't even show the option when creating the PR. I guess it makes sense because organizations have their own permissions/rules)

sharkdp:
> Well spotted, thanks! It is definitely useful to have these regression tests +1 I have cherry picked your commit and pushed some fixes. It uses some "heuristics" to determine if it is C code or not - namely presence of a block comment or `#include` directive - if you know of any man pages where this won't work, please let me know!

That sounds great :+1: 

> (about not being able to push to my fork: sorry about that. I like to keep forks separate from my own repos, so I created a GitHub org for them, but unfortunately GitHub since stopped allowing maintainers push access - it doesn't even show the option when creating the PR. I guess it makes sense because organizations have their own permissions/rules)

No worries. You can also create branches in this repo, if you prefer that.

sharkdp:
Thank you very much for the improvements and the detailed tests!

sharkdp:
There's actually a large C code block at the bottom of `select(2)` which is not highlighted (was not highlighted before either):

```
EXAMPLES
       #include <stdio.h>
       #include <stdlib.h>
       #include <sys/select.h>

       int
       main(void)
       {
           fd_set rfds;
           struct timeval tv;
           int retval;

           /* Watch stdin (fd 0) to see when it has input. */

           FD_ZERO(&rfds);
           FD_SET(0, &rfds);
…
```

Should that have been highlighted by your change? If not, that's also fine.

keith-hall:
It shouldn't have been - I didn't notice it and didn't change under which sections it applies. I have now pushed a new commit to the manpage-improvements branch addressing this.
There are a few changes in the regression test which I wasn't expecting, however, like
https://github.com/sharkdp/bat/commit/e0b7b8051b418106c527ece816059b3a30894786#diff-c601da59bcdb7dd775083ccc40783338d94e9d2ed4fefc316b698189e654addfL149

sharkdp:
There is a "visual diff" in the GA output: https://github.com/sharkdp/bat/runs/1268838933?check_suite_focus=true

Can't really tell why these changes happen though.

I guess the manpage syntax will always stay on a "best effort" basis, as there is no well-defined "syntax", to my knowledge. So I guess I'm fine with most changes (even if there are a few regressions) as long as there is an overall improvement.

PR #1316, opened 2020-10-15
Add Common Lisp syntax highlighting test files

bsinky:
For #1213 Lisp syntax highlighting tests. I'm not a Lisp person myself, but I believe this flavor is Common Lisp. I assume that's the flavor bat aims to provide highlighting support for?


sharkdp:
Yes - thank you very much for your contribution!

PR #1318, opened 2020-10-16
Added Makefile to syntax tests

MattMattV:
Added a "complex Makefile" as described here : https://www.gnu.org/software/make/manual/html_node/Complex-Makefile.html

It should cover most of Makefile syntax !


sharkdp:
@MattMattV Thank you very much for your contribution.

It would be great if we could find a Makefile that does not have a copyleft LICENSE attached to it. It's probably fine if we just add this one document, but I'm not really sure what that means for `bat` as a whole (do we "depend" on a GPL resource?).

I should have probably added this in the instructions in #1213, sorry!

MattMattV:
Oh yes, sorry, I did not thought at all about this. I will find another one with another license.

MattMattV:
I think this Makefile will be better, it's stating :

> Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.

Which are included as asked in #1213 :slightly_smiling_face:

sharkdp:
Thank you!

PR #1319, opened 2020-10-16
Adding reStructuredText example file for syntax tests

MattMattV:
Using the [reStructuredText quick reference](https://docutils.sourceforge.io/docs/user/rst/quickref.html) I crafted a sample file to be used in syntax regression tests.


sharkdp:
Thank you!

PR #1323, opened 2020-10-17
DotENV Syntax Sample File

zjr:
RE: #1213 

So, python's `glob.glob` won't pick up any dotfiles by default, but the DotENV files are, of course, dotfiles (e.g. `.env`).

I added some ugly junk in the create_highlighted_versions.py to inlcude these. I didn't spend much time on it because I don't know how you'd prefer to actually handle this situation and I'm no python expert. Happy to remove and/or adjust!


sharkdp:
Thank you!

> RE: #1213 
> 
> So, python's `glob.glob` won't pick up any dotfiles by default, but the DotENV files are, of course, dotfiles (e.g. `.env`).

No, that looks good. Will be useful for other syntaxes as well 👍🏼

PR #1325, opened 2020-10-19
Add test file for `/etc/group`

henil:
re: #1213 

I have added this file from my own system so License is not needed.

Also Due to lack of a better name for the folder I have named `EtcGroup`. If you have a better name feel free to suggest.


henil:
Test failed due to changes in the highlight of `ManPage`. Should I add the new highligted `ManPage` in this PR?

keith-hall:
Hi, thanks for your contribution. Maybe I'm missing something, but it looks like the test passed the Manpage regression tests with no differences? 
https://github.com/sharkdp/bat/pull/1325/checks?check_run_id=1274022743#step:8:159

Instead, the differences seem to come from the highlighted version using `panda` as a username, and the source using `username`?

henil:
Oh my bad, I forgot to update highlight files after changing the username. I thought it was related to man page because I have modified `ManPage` syntax file.

PR #1326, opened 2020-10-19
Manpage improvements

keith-hall:
Resolves https://github.com/sharkdp/bat/pull/1315#issuecomment-711024346

Now the Manpage syntax detects C code in more places, and also stops highlighting it as C code when a new sentence starts.


sharkdp:
Awesome - thank you very much!

PR #1328, opened 2020-10-19
Add syntax test for d 1213

frankschmitt:
I'm not 100% sure how testing this is supposed to work - should I just run cargo test (which completed without errors) ?


sharkdp:

The relevant test is not a Rust unit test, but a separate python/bash script (see `tests/syntax-tests/regression_test.sh`). The test will pass if you have just (re-)created the highlighted files, but could fail in the future if we update syntax descriptions.

Issue #1211, opened 2020-10-03
Documentation improvements


sharkdp:
*With all the recent news about Hacktoberfest I thought it would be a good idea to point out good beginner issues that would be actually helpful for `bat`. In the past years, I have actually experienced Hacktoberfest as a really great event - both as a contributor as well as a maintainer.*

One thing that I would like to focus on for the next `bat` release is documentation. This includes the README, the [man page](https://github.com/sharkdp/bat/tree/master/assets/manual) as well as the [`bat -h` and `bat --help` texts](https://github.com/sharkdp/bat/blob/master/src/bin/bat/clap_app.rs). It also includes documentation for contributors/developers like the [issue templates](https://github.com/sharkdp/bat/tree/master/.github/ISSUE_TEMPLATE).

In my mind, excellent documentation is something that separates good projects from really great ones. A lot of the issues that are opened in the `bat` repository could have been prevented if the documentation would have been clearer. And for each such issue, there are probably several more users that are already too frustrated with the tool and don't even bother opening a ticket on GitHub.

We can list potential TODOs here, but there is also a list of open issues that deal with documentation which an be found here:

[List of `documentation`-labeled issues](https://github.com/sharkdp/bat/issues?q=is%3Aopen+is%3Aissue+label%3Adocumentation).

Apart from that, a few things that come to my mind:

- [x] I think we should merge the `assets/README.md` file into the main `README.md` or maybe link to it from the main README. In general, the whole process around contributing new syntaxes (or themes) should be described in detail. Note that there is some information regarding this (policies) in the issue templates as well.
- [x] For contributors, we should clearly document that they should please update the `CHANGELOG.md` file as well. This makes the release of new `bat` versions much easier.
- [x] The short `bat -h` text should refer to the longer and more detailed `bat --help` text (and vice versa). See my `fd` project as an example for how this can be done with `clap`.
- [x] Add the step-by-step guide from #1213 to https://github.com/sharkdp/bat/blob/master/doc/assets.md as a requirement for the addition of new syntaxes.
- [x] Update the man page to be up-to-date with the `--help` text.

If you have ideas on how to improve the documentation, please post them here or open a new issue if it's a larger scale change.

trncb:
@sharkdp if no-one has taken it one, I can find a way to integrate `assets/README.md` into `README.md`

sharkdp:
> @sharkdp if no-one has taken it one, I can find a way to integrate `assets/README.md` into `README.md`

[...]

PR #1329, opened 2020-10-19
Add Coldark themes

ArmandPhilippot:
Added both versions of [Coldark](https://github.com/ArmandPhilippot/coldark).

I am not used to submodules. Hope I did it right. The linked repo contains two themes: the light version and the dark version.

There are some minor changes between the VS Code and bat versions since the tokens differ a bit. Is there a list of scopes somewhere? I searched but found nothing ... For example, how to color the keys in a json file?


ArmandPhilippot:
Sorry, I didn't do the manipulations in "Development". I didn't think it was about adding a theme ... now that I see cargo failing, I imagine it came from there.

keith-hall:
Thanks for your contribution. It looks like the test is failing because your new color schemes need to be added to the unit test:
https://github.com/sharkdp/bat/blob/2ffbe08eb171ef71ee08b4da38c87ffa75934f72/tests/assets.rs#L14

ArmandPhilippot:
It seems that this is not enough, and even worse ...

I also tried to start from 0 by following the instructions in [Development](https://github.com/sharkdp/bat#development) but there are 2 git submodules that I cannot clone (`sublimehq/Packages` and `elixir-sublime-syntax`). I do not understand why.

If I ignore this problem, and use `bash assets/create.sh`, `.rej` files are created related to the Sublime packages.

```
--- syntaxes/01_Packages/C#/C#.sublime-syntax
+++ syntaxes/01_Packages/C#/C#.sublime-syntax
@@ -1312,7 +1312,7 @@ contexts:
             2: punctuation.separator.cs
             3: punctuation.section.brackets.end.cs
             4: keyword.operator.pointer.cs
-    - match: \((?=(?:[^,)(]*|\([^\)]*\))*,)
+    - match: \((?=(?:[^,)(]|\([^\)]*\))*,)
       scope: punctuation.section.group.begin.cs
       push:
         - meta_scope: meta.group.tuple.cs
```
And `cargo test` always fails (the same problem as the original). Even adding the themes to the file `tests/assets.rs`.

I don't know if the initial problem is with the submodules that don't want to clone or if I'm just doing it wrong. Sorry for the inconvenience, the procedure did not seem complicated to me before trying.

ArmandPhilippot:
I don't understand why it doesn't work. Mainly, I don't understand why tests that worked the first time are failing now ...

This time, I had no submodules issues, no `.rej` files, everything was working fine. When I run `cargo test` locally, I have no errors. On the other hand, I did not add these two modified files `assets/syntaxes.bin` and `assets/themes.bin` as I did not see them in other pull requests.

Have I forgotten something? I did a push force. Should I have removed the fork and made a clean one again?

sharkdp:
Thank you for your contribution and sorry about that.. the current progress is kind of confusing when adding new themes:

We have some CI tests that simply run `cargo test` on the current state of the repository. These will fail when you add a new theme to the unit test, because `assets/themes.bin` is not included in this PR (which means that these tests will not see the new theme).

We also have a special CI test ("Test with new syntaxes and themes") which runs `assets/create.sh` before running `cargo test`. This one will fail if you do *not* include the new theme in the unit test.

To resolve this, you will need to commit the new `themes.bin` along with the submodule addition.

sharkdp:
I think there is a way to solve this for the future: We should only run the `all_themes_are_present` test in the "Test with new syntaxes and themes" pipeline. I guess we could somehow disable it by default and then specifically enable the test for that pipeline.

ArmandPhilippot:
Thank you for these explanations. I had looked at the base16-256 pull request, I didn't notice `themes.bin` therefore, I preferred not to add it. I figured it would generate the next bat update.

I just added the `themes.bin` file.

It might be interesting to update the `README.md` either directly under "Development" or in a sub-section to specify the files to commit for future contributors?

Something like :

```
### Add new theme

Follow the first 4 steps of [Development](https://github.com/sharkdp/bat#development). Then add your theme in `assets/themes`. Also edit the `tests/assets.rs` file to add your theme to the list.

Now you can follow the 5th step of "Development". You may need to re-run `cargo test` to make sure there are no errors.

Finally, you must commit:
* your theme
* .gitsubmodules
* unit.rt
* themes.bin
```

What do you think?

sharkdp:
> I just added the `themes.bin` file.

Sorry if I wasn't clear above. I already fixed the underlying problem on your branch, so committing `themes.bin` is *not* necessary anymore now.

> It might be interesting to update the `README.md` either directly under "Development" or in a sub-section to specify the files to commit for future contributors?

Yes, absolutely :+1:. See also: #1211. I think it should be in a separate file that we link to from the main README. Probably in `assets/README.md` as well.

ArmandPhilippot:
Ah ok! I thought it was just a correction of the tests. I did not understand that the file was no longer needed.

I just read the issue. So, I updated the Changelog. And yes, a separate file would be clearer; the README begins to be a bit long.

sharkdp:
Added in [bat v0.17](https://github.com/sharkdp/bat/releases/tag/v0.17.0).

PR #1335, opened 2020-10-23
Add Salt State (SLS) test file

ubaumann:
Adding Salt State file for syntax highlighting test #1213


sharkdp:

PR #1342, opened 2020-10-25
Add PowerShell syntax test

dp304:
Part of the work for issue #1213.

A short PowerShell script and the corresponding syntax-highlighted output
has been added. The script exhibits many of the peculiarities of the
PowerShell syntax.


dp304:
The added source file appears in the repo as binary, because it is in UTF-16 big endian format, complete with that silly BOM (U+FEFF) at the beginning.
This is how PowerShell ISE (on Win7) saved it, so I guess this is the way it's authentic, but please let me know if it should be converted.

sharkdp:
> The added source file appears in the repo as binary, because it is in UTF-16 big endian format, complete with that silly BOM (U+FEFF) at the beginning.
> This is how PowerShell ISE (on Win7) saved it, so I guess this is the way it's authentic, but please let me know if it should be converted.

Sounds good to me. `bat` can handle UTF-16 just fine, so this seems like a good additional test.

dp304:
Thanks for merging! Can you please add the `hacktoberfest-accepted` label?

Issue #1334, opened 2020-10-23
Adaptive numbers style based on length/pagination


xerus2000:
Hey, I am really enjoying `bat` for peeking at everything, but here is one little pet peeve I recently encountered:
I only want line numbers when the auto-pagination sets in. Beforehand, the files are so small that it doesn't matter, and having no numbers makes copying much easier. But in the pager, they are super useful!

I tried to create a workaround by using bat as a pager for itself, but something doesn't seem to work here:
```sh
❯ bat --style header --pager "bat --style numbers" "/tmp"
There is no style option ("less --help" for help)
numbers: No such file or directory
```

sharkdp:
> I tried to create a workaround by using bat as a pager for itself, but something doesn't seem to work here:

Ahh, we disallow usage of `bat` as a pager in order to avoid infinite recursion. It seems like the implementation is currently less than ideal and also buggy. Instead of printing a warning ("You can not use `bat` as a pager for `bat`…"), we silently replace `bat` by `less` and keep all command-line arguments (which doesn't make sense, because they are obviously not applicable to `less`). This should be fixed.

As for your use case: `bat` *always* calls the pager (unless specifically configured otherwise), even for small files. The pager then decides to quit itself if the file is small enough. At that point, we can not control whether or not to show line-numbers anymore. I think this would have to be solved in a separate wrapper script.

xerus2000:
Oh wait, `less` does that? Uh.
But what does `--pagination auto` (which seems to be the default) even do then?
```sh
       --paging <when>

              Specify  when  to use the pager. To disable the pager, use '--paging=never' or its alias, -P. To disable the pager permanently, set BAT_PAGER to an empty string. To con‐
              trol which pager is used, see the '--pager' option. Possible values: *auto*, never, always.
```

xerus2000:
> I think this would have to be solved in a separate wrapper script.

So I have to first cat the file, measures its length (`wc -l`) and then pipe it into bat with the appropriate arguments? Considering that I also want the bat header style, this doesn't sound great tbh - I'd have to access the file at least twice in what I came up with.

sharkdp:
Removing the "bug" label, as #1343 has been merged.



> So I have to first cat the file, measures its length (`wc -l`) and then pipe it into bat with the appropriate arguments? Considering that I also want the bat header style, this doesn't sound great tbh - I'd have to access the file at least twice in what I came up with.

[...]

PR #1343, opened 2020-10-25
Throws an error when `bat` is being user as pager.

adrian-rivera:
As mentioned on #1334 `bat` should not be used as a value for `pager`,
this change checks both the balue of `bat` provided as a parameter or
as an environment variable.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -20,6 +20,10 @@ error_chain! {
             description("unknown syntax"),
             display("unknown syntax: '{}'", name)
         }
+        InvalidPagerValueBat {
+            description("invalid value `bat` for pager property"),
+            display("Use of bat as a pager is disallowed in order to avoid infinite recursion problems")
+        }
     }
 }
 
--- a/src/output.rs
+++ b/src/output.rs
@@ -85,10 +85,10 @@ impl OutputType {
 
         match pagerflags.split_first() {
             Some((pager_name, args)) => {
-                let mut pager_path = PathBuf::from(pager_name);
+                let pager_path = PathBuf::from(pager_name);
 
                 if pager_path.file_stem() == Some(&OsString::from("bat")) {
-                    pager_path = PathBuf::from("less");
+                    return Err(ErrorKind::InvalidPagerValueBat.into());
                 }
 
                 let is_less = pager_path.file_stem() == Some(&OsString::from("less"));
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ## Bugfixes
 
+- Throws an error when `bat` is being used as `pager`, see #1343 (@adrian-rivera)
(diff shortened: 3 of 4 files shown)


src/output.rs
```diff
@@ -85,10 +85,10 @@ impl OutputType {
 
         match pagerflags.split_first() {
             Some((pager_name, args)) => {
-                let mut pager_path = PathBuf::from(pager_name);
+                let pager_path = PathBuf::from(pager_name);
 
                 if pager_path.file_stem() == Some(&OsString::from("bat")) {
-                    pager_path = PathBuf::from("less");
+                    panic!("You must not user bat as a pager for bat");
```
henil, line 91: typo `user` -> `use`

sharkdp: You typically want to avoid using `panic` for error handling. This function returns a `Result<…>`, so we should be able to return a proper error here (i.e. `Err(…)`) that would be reported using `bat`s error handling.

You might want to take a look at https://docs.rs/error-chain/0.12.4/error_chain/ which is used for error handling in `bast`

sharkdp: Also, we should aim to provide an explanation to the user.

adrian-rivera:
I made some changes the new message goes like this:
```
$ bat --pager="bat --style numbers" myfile.txt
[bat error]: the use of bat as a pager is disallow in order to avoid infinite recursion problems
```

src/output.rs
adrian-rivera: I made some changes the new message goes like this:
```
$ bat --pager="bat --style numbers" myfile.txt
[bat error]: the use of bat as a pager is disallow in order to avoid infinite recursion problems
```

src/error.rs
```diff
@@ -20,6 +20,10 @@ error_chain! {
             description("unknown syntax"),
             display("unknown syntax: '{}'", name)
         }
+        InvalidPagerValueBat {
+            description("invalid value `bat` for pager property"),
+            display("the use of bat as a pager is disallow in order to avoid infinite recursion problems")
```
keith-hall, line 25:
```suggestion
            display("the use of bat as a pager is disallowed in order to avoid infinite recursion problems")
```

henil: I also think `the` should be removed from the beginning. And `use` should be capitalized. i.e. `Use`

adrian-rivera: Fix the typo:
```
$ bat --pager="bat --style numbers" myfile.txt
[bat error]: Use of bat as a pager is disallowed in order to avoid infinite recursion problems
```

keith-hall (approved):

sharkdp (approved):
Thank you!

sharkdp:
Fixed in [bat v0.17](https://github.com/sharkdp/bat/releases/tag/v0.17.0).

Issue #448, opened 2018-12-14
Section about "Integration with other tools" (fzf, fd, find, git, ..)


sharkdp:
We should add a "Integration with other tools" section to the README to document some of the ways in which `bat` can be combined with other tools:

- [x] Use `bat` as a `--preview` program with `fzf`.
- [x] Execute `bat` on the search results of `fd`/`find` via `fd pattern -X bat`
- [x] Combine it with `git show` to view older versions of files: `git show v0.6.0:src/main.rs | bat -l rs`
- [x] Use it with `xclip` to copy a file to the console
- [x] Continuously monitor a growing file with `tail -f some-file.ext | bat -l ext --paging=never` with syntax highlighting.
- [x] Combine with `git diff --name-only` (see #940)
- etc.

erusev:
When I do `fd pattern -x bat`, the result looks like `cat` result: no highlighting, line numbers, etc.

Is this expected?

sharkdp:
That is expected, yes. You need to use `-X`/`--exec-batch` instead of `-x`/`--exec`. This option in `fd` is new and I haven't even released a new version that includes this feature :smile:. Sorry for the confusion.

The difference between the two is that `-x` will basically execute `bat file1`, `bat file2`, .. `bat fileN` whereas `-X` executes `bat file1 file2 ... fileN`. The former (`-x`) does that by executing many jobs in parallel in the background without being directly connected to the terminal (otherwise, you would just get an interleaved output mess). This is why `bat` does not detect an interactive terminal and will switch off colorization etc. by default. The latter `-X` executes just a single job at once and can therefore directly pipe the output to the terminal.

To use `bat` with the `-x` option, you would have to force a colorized output (`fd pattern -x bat --color=always`).

erusev:
Awesome, thanks for the thoughtful response and for your work on `bat` and `fd` :)

FinallyFinancialFreedom:
>Use bat as a --preview program with fzf.

you can config this feature with fzf. 
What i want is bat accept a linenumber and hightlight that line in the ouput . i know it's not common, but it's useful for me, how can i get this feature?

sharkdp:
Added four of the five points listed above. If somebody has more ideas, please let me know.

Boot-Error:
> * [ ]  Use `bat` as a `--preview` program with `fzf`.

This doesn't work. I guess the reason being `bat` by default pipes out plain text.

sharkdp:
[...]

PR #1350, opened 2020-10-27
Add git diff example to Integration docs

JesseVermeulen123:
Part of https://github.com/sharkdp/bat/issues/448

Thanks to https://github.com/sharkdp/bat/pull/940

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,13 @@ highlighting:
 git show v0.6.0:src/main.rs | bat -l rs
 ```
 
-Note that syntax highlighting within diffs is currently not supported. If you are looking for this, check out [`delta`](https://github.com/dandavison/delta).
+#### `git diff`
+
+You can combine `bat` with `git diff` to view lines around code changes with proper syntax
+highlighting:
+```bash
+git diff --name-only --diff-filter=d | xargs bat --diff
+```
 
 #### `xclip`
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -131,7 +131,13 @@ highlighting:
 git show v0.6.0:src/main.rs | bat -l rs
 ```
 
-Note that syntax highlighting within diffs is currently not supported. If you are looking for this, check out [`delta`](https://github.com/dandavison/delta).
```
SuperSandro2000, line 134: I really would love to keep the delta recommendation.

JesseVermeulen123: Sounds good. I considered keeping it.

README.md
```diff
@@ -131,7 +131,15 @@ highlighting:
...
+#### `git diff`
+
+You can combine `bat` with `git diff` to view lines around code changes with proper syntax
+highlighting:
+```bash
+git diff --name-only --diff-filter=d | xargs bat --diff
+```
```
sharkdp, line 140: @eth-p Probably also a good place to refer to `batdiff` from `bat-extras`, right?

sharkdp:

README.md
sharkdp: But we can already merge this as-is.

eth-p: This would be a good place to refer to it, yeah.

PR #1353, opened 2020-10-29
Add extra themes help to output

### LINKED ISSUE #1162: Mention how to add themes in terminal help
I tried bat about a year ago and I remember checking the available themes, seeing no evidence that they could be changed without making a custom build, and giving up on using it because, with my urxvt setup, every theme was either too intense, too pale, or, in the case of `ansi-dark` and `base16`, they didn't have the tweak to dark blue that I'd applied to other things to make it more readable.

(I can't just change the terminal's definition of dark blue because that throws off the contrast ratio for things which want to use it as a background.)

A brief mention of where to find more information on adding custom themes in either `--help` or `--list-themes` (or at least a mention that it *can* be done) would have prevented me from assuming that bat was unfit for my needs.

nicholas-l:
This outputs some help text regarding adding extra themes when --list-themes is run. Initially thought it would go under help for --list-themes or --help but believe this is the best place for it. However happy to move it to a more appropriate location.

Closes #1162

### DIFF
--- a/src/bin/bat/main.rs
+++ b/src/bin/bat/main.rs
@@ -205,6 +205,14 @@ pub fn list_themes(cfg: &Config) -> Result<()> {
             writeln!(stdout, "{}", theme)?;
         }
     }
+    writeln!(
+        stdout,
+        "Further themes can be installed to '{}/themes', \
+        and are added to the cache with `bat cache --build`. \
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you very much!

PR #1355, opened 2020-10-30
warning: fix; matches, closures, etc..

Horki:
@sharkdp fix warnings, please test

### DIFF
--- a/src/bin/bat/app.rs
+++ b/src/bin/bat/app.rs
@@ -82,10 +82,9 @@ impl App {
             Some("always") => PagingMode::Always,
             Some("never") => PagingMode::Never,
             Some("auto") | None => {
-                if self.matches.occurrences_of("plain") > 1 {
+                if self.matches.occurrences_of("plain") > 1
+                    || self.matches.is_present("no-paging") {
                     // If we have -pp as an option when in auto mode, the pager should be disabled.
-                    PagingMode::Never
-                } else if self.matches.is_present("no-paging") {
                     PagingMode::Never
                 } else if inputs.iter().any(Input::is_stdin) {
                     // If we are reading from stdin, only enable paging if we write to an
@@ -169,7 +168,7 @@ impl App {
                 || match self.matches.value_of("color") {
                     Some("always") => true,
                     Some("never") => false,
-                    Some("auto") | _ => {
+                    _ => {
                         env::var_os("NO_COLOR").is_none() && self.interactive_output
                     }
                 },
@@ -226,17 +225,14 @@ impl App {
             style_components,
             syntax_mapping,
             pager: self.matches.value_of("pager"),
-            use_italic_text: match self.matches.value_of("italic-text") {
-                Some("always") => true,
-                _ => false,
-            },
+            use_italic_text: matches!(self.matches.value_of("italic-text"), Some("always")),
             highlighted_lines: self
                 .matches
                 .values_of("highlight-line")
                 .map(|ws| ws.map(LineRange::from).collect())
                 .transpose()?
                 .map(LineRanges::from)
-                .map(|lr| HighlightedLineRanges(lr))
+                .map(HighlightedLineRanges)
                 .unwrap_or_default(),
         })
     }
--- a/src/input.rs
+++ b/src/input.rs
@@ -135,17 +135,12 @@ impl<'a> Input<'a> {
     }
 
     pub fn is_stdin(&self) -> bool {
-        if let InputKind::StdIn = self.kind {
-            true
-        } else {
-            false
-        }
+        matches!(self.kind, InputKind::StdIn)
     }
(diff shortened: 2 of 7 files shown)


Horki:
@sharkdp is backward compatibility with v1.40 mandatory?

sharkdp:
> @sharkdp is backward compatibility with v1.40 mandatory?

Not mandatory, but desirable. I'm okay with bumping the min. supported version if there is a good reason to do so.

src/bin/bat/app.rs
```diff
@@ -82,11 +82,10 @@ impl App {
             Some("always") => PagingMode::Always,
             Some("never") => PagingMode::Never,
             Some("auto") | None => {
-                if self.matches.occurrences_of("plain") > 1 {
+                if self.matches.occurrences_of("plain") > 1
+                    || self.matches.is_present("no-paging") {
```
sharkdp, line 86: I don't think that's really better. The comment does not apply to `--no-paging`.

src/bin/bat/app.rs
```diff
@@ -169,7 +168,7 @@ impl App {
                 || match self.matches.value_of("color") {
                     Some("always") => true,
                     Some("never") => false,
-                    Some("auto") | _ => {
+                    _ => {
```
sharkdp, line 171: Yes, this is the same. But I still like the previous version better.

src/bin/bat/app.rs
```diff
@@ -226,17 +225,14 @@ impl App {
             style_components,
             syntax_mapping,
             pager: self.matches.value_of("pager"),
-            use_italic_text: match self.matches.value_of("italic-text") {
-                Some("always") => true,
-                _ => false,
-            },
+            use_italic_text: matches!(self.matches.value_of("italic-text"), Some("always")),
```
sharkdp, line 228: Couldn't we simply use
```
self.matches.value_of("italic-text") == Some("always")
```
?

We're not really using the powerful feature of the `matches!` macro here.

src/input.rs
```diff
@@ -135,17 +135,12 @@ impl<'a> Input<'a> {
     }
 
     pub fn is_stdin(&self) -> bool {
-        if let InputKind::StdIn = self.kind {
-            true
-        } else {
-            false
-        }
+        matches!(self.kind, InputKind::StdIn)
```
sharkdp, line 138: Similar here.
```
if InputKind::StdIn = self.kind {
```
should work just fine?

src/pretty_printer.rs
```diff
@@ -54,6 +54,12 @@ impl<'a> PrettyPrinter<'a> {
             active_style_components: ActiveStyleComponents::default(),
         }
     }
+}
+
+impl<'a> PrettyPrinter<'a> {
+    pub fn new() -> Self {
+        PrettyPrinter::default()
+    }
```
sharkdp, line 62: welll.. either `new` or `default`. But not both?

sharkdp (changes requested):

src/input.rs
Horki: Please set disable linters that you dont like. Like this now is confusing and annoying

sharkdp: What do you mean? Why would we have to do that? It's not like we are enforcing zero clippy warnings anywhere.

Also, It's not that I don't like `clippy`. I think it's a great tool. But a linter does not *always* have the best suggestions. Which is why I added review comments to a few of your changes.

Horki: @sharkdp me personally prefer the idiomatic rust way.
It does the same thing in (mostly) fewer lines of code, and there is no down|upgrade in perfomance; also warning are really annoying.

I love using this tool; and looking forward seeing new updates.

sharkdp: > me personally prefer the idiomatic rust way

I, too, like idiomatic Rust. That's why I prefer
``` rust
     pub fn is_stdin(&self) -> bool { 
         self.kind == InputKind::StdIn
     } 
```
over
``` rust
     pub fn is_stdin(&self) -> bool { 
         matches!(self.kind, InputKind::StdIn)
     } 
```
Not only is it shorter, but also easier to read, especially for someone who doesn't know what `matches!(…)` does - which is rather new.

That's also why I prefer to have one single function that creates a `PrettyPrinter`. Either `default()`, or `new()`. Having both of them with the same functionality is not just confusing but also more code.

I added comments to a few of your changes. If you don't agree, please argue why you like your version better. Simply adding :-1: to all of my posts without giving any reasons is not really the greatest way to interact.

> also warning are really annoying

If there are compiler warnings, we should fix them, as they typically point to underlying problems. But what you are doing here is to simply copy & paste the `clippy` suggestion (*"help: try this: `matches!(self.kind, InputKind::StdIn)`"*), which is not always the best solution.

PR #1367, opened 2020-11-01
Import syntax regression test for Ruby Haml

byronalley:
The Haml for this PR was created custom for this regression test, but while looking at the Haml docs to try to ensure it covered typical syntax cases.

Haven't updated the changelog because this isn't a feature change, assuming that's right?

#1213


sharkdp:
> Haven't updated the changelog because this isn't a feature change, assuming that's right?

yes - thanks you!

PR #1368, opened 2020-11-01
Add sshd_config syntax highlighting test

byronalley:
Not copied from any single sshd_config file, but mostly based on docs.
#1213


sharkdp:

PR #1369, opened 2020-11-05
fix some clippy warnings

MarcoIeni:


MarcoIeni:
It looks like the `matches!` macro was introduced in rust 1.42.
We can either:
- increase minimum rust version to 1.42
- revert the changes regarding the `matches!` macro and opt out those clippy warnings in order to never appear again when you run `cargo clippy`

src/bin/bat/app.rs
```diff
@@ -169,9 +169,7 @@ impl App {
                 || match self.matches.value_of("color") {
                     Some("always") => true,
                     Some("never") => false,
-                    Some("auto") | _ => {
-                        env::var_os("NO_COLOR").is_none() && self.interactive_output
-                    }
+                    _ => env::var_os("NO_COLOR").is_none() && self.interactive_output,
```
eth-p, line 172: Honestly, I prefer the other way. By removing the `Some("auto")`, it no longer explicitly covers every possible option.

It may be redundant, but it's more clear.

eth-p (changes requested):
I like most of these changes, but I'm not a fan of using `matches!` everywhere. It has its uses, but I find it makes it harder to immediately understand what's happening. It's also not something I would consider worth bumping the minimum version for.

I've seen other projects disable that Clippy rule, and I would be in favor of doing it for bat. @sharkdp @keith-hall, your thoughts?

src/bin/bat/app.rs
```diff
@@ -226,17 +224,14 @@ impl App {
             style_components,
             syntax_mapping,
             pager: self.matches.value_of("pager"),
-            use_italic_text: match self.matches.value_of("italic-text") {
-                Some("always") => true,
-                _ => false,
-            },
+            use_italic_text: matches!(self.matches.value_of("italic-text"), Some("always")),
```
sharkdp, line 227: This could be:
``` rust
self.matches.value_of("italic-text") == Some("always")
```
instead, right?

src/input.rs
```diff
@@ -135,17 +135,12 @@ impl<'a> Input<'a> {
     }
 
     pub fn is_stdin(&self) -> bool {
-        if let InputKind::StdIn = self.kind {
-            true
-        } else {
-            false
-        }
+        matches!(self.kind, InputKind::StdIn)
```
sharkdp, line 138: This could be `self.kind == InputKind::StdIn`.

src/output.rs
```diff
@@ -150,11 +150,7 @@ impl OutputType {
 
     #[cfg(feature = "paging")]
     pub(crate) fn is_pager(&self) -> bool {
-        if let OutputType::Pager(_) = self {
-            true
-        } else {
-            false
-        }
+        matches!(self, OutputType::Pager(_))
```
sharkdp, line 153: This is an actual (non-trivial) use case of the `matches!` macro. I'm not strictly against using it, but I'd rather still support older Rust versions for now.

src/input.rs
MarcoIeni: Just tried. You can't do it because InputKind doesn't implement `PartialEq`.
And you can't derive it because there is a `Box`.
image

Maybe I am missing something..

src/bin/bat/app.rs
MarcoIeni: Yes, good catch, thank you!
changed in ba4c5d03463b74

MarcoIeni:
Ok, so both eth-p and sharkdp seems to agree they want to go for option 2:

> * revert the changes regarding the `matches!` macro and opt out those clippy warnings in order to never appear again when you run `cargo clippy`

If you confirm that I will do it :)

src/bin/bat/app.rs
MarcoIeni: There are different solutions:
- say it with a comment, something like:
  ```
                      _ => {
                        // this case includes also Some("auto")
                        env::var_os("NO_COLOR").is_none() && self.interactive_output
                    }
  ```
- throw an error:
```
                    Some("auto") => env::var_os("NO_COLOR").is_none() && self.interactive_output,
                    _ => return Err("unvalid color option".into()),
```
- disable local clippy warning and revert the change:
```
    #[allow(clippy::clippy::wildcard_in_or_patterns)]
    pub fn config(&self, inputs: &[Input]) -> Result<Config> {
```

sharkdp:
> Ok, so both eth-p and sharkdp seems to agree they want to go for option 2:
> 
> > * revert the changes regarding the `matches!` macro and opt out those clippy warnings in order to never appear again when you run `cargo clippy`
> 
> If you confirm that I will do it :)

I'd rather not add any clippy-annotations.

If we can keep backwards compatibility for now, that would be great.

src/input.rs
sharkdp: Oh, no. You are right. Thanks.

src/bin/bat/app.rs
sharkdp: If all cases are definitely covered otherwise, we can also do
```
Some("auto") => …
_ => unreachable!("…")
```

MarcoIeni:
> I'd rather not add any clippy-annotations.

Why? If we add them we can add clippy to the CI.

src/bin/bat/app.rs
MarcoIeni: Here we do something very similar:
https://github.com/sharkdp/bat/blob/2b82203041e0a4242659b95a4fa2ae6454497637/src/bin/bat/app.rs#L105
If you prefer we can do something like:

```rust
Some("auto") => …
_ => unreachable!("other values for --color are not allowed")
```

sharkdp: I would prefer that. I added a commit

sharkdp (approved):
Thank you!

sharkdp:
> > I'd rather not add any clippy-annotations.
> 
> Why? If we add them we can add clippy to the CI.

I don't want to force contributors to fix clippy suggestions in order to get a successful CI build.

PR #1373, opened 2020-11-06
Update bug_report.md

xerus2000:
Also, I was a bit confused that the info-instructions where at the top, while it told me to paste them at the bottom. I think this should be improved as well.


sharkdp:
Thank you!



> Also, I was a bit confused that the info-instructions where at the top, while it told me to paste them at the bottom. I think this should be improved as well.

Happy for any suggestions!

PR #1374, opened 2020-11-07
Simplify comments in bug_report issue template

xerus2000:


sharkdp:

I'd like this note to be at the top of the report, as we want users to really give us that kind of information. Running the script will also highlight some things (a config file, env variables) that the user might not be aware of (forgotten) which could even help solving the issue on their own.

xerus2000:
yeah, I was unsure about this. But that is something that can be fixed, there's more in this PR ;)

PR #1381, opened 2020-11-14
Syntax test files for some languages

henil:
re: #1213 

After hacktoberfest I didn't see any activity on the issue, so I have added basic test files for *bolded* languages on the issue.

Now only remaining is `Regular Expression`. What kind of structure of the file should be (as in just regular expression each on a new line)? and should it be PCRE, or javascript one?


sharkdp:
> I have added basic test files for _bolded_ languages on the issue.

Awesome - thank you! Where do these files come from? Did you write them yourself? Or do we need LICENSE files for them?



> What kind of structure of the file should be (as in just regular expression each on a new line)? and should it be PCRE, or javascript one?

The syntax file says:
```yaml
%YAML 1.2
---
name: Regular Expression
comment: Matches PCRE's regexp syntax, used in various languages and in
  Sublime Text for the Find panels
file_extensions:
  - re
```

So I think either a single line PCRE regex or maybe a multi-line regex, but not multiple regexes on multiple lines.

henil:
> Awesome - thank you! Where do these files come from? Did you write them yourself? Or do we need LICENSE files for them?

I have hacked them from either language's documentation and/or took reference from `learningxinyminutes`, so I don't think that would require any license.

> So I think either a single line PCRE regex or maybe a multi-line regex, but not multiple regexes on multiple lines.

So email regex from [this](https://stackoverflow.com/a/14075810/11894375), should work? 
I have validated it, you can check it out: https://regex101.com/r/UQ2Sap/2

henil:
Syntax's for which test are added:

- ASP
- AppleScript
- Objective-C
- Objective-C++
- Regular Expression
- Ruby on Rails
- Tcl
- Textile
- Fish

sharkdp:
> I have hacked them from either language's documentation and/or took reference from `learningxinyminutes`, so I don't think that would require any license.

:+1: 

> So email regex from [this](https://stackoverflow.com/a/14075810/11894375), should work?

In principle, yes. But note that StackOverflow has a non-trivial license file. We can't just copy that regex into our code base without attribution.

A simpler self-written regex (maybe from our own Manpage sublime syntax) would also be a good start.

henil:
I wanted to use as many regex identifiers(not sure if that's the correct word) as possible but then it's hard to come up with a useful regex. Although I think usefulness is not a concern here so I have came up with this:
> ^\[START\]\d\D\h\H\s\S[a-z]\v\V\w\W.([a-z]){3,5}\[END\]$

And this matches
> [START]1a b cd
> ef$ghijhk[END]

Would that be okay?

sharkdp:
> Would that be okay?

Yes, thanks!

> PS: Check failed due to unrelated reasons

Yes. That was my bad. It should hopefully work if you rebase on master.

sharkdp (approved):

sharkdp:
Thank you!

PR #1389, opened 2020-11-20
explain how to use bat as cat (no paging) (closes #1383)

### LINKED ISSUE #1383: `cat` doesn't page the output... how can I make `bat` work the same?
So, I feel like I must be an idiot because I can't find anything about this anywhere.

In the README, it states,

> * Be a drop-in replacement for (POSIX) `cat`

as a goal... but it clearly isn't. `cat` doesn't page output. I want `bat` to work the same way by default. Is there some way to achieve this behavior? To be clear, I want to **disable automatic paging** because I use tools which depend on `cat` output not being paged (e.g. `yay`/`paru`).

nstickney:
I added / reworded the [automatic paging](https://github.com/sharkdp/bat#automatic-paging) section of the README to include disabling the pager.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,14 +38,14 @@ characters:
 
 ### Automatic paging
 
-`bat` can pipe its own output to a pager (e.g `less`) if the output is too large for one screen.
+By default, `bat` pipes its own output to a pager (e.g `less`) if the output is too large for one screen.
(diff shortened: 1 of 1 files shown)


eth-p (approved):
Thanks for the contribution! Looks good!

keith-hall (approved):

PR #1395, opened 2020-11-23
Add --no-single-file-header flag

### LINKED ISSUE #1311: Show headers only when multiple files are passed
Sorry if this has already been asked, but I couldn't find it.

Could it be made possible to configure `bat` to show headers (`--style=header`) only when more than one file is passed? I like to keep my `--style` quite minimal, and when viewing a single file, the filename is usually just extra noise to me, but with multiple files it's practically a necessity. Thanks!

niklasmohrin:
Closes #1311

Hey there! This was a pretty straight-forward change, let me know if I missed anything or if you want to have this handled differently :)

### DIFF
--- a/src/bin/bat/app.rs
+++ b/src/bin/bat/app.rs
@@ -76,7 +76,8 @@ impl App {
     }
 
     pub fn config(&self, inputs: &[Input]) -> Result<Config> {
-        let style_components = self.style_components()?;
+        let single_file = inputs.len() == 1;
+        let style_components = self.style_components(single_file)?;
 
         let paging_mode = match self.matches.value_of("paging") {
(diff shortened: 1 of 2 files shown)


sharkdp:

@eth-p @keith-hall I'm not really convinced we should a new command-line option for this. Having a header for a single file is not really useful if someone is really just calling `bat example.rs`. But it's useful in a lot of other scenarios, e.g. when using `bat` to copy&paste code (or a screenshot of it) somewhere. So I don't want to disable the single file header by default.

I wonder if we could extend the `--style` option (again) to support this somehow?

sharkdp:
I'd rather not implement this for now.

Issue #1396, opened 2020-11-24
Running `bat` without arguments fails (output file is also an input)


sharkdp:
**What version of `bat` are you using?**
0.17.0

**Describe the bug you encountered:**
Running `bat` without any arguments results in:
```
▶ bat                         
[bat error]: The output file is also an input!
```

**What did you expect to happen instead?**
`bat` reads from (and writes to) STDIN

FYI: @niklasmohrin

niklasmohrin:
Oh, that's not good. I am on it

sharkdp:
No worries :smile:. If you would like to fix this - great! But please don't feel obliged. Also, there is no need to hurry!

That being said, I would like to get out a 0.17.1 release that fixes this. We can simply revert your PR for now and then work on a proper solution.

We should also check if this same bug appears if we read/write from/to a FIFO different from `STDIN`/`STDOUT`.

niklasmohrin:
That sounds fair, I also got a bug report on the repo itself about `bat` builds failing on the `i686-linux-android` target, so it would be nice to have this fixed until I figure out how to get the correct types. I could release a version where the Identifier just wraps the whole `FileStat` struct so that no types are named, which should then compile everywhere and then work on a fix.

niklasmohrin:
@sharkdp After some investigation, I think that it would be best to never hand out `Identifier`s to TTYs, to mitigate these clashes that really are none. Do you think this could break anything? Is there ever a conflict in reading and writing to the same tty at the same time?

niklasmohrin:
@sharkdp Everything is set, I opened #1398 to sync up with my changes on the `clircle` repository. I have some ideas for little additions over there too, but you can also merge if you want :sweat_smile:

PR #1398, opened 2020-11-24
Bring back clircle

niklasmohrin:
Closes #1193

After #1396, the cycle detection introduced originally in #1197 was reverted again. Clircle has been updated to behave as expected and may now be pulled back in.

The two commits are the same changes as in #1197, only the version number and lock file are different. I quickly tried to implement an integration test for the problem from #1396, but that didn't work. I don't know how to mock stdio in a way that clircle conceives them as a tty, so that the new behavior can be tested.

Maybe this functionality should be a compilation feature that is enabled by default, so that in case some dependency of clircle breaks or so, crates that depend on `bat` can opt out. Some users might want to opt out too.

Let me know what you think!


eth-p:
Can confirm the fix works on Mac OS.

eth-p:
>  I don't know how to mock stdio in a way that clircle conceives them as a tty, so that the new behavior can be tested.

You would likely need to create a pty to do that. I'm not entirely sure that it would be a great idea to pull more dependencies into `bat` just for testing that, but it would probably be useful for testing within `clircle` itself.

niklasmohrin:
I came up with a new integration test here and not in the `clircle` crate, because it was easier to work with your test suite than setting it all up again. I would also have to create a basic `bat` clone to hook up the pty to the stdio, because `clircle` doesn't expose something like `FromRawFd`. I also didn't need to pull in any extra dependencies, because I used the re-exported `nix`.

The test I added here fails on `bat` with `clircle` 0.1.3, but passes with v0.2.0, as expected. I assumed that three seconds are way more than a typical start up time of `bat`, but I could also increase the duration even more.

sharkdp:
Thank you very much. This looks great.

Concerning the regression test: I have had very bad experiences with tests that rely on `sleep` timers like this in the past. Sure, 3 seconds *seems* long enough. But we do rely on non-deterministic behavior, which is never good for testing.

Couldn't we simply call `bat` without args, then write `EOF` to `stdin`, and expect it to exit with code 0?

niklasmohrin:
@sharkdp I tried to think of a way to get rid of the timeouts, but I could only come up with a more elaborate test that includes checking that `bat` exits after `stdin` is fully read. Ironically, this now involves two timers. However, they _could_ both be removed with some drawbacks:

---

The first timer _seems_ to be unneeded, at least on my system, as calling `child.try_wait()` returns `Ok(None)` instantly, but I cannot be sure that this is always the case. I think that in a rare case, the process could not have been created yet and `try_wait` (which uses `libc::waitpid` internally) would error, resulting in a test failure. `man 3p wait`:

> The waitpid() function shall fail if:
>
>   ECHILD The process specified by pid does not exist or is not a child of the calling process, or the process group specified by pid  does not exist or does not have any member process that is a child of the calling process.
> ...

The step of checking that the program is waiting for input could be removed altogether, which would include removing this timeout, but I think it should stay in.

---

The second timeout is more critical. The test has to check, if `bat` has exited, but it has no way of knowing when this is the case. There are two options: a) calling `child.wait()`, which will wait until the process has exited or b) waiting for some time and then check if the process has exited at that point in time.

While approach a will never give false alarms (maybe `bat` just took a really long time to run), it could get stuck waiting, if `bat` itself would not exit for some reason. This cannot be mitigated unless an estimation for the time from "stdin eof" to "bat exited" is done.

If you think having the tests get stuck is fine, calling `wait` is definitely is the better option, but I just assumed that this should be avoided. To make increasing the expected exit time easier and have one agreed time, I added the `Duration` as a constant.

---

I know that this is currently quite the opposite of what you imagined. I know that having these constant durations could be trouble, but I think the alternatives are not strict enough. To be safe, all the durations could just be rather lenghty, so that the chance of the thread scheduler / CPU speed changing the outcome of the tests would be close to zero.

Let me know what you think and which approach you prefer!

sharkdp (commented):
Thank you for the detailed explanation and the code updates.

Let's try this with the current approach!

Concerning the wait times: we don't want them to be too long in order not to increase unit testing times unnecessarily. I would propose that we set determine a proper wait time by:

- reproducing the expected test failure for very small wait times. A good way to do this is to run `stress` in the background. Probably with a lot of `--io` or `--hdd` workers.
- increasing the time until it's not possible to reproduce the failure (while still running a reasonable amount of background `stress`).

I would guess that times below 1s would easily be enough, but 3s is also fine of course. But it's probably a good idea to measure.

Also, could you please rebase your branch on master, in order to run the new CI checks?

niklasmohrin:
Okay, I did some testing on my laptop (Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz and Micron 2200S NVMe SSD). While running `stress --cpu 4 --io 4 --hdd 4`, I ran `cargo test no_args_doesnt_break` (so not a release build, but CI doesn't pass `--release` either as far as I can tell).

Surprisingly, the test failed almost every time with `SAFE_EXIT_TIME` being 1.5 seconds. Giving it 1.7 seconds brings this down to only rare failures and with 2 seconds, tests never failed on my machine.

This might be due to CPU load though, so I tested again with `stress --io 16 --hdd 16`. This time, I got one failure in ~20 runs giving it 1.2 seconds, with 1.3 it didn't happen though.

Overall, I would go with at least 3 seconds. After all, this is a 6 core computer with an nvme ssd and I don't know what the specs for CI are.

---

The other time can be greatly reduced though, I moved it into another constant and set it down from 1 second to 100 milliseconds (which is probably still an eternity for forking off a process, but I would rather be safe here and the 100ms is irrelevant to the overall time).

niklasmohrin:
Well well, this did not work out.

First of all, 3 seconds are not enough, CI on nightly macos panicked on the `.expect` in question. I am gonna increase it to 5 seconds.

Secondly, the build failed on Rust 1.40, due to the dependency `semver-parser` v0.10.1 (the old lockfile had v0.10.0 specified). I generated a new lockfile as a good practice, but actually broke the build :D. `cargo tree` tells me, that this comes from `semver` v0.11:

```
...
├── semver v0.11.0
│   └── semver-parser v0.10.1
...
```

On [their cargo.toml](https://github.com/steveklabnik/semver/blob/master/Cargo.toml), they specified only v0.10.0, so this is odd. Do you know what to do here, except for relying on the `Cargo.lock`? If I recall correctly, this may happen to users building with `cargo install` too, right?

sharkdp:
> On [their cargo.toml](https://github.com/steveklabnik/semver/blob/master/Cargo.toml), they specified only v0.10.0, so this is odd.

Specifying `0.10.0` is equivalent to `^0.10.0`, which means that updates to `0.10.x` are allowed: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

> If I recall correctly, this may happen to users building with cargo install too, right?

Not if they use the `--locked` option, as suggested in the README:
- https://github.com/sharkdp/bat#from-source
- https://github.com/sharkdp/bat#development

Still, we need to address this at some point. I'm okay with increasing the minimum required Rust version to 1.44. But let's do that in another PR.

> First of all, 3 seconds are not enough, CI on nightly macos panicked on the `.expect` in question. I am gonna increase it to 5 seconds.

Uh :-/. I'm okay with trying this out, but if we see random test failures in the future, I'm probably going to disable or `#[ignore]` the test.

niklasmohrin:
I reverted the lockfile back to what happens if I build after adding the new dependency and all the versions from the original lockfile are back in, so that should not be a problem for this PR. I will open an issue about this later.

I had another idea about how to cut down the waiting times for the test, but I am not sure it will work, so I will write back here when I am ready.

niklasmohrin:
I turns out there is the [wait-timeout crate](https://crates.io/crates/wait-timeout) which provides an extension trait to do just what is needed here. Now, tests will not have to wait if the subprocess has already finished, but will also terminate if bat does not exit at all. I increased the timeout to 15 seconds, because this is only a cost that applies if bat would be broken again, but should probably never give false positives.

With this, I think this should be all there is to this PR, let me know if anything else needs to happen

sharkdp (approved):
Awesome, thank you very much for all of this! I think this is a great solution now. Spurious failures are still possible, but should be very unlikely. And the tests should finish as fast as they can.

PR #1399, opened 2020-11-25
only try to publish one set of archives/packages for linux x64

chris48s:
This will resolve the race condition where both the ubuntu 16.04 x64 and ubuntu 18.04 x64 builds generate a `bat-VERSION-x86_64-unknown-linux-gnu.tar.gz` and `bat_VERSION_amd64.deb` and try to publish them. We favour the older release for compatibility.


sharkdp:
Thank you very much!

PR #1401, opened 2020-11-26
adding snap installtion command

purveshpatel511:
Released this package as snap package. adding snap installtion method in readme and added it into chagelog files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@
 
 ## Packaging
 
-- `bat` is now available on snapstore with package name called `batwing`, see #1400 (@purveshpatel511)
+- `bat` is now available on snapstore with package name called `batwing`, see #1401 (@purveshpatel511)
 
 # v0.17.0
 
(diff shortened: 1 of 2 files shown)


sharkdp:
Thank you for creating a package.

We have package/executable names "bat" and "batcat" already. Would it be possible to prevent yet another new name? Is `bat` already taken?

purveshpatel511:
In snapstore, `bat` is not available for register. And thanks for pointing out the name `batcat`. I register snap with that name and package have been released on snapstore.

CHANGELOG.md
```diff
@@ -21,6 +21,10 @@
 
 - Running `bat` without arguments fails ("output file is also an input"), see #1396
 
+## Packaging
+
+- `bat` is now available on snapstore with package name called `batcat`, see #1401 (@purveshpatel511)
+
```
sharkdp, line 27: Could you please move that up to the "unreleased" section? Right now, it's in 0.17.1, which has already been released

purveshpatel511: @sharkdp sure.

purveshpatel511:
Packaging moved to "unreleased" section.

PR #1404, opened 2020-11-27
Simplify comments in bug_report issue template

xerus2000:
Adjusted as requested in #1374 :)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -8,22 +8,12 @@ assignees: ''
 ---
 
 <!--
-Hey there, thanks for creating an issue!
-
-In order to reproduce your issue, we might need to know a little bit more about the environment
-which you're running `bat` on.
-
-If you're on Linux or MacOS:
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you for your contribution. To be honest, I'm not really sure it improves the template. I'd love to hear @eth-p's opinion. Also, note that this template might change if we decide to implement #1405.

xerus2000:
One of the important things is the use of comments, because without them users have to replace text in the template, which is tedious. Everything else are basically little wording inconsistencies.

eth-p:
I could go either way. Both the current template and these changes would work for the template.

As for my opinion on the changes:

- They're more concise.
- The wording is more formal.

My only issue would be with replacing the part about Linux/MacOS with "Unix-like". I haven't tested the diagnostics script outside of MacOS or Linux, and I'm not confident it will work on other platforms like BSD.

xerus2000:
Fair, changed it back :)

sharkdp:
Thank you for the updates. I took your changes and added a few improvements (I hope) on top.

xerus2000:
I don't see how lots of empty lines are improvements, but hey - your call ;)

PR #1411, opened 2020-11-29
ShellCheck

VeryTastyTomato:
Fix the following warnings.

[SC2155](https://github.com/koalaman/shellcheck/wiki/SC2155): Declare and assign separately to avoid masking return values.
[SC2164](https://github.com/koalaman/shellcheck/wiki/SC2164): Use cd ... || exit in case cd fails.
[SC2230](https://github.com/koalaman/shellcheck/wiki/SC2230): which is non-standard. Use builtin 'command -v' instead.

### DIFF
--- a/diagnostics/info.sh
+++ b/diagnostics/info.sh
@@ -58,7 +58,8 @@ _bat_:run() {
 	_out "$BAT" --version
 	_out env | grep '^BAT_\|^PAGER='
 
-	local cache_dir="$($BAT --cache-dir)"
+	local cache_dir
+	cache_dir="$($BAT --cache-dir)"
 	if [[ -f "${cache_dir}/syntaxes.bin" ]]; then
 		_print_command "$BAT" "--list-languages"
 		echo "Found custom syntax set."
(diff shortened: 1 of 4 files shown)


VeryTastyTomato:
https://github.com/sharkdp/bat/blob/2da11097f945a156a20a38bed345f2c4765dd78a/diagnostics/info.sh#L207
The code above triggers warning [SC2001](https://github.com/koalaman/shellcheck/wiki/SC2001).
I think we can rewrite it like this:
`read -er -p "Collect ${"$1"//_/ } data? [Y/n] " response`
but I am not sure.

There are some files which triggers many errors and warnings in ShellCheck, but it may be intended. Those are under https://github.com/sharkdp/bat/tree/master/tests/syntax-tests/source/Bash.

eth-p (approved):
The fixes look good, thanks.

> The code above triggers warning SC2001.
> I think we can rewrite it like this:
> read -er -p "Collect ${"$1"//_/ } data? [Y/n] " response
> but I am not sure.

I don't think it matters too much about fixing SC2001 in that instance, to be honest. That line isn't run very frequently, and the fork()+exec() overhead isn't enough to be concerned about in a diagnostic script.

> There are some files which triggers many errors and warnings in ShellCheck, but it may be intended. Those are under https://github.com/sharkdp/bat/tree/master/tests/syntax-tests/source/Bash.

Those are the syntax tests. We don't ever run them, so it's fine to leave them as is.

VeryTastyTomato:
You’re welcome.

I agree that replacing the `sed` command in this instance won’t really matter. Would it be _ideal_ to do it anyway? I don’t mind doing it at all. The reason I didn’t is that I am not sure of my proposal.

That’s what I thought for the syntax tests. Same goes here, what would you prefer *ideally*? I don’t mind all going through the checks and fixes. Nevertheless, I would understand if you want to avoid to burden to review those. Also, they may be good test cases for `bat`.

sharkdp:
Thank you!

PR #1412, opened 2020-11-29
Add ansi theme to replace ansi-light and ansi-dark

### LINKED ISSUE #1104: Default theme compatible with light/dark terminals
Hi! 👋
First of all, thanks for making such a great tool!

One reason preventing me from using `bat` much is because there's no theme that's readable on both light and dark terminals. For instance, the default theme shows text as white, which is invisible on a white background.

I think this might be fairly common use case, as macOS' default terminal theme automatically switches between light/dark as the system theme changes.

I wonder if there's any plans of making the default theme compatible with both types of terminals? I think one major improvement would be to show text in the default text color, instead of white.

I'd be happy to help if this would seem reasonable.

mk12:
Fixes #1104

Combines `ansi-light` and `ansi-dark` into a single theme `ansi`:

* Works for terminals with light or dark backgrounds.

* To request "no color", themes set the alpha component to 01, i.e. `#00000001`. This results in the terminal's default foreground/background color being used.

* Rather than specify "no color" in many places, the new ansi theme just sets theme's overall foreground/background colors.

    * Exception: We set the gutter color explicitly, since printer.rs falls back to `Fixed(238)` for themes with no gutter color.

* On theme lookup failure for `ansi-light` or `ansi-dark` (i.e. user has no custom theme with this name), prints a deprecation message and uses `ansi`.

    * I think this is preferable to copying/symlinking files since some users might have manual workarounds to switch between `ansi-light` and `ansi-dark`, and it would be good for them to learn this is no longer needed.

As usual, PR does not include themes.bin regeneration.

Here are screenshots from macOS Terminal.app's default theme in system light/dark mode:

light

dark


sharkdp:
Excellent change, description and documentation. Plus a deprecation warning. Thank you very much!

wbolster:
thanks! this finally seems to address this issue that was mostly misunderstood here:

https://github.com/sharkdp/bat/issues/490#issuecomment-478067149
https://github.com/sharkdp/bat/issues/490#issuecomment-503649571

dandavison:
Thanks @mk12  and @sharkdp for this work. Is `to_ansi_color` something it could make sense to expose in the crate? It's one of the functions that delta vendors from bat (since delta uses bat's binary assets, delta also follows bat's color encoding conventions; an arrangement that is thoroughly beneficial to delta).

sharkdp:
@dandavison Yes - absolutely! It would be great if you could open a new ticket to discuss the details on how we want to expose this function.

PR #1439, opened 2020-12-16
only print themes hint in interactive mode

### LINKED ISSUE #1437: extra themes help text to stderr
Can the help text for `--list-themes` from #1353 be written to stderr?
I'm using the command to retrieve themes for [completion](https://github.com/rsteube/carapace-bin/blob/master/completers/bat_completer/cmd/root.go#L86) and that interfers a bit.

rsteube:
fixes #1437

### DIFF
--- a/src/bin/bat/main.rs
+++ b/src/bin/bat/main.rs
@@ -200,19 +200,19 @@ pub fn list_themes(cfg: &Config) -> Result<()> {
                 .ok();
             writeln!(stdout)?;
         }
+        writeln!(
+            stdout,
+            "Further themes can be installed to '{}', \
+            and are added to the cache with `bat cache --build`. \
+            For more information, see:\n\n  \
+            https://github.com/sharkdp/bat#adding-new-themes",
+            config_file().join("themes").to_string_lossy()
+        )?;
     } else {
         for theme in assets.themes() {
             writeln!(stdout, "{}", theme)?;
         }
     }
-    writeln!(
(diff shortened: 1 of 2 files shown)


src/bin/bat/main.rs
```diff
@@ -200,19 +200,19 @@ pub fn list_themes(cfg: &Config) -> Result<()> {
                 .ok();
             writeln!(stdout)?;
         }
+        writeln!(
+            stdout,
+            "Further themes can be installed to '{}', \
+        and are added to the cache with `bat cache --build`. \
```
keith-hall, line 206: I guess this should be indented an extra level, so everything is the same level like it was before

rsteube: oh, thought i had executed `:Format` - will have a look at it

sharkdp: I could imagine `cargo fmt` doesn't fix that because it's a multiline string

rsteube: ah ok - fixed it

sharkdp (approved):

keith-hall (approved):
Thanks for your contribution!

keith-hall:
Is it worth including this in the changelog? I was thinking yes, personally :)

sharkdp:
Right, that would be great: https://github.com/sharkdp/bat/blob/master/CONTRIBUTING.md#add-an-entry-to-the-changelog

Issue #299, opened 2018-09-10
last decoration line is not formatted properly with `--wrap never` 


Rogach:
Example:
```bash
$ echo -n deadbeef > test.txt
$ bat --wrap=never test.txt 
───────┬─────────────────────────────────
       │ File: test.txt
───────┼─────────────────────────────────
   1   │ deadbeef───────┴─────────────────────────────────
```

Compare that with `--wrap=character` (same input file):
```bash
$ bat test.txt
───────┬─────────────────────────────────
       │ File: test.txt
───────┼─────────────────────────────────
   1   │ deadbeef
───────┴─────────────────────────────────
```

Rogach:
I'm using 0.6.1 on Arch Linux:
```
$ bat --version
bat 0.6.1
```

Rogach:
It seems that this section is the culprit:
https://github.com/sharkdp/bat/blob/master/src/printer.rs#L230-L245

It simply prints the line as is, expecting that every line will be ended in a newline. But the last line can end without a newline, resulting in broken formatting.

It is possible to look at the last character and print the missing newline if there isn't one:
```diff
impl<'a> Printer for InteractivePrinter<'a> {
                     )).collect::<Vec<_>>()
                     .join("")
             )?;
+
+            if line.bytes().next_back() != Some(0x0a) {
+                write!(handle, "\n")?;
+            }
         } else {
             for &(style, region) in regions.iter() {
                 let mut ansi_iterator = AnsiCodeIterator::new(region);
```

If you are okay with such solution, I would be happy to create a pull request!

Rogach:
Here you go: #300

sharkdp:
Released in [v0.7.0](https://github.com/sharkdp/bat/releases/tag/v0.7.0).

PR #1440, opened 2020-12-16
Don't add artificial newline to last line if --style=plain

### LINKED ISSUE #1438: Newline can be added even if --style=plain
In #975 @sharkdp said this:

> The fact that bat -p prints a trailing newline (even if there is no newline at the end of the file) should be considered a bug, I believe (but we should check older tickets, if there was a reason for this behavior).

And I completely agree. The reason the artificial newline was added in the first place was to fix #299, but there should be no risk of regressions since --style=plain excludes any kind of decorations that bug was about.

Note that a pager such as `less` will add an artificial newline too, so one either has to disable paging or use a "pager" that does not add a newline to get the benefit of fixing this bug.

**Step-by-step:**
1.  echo -n nonewline > nonewline.txt
2. bat --style=plain --paging=never nonewline.txt
3. bat --style=plain --pager=cat nonewline.txt

**Expected result:**
After 2. and 3. the line shall not contain a newline. I personally use `zsh` to check this since by default `zsh` will print e.g. `%` as the last char if there is a missing newline.

**Actual result:**
A newline is printed with `bat 0.17.1`.

(I will soon open a PR with a proposed fix)

Enselic:
This fixes #1438.

Note however, that using a pager such as less will add a newline itself.
So to actually not print a newline for such files, you need to either
disable paging:

  bat --style=plain --paging=never no-newline-at-end-of-file.txt

or use a "pager" that does not add a newline:

  bat --style=plain --pager=cat no-newline-at-end-of-file.txt


Enselic:
I'd say the fix definitely should include some integration tests too, but before I put too much time into that, I just wanted to check if this fix looks reasonable to you?

sharkdp:



> I'd say the fix definitely should include some integration tests too, but before I put too much time into that, I just wanted to check if this fix looks reasonable to you?

It does! :-)

The "Test with new syntaxes and themes" currently fails in the syntax regression tests, as there are some files without newlines at the end. This should be fixed anyway (trailing newlines should be added).

Enselic:
Just a quick status update. I've pushed a set of commits that (should) pass all existing tests. Feel free to code-review if you wish.

However, to add new tests for this usecase, I need to come up with a way to test tty output.

From what I can tell, only non-tty output is tested in `tests/integration_tests.rs`. And after a quick glance at the `assert_cmd` crate, tty output/simulation does not seem to be not supported.

So it seems to me as if the right course of action is to add support for tty-output-testing to `assert_cmd`, bump our dependency, and then add tests to this fix. So it will take a while to get in place.

If you have idea for other approaches, I am all ears.

sharkdp:
> If you have idea for other approaches, I am all ears.

I don't think we need to mock a complete terminal to test this. We can force `bat` to run in "interactive mode" (as if it was connected to a terminal) by using either `--color=always` or `--decorations=always`.

I added a test that fails on master but succeeds on your branch.

The changes here look great - thank you very much!

Enselic:
Thanks a lot for the help with adding a test for this fix!

I found no dedicated regression test for the fix for #299, so I added such a test in PR #1445. The new test fails if the fix for #299 is reverted from master, but of course pass on unmodified master.

It is true that we can force bat options to behave as if a tty was used, but I still think tests for tty defaults has some merit. There is currently no test that fails if we mess up the code paths that depends on e.g. `atty::is(Stream::Stdout)` being `true`, though such regressions arguable would affect a lot of users, and the involved code is rather complex. Making such tests possible to write in a maintainable manner is a big topic of its own, so let's defer that discussion to some other occasion though :)

sharkdp:
> I found no dedicated regression test for the fix for #299, so I added such a test in PR #1445. The new test fails if the fix for #299 is reverted from master, but of course pass on unmodified master.

:+1: 



> It is true that we can force bat options to behave as if a tty was used, but I still think tests for tty defaults has some merit. There is currently no test that fails if we mess up the code paths that depends on e.g. `atty::is(Stream::Stdout)` being `true`, though such regressions arguable would affect a lot of users, and the involved code is rather complex.

Interesting thoughts. I agree in general. It would be great to see an explicit example where such a test would be superior over forcing colorized/decorated output.. even if constructed.

> Making such tests possible to write in a maintainable manner is a big topic of its own, so let's defer that discussion to some other occasion though :)

Yes, I think that should be discussed in a new ticket :+1:

PR #1441, opened 2020-12-17
Add `-s`/`--squeeze` and `--squeeze-limit`.

eth-p:
After some reconsideration towards #762, I decided to try adding support for `bat -s`.

While I personally think that having the option to squeeze empty lines goes against the Unix philosophy, `cat` itself has the feature on every distro. For the sake of maintaining drop-in replacement compatibility, I added support to bat.

Another benefit of including it rather than relying on sed would be to show the correct line numbers in the sidebar.

---

This pull request adds:

- `bat --squeeze`/`bat -s`
- `bat --squeeze-limit=<number>` -- to set the maximum number of consecutive empty lines when using `-s`
- `PrettyPrinter::squeeze_empty_lines`

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -84,7 +84,7 @@ pub struct Config<'a> {
     pub highlighted_lines: HighlightedLineRanges,
 
     /// The maximum number of consecutive empty lines to display
-    pub squeeze_lines: usize,
+    pub squeeze_lines: Option<usize>,
 }
 
 #[test]
--- a/src/config.rs
+++ b/src/config.rs
@@ -82,6 +82,9 @@ pub struct Config<'a> {
 
     /// Ranges of lines which should be highlighted with a special background color
     pub highlighted_lines: HighlightedLineRanges,
+
+    /// The maximum number of consecutive empty lines to display
+    pub squeeze_lines: usize,
 }
 
 #[test]
--- a/src/bin/bat/app.rs
+++ b/src/bin/bat/app.rs
@@ -234,6 +234,11 @@ impl App {
                 .map(LineRanges::from)
                 .map(HighlightedLineRanges)
                 .unwrap_or_default(),
+            squeeze_lines: if self.matches.is_present("squeeze") {
+                1
+            } else {
+                0
+            },
         })
     }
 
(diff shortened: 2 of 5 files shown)


sharkdp:
Thank you for the implementation!

I haven't looked at `-s` in the past (for drop-in compatibility) because it's not part of the POSIX standard, see https://pubs.opengroup.org/onlinepubs/9699919799/utilities/cat.html (I just now saw the "RATIONALE" section: *"The -s option was omitted because it corresponds to different functions in BSD and System V-based systems."*).

That said, I am happy to consider adding this to `bat`. We should definitely add some tests though before merging this.

src/config.rs
```diff
@@ -82,6 +82,9 @@ pub struct Config<'a> {
 
     /// Ranges of lines which should be highlighted with a special background color
     pub highlighted_lines: HighlightedLineRanges,
+
+    /// The maximum number of consecutive empty lines to display
```
sharkdp, line 86: Can we expand this a bit to explain what happens if it is set to `None`? Also, maybe an example wouldn't hurt? My understanding would be that `Some(3)` would mean that
```
something
like



this
```
would be printed unchanged, while
```
something
like



this
```
would be squeezed (to one empty line or to three empty lines?).

eth-p:
> I haven't looked at `-s` in the past (for drop-in compatibility) because it's not part of the POSIX standard, see https://pubs.opengroup.org/onlinepubs/9699919799/utilities/cat.html (I just now saw the "RATIONALE" section: *"The -s option was omitted because it corresponds to different functions in BSD and System V-based systems."*).

It seems that `-s` on System V is just for silencing messages. I think it would be safe to implement the BSD/GNU behavior for bat.

Without `-s`, it would be equally incompatible with System V `cat` as it would be with the option. We won't really be losing anything from adding it, I think.

> That said, I am happy to consider adding this to `bat`. We should definitely add some tests though before merging this.

Will do.

src/config.rs
eth-p: To three empty lines.

I had a hard time trying to describe how it worked in only a few comments. The example would definitely be a good idea, thanks!

PR #1444, opened 2020-12-19
Fix #1443 macOS: regression_test.sh: mktemp: illegal option

### LINKED ISSUE #1443: macOS: regression_test.sh: mktemp: illegal option --suffix=.bat-syntax-regression-test
When working on tests pertaining to #1438 I bumped into a minor annoyance. (That seems easy to fix if we are a bit pragmatic; more info in soon-to-be-created MR.)

**Step-by-step:**
1. Use macOS Catalina 10.15.7 ([rumor](https://unix.stackexchange.com/questions/30091/fix-or-alternative-for-mktemp-in-os-x) has it that any version back to and including macOS 10.11 should also work)
2. Setup a `bat` development environment from `git`
3. Run `cargo install --locked --path . --force && bash assets/create.sh && cargo install --locked --path . --force && cargo test --release && cargo test --release -- --ignored && tests/syntax-tests/regression_test.sh` to simulate `Test with new syntaxes and themes` from `CICD.yml`

**Actual:**
```
% [...] tests/syntax-tests/regression_test.sh 
mktemp: illegal option -- -
usage: mktemp [-d] [-q] [-t prefix] [-u] template ...
       mktemp [-d] [-q] [-u] -t prefix 
```

**Expected:**
No issues

Enselic:
The macOS version of mktemp does not recognize the --suffix option.
Using pure -d should work since, it seems [1], macOS 10.11 however.

So to make the script work on macOS, stop using the --suffix option.

The downside is of course that the temporary dir will have an anonymous
name, but I see no risk of confusion given how short-lived the usage of
the dir is, and given the context it is used.

I will not be offended if you disagree, but it is nice to keep things simple.

[1] https://unix.stackexchange.com/questions/30091/fix-or-alternative-for-mktemp-in-os-x

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 
 ## Bugfixes
 
+- Make ./tests/syntax-tests/regression_test.sh work on recent versions of macOS, see #1443 (@Enselic)
+
 ## Other
 
 ## Syntaxes
--- a/tests/syntax-tests/regression_test.sh
+++ b/tests/syntax-tests/regression_test.sh
@@ -4,7 +4,7 @@ set -eou pipefail
 
 script_directory="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 
-output_directory=$(mktemp -d --suffix=.bat-syntax-regression-test)
+output_directory=$(mktemp -d)
 
 "$script_directory"/create_highlighted_versions.py --output="$output_directory"
 


eth-p:
After your issue about this, I actually ended up creating a fix for it with:

```bash
make_temp_dir() {
	local mktemp_info="$(mktemp --help 2>&1 || true)"
	if [[ "$mktemp_info" =~ .*GNU.* ]]; then
		# GNU mktemp
		mktemp -d --suffix=.bat-syntax-regression-test
	else
		# BSD mktemp
		mktemp -d -t bat-syntax-regression-test
	fi
}
```

But, your fix is certainly simpler. This PR has my support!

sharkdp:

PR #1445, opened 2020-12-21
Add test 'grid_for_file_without_newline' (for issue #299 fix)

Enselic:
This is a regression test for the fix for issue #299. If that fix is
reverted, currently only one test ('header_padding') fails. But that
test is for a different use case, so add a dedicated regression test for
the particular use case issue #299 is about.


sharkdp (approved):
Awesome - thank you very much!

PR #1447, opened 2020-12-24
Added NSI syntax submodule.

virtualritz:
NSI is a standard for text-based 3D scene description in VFX. It is similar to (and based on the same parser) as RenderMan RIB format.
See [nsi.readthedocs.io](https://nsi.readthedocs.io/en/latest/).


sharkdp:

Have you seen https://github.com/sharkdp/bat/blob/master/doc/assets.md#criteria-for-inclusion-of-new-syntaxes ?

PR #1451, opened 2020-12-27
Cargo.toml: remove "readme" field

MarcoIeni:
It is no more required. See [here](https://doc.rust-lang.org/cargo/reference/manifest.html#the-readme-field).


sharkdp:
Thank you

PR #1452, opened 2020-12-27
Introduce bat_warning! helper macro

Enselic:
This macro is intended to be package-internal and is not to be
considered part of the public lib API.

Use it in three places to reduce code duplication. However, main reason
for this refactoring is to allow us to fix #1063 without duplicating the
code yet another time.

The macro can also be used for the "Binary content from {} will not be
printed to the terminal" message if that message starts to use eprintln!
instead (if ever).

To trigger/verify the changed code, the following commands can be used:

    cargo run -- --theme=ansi-light tests/examples/single-line.txt
    cargo run -- --theme=does-not-exist tests/examples/single-line.txt
    cargo run -- --style=grid,rule tests/examples/single-line.txt

### DIFF
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -11,6 +11,7 @@ use syntect::parsing::{SyntaxReference, SyntaxSet, SyntaxSetBuilder};
 use path_abs::PathAbs;
 
 use crate::assets_metadata::AssetsMetadata;
+use crate::bat_warning;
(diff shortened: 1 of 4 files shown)


j0hnmeow (approved):

sharkdp:
Awesome, thank you!

PR #1453, opened 2020-12-28
bat cache --build: Print syntect error message for themes if any

### LINKED ISSUE #614: ThemeSet always empty
**Environment**

bat 0.11.0 on macOS using the default config path

**Description**

I tried to add a custom theme using the following directory structure:
```console
$ tree /Users/me/.config/bat/themes
/Users/me/.config/bat/themes
└── solarized-dark
    └── solarized-dark.tmTheme
```

but `bat cache -b` does not find my theme:

```console
$ bat cache --blank -b
No themes were found in '/Users/me/.config/bat/themes', using the default set
No syntaxes were found in '/Users/me/.config/bat/syntaxes', using the default set.
Writing theme set to /Users/me/.cache/bat/themes.bin ... okay
Writing syntax set to /Users/me/.cache/bat/syntaxes.bin ... okay
```

I quickly checked the following code, and from my understanding of the [`ThemeSet::add_from_folder()`](https://docs.rs/syntect/3.2.0/src/syntect/highlighting/theme_set.rs.html#53-64) method (not familiar with Rust, bear with me) the directory tree shown above should be correct https://github.com/sharkdp/bat/blob/c9d7e3652bb0be4becdecfd8d2e7d4e795071c83/src/assets.rs#L42-L48

Enselic:
This will fix #614 by making it clear what is wrong by showing the
following error message:

    Failed to load one or more themes from
    '/Users/me/.config/bat/themes' (reason: 'Invalid syntax theme
    settings')

We also need to add a check if theme_dir.exists(), otherwise an absent
dir will seem like an error:

    Failed to load one or more themes from
    '/Users/me/.config/bat/themes' (reason: 'IO error for
    operation on /Users/me/.config/bat/themes: No such file or
    directory (os error 2)')

(This is the same check we already have for syntax_dir.)

### DIFF
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -38,9 +38,16 @@ impl HighlightingAssets {
         };
 
         let theme_dir = source_dir.join("themes");
-
-        let res = theme_set.add_from_folder(&theme_dir);
-        if res.is_err() {
+        if theme_dir.exists() {
+            let res = theme_set.add_from_folder(&theme_dir);
+            if let Err(err) = res {
+                println!(
+                    "Failed to load one or more themes from '{}' (reason: '{}')",
+                    theme_dir.to_string_lossy(),
+                    err,
+                );
+            }
+        } else {
             println!(
                 "No themes were found in '{}', using the default set",
                 theme_dir.to_string_lossy()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Only print themes hint in interactive mode (`bat --list-themes`), see #1439 (@rsteube)
 - Make ./tests/syntax-tests/regression_test.sh work on recent versions of macOS, see #1443 (@Enselic)
 - VimL syntax highlighting fix, see #1450 (@esensar)
+- Print an 'Invalid syntax theme settings' error message if a custom theme is broken, see #614 (@Enselic)
 
 ## Other
 


Enselic:
Oops, forgot a `CHANGELOG.md` entry, hang on

keith-hall (approved):

src/assets.rs
```diff
@@ -38,9 +38,16 @@ impl HighlightingAssets {
         };
 
         let theme_dir = source_dir.join("themes");
-
-        let res = theme_set.add_from_folder(&theme_dir);
-        if res.is_err() {
+        if theme_dir.exists() {
```
sharkdp, line 41: What if the theme dir exists but no themes are inside?

Enselic: Same behavior as before, namely that no additional themes are added to the `ThemeSet`, because the loop in `add_from_folder` is run 0 times:

```
    /// Load all the themes in the folder into this `ThemeSet`
    pub fn add_from_folder<P: AsRef<Path>>(&mut self, folder: P) -> Result<(), LoadingError> {
        let paths = Self::discover_theme_paths(folder)?;
        for p in &paths {
            let theme = Self::get_theme(p)?;
            let basename =
                p.file_stem().and_then(|x| x.to_str()).ok_or(LoadingError::BadPath)?;
            self.themes.insert(basename.to_owned(), theme);
        }

        Ok(())
    }
```

sharkdp:
Thank you!

PR #1454, opened 2020-12-28
bat --list-languages: remove unnecessary format!() call

Enselic:
To trigger/verify the changed code, run

    bat --list-languages   # or -L

This is the last clippy warning in the code that you get if you run

    cargo clippy --all-targets --all-features -- --allow clippy::style

so by fixing it it becomes easier to spot when a new warning is
introduced (that does not belong to the clippy category clippy::style).
And by making it easy to spot new warnings, we increase chance of such
regressions not ending up in the code base.


sharkdp (approved):
Thank you!

PR #1459, opened 2020-12-31
Add --diagnostic option to bat

### LINKED ISSUE #1405: Add --diagnostic option to bat itself
The bug report issue templates asks to run https://github.com/sharkdp/bat/blob/master/diagnostics/info.sh or add the Windows version, as well as add the bat version. It would be much easier to have a `--diagnostic` command in bat directly, like e.g. `jrnl` has it:
```
❯ jrnl --diagnostic
jrnl: v2.4.5
Python: 3.8.6 (default, Sep 30 2020, 04:00:38) 
[GCC 10.2.0]
OS: Linux 5.9.10-arch1-1
```

sharkdp:
This is a draft with a new `--diagnostic` option for `bat` that could eventually replace the `diagnostics/info.sh` script for bug reports (closes #1405). The main part of the work in is in [a new crate called `bugreport`](https://github.com/sharkdp/bugreport), which is in a early stage as well.

I'd be glad for any kind of feedback!

### example output

*(see CI output for more examples)*

```
▶ alias bat="bat --style='numbers,grid,header'"     
▶ export BAT_STYLE=numbers,grid
▶ bat --pager="less -RF" src/macros.rs --diagnostic
```

#### Software version

bat 0.17.1

#### Operating system

Linux 5.9.14-arch1-1

#### Command-line

```bash
bat --style=numbers,grid,header '--pager=less -RF' src/macros.rs --diagnostic 
```

#### Environment variables

```bash
SHELL=/usr/bin/zsh
PAGER=less
BAT_PAGER=<not set>
BAT_CONFIG_PATH=<not set>
BAT_STYLE=numbers,grid
BAT_THEME=<not set>
BAT_TABS=<not set>
```

#### Config file

```
# Use italic text on the terminal (not supported on all terminals)
--italic-text=always

# Syntax mappings
--map-syntax .ignore:"Git Ignore"
--map-syntax .fdignore:"Git Ignore"
--map-syntax '*.ino':"C++"
```

#### Compile time information

- Profile: release
- Family: unix
- OS: linux
- Architecture: x86_64
- Pointer width: 64
- Endian: little
- CPU features: fxsr,sse,sse2

#### Less version

```
> less --version 
less 563 (PCRE regular expressions)
Copyright (C) 1984-2020  Mark Nudelman

Home page: http://www.greenwoodsoftware.com/less
```

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -226,7 +226,13 @@ jobs:
       with:
         use-cross: ${{ steps.vars.outputs.CARGO_USE_CROSS }}
         command: run
-        args: --target=${{ matrix.job.target }} ${{ matrix.job.cargo-options }} -- --paging=never --color=always --theme=ansi-dark Cargo.toml src/config.rs
+        args: --target=${{ matrix.job.target }} ${{ matrix.job.cargo-options }} -- --paging=never --color=always --theme=ansi Cargo.toml src/config.rs
+    - name: bat diagnostics output
+      uses: actions-rs/cargo@v1
+      with:
+        use-cross: ${{ steps.vars.outputs.CARGO_USE_CROSS }}
+        command: run
+        args: --target=${{ matrix.job.target }} ${{ matrix.job.cargo-options }} -- --paging=never --color=always --theme=ansi Cargo.toml src/config.rs --diagnostic
     - name: Check features regex-onig
       uses: actions-rs/cargo@v1
       with:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -197,8 +197,6 @@ dependencies = [
 [[package]]
 name = "bugreport"
 version = "0.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "824268bc0fcb3b94da597b81fa9dd52175cfde5522f4892a0cff3932802ff805"
 dependencies = [
  "snailquote",
  "sys-info",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -196,7 +196,9 @@ dependencies = [
 
 [[package]]
 name = "bugreport"
-version = "0.1.0"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "116762017f173ea5d8103e75533f6bdb62cb6f2257c9c95672beb085a70daed8"
 dependencies = [
  "snailquote",
  "sys-info",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -98,6 +98,7 @@ dependencies = [
  "ansi_term 0.12.1",
  "assert_cmd",
  "atty",
+ "bugreport",
  "clap",
  "clircle",
  "console",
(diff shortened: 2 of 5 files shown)


eth-p (approved):
Looks great!

<details>
<summary>My output on MacOS:</summary>
<div>

#### Software version

bat 0.17.1

#### Operating system

Darwin 18.7.0

#### Command-line

```bash
target/debug/bat --diagnostic
```

#### Environment variables

```bash
SHELL=/usr/local/bin/fish
PAGER=/usr/local/bin/less
BAT_CACHE_PATH=<not set>
BAT_CONFIG_PATH=<not set>
BAT_OPTS=<not set>
BAT_PAGER=<not set>
BAT_STYLE=<not set>
BAT_TABS=<not set>
BAT_THEME=<not set>
XDG_CONFIG_HOME=<not set>
XDG_CACHE_HOME=<not set>
COLORTERM=truecolor
NO_COLOR=<not set>
```

#### Config file

```
--tabs 4

# Syntaxes
#--map-syntax From:to
```

#### Compile time information

- Profile: debug
- Family: unix
- OS: macos
- Architecture: x86_64
- Pointer width: 64
- Endian: little
- CPU features: cmpxchg16b,fxsr,sse,sse2,sse3,ssse3
- Host: x86_64-apple-darwin

#### Less version

```
> less --version
less 563 (PCRE regular expressions)
Copyright (C) 1984-2020  Mark Nudelman

Home page: http://www.greenwoodsoftware.com/less
```

</div>
</details>

eth-p:
I just had an additional thought:

It might be a good idea to add `--diagnostics` as an alias. In this context, both "diagnostic" and "diagnostics" should mean the same thing. And while "diagnostic" is apparently more common, it would probably be more user friendly to just accept both forms.

sharkdp:
> It might be a good idea to add `--diagnostics` as an alias. In this context, both "diagnostic" and "diagnostics" should mean the same thing. And while "diagnostic" is apparently more common, it would probably be more user friendly to just accept both forms.

I tried to find other programs that implement similar options. So far, I only found
- `git bugreport`
- `ffmpeg -report`
- `jrnl --diagnostic` (example from #1405)
- `vim`s command to generate a bug report: `:so $VIMRUNTIME/bugreport.vim`.

Looking at these, I would suggest we call the new option `--bugreport`? Using a subcommand like `git` is not an option because we want the original command line to be retained.

Some other programs display information similar to this when running `-V`/`--version`. Like `perl -V`.

There is also `ripgrep`s `--debug` and `find`s `-D debugopts` options, but they have a slightly different purpose: they enable debug output (as compared to information about the system) while running the actual command.

sharkdp:
> In the future, would you want me to add support for extracting `alias cat=bat` to the `bugreport` crate?

Possibly. There is also a way to add custom "collectors": https://github.com/sharkdp/bugreport/blob/master/examples/custom_collector.rs This way, we could implement `bat`-specific things here.

Do we need the `alias cat=bat` detection? In which situation would it help?

keith-hall:
> Looking at these, I would suggest we call the new option `--bugreport`?

That to me implies it would automatically submit a bug report, to my eye the current "diagnostic" switch is clearer in intention :)

PR #1476, opened 2021-01-05
Add integration test for unparsable pager

Enselic:
Right now all tests pass even if we e.g. return

    Ok(OutputType::stdout())

instead of doing

    .chain_err(|| "Could not parse pager command.")?

so add a test to make sure this functionality don't break.

This is mainly to harden against regressions in #1402, but is self-contained and can be merged right away. As a side-note, we already have a test for `BAT_PAGER=""` so there is no need to add another one.


sharkdp:
Thank you!

Issue #1474, opened 2021-01-03
Simplify, refactor and clean up our GitHub Actions script


sharkdp:
@rivy has done a great job of providing a GitHub Actions configuration for `bat`:

- [`/.github/workflows/CICD.yml`](https://github.com/sharkdp/bat/blob/master/.github/workflows/CICD.yml)
- [Latest run on `master`](https://github.com/sharkdp/bat/actions/runs/458760265)

It works great (both for CI as well as for CD) and is a huge improvement over our previous Travis+Appveyor setup. However, I can't help but feel that it could be improved.

As for myself, I am currently starting to get into GitHub Actions (somewhat involuntarily because I need to move a lot of my projects away from Travis CI). So I would really appreciate some help. Here are some ideas:

- [x] ~~The whole Debian-package stuff feels like it should be in a separate GitHub action that we could re-use here.~~ (removed for now, see below)
- [x] The following section is particularly hard to understand and contains a lot of duplication: https://github.com/sharkdp/bat/blob/4fe5497d19914c6d912b3cc4586134981b186b3b/.github/workflows/CICD.yml#L109-L210
- [x] The naming of the build jobs and the naming of some steps in the process could be improved, in order to make it easier to find something in the CI output.
- [x] It seems like we are unnecessarily(?) using the full `cross`-compiling toolchain for builds that are native to the host OS: https://github.com/sharkdp/bat/blob/4fe5497d19914c6d912b3cc4586134981b186b3b/.github/workflows/CICD.yml#L86-L98 This should be checked.
- [x] We are currently not using the code coverage part. This is mainly because we deactivated the automatic posting of coverage results in the PR (which I don't really want to re-activate with the full output). Maybe someone has an idea on how to improve this situation.

Any help - especially in the form of *small* PRs that are easy to review for me as a beginner - would be very much appreciated!

Enselic:
Now that the second bullet point is checked, I thought it would be good time to elaborate a bit on the remaining points, and perhaps reconsider the importance of them.

- The whole Debian-package stuff

I took a [quick look](https://github.com/marketplace?type=actions&query=debian) but could not find any mature GitHub Action for creating Debian packages. Now that the Debian Package is its own step, I personally do not think it is a pain point any longer. It's pretty straightforward what it does, and it gives us a lot of flexibility to have it setup the way it is.

So my recommendation is to remove this bullet point from the list. (Or at least not have have it in a position as prominent as in a pinned issue.)

[...]

PR #1478, opened 2021-01-05
CICD: Code Coverage: Use matrix.job.toolchain directly

Enselic:
No need for complicated var and set-output logic when we can simply use
matrix.job.toolchain directly.

This is a small and self-contained PR to simplify CICD as requested in #1474.

(Disclaimer: I am new to GitHub Actions myself, but I am comfortable with CI/CD in general.)


sharkdp:
Perfect - thank you!

PR #1479, opened 2021-01-06
CICD: Remove duplicate set-output calls

Enselic:
The ::set-output syntax is the correct one according to
https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-output-parameter,
so remove duplicates without that syntax.

(I also tested to remove all ::set-output instead as a control, and then the build failed miserably: https://github.com/Enselic/bat/actions/runs/465537465)

This PR is for #1474.


sharkdp:
Thank you!

The initial idea was to have this as a debug output for humans, I think. But I'd rather get rid of the duplication :+1:

sharkdp (approved):

PR #1480, opened 2021-01-06
CICD: Remove use-cross when host == target

Enselic:
I tried to remove all `use-cross` occurrences, but the two targets this PR were the only ones that builds without `use-cross`.

For reference, here is what happens if you try to remove `use-cross` for all builds: https://github.com/Enselic/bat/actions/runs/464151908

This PR is for #1474.


sharkdp:
Thank you for looking into this!

This change potentially impacts continuous deployment. We should test this first. (I've experienced too many deployment failures when I finally have the time to make a release).

I'm never sure how to best do this. You could either tag a test-release on your fork. Or we could probably do the same on the branch in this repository.

sharkdp (approved):

Enselic:
I tagged the PR commit `93e802c` with `v0.0.1` which I pushed to my fork, and got a [green job](https://github.com/Enselic/bat/actions/runs/466229625) and a [release with downloads](https://github.com/Enselic/bat/releases/tag/v0.0.1), so seems like things didn't break?

sharkdp:
Perfect - thank you!

PR #1484, opened 2021-01-06
CICD: Code Coverage: Disable for now

Enselic:
Accoding to #1474:
> We are currently not using the code coverage part.

Since the Code Coverage job usually is the last to finish, I propose we disable it until someone makes it useful to run again. That will reduce the time it takes to complete the CICD pipeline and thus increase productivity for everyone!

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -380,7 +380,7 @@ jobs:
 
   coverage:
     name: Code Coverage
-    if: "false"
+    if: "false" # Temporarily disabled until made functional again, see https://github.com/sharkdp/bat/pull/1484
     runs-on: ${{ matrix.job.os }}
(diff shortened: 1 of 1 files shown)


.github/workflows/CICD.yml
```diff
@@ -380,6 +380,7 @@ jobs:
 
   coverage:
     name: Code Coverage
+    if: "false"
```
sharkdp, line 383: Can we maybe add a comment with a reference to this PR?

Enselic: Good idea. Done!

sharkdp (approved):

Enselic:
This is for https://github.com/sharkdp/bat/issues/1474 (mentioning it here to get a link from that issues to this PR)

Issue #1426, opened 2020-12-07
File content end part is not showing on plain mode


dalisoft:
**What version of `bat` are you using?**
`bat 0.17.1`

**Describe the bug you encountered:**
<img width="1931" alt="image" src="https://user-images.githubusercontent.com/3511344/101402691-e5e40100-38f5-11eb-9abc-682acd34bd43.png">

**What did you expect to happen instead?**
<img width="1929" alt="image" src="https://user-images.githubusercontent.com/3511344/101402759-001ddf00-38f6-11eb-8e18-295704e05e65.png">

**How did you install `bat`?**
`brew install bat`

**Workaround**
Passing `--wrap=character` as argument, but i want it should be by default and like it's performance bit slower

---

- OS: macOS Catalina 10.15.7
- Terminal: Apple Terminal
- Prompt: fish + starship
- Package manager: fish

Executing `https://github.com/sharkdp/bat/blob/master/diagnostics/info.sh` failed, says 
```
❯ ./info.sh
```

system
------

**$ uname -srm**
Darwin 19.6.0 x86_64  

**$ sw_vers**
ProductName:	Mac OS X  
ProductVersion:	10.15.7  
BuildVersion:	19H15  

bat
---

**$ bat --version**
bat 0.17.1  

**$ env**
BAT_THEME=Monokai Extended Bright  

bat_config
----------

bat_wrapper
-----------

No wrapper script for 'bat'.

bat_wrapper_function
--------------------

No wrapper function for 'bat'.

**$ /usr/local/bin/fish --login -i -c functions cat**
```
# Defined in - @ line 1
function cat --wraps=bat --description 'alias cat=bat'
  bat  $argv;
end
```

tool
----

**$ less --version**
less 487 (POSIX regular expressions)  

~/Downloads took 22s 
❯ 

```

**Comment by sharkdp** [comment]
> Executing `https://github.com/sharkdp/bat/blob/master/diagnostics/info.sh` failed, says
> 
> ```
> ❯ sh info.sh
> info.sh: line 7: syntax error near unexpected token `newline'
> info.sh: line 7: `<!DOCTYPE html>'
> ```

Looks like you downloaded the HTML page. Please try
```
wget https://raw.githubusercontent.com/sharkdp/bat/master/diagnostics/info.sh
```

sharkdp:
Ok, I think I see what is going on here.

[...]

PR #1486, opened 2021-01-07
fix bug where long lines were truncated in plain mode without wrap=never not being set

samuwen:
Added a boolean to the NoWrapping enum to detail if wrap=never is explicitly set or not.


CHANGELOG.md
```diff
@@ -35,14 +35,12 @@
 
 - `bat` is now available on snapstore with package name called `batcat`, see #1401 (@purveshpatel511)
 
-
```
sharkdp, line 38: Seems like your editor completely reformatted our CHANGELOG.md file.

sharkdp:

Adding a link to #1426

src/wrapping.rs
```diff
@@ -1,11 +1,11 @@
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum WrappingMode {
     Character,
-    NoWrapping,
+    NoWrapping(bool), // explicitly opted in or not
```
sharkdp, line 4: Maybe: "The bool specifies whether wrapping has been explicitly disabled by the user via `--wrap=never`"?

src/printer.rs
```diff
@@ -424,7 +424,9 @@ impl<'a> Printer for InteractivePrinter<'a> {
         }
 
         // Line contents.
-        if self.config.wrapping_mode == WrappingMode::NoWrapping {
+        if self.config.wrapping_mode == WrappingMode::NoWrapping(false)
+            || self.config.wrapping_mode == WrappingMode::NoWrapping(true)
+        {
```
sharkdp, line 429: I think this could be:
```rust
if let WrappingMode::NoWrapping(_) = self.config.wrapping_mode {
  // …
```

Not really easier to read though. Another option would be to use the `matches!` macro. But we would need to bump the minimum supported Rust version for that. I'd be okay with this. We could use the `matches!` macro in other places as well.

CHANGELOG.md
samuwen: ...stupid vscode. reverting.

src/wrapping.rs
samuwen: I like that. Moving to a line above as it is a little easier to read that way.

src/printer.rs
samuwen: Yeah that looks a lot cleaner. Thanks!

sharkdp (approved):
Thank you for the updates!

PR #1487, opened 2021-01-07
CICD: Build: Remove unused REF_* outputs

Enselic:
There is no usage of `steps.vars.outputs.REF_*` so no need to setup such
things. Also remove setting up REF_NAME and REF_BRANCH env vars, since
they are never read.

For https://github.com/sharkdp/bat/issues/1474

Successful deploy: https://github.com/Enselic/bat/releases/tag/v0.0.4


sharkdp (approved):
Thank you!

PR #1488, opened 2021-01-07
CICD: Build: Use matrix.job.use-cross directly

Enselic:
There is no need for a var and set-output detour.

For  https://github.com/sharkdp/bat/issues/1474

Successful deploy: https://github.com/Enselic/bat/releases/tag/v0.0.3


sharkdp (approved):
Thanks! Much better.

PR #1490, opened 2021-01-07
CICD: Build: Stop building on Ubuntu 16.04

Enselic:
This results in a nicer workflow file that is easier to follow.

Also remove the unneccesary doc row that repeats what is already in the
matrix and that is annoying to keep up to date.

For #1474

Successful deploy: https://github.com/Enselic/bat/releases/tag/v0.0.6

Btw, when deploy failed but build worked, what was wrong? Would be nice to be able to add some kind of deploy-dry-run so we don't have to manually push tags to verify deploy.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -87,9 +87,8 @@ jobs:
           - { os: ubuntu-18.04   , target: aarch64-unknown-linux-gnu   , use-cross: true }
           - { os: ubuntu-18.04   , target: i686-unknown-linux-gnu      , use-cross: true }
           - { os: ubuntu-18.04   , target: i686-unknown-linux-musl     , use-cross: true }
-          - { os: ubuntu-18.04   , target: x86_64-unknown-linux-gnu                      , disable-deploy: true }  ## deployed from ubuntu-16.04
+          - { os: ubuntu-18.04   , target: x86_64-unknown-linux-gnu    }
           - { os: ubuntu-18.04   , target: x86_64-unknown-linux-musl   , use-cross: true }
-          - { os: ubuntu-16.04   , target: x86_64-unknown-linux-gnu    }
           - { os: macos-latest   , target: x86_64-apple-darwin         }
           # - { os: windows-latest , target: i686-pc-windows-gnu         }  ## disabled; error: linker `i686-w64-mingw32-gcc` not found
           - { os: windows-latest , target: i686-pc-windows-msvc        }
@@ -357,7 +356,7 @@ jobs:
         fi
     - name: Publish archives and packages
       uses: softprops/action-gh-release@v1
-      if: steps.vars.outputs.DEPLOY && !matrix.job.disable-deploy
+      if: steps.vars.outputs.DEPLOY
       with:
         files: |
           ${{ steps.vars.outputs.STAGING }}/${{ steps.vars.outputs.PKG_NAME }}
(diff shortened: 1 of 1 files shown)


sharkdp:
> Btw, when deploy failed but build worked, what was wrong?

Well.. some of the deployment steps don't even run when the build is not triggered by a pushed Git tag. So any kind of error in the deployment part would only be visible once we make a release.

> Would be nice to be able to add some kind of deploy-dry-run so we don't have to manually push tags to verify deploy.

Absolutely! I wonder how other projects deal with this. Maybe all the artifacts should be built for every single build and only the upload to GitHub would be disabled? It would certainly increase the CI time but maybe it's worth it.

.github/workflows/CICD.yml
```diff
@@ -83,12 +83,11 @@ jobs:
       fail-fast: false
       matrix:
         job:
-          # { os, target, cargo-options, features, use-cross }
           - { os: ubuntu-latest  , target: arm-unknown-linux-gnueabihf , use-cross: true }
           - { os: ubuntu-18.04   , target: aarch64-unknown-linux-gnu   , use-cross: true }
           - { os: ubuntu-18.04   , target: i686-unknown-linux-gnu      , use-cross: true }
           - { os: ubuntu-18.04   , target: i686-unknown-linux-musl     , use-cross: true }
-          - { os: ubuntu-18.04   , target: x86_64-unknown-linux-gnu    }
+          - { os: ubuntu-18.04   , target: x86_64-unknown-linux-gnu                      , disable-deploy: true }  ## deployed from ubuntu-16.04
```
sharkdp, line 90: The change looks great. But it makes me wonder. Why do we even run the exact same job on Ubuntu 18.04 and 16.04? It's not really our job to make sure that everything still works on 16.04. The Rust toolchain should be exactly the same (?) so I'm not sure if there's any benefit in the 16.04 job.

What do you think?

Enselic:
> Well.. some of the deployment steps don't even run when the build is not triggered by a pushed Git tag. 

Ah yes, I see now that e.g. all the Debian package stuff only runs for release tags. I first thought the only git-tag-specific logic was in `name: Publish archives and packages`  (the only one with `if: steps.vars.outputs.DEPLOY`) which is a very simple step.

I'll see what I can do to in upcoming PRs to make normal CI runs closer to a real deploy.

> I wonder how other projects deal with this. Maybe all the artifacts should be built for every single build and only the upload to GitHub would be disabled?

Yes, that's the only way you can do it. The only way to know that PRs don't break deploy, is to do a deploy for every PR :) (but of course skip just the last "upload" step like you say)

> It would certainly increase the CI time but maybe it's worth it.

Speaking of, did we ever hit the CICD minutes quota? From what I know it is limited to I think 2000 minutes per month for the free tier.

.github/workflows/CICD.yml
Enselic: Considering that 16.04 reaches end-of-life in 3 months, I agree it makes sense to stop building for it and just use 18.04 instead. I'll update this PR.

sharkdp:
> Speaking of, did we ever hit the CICD minutes quota? From what I know it is limited to I think 2000 minutes per month for the free tier.

Looks like that applies to private repos only? https://docs.github.com/en/free-pro-team@latest/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions#about-billing-for-github-actions

Enselic:
Test deploy of latest commit looks good: https://github.com/Enselic/bat/releases/tag/v0.0.7

sharkdp (approved):

PR #1491, opened 2021-01-07
Fix theme dir hint in --list-themes output

csabahenk:


Enselic:
Looks like a reasonable patch to me. Before patch:

```
% bat --list-themes
...
Further themes can be installed to '/Users/martinhobby/.config/bat/config/themes', and are added to the cache with `bat cache --build`. For more information, see:
...
```

after patch:

```
Further themes can be installed to '/Users/martinhobby/.config/bat/themes', and are added to the cache with `bat cache --build`. For more information, see:
```

sharkdp:
Thank you very much!

PR #1494, opened 2021-01-09
Fix remaining clippy suggestions

sharkdp:
Now that we decided to move to Rust 1.42 as the min. required version, we can fix the remaining clippy "style" suggestions.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -27,7 +27,7 @@ jobs:
       uses: actions-rs/cargo@v1
       with:
         command: clippy
-        args: --all-targets --all-features -- --allow clippy::style
+        args: --all-targets --all-features
     - name: Test
       uses: actions-rs/cargo@v1
       with:
--- a/src/bin/bat/input.rs
+++ b/src/bin/bat/input.rs
@@ -2,7 +2,7 @@ use bat::input::Input;
 use std::ffi::OsStr;
 
 pub fn new_file_input<'a>(file: &'a OsStr, name: Option<&'a OsStr>) -> Input<'a> {
-    named(Input::ordinary_file(file), name.or_else(|| Some(file)))
+    named(Input::ordinary_file(file), name.or(Some(file)))
 }
 
 pub fn new_stdin_input(name: Option<&OsStr>) -> Input {
--- a/src/input.rs
+++ b/src/input.rs
@@ -137,11 +137,7 @@ impl<'a> Input<'a> {
     }
 
     pub fn is_stdin(&self) -> bool {
-        if let InputKind::StdIn = self.kind {
-            true
-        } else {
-            false
-        }
+        matches!(self.kind, InputKind::StdIn)
     }
 
     pub fn with_name(mut self, provided_name: Option<&OsStr>) -> Self {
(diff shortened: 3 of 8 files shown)


sharkdp:
I actually changed my mind about concerning clippy suggestions. Should we remove the `--allow clippy::style` argument here?

https://github.com/sharkdp/bat/blob/de6cb75f4bb5d6e65301b5e9a9890228760da4c4/.github/workflows/CICD.yml#L30

FYI: @Enselic @eth-p

Enselic:
The change itself looks good to me. but II still get a handful of remaining warnings, for example

    warning: you should consider adding a `Default` implementation for `BatTester`

maybe you don't run with ` --all-targets --all-features`? Also, I've noticed that clippy does not work well with incremental builds, so what I've ended up doing is to clean bat before running clippy, so this is the command I run locally to check for clippy lints:

    cargo clean -p bat && cargo clippy --all-targets --all-features

I support the removal of `--allow clippy::style` 👍, probably a good idea to do that change in this PR to make sure you don't miss any lints

sharkdp:
> maybe you don't run with ` --all-targets --all-features`?

Ah, right. I knew I had to clean, but forgot about these options. Thanks!

Enselic:
Just wanted to check if you deliberately left one warning around? The following warning remain:

    warning: you seem to be trying to match on a boolean expression
       --> src/bin/bat/app.rs:207:28

which can be seen in the clippy job for this PR found here: https://github.com/sharkdp/bat/pull/1494/checks?check_run_id=1674286630

IIRC it is tricky to fix, so maybe we should suppress that one? If we keep it around it will be somewhat annoying to make sure we don't introduce new ones, since it makes it harder to detect new ones due to noise when you run the clippy command.

(Keep mind that we only fail the CI job for clippy::correctness, and only warn (print in the log) for clippy:perf, clippy::style and clippy::complexity, i.e. the default behavior)

sharkdp:
> Just wanted to check if you deliberately left one warning around?

Yeah, I saw that as well. It was not deliberate :-). I did run the lastest version of `clippy` locally and - interestingly - it is not reported there. Maybe clippy is now smart enough to see that this whole construction makes sense due to the `#[cfg(…)]` annotation. I tried a few things to fix it but could not come up with a solution I liked. Unfortunately, I also didn't manage to silence the warning.



> If we keep it around it will be somewhat annoying to make sure we don't introduce new ones, since it makes it harder to detect new ones due to noise when you run the clippy command.

Agreed. So how do we fix it?

1. Switch off that particular warning globally.
2. Try to somehow switch it off locally. I didn't find a way to do that, but haven't tried for long.
3. Increase the minimum supported Rust version to a version where the clippy warning does not appear.
4. Use clippy on the latest Rust toolchain instead of the minimum supported.

I think I would vote for 3, but unfortunately, it could lead to broken builds whenever new Rust versions appear. But as we only break the build for correctness, that's probably fine?

Enselic:
Ah yes, I remember now, I looked that one up before, and they downgraded it to pedantic because it was too annoying: https://github.com/rust-lang/rust-clippy/commit/e1d13c34b0beaea9a5fbf13687672ef85e779d9f

I propose we handle it with 1, because it is easy and harmless. I can create a PR for that later. Then we can remove it when we upgrade MSRV for other reasons (which might take a while but that's OK)

PR #1500, opened 2021-01-10
CICD: Use fixed OS versions instead of 'latest' ones

Enselic:
This reduces the risk of the build suddenly breaking, and fixes this
current warning:

    Ubuntu-latest workflows will use Ubuntu-20.04 soon.
    For more details, see https://github.com/actions/virtual-environments/issues/1816

I've used the mapping found at
https://github.com/actions/virtual-environments, so there should be no
actual change in OS versions, only semantically so.

Test deploy looks good: https://github.com/Enselic/bat/releases/tag/v0.0.11

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -12,7 +12,7 @@ on: [push, pull_request]
 jobs:
   min_version:
     name: Minimum supported rust version
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
     steps:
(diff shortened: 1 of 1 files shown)


sharkdp:
I think that's a good idea. I did the same thing in another project of mine: https://github.com/sharkdp/dbg-macro/blob/master/.github/workflows/ci.yml (did not fix macOS version because I don't really expect macOS to be any different from Linux for that project).

PR #1501, opened 2021-01-10
Also build Debian packages in PR CI workflow

Enselic:
This PR makes Debian packages be built and published as artifacts for regular PR workflows.

It consists of four steps (commits) that can be code review individually together with respective commit message if the full diff is too confusing.

I think the only controversial change is the addition of `PROJECT_VERSION: "0.17.1"`, but it is already duplicated a lot, and I could not figure out a way to parse it from Cargo.toml in the CI script that was less annoying than duplicating it another time. I am open for ideas though.

The reason we need to do it is that Debian packages can not be built without a version, and since a PR does not contain a version (unlike a tag) we need to use the Cargo.toml version. If we think it is too confusing to reuse the same version string for different builds, we could perhaps suffix the git SHA at the end, but personally I think that is not necessary at this stage.

This PR makes the following artifacts be built and uploaded for each PR:

```
bat-musl_0.17.1_amd64.deb
bat-musl_0.17.1_i686.deb
bat-v0.17.1-aarch64-unknown-linux-gnu.tar.gz
bat-v0.17.1-arm-unknown-linux-gnueabihf.tar.gz
bat-v0.17.1-i686-pc-windows-msvc.zip
bat-v0.17.1-i686-unknown-linux-gnu.tar.gz
bat-v0.17.1-i686-unknown-linux-musl.tar.gz
bat-v0.17.1-x86_64-apple-darwin.tar.gz
bat-v0.17.1-x86_64-pc-windows-gnu.zip
bat-v0.17.1-x86_64-pc-windows-msvc.zip
bat-v0.17.1-x86_64-unknown-linux-gnu.tar.gz
bat-v0.17.1-x86_64-unknown-linux-musl.tar.gz
bat_0.17.1_amd64.deb
bat_0.17.1_arm64.deb
bat_0.17.1_armhf.deb
bat_0.17.1_i686.deb
```

so now the only entries missing compared to a real release is:

```
Source code (zip)
Source code (tar.gz)
```

Test deploy looks good: https://github.com/Enselic/bat/releases/tag/v0.0.12

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -124,7 +124,7 @@ jobs:
         echo ::set-output name=PKG_BASENAME::${PKG_BASENAME}
         echo ::set-output name=PKG_NAME::${PKG_NAME}
         # deployable tag? (ie, leading "vM" or "M"; M == version number)
-        unset DEPLOY ; if [[ $REF_TAG =~ ^[vV]?[0-9].* ]]; then DEPLOY='true' ; fi
+        unset DEPLOY ; if [[ $REF_TAG =~ ^v[0-9].* ]]; then DEPLOY='true' ; fi
         echo ::set-output name=DEPLOY::${DEPLOY}
         # DPKG architecture?
         unset DPKG_ARCH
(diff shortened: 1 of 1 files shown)


sharkdp:
Fantastic - thank you very much. I could imagine that we could use [`cargo get`](https://crates.io/crates/cargo-get) to extract the crate version (and maybe some other things) programatically.  I will look into that.

PR #1502, opened 2021-01-10
CICD: Use 'cargo get' to extract crate metadata

sharkdp:
As discussed in #1501: try to use `cargo get` to extract the crate metadata from `Cargo.toml` automatically.

GitHub Actions documentation on setting env variables dynamically: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable

Ref: #1474 

Edit: Looks like this works as expected. This is how the environment looks like *after* the "Extract crate information" step:
```yaml
  env:
    MIN_SUPPORTED_RUST_VERSION: 1.42.0
    PROJECT_NAME: bat
    PROJECT_VERSION: 0.17.1
    PROJECT_MAINTAINER: David Peter <mail@david-peter.de>
    PROJECT_HOMEPAGE: https://github.com/sharkdp/bat
```

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -105,6 +105,7 @@ jobs:
       with:
         crate: cargo-get
         version: latest
+        use-tool-cache: true
     - name: Extract crate information
       shell: bash
       run: |
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -1,11 +1,6 @@
 name: CICD
 
 env:
-  PROJECT_NAME: bat
-  PROJECT_VERSION: "0.17.1"
-  PROJECT_DESC: "A `cat` clone with wings"
-  PROJECT_MAINTAINER: "David Peter <mail@david-peter.de>"
-  PROJECT_HOMEPAGE: "https://github.com/sharkdp/bat"
   MIN_SUPPORTED_RUST_VERSION: "1.42.0"
 
 on: [push, pull_request]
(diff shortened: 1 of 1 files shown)


.github/workflows/CICD.yml
```diff
@@ -105,6 +100,18 @@ jobs:
           arm-unknown-linux-gnueabihf) sudo apt-get -y update ; sudo apt-get -y install gcc-arm-linux-gnueabihf ;;
           aarch64-unknown-linux-gnu) sudo apt-get -y update ; sudo apt-get -y install gcc-aarch64-linux-gnu ;;
         esac
+    - name: Install 'cargo get'
+      uses: actions-rs/install@v0.1
+      with:
+        crate: cargo-get
+        version: latest
```
sharkdp, line 107: Maybe that should be set to a fixed version.

.github/workflows/CICD.yml
```diff
@@ -1,11 +1,6 @@
 name: CICD
 
 env:
-  PROJECT_NAME: bat
-  PROJECT_VERSION: "0.17.1"
-  PROJECT_DESC: "A `cat` clone with wings"
```
sharkdp, line 6: PROJECT_DESC was unused. We need different descriptions for the Debian package due to packaging policies ("Should not start with 'A …'").

.github/workflows/CICD.yml
Enselic: I think so, yes (but I won't be upset if you keep it as is).

Enselic:
Nice! 👍

Enselic (approved):

sharkdp:
Unfortunately, it adds 1 min 30 sec to the CI time on all Build jobs. This could be improved by using cached binaries, but unfortunately, `cargo get` is not yet part of the cached crates. I guess we could ask at https://github.com/actions-rs/tool-cache for it to be included.

sharkdp:
As a test, I changed to a rather ugly version that simply uses `sed` to extract the information from the TOML file. Not sure if this will work on Windows.

But 1m30s is a bit too long for this, I think.

Enselic:
I actually think this is even better :)

Had I known about the `>> $GITHUB_ENV` trick I would have done it like this myself.

sharkdp:
> Seems to work fine on windows.

Indeed. Apparently they make sure to set up some basic Unix tools on the CI image, which is great.

> I actually think this at least as good :)

Ok :smile:. Let's go with the `sed` version for now. I agree: if it causes problems later we can switch to the `cargo get` version.

PR #1506, opened 2021-01-10
Fix #1413 Can't use `bat` at all! (Error: Use of bat as a pager is di…

### LINKED ISSUE #1413: Can't use `bat` at all! (Error: Use of bat as a pager is disallowed...)
bat 0.17.1

Whenever I use `bat`, I'm shown the following error message:

> [bat error]: Use of bat as a pager is disallowed in order to avoid infinite recursion problems

I believe this started happening after a certain update but can't exactly pinpoint which one. I've uninstalled and reinstalled `bat` again but with no luck.

For `bat` to display the contents as usual.

Homebrew

---

system
------

**$ uname -srm**
Darwin 20.1.0 x86_64  

**$ sw_vers**
ProductName:	macOS  
ProductVersion:	11.0.1  
BuildVersion:	20B29  

bat
---

**$ bat --version**
bat 0.17.1  

**$ env**
PAGER=bat  

bat_config
----------

bat_wrapper
-----------

No wrapper script for 'bat'.

bat_wrapper_function
--------------------

tool
----

**$ less --version**

Enselic:
…sallowed...)

Fixed by implementing the proposal by sharkdp:

* Allow PAGER=bat, but ignore the setting in bat and simply default to
  less. Unless of course, BAT_PAGER or --pager is used to overwrite the
  value of PAGER.

* Disallow the usage of bat within BAT_PAGER and --pager.

### DIFF
--- a/src/pager.rs
+++ b/src/pager.rs
@@ -98,20 +98,18 @@ pub(crate) fn get_pager(config_pager: Option<&str>) -> Result<Option<Pager>, Par
         Some((bin, args)) => {
             let kind = PagerKind::from_bin(bin);
 
-            let use_less_instead = match source {
-                PagerSource::EnvVarPager => {
-                    // 'more' and 'most' do not supports colors; automatically use 'less' instead
-                    // if the problematic pager came from the generic PAGER env var
-                    let no_color_support = kind == PagerKind::More || kind == PagerKind::Most;
+            let use_less_instead = match (&source, &kind) {
+                // 'more' and 'most' do not supports colors; automatically use 'less' instead
+                // if the problematic pager came from the generic PAGER env var
+                (PagerSource::EnvVarPager, PagerKind::More) => true,
+                (PagerSource::EnvVarPager, PagerKind::Most) => true,
 
-                    // If PAGER=bat, silently use 'less' instead to prevent recursion ...
-                    let is_self = kind == PagerKind::Bat;
+                // If PAGER=bat, silently use 'less' instead to prevent recursion ...
+                (PagerSource::EnvVarPager, PagerKind::Bat) => true,
 
-                    no_color_support || is_self
-                },
-                // Never silently replace with less if bat-specific means to set a pager is used
-                _ => false
-            } ;
+                // Never silently use less if BAT_PAGER or --pager has been specified
+                _ => false,
+            };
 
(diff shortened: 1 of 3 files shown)


src/pager.rs
```diff
@@ -98,10 +98,20 @@ pub(crate) fn get_pager(config_pager: Option<&str>) -> Result<Option<Pager>, Par
...
+                    let no_color_support = kind == PagerKind::More || kind == PagerKind::Most;
+
+                    // If PAGER=bat, silently use 'less' instead to prevent recursion ...
+                    let is_self = kind == PagerKind::Bat;
+
+                    no_color_support || is_self
+                },
+                // Never silently replace with less if bat-specific means to set a pager is used
+                _ => false
+            } ;
```
sharkdp, line 114: Maybe this would be a bit clearer? (with added comments)
``` rust
let use_less_instead = match (source, kind) {
  (PagerSource::EnvVarPager, PagerKind::More) => true,
  (PagerSource::EnvVarPager, PagerKind::Most) => true,
  (PagerSource::EnvVarPager, PagerKind::Bat) => true,
  _ => false,
};
```

src/pager.rs
```diff
@@ -98,10 +98,20 @@ pub(crate) fn get_pager(config_pager: Option<&str>) -> Result<Option<Pager>, Par
...
+                    // 'more' and 'most' do not supports colors; automatically use 'less' instead
+                    // if the problematic pager came from the generic PAGER env var
+                    let no_color_support = kind == PagerKind::More || kind == PagerKind::Most;
+
+                    // If PAGER=bat, silently use 'less' instead to prevent recursion ...
+                    let is_self = kind == PagerKind::Bat;
+
+                    no_color_support || is_self
+                },
+                // Never silently replace with less if bat-specific means to set a pager is used
```
sharkdp, line 112: Took my a while to understand this comment. Maybe:
"Never silently use less if BAT_PAGER or --pager has been specified"

sharkdp:
Great how this can now be easily solved with the previous changes!

src/pager.rs
Enselic: Agreed, I will change to something like that. I especially like that the deepest level of indentation is reduced by one.

sharkdp (approved):

PR #1507, opened 2021-01-10
CICD: Build: Split up into separate 'Debian package' step

Enselic:
Make sure to ignore whitespace changes when diffing.

For #1474

(I'm skipping to test-deploy a release from now on, since we have started to build all release/deploy artifacts with every PR. One can manually sanity-check that deploy still works by checking for 16 artifacts that seems to have correct names and file sizes)


sharkdp:
Thanks again!



> (I'm skipping to test-deploy a release from now on, since we have started to build all release/deploy artifacts with every PR. One can manually sanity-check that deploy still works by checking for 16 artifacts that seems to have correct names and file sizes)

:+1:

Issue #1098, opened 2020-07-06
Bottom border is cut off when pager is active on Windows


brainplot:
**What version of `bat` are you using?**
`bat 0.15.4`

**Describe the bug you encountered:**
It looks like the bottom border `bat` displays by default is cut off on Windows, when the pager is active, as shown in this screenshot:

#### _With pager_
![image](https://user-images.githubusercontent.com/28014843/86616902-c9dedb00-bfb6-11ea-843a-10374dea75e1.png)

#### _Without pager_
![image](https://user-images.githubusercontent.com/28014843/86617326-59848980-bfb7-11ea-993d-09df584b008b.png)

I have `bat` also installed in WSL on Ubuntu 20.04 and this issue isn't present at all, which makes me think there's something wrong with how Windows handles `less`.

**Describe what you expected to happen?**
Bottom border not being cut off like on other platforms :)

**How did you install `bat`?**
`scoop install bat`

**EDIT**: I'll add my `less` version for completeness
```
less 561 (Spencer V8 regular expressions)
Copyright (C) 1984-2020  Mark Nudelman
```

---

Windows version; Windows 10 2004 (OS Build 19041.329)

sharkdp:
This does look like a problem with the `--quit-if-one-screen` option in your version of `less`.

As a workaround, you could try to use `--pager="less -R"` in your `bat` config file to disable `--quit-if-one-screen`.

Does this happen for all terminal sizes?

brainplot:
Sadly that doesn't seem the cause of the issue:

![image](https://user-images.githubusercontent.com/28014843/88466247-a4186800-ceca-11ea-81d1-3e3d6e839108.png)

sharkdp:
Thank you for the feedback.

Could you please try to output `bat`s output to a file (`bat --color=always .\test.txt > output.txt`) and display it with `less` directly (or `less -R`)? Does this lead to the same problem?

brainplot:
I tried that but it doesn't quite show the intended output:

![image](https://user-images.githubusercontent.com/28014843/88467184-b26b8180-ced4-11ea-9205-1a16da885c59.png)

However, even though the output is jumbled, it looks as though the bottom border was preserved (or at least it does so to me 😄)

sharkdp:
It looks like `less` 562 fixes a lot of bugs on Windows: http://www.greenwoodsoftware.com/less/news.562.html

[...]

PR #1509, opened 2021-01-10
Fixes to use the user-specified pager for less version check and diagnostic.

eth-p:
As we briefly mentioned in #1098, hardcoded paths to `less` were a bit of a pain point for both the version check and the `--diagnostic` option. This is a quick pull request that fixes both of them.

### DIFF
--- a/src/bin/bat/main.rs
+++ b/src/bin/bat/main.rs
@@ -229,6 +229,8 @@ fn run() -> Result<bool> {
 
     if app.matches.is_present("diagnostic") {
         use bugreport::{bugreport, collector::*, format::Markdown};
+        let pager = bat::config::get_pager_executable(app.matches.value_of("pager"))
+            .unwrap_or("less".to_owned()); // FIXME: Avoid non-canonical path to "less".
 
         bugreport!()
             .info(SoftwareVersion::default())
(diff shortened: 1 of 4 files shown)


sharkdp:
Thank you.

Unfortunately, this will not work for `PAGER` and `BAT_PAGER`, right?

```
BAT_PAGER="path/to/custom/less" …
```

sharkdp:
Also, it will not apply `shell_words::split(…)` like in `src/pager.rs`. So it will not work in a case like this:
```
bat --pager="path/to/custom/less -RF" …
```

eth-p:
I... forgot to use `get_pager`. I was looking at the function, but it slipped my mind when actually committing it. My bad haha... 

> Also, it will not apply shell_words::split(…) like in src/pager.rs. So it will not work in a case like this:
That one I didn't realize. Thanks!

I'll fix these.

eth-p:
Fixed. I had to expose some internal API to be able to access `bat::pager::get_pager` function, but I hid it behind the `application` feature to prevent it from being used in bat-as-a-library.

sharkdp:
Thank you for the updates! I need to think about this some more. Will get back to you soon.

sharkdp:
Looks great - thanks!

PR #1512, opened 2021-01-11
CICD: Build: Split out dir creation into the individual package steps

Enselic:
It's cleaner because we can reuse helper vars and create the dirs closer
in time to when they are needed.

For #1474


.github/workflows/CICD.yml
```diff
@@ -260,6 +256,7 @@ jobs:
       run: |
         COPYRIGHT_YEARS="2018 - "$(date "+%Y")
         DPKG_DIR="${{ steps.vars.outputs.STAGING }}/dpkg"
+        mkdir -p "${DPKG_DIR}/dpkg"
```
Enselic, line 259: shall be `mkdir -p "${DPKG_DIR}"`

Enselic:
(fixed) commit included in https://github.com/sharkdp/bat/pull/1513

PR #1514, opened 2021-01-11
Add support for env vars and manpage refs with dots to Manpage syntax

### LINKED ISSUE #1511: Name of man pages with dots is not fully highlighted
# Description
Names like `pacman.conf` or `systemd-system.conf` does not get highlighted correctly, only the part after the dot is highlighted.

image

Command used to produce the image above:

```bash
MANPAGER="sh -c 'col -bx | bat -l man -p'" man systemd
```

# Expected result
The whole name (e.g: `systemd-system.conf`) is highlighted instead of just the part after the dot.

# Environment
`bat` is installed using the package in Arch official repo.

`info.sh` output:

```
system
------

**$ uname -srm**
Linux 5.9.14-arch1-1 x86_64

**$ lsb_release -a**
LSB Version:	1.4
Distributor ID:	Arch
Description:	Arch Linux
Release:	rolling
Codename:	n/a

bat
---

**$ bat --version**
bat 0.17.1

**$ env**
PAGER=less

bat_config
----------

bat_wrapper
-----------

No wrapper script for 'bat'.

bat_wrapper_function
--------------------

tool
----

**$ less --version**
less 563 (PCRE regular expressions)
```

keith-hall:
This PR updates the Manpage syntax to fix #1511 and also highlight environment variables, i.e. like those beginning with a `$` sign in `man systemd`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - If plain mode is set and wrap is not explicitly opted in, long lines will no be truncated, see #1426
 - If `PAGER` (but not `BAT_PAGER` or `--pager`) is `more` or `most`, silently use `less` instead to ensure support for colors, see #1063 (@Enselic)
 - If `PAGER` is `bat`, silently use `less` to prevent recursion. For `BAT_PAGER` or `--pager`, exit with error, see #1413 (@Enselic)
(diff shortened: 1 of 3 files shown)


sharkdp (approved):
Thanks!

PR #1423, opened 2020-12-02
Debian package fixes (2)

chris48s:
Refs https://github.com/sharkdp/bat/issues/1352#issuecomment-737054985

This replaces the commits reverted in https://github.com/sharkdp/bat/pull/1393 which are not related to the issue reported in #1352

PR #1516, opened 2021-01-12
CICD: Build: Introduce and use new 'Strip release bin' step

Enselic:
So that we don't have to duplicate that logic in both 'Debian package'
and 'Package' steps.

For #1474

I have sanity-checked/compared a selection of extracted Debian and regular packages produced from these changes with the ones from the official 0.17.1 release, and seems like nothing is broken.

Only differences were intentional and not related to my changes (Using `/usr/share/doc/bat-musl/` instead of `/usr/share/doc/bat/` in musl .deb, and the addition of the `CHANGELOG.md` in both package types)

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -116,9 +116,6 @@ jobs:
         # staging directory
         STAGING='_staging'
         echo ::set-output name=STAGING::${STAGING}
-        # determine EXE suffix
-        EXE_suffix="" ; case ${{ matrix.job.target }} in *-pc-windows-*) EXE_suffix=".exe" ;; esac;
-        echo ::set-output name=EXE_suffix::${EXE_suffix}
         unset IS_RELEASE ; if [[ $GITHUB_REF =~ ^refs/tags/v[0-9].* ]]; then IS_RELEASE='true' ; fi
         echo ::set-output name=IS_RELEASE::${IS_RELEASE}
         # target-specific options
@@ -126,9 +123,6 @@ jobs:
         unset CARGO_TEST_OPTIONS
         unset CARGO_TEST_OPTIONS ; case ${{ matrix.job.target }} in arm-* | aarch64-*) CARGO_TEST_OPTIONS="--lib --bin ${PROJECT_NAME}" ;; esac;
         echo ::set-output name=CARGO_TEST_OPTIONS::${CARGO_TEST_OPTIONS}
-        # * executable for `strip`?
-        STRIP="strip" ; case ${{ matrix.job.target }} in arm-unknown-linux-gnueabihf) STRIP="arm-linux-gnueabihf-strip" ;; aarch64-unknown-linux-gnu) STRIP="aarch64-linux-gnu-strip" ;; *-pc-windows-msvc) STRIP="" ;; esac;
-        echo ::set-output name=STRIP::${STRIP}
     - name: Install Rust toolchain
       uses: actions-rs/toolchain@v1
       with:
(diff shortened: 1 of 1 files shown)


sharkdp:
Thanks!

> Using `/usr/share/doc/bat-musl/` instead of `/usr/share/doc/bat/` in musl .deb

Wait.. is that really what we want? Or rather.. was this really intentional? I must have missed that.

Enselic:
Not sure if that is what we want, but that's what we committed: https://github.com/sharkdp/bat/commit/c1e4746d509b5d5b3a08d88cc69833b77116e72b

sharkdp:
Oh, thank you for tracking that down. Should have payed more attention in the review. @chris48s could you comment on this? We probably do this to be in line with Debian policies? In that case, do we really want that package to be called `bat-musl`? Or should we just give it the same name as the libc package?

Enselic:
Do you think this PR looks good though?

I would prefer to not block this PR on discussions about unrelated Debian packaging 😬 I hope that's OK!

Enselic:
Probably better to ping Chris in the original PR? https://github.com/sharkdp/bat/pull/1423

sharkdp:
> Do you think this PR looks good though?

Will take a closer look soon, haven't looked at the details so far.

chris48s:
In terms of where the docs live, installed documentation files should live in `/usr/share/doc/<package name>`
https://www.debian.org/doc/debian-policy/ch-docs.html#additional-documentation

---

In terms of renaming the package, I think if you just call both packages just "bat" then they become interchangeable so for example [bat (libc) 17.0] --> [bat (musl) 17.1] would be considered a valid upgrade path.

At the moment, bat-musl/bat conflict with each other i.e:

* In the debian control file for bat-musl you've got the line `Conflicts: bat`
* In the debian control file for bat you've got the line `Conflicts: bat-musl`

so if you've got the musl or libc version installed already then you have to upgrade libc --> libc / musl --> musl. If both packages are called bat, you would have to drop the conflict. Bat can't conflict with bat.

I'm not sure if there is any guidance on how to do this if you support multiple complier versions like this but the way you have it set up now (with the naming/conflicts) seems like it is probably done like that to allow you to enforce a useful constraint.

.github/workflows/CICD.yml
```diff
@@ -151,6 +145,41 @@ jobs:
...
+        # Figure out what strip tool to use if any
+        STRIP="strip"
+        case ${{ matrix.job.target }} in
+          arm-unknown-linux-gnueabihf) STRIP="arm-linux-gnueabihf-strip" ;;
+          aarch64-unknown-linux-gnu) STRIP="aarch64-linux-gnu-strip" ;;
+          *-pc-windows-msvc) STRIP="" ;;
+        esac;
+
+        # Setup paths
+        BIN_DIR="_cicd-intermediates/stripped-release-bin/"
```
sharkdp, line 167: Any reason why this has a complicated path like this? Do `_*` folders have a special meaning?

sharkdp (approved):

sharkdp:
Looks good thanks!

.github/workflows/CICD.yml
Enselic: The main reason for the _ prefix is to follow the established naming convention is this file. Personally I kind of like the _ prefix, because it signals "temporary, short lived dir". If you annoys you, we can of course stop using that prefix :)

The reason for the /stripped-release-bin/ sub dir is for a long term plan of collecting everything CI-temporary under the same dir, so the sub dir is only for "isolation".

sharkdp: Makes sense, thanks.

Issue #1519, opened 2021-01-15
Permission denied to read file


barenko:
**What version of `bat` are you using?**
bat 0.17.1

**Describe the bug you encountered:**
I Have a file with read permission:
.rw-r--r-- 7,0k myuser 14 Jan 23:16 .zshrc

I tried to read normally (with the file owner)
batcat .zshrc
[bat error]: '.zshrc': Permission denied (os error 13)

I retried using sudo:
sudo batcat /home/myuser/.zshrc 
[bat error]: '/home/myuser/.zshrc': Permission denied (os error 13)

...

**What did you expect to happen instead?**
show the file contents
...

**How did you install `bat`?**
using the snap package

---

**Environment**

system
------

**$ uname -srm**
Linux 5.4.0-26-generic x86_64  

**$ lsb_release -a**
No LSB modules are available.  
Distributor ID:	Linuxmint  
Description:	Linux Mint 20  
Release:	20  
Codename:	ulyana  

bat
---

**$ batcat --version**
bat 0.17.1  

**$ env**
PAGER=less  

bat_config
----------

bat_wrapper
-----------

No wrapper script for 'bat'.

No wrapper script for 'batcat'.

bat_wrapper_function
--------------------

No wrapper function for 'bat'.

No wrapper function for 'cat'.

No wrapper function for 'batcat'.

tool
----

**$ less --version**
less 551 (GNU regular expressions)

Enselic:
Thanks for taking the time to report a bug!

I have a couple of questions:
1. Do you have the same problem with other files with the same permissions?
2. Do you have the same problem when you use `cat` instead of `bat`?
3. Are there any symbolic links involved? Please provide the full output of `ls -l /home/myuser/.zshrc` (you seem to have redacted some parts in the output you include, since the group info is missing)
4. Also provide the output of `whoami` please

sharkdp:
> **How did you install `bat`?**
> using the snap package

I'm pretty sure this is the problem. Others had similar issues when installing from snap (#1424).  FYI @purveshpatel511 

To be honest, I am considering to remove the snap install instructions from the README. I don't want users to go through this experience. I remember that ripgrep did something similar because they encountered too many bugreports concerning wrong permissions.

Unless that problem can be fixed in the snap package itself (classic confinement), I would discourage from installing via snap.

@barenko Could you try to install `bat` from the packages in this repository and see if that fixes your problem? https://github.com/sharkdp/bat/releases/tag/v0.17.1

[...]

PR #1524, opened 2021-01-22
README.md: Warn about buggy snap packages

Enselic:
As discussed in #1519, the batcat snap package is too problematic for
official endorsement, so withdraw recommendation from README.me (and
remove section from CHANGELOG.md since a release with it has not been
made yet).

For now this MR should just be seen as a sketch of how we could go about doing it, in case we end up choosing this path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,10 +42,6 @@
    - `input_from_reader_with_name`
    - `vcs_modification_markers` (if the `git` feature is not enabled)
 
-## Packaging
-
-- `bat` is now available on snapstore with package name called `batcat`, see #1401 (@purveshpatel511)
-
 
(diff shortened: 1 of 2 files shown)


eth-p (changes requested):
I think calling the existing packages "buggy" can come off as rude or callous.

It might be better to say "Existing packages may be available, but are not officially supported and may contain issues."

With the link on the word "issues".

sharkdp:
Thanks!

sharkdp:
Unfortuntely, Ubuntu shows this:
```
$ sudo apt install batcat
Reading package lists... Done
Building dependency tree       
Reading state information... Done

No apt package "batcat", but there is a snap with that name.
Try "snap install batcat"
```

PR #1541, opened 2021-02-16
CICD: Build: Use package-specific staging dir

Enselic:
A common staging dir confuses more than it helps, so let each package
step take care of its own staging dir.

For #1474

Test deploy (with equivalent commit tree, I just modified commit msg) looks good: https://github.com/Enselic/bat/releases/tag/v0.0.13


sharkdp (approved):
Thanks!

PR #1544, opened 2021-02-17
CICD: Build: Split up ‘Initialize workflow variables’ step

Enselic:
Here are the 3 last commits towards the goal of getting rid of the ‘Initialize workflow variables’ step mentioned in the second bullet in #1474.

Full diff turned out a bit confusing, but reviewing each commit individually should make it crystal clear what is happening.

Test deploy look good: https://github.com/Enselic/bat/releases/tag/v0.0.14


.github/workflows/CICD.yml
```diff
@@ -376,9 +374,15 @@ jobs:
       with:
         name: ${{ steps.debian-package.outputs.DPKG_NAME }}
         path: ${{ steps.debian-package.outputs.DPKG_PATH }}
+    - name: Check is release
```
sharkdp, line 376: Maybe "Check for release"? Or "Check for release version tag"?

sharkdp (approved):
Thank you!

.github/workflows/CICD.yml
Enselic: I agree "Check is release" sounds weird if one reads it like regular English (which is a reasonable thing to do) rather than dry systematic specification of a build step, which is what it currently is.

I'll change to "Check for release",

PR #1546, opened 2021-02-20
CICD: Naming (jobs, stages)

sharkdp:
for #1474

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -76,7 +76,7 @@ jobs:
       run: bat --list-themes
 
   build:
-    name: Build
+    name: ${{ matrix.job.os }} (${{ matrix.job.target }})
     runs-on: ${{ matrix.job.os }}
     strategy:
       fail-fast: false


.github/workflows/CICD.yml
```diff
@@ -76,7 +76,7 @@ jobs:
       run: bat --list-themes
 
   build:
-    name: Build
+    name: ${{ matrix.job.os }} (${{ matrix.job.target }})
```
Enselic, line 79: Nice!

But, hmm, I wonder, wouldn't it make more sense to have `target` first, and `os` in parenthesis after, i.e. flip them around?

Quite often you don't see the full name in GitHub Actions UI, and since target is the most differentiating factor, it is more practical to see that first. I _think_ :thinking:

sharkdp: > But, hmm, I wonder, wouldn't it make more sense to have `target` first, and `os` in parenthesis after, i.e. flip them around?

I thought about it, but I think this way the jobs would be group by architecture (e.g. i686), not by OS. Since most of the build failures affect a certain OS, I guess it makes more sense to have that first.

> Quite often you don't see the full name in GitHub Actions UI, and since target is the most differentiating factor, it is more practical to see that first. I _think_ thinking

Since I removed "Build", we can now see the full remaining string.

Enselic: Sounds good, please go ahead :+1:

sharkdp:
I also tried to integrate the min. supported Rust version in the job name, but failed (can't use `env.*` vars).

PR #1549, opened 2021-02-21
CICD: Don't run jobs twice in PRs

Enselic:
Without this change, creating a PR triggers all jobs to run twice. Once
due to a push event and once due to a pull_request event.

Change to only trigger jobs due to push when pushing a tag or to the
master branch, to avoid duplicate jobs for each PR.

Triggering for tags confirmed to still work: https://github.com/Enselic/bat/actions/runs/586250776

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -4,7 +4,13 @@ env:
   MIN_SUPPORTED_RUST_VERSION: "1.42.0"
   CICD_INTERMEDIATES_DIR: "_cicd-intermediates"
 
-on: [push, pull_request]
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+    tags:
+      - '*'
 
 jobs:
   min_version:


sharkdp:
> Change to only trigger jobs due to push when pushing a tag or to the
> master branch, to avoid duplicate jobs for each PR.

Are you sure this will work for (normal) PRs from forks? I think they only trigger the pull_request event.

In the current situation, we only have duplicated builds for PRs from a local branch. I would definitely like to get rid of this. That being said, I'm not really sure if the two are actually 100% the same. In some CI setups, a "push" build will only build the local branch, but a "PR" build would first merge the (remote) branch to master and then perform a build. Both types of checks can be useful and are generally not the same. But I haven't checked if that's really the case for GitHub actions.

Enselic:
> Are you sure this will work for (normal) PRs from forks? I think they only trigger the pull_request event.

Hm I don't quite follow your reasoning I'm afraid. Would you like Actions to trigger for all pushed branches? In general, you don't want to trigger CICD for arbitrary branches, since they are usually work-in-progress.

It could be useful to know what GitHub themselves recommend. If you let GitHub create a template for you, this is what you get:

```
# This is a basic workflow to help you get started with Actions

name: CI

# Controls when the action will run. 
on:
  # Triggers the workflow on push or pull request events but only for the master branch
  push:
    branches: [ master ]
  pull_request:
    branches: [ master ]

  # Allows you to run this workflow manually from the Actions tab
  workflow_dispatch:
```

so the vanilla GitHub approach is to only trigger Actions for the main branch, There are of course exceptions, but I don't think we are an exception?

Issue #952, opened 2020-04-26
bat is not working on the Raspberry Pi Zero


HattDroid:
Hi,

First thank you for your tool, it's been serving me very well since 2018.

I have been using on x86_64 for years now and got used to it. I saw you distribute ARM binaries and decided to give it a go.

Trying both packages

```
bat-v0.15.0-arm-unknown-linux-gnueabihf.tar.gz
bat_0.15.0_armhf.deb
```

It every time gives me the following error when trying to launch bat binary:

```Illegal instruction```

A quick 'file' on the binary shows nothing unusual for the Pi Zero:

```bat: ELF 32-bit LSB shared object, ARM, EABI5 version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-armhf.so.3, for GNU/Linux 3.2.0, BuildID[sha1]=5545ace89bb1ef2c3b0e80caf53964cdb762ca36, stripped```

Would you know what is wrong with the release? Are you assuming a specific instruction set (or extension) when cross-compiling?

Thanks.

**Comment by sharkdp** [comment]
I'm not an expert on the difference between various ARM architectures, but it seems that the Pi Zero with its ARMv6 architecture is not compatible with the `arm-unknown-linux-gnueabihf` target triple, which is compatible with ARMv7 and upwards [1].

aarch64 is certainly not compatible as the RPi Zero has a 32bit architecture.

Cross compiling for ARMv6 seems nontrivial. [This project](https://github.com/rust-embedded/cross) provides Docker images, if you want to do this locally.

And you could always try to compile it on the Pi via `cargo install bat`.

[1] https://www.reddit.com/r/rust/comments/9io0z8/run_crosscompiled_code_on_rpi_0/e6l9hmc?utm_source=share&utm_medium=web2x

**Comment by sharkdp** [comment]
Yes, it is non-trivial :smile: 

Even with the cross-compilation Docker images, it doesn't work out of the box due to 
[bat-armv5.zip](https://github.com/sharkdp/bat/files/4536434/bat-armv5.zip)

#650.

We can create a custom docker which includes `libclang`:
``` Dockerfile
FROM rustembedded/cross:armv5te-unknown-linux-musleabi

RUN apt-get update && \
    apt-get install -y clang libclang-dev libc6-dev-i386
```

Next, build the Docker image:
```
docker build --tag arm-cc:v1 path/to/folder/with/dockerfile
```

and create a `Cross.toml` file with this content:
``` toml
[...]

PR #1552, opened 2021-02-24
CICD: Add aarch64-apple-darwin

### LINKED ISSUE #1551: Release binary for aarch64-apple-darwin
Awesome project :) Could you please release a binary for aarch64-apple-darwin (M1 Macs)?

I've compiled from source and that works well though would be handy to have a ready made binary for download

Enselic:
This is a Rust Tier 2 platform. We already build other Tier 2 targets,
for example arm-unknown-linux-gnueabihf.

The tracking issue for this target is here:
https://github.com/rust-lang/rust/issues/73908

closes #1551

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -95,6 +95,7 @@ jobs:
           - { os: ubuntu-18.04 , target: x86_64-unknown-linux-gnu    }
           - { os: ubuntu-18.04 , target: x86_64-unknown-linux-musl   , use-cross: true }
           - { os: macos-10.15  , target: x86_64-apple-darwin         }
+          - { os: macos-10.15  , target: aarch64-apple-darwin        , use-cross: true }
           # - { os: windows-2019 , target: i686-pc-windows-gnu         }  ## disabled; error: linker `i686-w64-mingw32-gcc` not found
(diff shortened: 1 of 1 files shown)


Enselic:
Build fails, so is not as easy as adding a line to our CICD config.

With some luck this might start working in the future as more work is put into the target by the Rust team and others.

Or maybe the CICD needs to be tweaked a bit bot make it work. (I did not analyze the build failure in depth).

Either way, setting needs-work label on this one for now.

sharkdp:
Looks like a C dependency fails to build (libz, which is built by the `libz-sys` crate). That dependency is pulled in through `git2 => libgit2-sys => libz-sys`. It's a mandatory dependency of `git2`, but we *could*, for now, build a `bat` release without git support for this target. That could work similar to what is done in https://github.com/sharkdp/bat/issues/952#issuecomment-633196350

Apart from that, @jpillora, did you do anything special when building `bat`? Did you build it natively on an M1 processor? Or via cross compiling?

Also, it would be great to report that build failure upstream in https://github.com/rust-lang/libz-sys

jpillora:
Nothing special, built natively from an M1

Enselic:
Closing this MR for now as I see no reason to keep it open

PR #1556, opened 2021-02-27
Do not ignore non-existent BAT_CONFIG_PATH

### LINKED ISSUE #1550: BAT_CONFIG_PATH ignored by --generate-config-file

Newest

image

use custom location instead

**Windows version:**

image

sharkdp:
Do not ignore `BAT_CONFIG_PATH` if it doesn't exist. Both when
generating a new config file with `--generate-config-file` and
when attempting to read the config.

Also, provide a better error message in case the file can not
be created.

closes #1550


sharkdp:
This will not (yet) print a warning if `BAT_CONFIG_PATH` does not exist. It will simply use an empty config. We might change that later, but this definitely fixes #1550 and is better than what we had before, where it would default to the system-defined config path (instead of an empty config).

Enselic:
Nice and clean fix! Only thing missing IMHO is a test that actually uses `--generate-config-file` to make sure the bug does not come back.

I created a PR for it here: https://github.com/sharkdp/bat/pull/1559

Issue #1400, opened 2020-11-25
JSON: attributes are not distinguished from strings


csabahenk:
**What version of `bat` are you using?**

0.17.1

**Describe the bug you encountered:**

The highlighted output for JSON looks like this:

![image](https://user-images.githubusercontent.com/43937/100288511-226b4080-2f77-11eb-9285-b245ad1609ac.png)

according to the https://github.com/sharkdp/bat/blob/v0.17.1/tests/syntax-tests/highlighted/JSON/test.json test data file.

I haven't run the test suite, but verified that the executable produces the same result:

```
~/rust/src/bat $ /tmp/bat-v0.17.1-x86_64-unknown-linux-gnu/bat tests/syntax-tests/source/JSON/test.json -p --color=always | md5sum - tests/syntax-tests/highlighted/JSON/test.json

7351ba203005a061be42e64995b9982c  -
7351ba203005a061be42e64995b9982c  tests/syntax-tests/highlighted/JSON/test.json
```

**What did you expect to happen instead?**

Something like we get from an alternative highlighter on the source of the test (https://github.com/sharkdp/bat/blob/v0.17.1/tests/syntax-tests/source/JSON/test.json):

![image](https://user-images.githubusercontent.com/43937/100289002-37949f00-2f78-11eb-8559-1b1326648e66.png)

Ie. attributes (keys) are distinguished from strings (scalar values). As these two are semantically different, I expect them to look different.

**How did you install `bat`?**

Using the binary from Github release v0.17.1

csabahenk:
Tthis might be deemed to be a duplicate of #1099; I filed a new issue because
- wanted to be more specific about my gripe with JSON than "doesn't look right"
- https://github.com/sharkdp/bat/issues/1099#issuecomment-667918309 seems to indicate that this issue is tacitly abandoded, even if kept open.

keith-hall:
Thanks for reporting. It may be related to https://github.com/sublimehq/Packages/pull/862, where some scopes were changed - maybe most color schemes aren't targeting these still?

sharkdp:
How can we move forward with this? This has been requested a few times and I also think it would be great if keys would be highlighted differently from string values.

I guess we would have to manually review the existing themes and propose upstream changes / patch the themes locally?

keith-hall:
> I guess we would have to manually review the existing themes and propose upstream changes / patch the themes locally?

Yep. As an experiment, I tried adding the following to `Monokai Extended.tmTheme`:

[...]

PR #1557, opened 2021-02-27
Improve Monokai Extended JSON key color with a patch

keith-hall:
Added a patch for the `Monokai Extended.tmTheme` color scheme, to differentiate the colors between a JSON [string] key and string value. Relates to #1400.


sharkdp:
Thank you!

I'll update the ChangeLog separately

PR #1559, opened 2021-02-28
integration_tests.rs: Add test config_location_when_generating

Enselic:
As a complement to #1556 to prevent #1550 from coming back.

The test purposefully fails if ca60937c2e is reverted.

### DIFF
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -662,6 +662,30 @@ fn config_location_test() {
 }
 
 #[test]
+fn config_location_when_generating() {
+    let tmp_config_path = std::path::PathBuf::from("/tmp/should-be-created.conf");
+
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you!

I think it would be cleaner to use tempfile to create a temporary *directory*, which we could safely use to test config file generation. This would make the prior existence check and the cleanup unnecessary. Also, hardcoding /tmp could be problematic on some platforms. Apparently it just works on Windows (probably because we create non-existent directories).

Enselic:
Good idea, I'll change it to create a temporary dir :+1:

sharkdp (approved):
Thanks!

Enselic:
Commit pushed. Significant improvement, thanks for the tip :+1:

PR #1571, opened 2021-03-02
Improve readability

matklad:
Using `Path`s for paths expresses intent more clearly.

Drive by change mostly, I just accidentally fell into non-utf8 arguments pit :)

### DIFF
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -327,7 +327,7 @@ mod tests {
                 writeln!(temp_file, "{}", first_line).unwrap();
             }
 
-            let input = Input::ordinary_file(file_path.as_os_str());
+            let input = Input::ordinary_file(&file_path);
             let dummy_stdin: &[u8] = &[];
             let mut opened_input = input.open(dummy_stdin, None).unwrap();
 
@@ -341,7 +341,7 @@ mod tests {
         fn syntax_for_file_with_content_os(&self, file_name: &OsStr, first_line: &str) -> String {
             let file_path = self.temp_dir.path().join(file_name);
             let input = Input::from_reader(Box::new(BufReader::new(first_line.as_bytes())))
-                .with_name(Some(file_path.as_os_str()));
+                .with_name(Some(&file_path));
             let dummy_stdin: &[u8] = &[];
             let mut opened_input = input.open(dummy_stdin, None).unwrap();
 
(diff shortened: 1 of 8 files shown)


src/assets.rs
```diff
@@ -341,7 +341,7 @@ mod tests {
         fn syntax_for_file_with_content_os(&self, file_name: &OsStr, first_line: &str) -> String {
             let file_path = self.temp_dir.path().join(file_name);
             let input = Input::from_reader(Box::new(BufReader::new(first_line.as_bytes())))
-                .with_name(Some(file_path.as_os_str()));
+                .with_name(Some(&file_path));
```
matklad, line 344: Note that here we don't need a conversion back from `Path`.

src/bin/bat/app.rs
```diff
@@ -248,16 +248,19 @@ impl App {
             }
             _ => {}
         }
-        let filenames: Option<Vec<&str>> = self
+        let filenames: Option<Vec<&Path>> = self
             .matches
-            .values_of("file-name")
-            .map(|values| values.collect());
+            .values_of_os("file-name")
```
matklad, line 253: This adds support for non-utf8 names

Enselic:
I haven't done a detailed review yet, but am in general positive to this change after skimming it through.

I'm curious what problem you saw with non-UTF-8 file names though? Works OK for me at least, using official build of `bat 0.18.0` on `x86_64-unknown-linux-gnu`:

```
% echo testing > $(printf "\x$(printf %x 130)")       
% bat $(printf "\x$(printf %x 130)") 
───────┬────────────────────────────────
       │ File: �
───────┼────────────────────────────────
   1   │ testing
───────┴────────────────────────────────
% bat --version                                                 
bat 0.18.0
```

matklad:
@Enselic you need to check with the `file-names` argument

sharkdp:

The changes look good. This is (at least technically) a breaking change, so it should be documented as such in the "`bat` as a library" section of the CHANGELOG.md. Changing `AsRef<OsStr>` to `AsRef<Path>` is unlikely to cause any problems (I guess), but the changes for [`Input::ordinary_file`](https://docs.rs/bat/0.18.0/bat/input/struct.Input.html#method.ordinary_file) and [`Input::with_name`](https://docs.rs/bat/0.18.0/bat/input/struct.Input.html#method.with_name) could lead to errors in client code. Maybe we could accept `AsRef<Path>` for these methods as well? That would still allow users to pass in `OsStr`s, I believe.

Enselic:
Thanks, with `--file-names` I can confirm that it crashes:

```
% bat --file-names $(printf "\x$(printf %x 130)")      
thread 'main' panicked at 'unexpected invalid UTF-8 code point', /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/clap-2.33.3/src/app/parser.rs:1683:33
```

matklad:
Yeah, makes sense, updated!

PR #1577, opened 2021-03-07
CICD: Remove explicit allow of clippy::match_bool since MSRV 1.45 bump

Enselic:
Now that we are on MSRV 1.45, there is no need to "backport" the change
that reclassified clippy::match_bool as "pedantic".


sharkdp (approved):
Ah, I forgot about that. Thanks!

Issue #1580, opened 2021-03-11
Add SystemVerilog syntax support


SeanMcLoughlin:
This is similar to #1535 but I'm opening a new bug because...
1. I have a PR I'm about to submit for this, so I can reference this issue in it
2. The PR is just for SystemVerilog and not VHDL

PR #1581, opened 2021-03-11
Add SystemVerilog syntax

SeanMcLoughlin:
See #1580.


.gitmodules
```diff
@@ -216,3 +216,6 @@
 [submodule "assets/syntaxes/02_Extra/gnuplot"]
 	path = assets/syntaxes/02_Extra/gnuplot
 	url = https://github.com/hesstobi/sublime_gnuplot
+[submodule "assets/syntaxes/02_Extra/SystemVerilog"]
+	path = assets/syntaxes/02_Extra/SystemVerilog
+	url = git@github.com:TheClams/SystemVerilog.git
```
sharkdp, line 221: All other submodules have a `https` link. Not sure if that could cause any problems, but might make sense to change it even if only for the sake of consistency.

tests/no_duplicate_extensions.rs
```diff
@@ -17,6 +17,9 @@ fn no_duplicate_extensions() {
         // The '.fs' extension appears in F# and GLSL.
         // We default to F#.
         "fs",
+        // SystemVerilog and Verilog both use .v files.
+        // We default to Verilog.
+        "v",
```
sharkdp, line 22: Could you please explain shortly what Verilog and SystemVerilog are, why we need syntaxes for both, and why a SystemVerilog syntax would still be valuable if `.v` files are highlighted as Verilog, by default.

sharkdp:
Thank you for you contribution.

tests/no_duplicate_extensions.rs
SeanMcLoughlin: Not sure if you mean to add the explanation in the comments of this file or here, but I'll add it here first:

[Verilog](https://en.wikipedia.org/wiki/Verilog) is a hardware description language used to describe circuits at the register transfer level. It's used in the semiconductor, ASIC, and FPGA industries.

[SystemVerilog](https://en.wikipedia.org/wiki/SystemVerilog) is a superset of Verilog, which provides many more features, including object-oriented programming, a random constraint solver, an assertion language, functional coverage constructs, and several others.

Unless required to for tool reasons, most developers these days will be using SystemVerilog instead of Verilog. Most of the logic in your CPU/GPU is most likely programmed in SystemVerilog.

The reason I have `.v` for SystemVerilog and Verilog is because that's what the Sublime Text submodule I added has them defined as, and I didn't see a way to get around that without updating the submodule. Personally, I have never seen a SystemVerilog file defined with `.v`, and the Wikipedia article I linked above seems to agree. If you know of a way to avoid that without updating the submodule, I can add it to this PR to avoid this extension conflict.

sharkdp: The explanation was only for me, thank you! :-)

I somehow thought `.v` was commonly used by both languages, but if there is `.sv` for SystemVerilog, your solution here is completely fine.

.gitmodules
g2p: I think it should be changed away from SSH, yes.

PR #1584, opened 2021-03-12
Add a system wide configuration file #668

akinnane:
This adds a system wide configuration file `/etc/bat/config` as requested in #668.

In the thread, there was discussion around the correct paths for this file on different platforms. For Unix I've used `/etc/bat/config`, however, I've not added a location for Windows as I'm not sure what the correct path would be. I'm happy to update this if someone can suggest locations for either platform.

There was also discussion about making the path configurable for package maintainers. I had a look in `build.rs` but it wasn't clear how to feed through into the binary. Can you provide guidance on how that would be done?

### DIFF
--- a/src/bin/bat/directories.rs
+++ b/src/bin/bat/directories.rs
@@ -9,6 +9,7 @@ use lazy_static::lazy_static;
 pub struct BatProjectDirs {
     cache_dir: PathBuf,
     config_dir: PathBuf,
+    system_config_dir: Option<PathBuf>,
 }
 
 impl BatProjectDirs {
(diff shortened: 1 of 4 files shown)


src/bin/bat/directories.rs
```diff
@@ -9,6 +9,7 @@ use lazy_static::lazy_static;
 pub struct BatProjectDirs {
     cache_dir: PathBuf,
     config_dir: PathBuf,
+    system_config_dir: Option<PathBuf>,
```
Enselic, line 12: Thanks for taking the time to make a contribution!

I understand this is temporary, but making a comment so we don't forget:

There shall always be a system_config_dir, so we should make it a non-`Option`

sharkdp:
I'm closing this for now due to inactivity. Please feel free to comment in case it should be re-opened.

PR #1591, opened 2021-03-18
Change TheClams/SystemVerilog submodule to use HTTPS

msnspk:
Using SSH to fetch the submodule was causing build failures for me, and as mentioned in [#1581](https://github.com/sharkdp/bat/pull/1581), this should be changed to HTTPS


keith-hall (approved):

sharkdp:
Thank you!

Issue #971, opened 2020-05-02
"ls" functionality when called with a directory


samuelcolvin:
_(I'm sorry if this is a duplicate, I can't imagine it's not, yet I've search a lot and can't find anything)_

It would be wonderful if bat behaved like `ls -lh ...` when executed with a directory as an arg.

This would be particularly useful when looking for a file to display; currently I often end up replacing `bat` with `ls -lh` at the beginning of a long path.

Obviously it would be useful if the output was highlighted unlike normal `ls`.

The main downside I can imagine would be confusion when "conbatenating" multiple files: if I accidentally included a directory instead of a file as one of the paths, however:
1. unlike `cat` I imagine `bat` is rarely used for this, especially in scripts
2. `bat` could continue to use a non-zero exit code when executed with a path to show it's not the primary intended purpose, while still showing the list of files - I'm not advocating this, just a potential work around

If this is something you'd be interested in, I'd be willing to have a crack at a PR, through I a somewhat novice rust developer.

eth-p:
This is a pretty cool idea, but I think it could also be solved in a few easier ways too:

1. If we add support for lesspipe, it could probably be configured to `ls` and highlight any directories passed as an input.

2. A wrapper script that combines `bat` and [exa](https://github.com/ogham/exa). Exa is already a well-established project that provides exactly what you're looking for, and it shouldn't be much harder to implement than adding this to your `.bash_profile`:

```bash
bat() {
    local arg
    local files=()
    local dirs=()

    for arg in "$@"; do
        if [ "${arg:0:1}" = "-" ]; then continue; fi
        if [ -d "$arg" ]; then
            dirs+=("$arg")
        elif [ -e "$arg" ]; then
            files+=("$arg")
        fi
    done

    if [ "${#files[@]}" -gt 0 ] && [ "${#dirs[@]}" -gt 0 ]; then
        printf "\x1B[31m[%s error]: %s\x1B[0m%s\n" "$0" "Cannot concatenate files and directories"
        return 1
    fi

    if [ "${#dirs[@]}" -gt 0 ]; then
        exa "${dirs[@]}"
        return $?
    fi

    bat "$@"
    return $?
}
```

> The main downside I can imagine would be confusion when "conbatenating" multiple files: if I accidentally included a directory instead of a file as one of the paths

I would just consider it an error, to be honest. There's such a low likelihood of it being intentional that it would make more sense to err on the side of caution and make it a clear error.

samuelcolvin:
[...]

Issue #748, opened 2019-11-10
Feature request: automatic decompression


wwood:
Hi,

I use `bat` as a nice replacement for `less` and one thing I see bat not doing is decompressing files that are compressed plain text e.g. gzip. For instance if I run `less a.gz` I see the contents of a.gz, while `bat a.gz` just tells me it is binary. Of course the workaround is just `zcat a.gz |bat` but I'm lazy.

I suppose this is related: https://github.com/wofr06/lesspipe

Thanks, ben

sharkdp:
please see #237 and #642

Issue #642, opened 2019-08-30
Handle compressed files


g2p:
This could be done by integrating with lesspipe / LESSOPEN at first.

sharkdp:
Please see #237

If you think there is a nice way to build a wrapper script that supports that, I'm happy to add it to the repository.

anarcat:
> If you think there is a nice way to build a wrapper script that supports that, I'm happy to add it to the repository.

i did https://github.com/sharkdp/bat/pull/3177

PR #1597, opened 2021-03-24
Add support for LESSOPEN and LESSCLOSE.

### LINKED ISSUE #237: zcat like functionality
It would be great if bat would detect if a file or stream is gzipped and would display it like zcat does.

eth-p:
This pull request adds opt-in support for preprocessors using the `LESSOPEN` and `LESSCLOSE` environment variables.

To enable the preprocessor, `--preprocessor=lessopen` must be set either be passed as a command line argument or set in the config file. This was made opt-in to avoid breaking default installations that provide an incompatible `LESSOPEN` program.

image

When used with something like `lesspipe.sh`, this can potentially fix #237, #642, #748, and #971.

## Implementation (Preprocessor Abstraction)

The preprocessor implementation is basically a `map` operation on `Input<'a> -> OpenedInput<'a>`.
It can either open the source input and return that directly (e.g. nothing to be done), or create a new `Input<'a>` and return that instead.

Preprocessors (or inputs like `InputKind::StdIn`) can attach "handles" to `OpenedInput<'a>` structs.
Handles can be used to perform cleanup actions (like deleting temporary files), or in the case of `InputKind::StdIn`, used to (`unsafe`ly) keep a reference to a `Stdin` object alive as long as its corresponding `InputReader`.

Note: handles will either be closed automatically when dropped (panicking on errors), or explicitly closed with `OpenedInput::close`.

## Implementation (Lessopen)

Using the new preprocessor abstraction, basic `LESSOPEN`/`LESSCLOSE` support was added to `bat`.
**Note:** This only supports file inputs.

1. If `LESSOPEN` is defined and invalid (as determined by `shell_words::split`), `bat` will silently do nothing.
  This doesn't follow the behavior of `less`, but it's more consistent with how `bat` handles the pager.

2. If `LESSOPEN` is defined but the process could not be spawned, `bat` will print a warning.
  For consistency with `less` (which simply ignores the exit code), this does not include the exit code of an unsuccessfully-exited command.

3. If `LESSOPEN` is defined and starts with `"|"`, `bat` will use its standard output as the preprocessed file contents.
  If no output is provided by the LESSOPEN function, `bat` will use the original file.

4.  If `LESSOPEN` is defined and does not start with `"|"`, `bat` will read the first line of output as the path to a temporary file containing the preprocessed file contents.

5. If any preprocessor command was executed and `LESSCLOSE` is defined, `bat` will execute `LESSCLOSE`.
  If the process could not be spawned, `bat` will print a warning.

## Caveats

- Git changes will be displayed with LESSOPEN, even if it outputs something different than the original file.
  This is intentional, and it's an unfortunate side-effect of the way that `lesspipe.sh` will always `cat` the file, even if it doesn't need to preprocess it.
  Without this, git changes would not be displayed at all.

- Due to some rewrites with the `input` module in order to add support for preprocessors, two instances of `unsafe` were added. 
  - A comment with safety information was included.

- It is not 100% compatible with `less`'s implementation.
  - This implementation tries to avoid invoking the user's shell, which limits the support to a shell-quoted list of arguments.
  - This implementation only replaces substitutes `%s` if it is an entire argument. Something like `LESSOPEN='| echo File:%s'` will not work.

- No tests were added. I couldn't think of a way to add tests without writing to the filesystem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Features
 - Add SystemVerilog file syntax, see #1580 (@SeanMcLoughlin)
+- Add `--preprocessor=lessopen` support, see #1597 (@eth-p) 
 
 ## Bugfixes
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -99,6 +99,7 @@ dependencies = [
  "globset",
  "lazy_static",
  "nix",
+ "os_str_bytes",
  "path_abs",
  "predicates",
  "semver",
@@ -686,6 +687,12 @@ dependencies = [
  "cc",
  "pkg-config",
 ]
+
+[[package]]
+name = "os_str_bytes"
+version = "3.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e293568965aea261bdf010db17df7030e3c9a275c415d51d6112f7cf9b7af012"
 
 [[package]]
 name = "parking_lot"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,9 +24,12 @@ application = [
     "paging",
     "wild",
     "regex-onig",
+    "preprocessor-lessopen",
 ]
 git = ["git2"] # Support indicating git modifications
 paging = ["shell-words"] # Support applying a pager on the output
+preprocessor = []
+preprocessor-lessopen = ["preprocessor", "os_str_bytes"] # Support LESSOPEN and LESSCLOSE
 
 # You need to use one of these if you depend on bat as a library:
 regex-onig = ["syntect/regex-onig"] # Use the "oniguruma" regex engine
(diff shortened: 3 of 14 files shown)


Enselic:
We currently have no `unsafe` code in the app code, only in tests. Personally, I would consider the addition of `unsafe` to the "real" code to be a dramatic change for the project. Please accept my apologies in advance, if you have already discussed this and agreed that starting to use `unsafe` in the app code is no big deal.

I'm tentatively marking as **needs-work** for now. (There are also some conflicts by now that needs to be sorted out.)

eth-p:
Oh man, it's been a while. Honestly, this was more of a proof-of-concept idea/prototype than anything. Looking back at the implementation, I agree that I shouldn't have used `unsafe` here and that we don't need to introduce any unsafe code unless strictly necessary.

I'm actually going to close this. If we end up with more interest in supporting LESSOPEN thought, I would be happy to look into rewriting it at a later date when I have more free time.

PR #1606, opened 2021-04-04
Improve Syslog and general log file highlighting

### LINKED ISSUE #1605: syslogs
If there's one thing I really need bat for, it's to query logs, quickly. But it doesn't automatically recognize the syslog syntax and if I enter it manually (-m=syslog=log) it's pretty poor.

I think it is something that by default should be well polished in a tool of this kind.

keith-hall:
Replace the Syslog dependency with a custom syntax definition and add a syntax definition for highlighting general log files. Fixes #1605

Note: https://github.com/sublimehq/Packages/issues/1036#issuecomment-308998132 isn't officially standardised yet, so we might need to update the color schemes we use to cater for the `markup.error` and `markup.warning` scopes, potentially just by covering them in the existing `invalid.illegal` and `invalid.deprecated` rules (which I've done for Monokai by updating the patch) ... But, the good news is that the LSP plugin for Sublime Text is making use of them already, so many actively maintained color schemes likely should already highlight them or will soon if they don't already. :)


eth-p (approved):

fort068fibl (approved):

sharkdp:
Thank you, this looks great!

PR #1612, opened 2021-04-15
CI: check docs

MarcoIeni:
working example: https://github.com/LukeMathWalker/wiremock-rs/pull/66

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -36,6 +36,13 @@ jobs:
       uses: actions-rs/cargo@v1
       with:
         command: test
+    - name: Check documentation
+      env:
+        RUSTDOCFLAGS: -D warnings
+      uses: actions-rs/cargo@v1
+      with:
+        command: doc
+        args: --no-deps --document-private-items --all-features
 
   test_with_new_syntaxes_and_themes:
     name: Run tests with updated syntaxes and themes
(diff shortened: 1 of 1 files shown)


Enselic:
Thanks for taking the time to make a contribution!

Seems to me as if it does not hurt to continuously make sure docs build.

I do think however that the new  'Check documentation' step should be part of the existing 'Minimum supported rust version' job, to prevent getting build errors that we "can't fix". I'd say the new step fits best below the existing 'Run tests' step in that job.

Can you move it there please?

MarcoIeni:
Done. I also added the `--all-features` flag.

Enselic:
Thanks for the update!

I've tried to confirm that the new command will fail when there are documentation problems, but so far couldn't get it to fail. I've tried adding unmatched markdown tags and invalid references, but it just chugs along.

Do you know what it takes to make your `cargo doc` command to fail when e.g. `cargo test` doesn't fail? If `cargo doc` never fails, I'm not sure it is worth having in CI.

MarcoIeni:
[This](https://github.com/MarcoIeni/bat/pull/1) fails by running `cargo doc --no-deps --document-private-items --all-features` locally:
image

The problem is that rust 1.45.0 doesn't have [intra-doc](https://doc.rust-lang.org/rustdoc/linking-to-items-by-name.html) support. So maybe it's a better idea to use the latest rust for `cargo doc`. What do you think?

By the way the github action is working, check [this](https://github.com/MarcoIeni/rust-docs-experiment/commit/af57c9d5bafc97bb003abb07b48075deaf363ff8) experiment, for example.

Enselic:
Indeed, I can confirm that your command fails when there are invalid doc references when using 1.48.0. It was when I used 1.45.0 that it didn't fail.

Given this new information, I agree it makes sense to use latest rust for the `cargo doc` check. I don't see how it could fail without it being easy to fix in a MSRV-compatible way. This is in contrast to lint (clippy) failures. Fixing them properly often requires API that is not available in MSRV versions.

So it would be great if you could move your new step to the 'Run tests with updated syntaxes and themes' job, below the 'List of themes' job, i.e. add it as the last job. Sorry for the back and forth, but I think we now have figured out the final solution :) 

(Another option is the matrix job, but I don't think it makes sense to run the documentation check 10 times.)

Enselic (approved):
Looks good to me now, thanks :+1: 

(I'll let @sharkdp have the final say on this PR, and do the merge if he thinks it looks ok.)

sharkdp:
Looks good to me! Thank you @MarcoIeni and @Enselic.

PR #1629, opened 2021-05-02
Fix less version parsing for minor versions of less

aswild:
Less 581.2 is here, and it has a ".2" in the version string, which can't
be parsed as a usize.

Update the check to find a non-digit character rather than a space. This
ignores the minor version, but parses the major version correctly.


eth-p (approved):
Looks good, thanks!

sharkdp:
Thank you!

Issue #1545, opened 2021-02-18
Auto detect language by shebang


xiaozhuai:
Consider a file named `luci` and its content

```lua
#!/usr/bin/lua
require "luci.cacheloader"
require "luci.sgi.cgi"
-- luci.dispatcher.indexcache = "/tmp/luci-indexcache"
luci.sgi.cgi.run()
```

![image](https://user-images.githubusercontent.com/4773701/108337153-c6c7f080-720f-11eb-88d0-a7615a3bb50d.png)

Since this file has no extension name like `.lua`, syntax highlighting stop working unless providing option `-l lua`

These files usually are executable scripts, and begining with [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)) line

It's possible to guess language by shebang. 

And also syntax highlighting should ignore shebang line

xu-cheng:
FYI, this actually can be implemented using the existing [libmagic](https://github.com/file/file/tree/master/magic), which contains a database to detect the file type based on its content (not just shebang).

sharkdp:
Thank you for reporting this.

We already have shebang support (e.g. for Bash scripts or Python scripts), it only needs to be supported by the specific syntax. This needs to be fixed upstream in sublimehq/Packages.

> FYI, this actually can be implemented using the existing [libmagic](https://github.com/file/file/tree/master/magic), which contains a database to detect the file type based on its content (not just shebang).

I'd rather not add an external dependency for this. libmagic is powerful for detecting all kinds of binary formats, but those don't really matter for `bat`, right?

xiaozhuai:
@sharkdp 
Thank you!
And one more thing, bat should treat shebang as comment and ignore syntax highlighting on it.


![image](https://user-images.githubusercontent.com/4773701/108594098-dcc1e680-73b2-11eb-81bd-fe1d1a9d99d3.png)

![image](https://user-images.githubusercontent.com/4773701/108594060-a6846700-73b2-11eb-8f85-722e971ee3e5.png)

sharkdp:
Closing this for now, see reasons given here and in https://github.com/sharkdp/bat/pull/1664#issuecomment-886187100. Focusing on #1600 for now.

Issue #1600, opened 2021-03-27
Allow generic 'first line match' across all syntaxes


Grazfather:
Suggested by @keith-hall in #160 

The idea is to read the syntax from the first line of input, like a shebang, emacs syntax hint, or vim modeline.

Ideally this wouldn't have to literally be on the first line, but for technical/performance reasons it probably must.

Freed-Wu:
Excuse me.
How can we define the map of shebang and syntax? I know `bat -m'*.tlu:Lua a.tlu` can map tlu extension to Lua. However can we map a shebang to a syntax? Like:

```
#! /usr/bin/env texlua
...
```

Then we use `bat -Mtexlua:Lua` can view the file with the highlight of Lua?

Thanks.

glyh:
actually vim modelines can occur at the last line of the file... And consider this common lisp & bash hybrid script, how would you highlight it?
```lisp
#!/bin/sh
#|-*- mode:lisp -*-|#
#|
exec ros -Q -- $0 "$@"
|#
(progn ;;init forms
  (ros:ensure-asdf)
  #+quicklisp(ql:quickload '() :silent t))
  

(defpackage :ros.script.test.3852975384
  (:use :cl))
(in-package :ros.script.test.3852975384)

(defun main (&rest argv)
  (declare (ignorable argv)))
;;; vim: set ft=lisp lisp:

```

Grazfather:
Similar:

```bash
#!/usr/bin/env bb
#_" -*- mode: clojure; -*-"
```

eggbean:
Please implement this!

mmahmoudian:
> actually vim modelines can occur at the last line of the file... And consider this common lisp & bash hybrid script, how would you highlight it?
> ```lisp
> #!/bin/sh
> #|-*- mode:lisp -*-|#
> #|
> exec ros -Q -- $0 "$@"
> |#
> (progn ;;init forms
>   (ros:ensure-asdf)
>   #+quicklisp(ql:quickload '() :silent t))
>   
> 
> (defpackage :ros.script.test.3852975384
>   (:use :cl))
> (in-package :ros.script.test.3852975384)
> 
> (defun main (&rest argv)
>   (declare (ignorable argv)))
> ;;; vim: set ft=lisp lisp:
> 
> ```

[...]

PR #1664, opened 2021-05-27
Add libmagic extension guessing

hingev:
There was a suggestion in #1545 to use libmagic to guess file types.

Co-authored-by: In-line <inline0@protonmail.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Features
 
+- Add file type guessing using libmagic (#1545)
 
 ## Bugfixes
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -98,6 +98,9 @@ dependencies = [
  "git2",
  "globset",
  "lazy_static",
+ "magic",
+ "mime",
+ "mime_guess",
  "nix",
  "path_abs",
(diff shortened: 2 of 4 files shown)


Enselic:
Thanks for taking the time to make a contribution, but did you see [this comment](https://github.com/sharkdp/bat/issues/1545#issuecomment-782604457) where @sharkdp says he does not think libmagic is a good fit for bat?

hingev:
Yes, that's why it is packaged as an optional feature

Enselic:
I’ll let @sharkdp have the final say in this PR, but would still like to ask: In what way does making it an optional feature make it a better fit for bat?

hingev:
I was looking at open feature-request issues, and came around #1545 
Since @sharkdp does not want libmagic, the official builds can still be compiled without it, while other people who need it can compile with the feature enabled.

sharkdp:
@hingev Thank you very much for your contribution. I'd rather not include this in `bat`, and focus on #1600. Even if this is included as an optional feature (I appreciate the effort), we will have to maintain this code in the future, if we merge this PR. Unless someone can demonstrate that this would have significant advantages over #1600, I'd rather not merge this.

Issue #1010, opened 2020-05-21
Bash CLI completion script?


flipperbw:
Unless I'm missing something obvious, is there any autocomplete script for bash? I see one for fish, but couldn't find anything for bash.

I can take a pass at converting that one or writing my own, but wanted to check in first.

sharkdp:
We used to have automatically created shell completion files for Bash, Zsh, Fish and PowerShell. They were causing some problems though:
https://github.com/sharkdp/bat/blob/7e844774085f3ecd10182daea9a0122ee008bb82/build.rs#L1-L2

We have since included manually created Fish completion files (https://github.com/sharkdp/bat/issues/372)

Apart from the reasons given in https://github.com/sharkdp/bat/pull/524#issuecomment-480018300 I am open to including manually created shell completion files.

As a starting point, you could either take the Fish completion file (which is probably already out of date) or re-enable the automatic generation of completion files and manually fix the autogenerated Bash completion file.

flipperbw:
Hmm gotcha. I will take a look.

flipperbw:
What I wound up doing, in lieu of making a whole script, was just to add the line `complete -F _longopt bat` to my profile. It's not perfect but it's a quick and easy fix for some completion without the need for an entire file.

ernstki:
@flipperbw That's a great idea, and not a thing I even knew existed. Now I can stop re-inventing the wheel of parsing the output of a command's `--help` into a word list for `complete -W`. <tt>:)</tt>

Worth noting, though, that that's [part of bash-completion](https://github.com/scop/bash-completion/blob/0c69125d65faaa60b0829db58c88dc33366bb67c/bash_completion#L1969), which is often a separate package for people to install / set up.

bretty-mcbrettface:
True, though I suppose it could be tested for in an install script here, and if not, include that function definition.

flipperbw:
True, though I suppose it could be tested for in an install script here, and if not, include that function definition.

sharkdp:
We now have custom Zsh and Fish completion files. It would be great if someone could write a Bash completion file as well.

ds2606:
Just an FYI, filename completion no longer works using zsh with the re-dded scripts. For example, with directory `~/Misc`, `bat ~/mi<tab>` doesn't expand to the full directory, whereas it does for me with any other completion executable (e.g. `ls ~/mi<tab>` expands to `ls ~/Misc`)

scop:
[...]

PR #1678, opened 2021-06-03
Add bash completion

scop:
Closes https://github.com/sharkdp/bat/issues/1010

This implementation is self contained and does not require https://github.com/scop/bash-completion. If a dependency on it would be ok, adding it would make it possible to clean this up some, and would gain easy support for completing long options separated by `=` (which this one doesn't currently, and I'd rather not duplicate the code for that). Let me know if that would be desirable and I can adjust this.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -268,6 +268,7 @@ jobs:
         cp "README.md" "LICENSE-MIT" "LICENSE-APACHE" "CHANGELOG.md" "$ARCHIVE_DIR"
 
         # Autocompletion files
+        cp 'target/${{ matrix.job.target }}/release/build/${{ env.PROJECT_NAME }}'-*/out/assets/completions/bat.bash "$ARCHIVE_DIR/autocomplete/${{ env.PROJECT_NAME }}.bash"
         cp 'target/${{ matrix.job.target }}/release/build/${{ env.PROJECT_NAME }}'-*/out/assets/completions/bat.fish "$ARCHIVE_DIR/autocomplete/${{ env.PROJECT_NAME }}.fish"
         cp 'target/${{ matrix.job.target }}/release/build/${{ env.PROJECT_NAME }}'-*/out/assets/completions/bat.zsh "$ARCHIVE_DIR/autocomplete/${{ env.PROJECT_NAME }}.zsh"
 
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 **/*.rs.bk
 
 # Generated files
+/assets/completions/bat.bash
 /assets/completions/bat.fish
 /assets/completions/bat.zsh
 /assets/manual/bat.1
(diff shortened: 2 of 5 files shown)


sharkdp:

I am

eth-p:
@sharkdp I'm not too familiar with Bash completions, but I'll give it a look. I'll need to give it a proper review later, but this seems good from a quick glance.

> This implementation is self contained and does not require https://github.com/scop/bash-completion. If a dependency on it would be ok, adding it would make it possible to clean this up some, and would gain easy support for completing long options separated by `=` (which this one doesn't currently, and I'd rather not duplicate the code for that). Let me know if that would be desirable and I can adjust this.

I would actually say this is one of those cases where it's fine to add an external dependency. From my understanding, a ton of other programs use `bash-completion` for their completions as well, and there's probably a good chance it will be installed already.

In the cases where it isn't (e.g. Mac), it shouldn't be too hard for package managers to simply add a dependency.

@shark, thoughts?

scop:
FWIW I do support adding the dependency, but I may also be quite heavily biased. In any case, the changes that I'd make if the dependency would be desirable are quite small and easy to review, so if the decision is not immediately clear, starting to review this in its current shape wouldn't be throwaway work if the dep+changes were introuced later.

sharkdp:
Adding that dependency sounds good to me if it helps with the implementation.

scop:
Ok, implementation adjusted to depend on it.

sharkdp:
Thank you very much. This looks good to me.

I think this would be worth an entry in the CHANGELOG.md file. It would be great if you could add one (https://github.com/sharkdp/bat/blob/master/CONTRIBUTING.md#add-an-entry-to-the-changelog), otherwise I can also do it myself.

scop:
Sure, added to the Other section.

ds2606:
Just commented in #1010 that completions seem to re-surface the original issues for me (generic fliename completions are failing in zsh for bat) so I'll need to manually remove the completion scripts for the time being. Could we perhaps add a bat command to disable completion files?

scop:
For completeness, noting here too: as refined and discussed in #1010, this particular completion does not cause the above mentioned breakage to occur in bash.

sharkdp (approved):
Thank you!

eggbean:
Not a big problem, but curioius to know why I cannot get this to work. 

I added `bat.bash.in` to `etc/bash_completion.d/` (actually a `stow` symlink, as that is how I install bat) and sourced `~/.profile` but I get no indication anything has changed. This way works for several other programs I install using `stow`, so I suspect the unusual `.in` suffix has something to do with it?

sharkdp:
> I added `bat.bash.in` to `etc/bash_completion.d/`

Ah, that won't work. `bat.bash.in` is just a template for a bash completion file (hence the `.in` suffix). The actual `bat.bash` completion file will be generated when you build the project with `cargo`. Alternatively, download one of the tarballs / Debian packages from the release page. They also contain the generated files.

sharkdp:
Oh - they're actually not part of the Debian packages (but the tarballs). That should be fixed in `.github/workflows/CICD.yml `.

michaelblyons:
Is this also missing from Homebrew?

Issue #1666, opened 2021-05-28
Document that pagers `most` and `more` are not used


pgimalac:
As mentioned in https://github.com/sharkdp/bat/issues/1063, pagers `most` and `more` are not used by `bat` for compatibility reasons, silently falling back to `less` (when set with `PAGER` environment variable), but this is not documented anywhere but the release description https://github.com/sharkdp/bat/releases/tag/v0.18.0.

I believe it should at least be explained why in the README.

Enselic:
@pgimalac thanks for taking the time to report.

Is https://github.com/sharkdp/bat/blob/f366bbe34941f1c94851b4a91eef90b2dc3bc3ba/README.md#using-a-different-pager good enough from your perspective? If not, what would you change?

pgimalac:
@Enselic thanks to you for taking the time to document this :)

I let you judge but perhaps referencing the issue which led to this choice (https://github.com/sharkdp/bat/issues/1063) so people can easily find the details ? Otherwise I think it is good enough !

Enselic:
Thanks for the feedback!

I do think however that the README.md would become too noisy if we would start adding references, so let's keep it as is.

pgimalac:
@Enselic that's exactly what I was worried about, and why I said that I let you judge.
Thanks for taking the time to document this.

PR #1680, opened 2021-06-05
README.md: Mention that PAGER is ignored if set to more or most

Enselic:
Quick and dirty fix for #1666. I think the whole "Using a different pager" section would probably benefit from some restructuring and cleanup, but with this PR, we at least have _something_ in the readme that informs the user of PAGER + most/more, which decidedly is better than _nothing_.


sharkdp (approved):
Thank you!

PR #1683, opened 2021-06-10
Highlight mtab files with fstab highlighting

kidonng:
This is mentioned in https://github.com/sharkdp/bat/pull/696#issuecomment-544692236, but somehow not implemented.

### DIFF
--- a/assets/syntaxes/02_Extra/Fstab.sublime-syntax
+++ b/assets/syntaxes/02_Extra/Fstab.sublime-syntax
@@ -5,6 +5,7 @@ name: fstab
 file_extensions:
   - fstab
   - crypttab
+  - mtab
 scope: source.fstab
 
 contexts:


sharkdp:

I wasn't sure if it really was the same syntax. Do you have any information on that?

kidonng:
Quoting: https://github.com/karelzak/util-linux/blob/fb9ed3b409c05752a06619b2a5f02f5a21b3bb62/sys-utils/mount.8.adoc#the-files-etcfstab-etcmtab-and-procmounts:

> The programs mount and umount(8) traditionally maintained a list of currently mounted filesystems in the file /etc/mtab.

The difference is basically that `fstab` is managed by the user and `mtab` by the system.

sharkdp (approved):
Thank you!

PR #1685, opened 2021-06-13
Add groff syntax

scop:
The syntax is named "Man Page" upstream, but our man page syntax is different, it's for rendered man pages. Rename to Groff and remove `.man` from extensions.

I noticed the requirement of >= 10k downloads only after creating this; I personally find it quite useful and hope it can be included even though it has only 1k at the moment.

I've [upstreamed](https://github.com/carsonoid/sublime_man_page_support/pull/6) the addition of .8 and .9.


sharkdp:
Thank you for your contribution. I'm going to merge this as you did prepare this very carefully including all tests - thanks!

Let's try to stick to the 10k downloads requirement in the future (or discuss alternative policies, see #1262)

PR #1702, opened 2021-06-23
Manual and Zsh completion updates

Frederick888:
- Document --generate-config-file in manual
- Fix missing config/cache arguments in Zsh completion

---

Not sure if changelog entries are need?


Frederick888:
CI failure doesn't seem to be related to my changes?

Enselic:
> CI failure doesn't seem to be related to my changes?

Thanks for the heads-up. You are correct. This appears to be caused by a new `cargo doc` check introduced in the recently released Rust 1.53.0 that our CI job picks up.

I created a PR with a fix proposal that I will merge with some urgency: https://github.com/sharkdp/bat/pull/1707

Enselic:
@Frederick888 Fix is merged, so I propose that you rebase your PR to get a green CI build

Frederick888:
@Enselic Done :)

sharkdp (approved):

scop:
I see the bash completion is missing some as well. But I'm wondering if we should have completions offer options that are marked as hidden? If they're hidden for a reason, maybe they should be hidden from completions (in the sense that not being offered in them), too?

sharkdp:
The following options are *completely* hidden from all help output (short and long):
* `--no-paging`/`-P`, an alias for `--paging=never`. This option is hidden because it's mainly there due to it's short form `-P`, which is mentioned in the help text for `--paging`. I think `-P` could be included in the completion, but single-char flags are probably not very interesting for completion anyways.
* `--no-config`. This option mainly exists for integration testing. I don't think it hurts to have it in the autocompletion, but I guess it's very unlikely to be used in an interactive shell.
* `--config-file`, `--config-dir`, `--cache-dir` and `--generate-config-file`. These are mentioned in the documentation, but not in the `-h`/`--help` text. I don't really remember why. I think they could also be included in the long `--help` text. And in completions as well.

PR #1707, opened 2021-06-28
pretty_printer.rs: Don't use a URL as an example Input title

Enselic:
Otherwise Rust 1.53.0 gets confused during `cargo doc` because it thinks
we want an actual URL:

    warning: this URL is not a hyperlink
    --> src/pretty_printer.rs:331:40
        |
    331 |     /// The title for the input (e.g. "http://example.com/example.txt")
        |                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use an automatic link instead: `<http://example.com/example.txt>`
        |
        = note: `#[warn(rustdoc::bare_urls)]` on by default
        = note: bare URLs are not automatically turned into clickable links

It was perhaps also a bit confusing to give an URL as an example in the
first place, because according to our own API example
`examples/inputs.rs` it is meant to be more a free-text thing.

### DIFF
--- a/src/pretty_printer.rs
+++ b/src/pretty_printer.rs
@@ -328,7 +328,7 @@ impl<'a> Input<'a> {
         self
     }
 
-    /// The title for the input (e.g. "http://example.com/example.txt")
+    /// The title for the input (e.g. "Descriptive title")
     /// This defaults to the file name.
     pub fn title(mut self, title: impl Into<String>) -> Self {
         self.input.description_mut().set_title(Some(title.into()));


Enselic:
@eth-p This should be fine, right? I will merge this pretty soon since our CI currently is broken, but just wanted to quickly ping you since I am changing the docs you added in https://github.com/sharkdp/bat/commit/798b742617fe789647a34611716d60e5ece9ebf8#diff-aa44ca43c1509f18000ec6c02402b7d89332415e6531f845739acacc7992cf0cR355

sharkdp (approved):

PR #1724, opened 2021-07-12
Fix for Windows: do not run binaries from CWD

sharkdp:
This fixes a bug on Windows where `Command::new` would also run
executables from the current working directory, possibly resulting in
accidental runs of programs called `less`.


sharkdp:
Archive with a build from this branch:
[bat-v0.18.1-x86_64-pc-windows-gnu.zip.zip](https://github.com/sharkdp/bat/files/6803893/bat-v0.18.1-x86_64-pc-windows-gnu.zip.zip)

sharkdp:
[bat-v0.18.1-x86_64-pc-windows-gnu.zip.zip](https://github.com/sharkdp/bat/files/6804206/bat-v0.18.1-x86_64-pc-windows-gnu.zip.zip)

sharkdp:
Thanks a lot to @BurntSushi for providing this fix as a re-usable function in `grep-cli` (https://github.com/BurntSushi/ripgrep/commit/229d1a8d41b0023420e7815578fa0b39c0d5c2e4)!

Enselic:
For future reference, here is the PR that will fix this in the Rust std lib: https://github.com/rust-lang/rust/pull/87704

ChrisDenton:
I would however note that the std has to be cautious about (potentially) breaking compatibility with prior versions of Rust. This means that even if that PR gets merged, it will still do more than just search `PATH`. I would love to further simplify it in the future but in the meantime it may still be worth it for applications to use a third party resolver (such as `grep_cli`).

PR #1729, opened 2021-07-13
De-duplicate some themes.bin and syntaxes.bin related code

Enselic:
When prototyping for https://github.com/sharkdp/bat/issues/951 I play around with introducing a new asset called `lookup.bin`. I bumped into some inconvenient code duplication that I de-duplicate in this PR. I think it is a sensible PR regardless of how the prototyping falls out.

I might appear to have placed the helper functions in odd places, but it is to make the diff as easy to review as possible. We can always move them around once the initial code review is done, in case you find the current placement of them unacceptable. FWIW, I personally can live with their current placement.

The change should be low-risk since it tries to make no behavioural changes. I still plan to do a thorough verification of the changes after the first code review round though.

### DIFF
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -97,30 +97,9 @@ impl HighlightingAssets {
     }
 
     pub fn from_cache(cache_path: &Path) -> Result<Self> {
-        let syntax_set_path = cache_path.join("syntaxes.bin");
-        let theme_set_path = cache_path.join("themes.bin");
-
-        let syntax_set_file = File::open(&syntax_set_path).chain_err(|| {
-            format!(
-                "Could not load cached syntax set '{}'",
-                syntax_set_path.to_string_lossy()
-            )
-        })?;
-        let syntax_set: SyntaxSet = from_reader(BufReader::new(syntax_set_file))
-            .chain_err(|| "Could not parse cached syntax set")?;
-
-        let theme_set_file = File::open(&theme_set_path).chain_err(|| {
-            format!(
-                "Could not load cached theme set '{}'",
-                theme_set_path.to_string_lossy()
-            )
-        })?;
-        let theme_set: ThemeSet = from_reader(BufReader::new(theme_set_file))
-            .chain_err(|| "Could not parse cached theme set")?;
-
         Ok(HighlightingAssets {
-            syntax_set,
-            theme_set,
+            syntax_set: asset_from_cache(&cache_path.join("syntaxes.bin"), "syntax set")?,
+            theme_set: asset_from_cache(&cache_path.join("themes.bin"), "theme set")?,
             fallback_theme: None,
         })
     }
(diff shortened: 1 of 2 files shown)


src/assets.rs
```diff
@@ -146,32 +125,12 @@ impl HighlightingAssets {
...
-        );
-        dump_to_file(&self.syntax_set, &syntax_set_path).chain_err(|| {
-            format!(
-                "Could not save syntax set to {}",
-                syntax_set_path.to_string_lossy()
-            )
-        })?;
-        println!("okay");
+        asset_to_cache(&self.theme_set, &target_dir.join("themes.bin"), "theme set")?;
+        asset_to_cache(
```
Enselic, line 129: sometimes `cargo fmt` yields annoying inconcistencies, like here :grimacing: 

oh well...

Ordiifranko (approved):

sharkdp (approved):
This looks good - thank you for the refactoring!

I think the functions belong in the respective modules. I'm happy with the placement.

keith-hall (approved):

Enselic:
Thanks for the code reviews!

I have verified this now by comparing output before and after my change with these commands:
```bash
# zsh

# Trigger "Clearing ..."
mkdir -p /tmp/assets
ls -la /tmp/assets
cargo run -- cache --build --blank --source assets --target /tmp/assets
ls -la /tmp/assets
BAT_CACHE_PATH=/tmp/assets cargo run -- cache --clear
ls -la /tmp/assets

# Trigger "Could not parse  ..."
cargo run -- cache --build --blank --source assets --target /tmp/assets
echo garbage >! /tmp/assets/themes.bin
echo garbage >! /tmp/assets/syntaxes.bin
ls -la /tmp/assets
BAT_CACHE_PATH=/tmp/assets cargo run -- -A tests/examples/test.txt
ls -la /tmp/assets
BAT_CACHE_PATH=/tmp/assets cargo run -- cache --clear
ls -la /tmp/assets

# Trigger "Could not save ..."
cargo run -- cache --build --blank --source assets --target /no-permission

# Trigger "Could not load  ..."
BAT_CACHE_PATH=/gone cargo run -- -A tests/examples/test.txt
```

I had to have a small patch in both runs to actually get the error messages to show:

```diff
% git diff
diff --git a/src/bin/bat/assets.rs b/src/bin/bat/assets.rs
index ef6484a..b6be48d 100644
--- a/src/bin/bat/assets.rs
+++ b/src/bin/bat/assets.rs
@@ -54,5 +54,5 @@ pub fn assets_from_cache_or_binary() -> Result<HighlightingAssets> {
     }
 
     Ok(HighlightingAssets::from_cache(&cache_dir)
-        .unwrap_or_else(|_| HighlightingAssets::from_binary()))
+        .unwrap())
 }
```

The output was identical (except for timestamps and line numbers) so I feel confident I didn't break anything with this refactoring.

Merging 🎉

Issue #1733, opened 2021-07-14
Add bash completion files to Debian packages


sharkdp:
Follow the same route as for the fish/zsh completions in `.github/workflows/CICD.yml`.

PR #1734, opened 2021-07-15
Include bash completion in deb package

scop:
Get correct path to install in from bash-completion using
`pkg-config`, install as `steps.strip.outputs.BIN_NAME` as the filename
needs to match the called executable for bash-completion's on demand
loading to work.
    
This `BIN_NAME` needs to stay in sync with `env.PROJECT_EXECUTABLE`
which is replaced in the template, but not available for the deb build
at this stage.

Closes https://github.com/sharkdp/bat/issues/1733

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -348,6 +348,7 @@ jobs:
         gzip -n --best "${DPKG_DIR}/usr/share/man/man1/${{ env.PROJECT_NAME }}.1"
 
         # Autocompletion files
+        install -Dm644 'target/${{ matrix.job.target }}/release/build/${{ env.PROJECT_NAME }}'-*/out/assets/completions/bat.bash "${DPKG_DIR}/usr/share/bash-completion/completions/${{ steps.strip.outputs.BIN_NAME }}"
         install -Dm644 'target/${{ matrix.job.target }}/release/build/${{ env.PROJECT_NAME }}'-*/out/assets/completions/bat.fish "${DPKG_DIR}/usr/share/fish/vendor_completions.d/${{ env.PROJECT_NAME }}.fish"
         install -Dm644 'target/${{ matrix.job.target }}/release/build/${{ env.PROJECT_NAME }}'-*/out/assets/completions/bat.zsh "${DPKG_DIR}/usr/share/zsh/vendor-completions/_${{ env.PROJECT_NAME }}"
 


sharkdp:
Thank you, this looks good.

The only thing I'm not sure about is the `apt` calls (at these specific places). Maybe @Enselic could have a short look?

Enselic:
I must admit I'm a bit confused with the change.

Shouldn't you make it so that the generated .deb `Depends` on `bash-completion`, rather than installing `bash-completion` in the temporary CI VM/image?

And why would it not work to hardcode the installation path for the script like it's done for fish and zsh?

sharkdp:
> Shouldn't you make it so that the generated .deb `Depends` on `bash-completion`

maybe yes, but it would be good to model it as an optional dependency, if that's possible in Debian.

> rather than installing `bash-completion` in the temporary CI VM/image?

It looks like this is used to call `pkg-config --variable=completionsdir bash-completion`, which is (presumably) a portable way to get the bash completions directory. Since we are explicitly building a Debian/Ubuntu package on a fixed Ubuntu image here, I also think it might be okay to hard-code the path. What are your thoughts @scop?

scop:
With my bash-completion upstream hat on: querying pkg-config is the official, supported way to find the completions dir. But then again there are no plans to change it upstream, and I don't see why Debian and Ubuntu would start to differ, so hardcoding a path is likely to work as well. And as the deb is targeted to different distros, the build one is essentially making the decision of the location for the others too from its pkg-config, and in that sense it's not that different than hardcoding it for all.

On the other hand, I would strongly recommend against adding a dependency on bash-completion. It is by no means an essential thing for bat users, is bloat for smaller installations, and not everyone likes the behavioral changes caused by just installing it. I'm not sure if there's a written policy on this, but I'm fairly certain that Debian maintainers would consider a hard dependency a packaging bug, and I don't remember even Recommends/Suggests being used. @inconstante do you remember a Debian policy or something reference related to this?

sharkdp:
> With my bash-completion upstream hat on: querying pkg-config is the official, supported way to find the completions dir. But then again there are no plans to change it upstream, and I don't see why Debian and Ubuntu would start to differ, so hardcoding a path is likely to work as well. And as the deb is targeted to different distros, the build one is essentially making the decision of the location for the others too from its pkg-config, and in that sense it's not that different than hardcoding it for all.

Your solution is definitely the cleanest. I see two arguments for simply hardcoding the path:

- simplicity
- CI time

I am not sure how much, but I could imagine that a `apt update` and `apt install` add a significant amount of time (30 seconds?).

> On the other hand, I would strongly recommend against adding a dependency on bash-completion. It is by no means an essential thing for bat users, is bloat for smaller installations, and not everyone likes the behavioral changes caused by just installing it. I'm not sure if there's a written policy on this, but I'm fairly certain that Debian maintainers would consider a hard dependency a packaging bug, and I don't remember even Recommends/Suggests being used. @inconstante do you remember a Debian policy or something reference related to this?

Thank you for the explanation!

kuanyili:
For simplicity and consistency between the handling of different shells, I'm also for hardcoding the path and drop `apt` calls.

kuanyili:
After over a month of inactivity, I feel like creating another pull request.... What do you guys think?

Will go ahead and create one if nobody has any thoughts on how to move forward.

scop:
Nah, I kind of missed that the decision was in favor of the hardcoding and I was supposed to update this PR, but done now, switched to hardcoded path.

kuanyili:
Another pull request was already created 2 days ago, and to keep things consist, `${{ env.PROJECT_NAME }}` is used like the other shell completions. See #1848

I'll leave it to the author to decide which implementation is preferred then.

scop:
`env.PROJECT_NAME` is wrong though, for reasons specified in the commit message here, so it's not the right thing to be consistent about.

I don't know how it is for other shells, but I know very well that the name of the completion file needs to adhere to the rules how bash-completion derives it from the executable being completed to be automatically loaded, and the best thing we have available at the moment is `steps.strip.outputs.BIN_NAME`. IIRC `env.PROJECT_EXECUTABLE` would kind be the correct variable for this, but we don't have that available in CI at this point.

Of course if all those variables point to the same thing and always will, then it doesn't matter. But in that case we could just as well hardcode `bat` everywhere.

kuanyili:
I get your point and also think that `env.PROJECT_NAME` is not exactly appropriate.

But I would rather choose one the the following ways

1. use `steps.strip.outputs.BIN_NAME` everywhere
2. use `env.PROJECT_NAME` everywhere
3. hardcode `bat` everywhere

than using `steps.strip.outputs.BIN_NAME` for bash only.

Enselic:
Your code is identical except for `steps.strip.outputs.BIN_NAME` vs `env.PROJECT_NAME`.

Even though `steps.strip.outputs.BIN_NAME` might be more correct in theory, I do think consistency with surrounding code is more important. I admit it is in many ways an arbitrary opinion.

In any case, I vote for merging https://github.com/sharkdp/bat/pull/1848 over this PR. Any objections?

scop:
> I vote for merging #1848 rather than this PR. Any objections?

FWIW mine, for reasons stated. Guess that was kind of implicit though :)

But if that one is merged instead of this, the other one's commit message could borrow some of the elaboration about the pitfalls it knowingly introduces from the message we have here.

kuanyili:
Fair enough. I've updated my commit message in #1848 to elaborate the pitfalls.

IMHO, adding bash completion to deb package and switching from `env.PROJECT_NAME` to `steps.strip.outputs.BIN_NAME` are two things and shall be performed in different PRs. In a separate PR, implementation for every shell completion can be replaced altogether.

Enselic:
I went ahead and merged https://github.com/sharkdp/bat/pull/1848 now, so let's close this PR.

Thanks to you all for your efforts to make bat better.

PR #1736, opened 2021-07-18
HighlightingAssets: Encapsulate syntax_set behind a getter

Enselic:
Since we only modify `pub(crate)` items, the stable bat-as-a-library API is not
affected.

This takes us one step closer to making SyntaxSet lazy-loaded, which in turn
takes us one step closer to solving #951. This change has proven useful in the prototype I made to improve startup time, see https://github.com/sharkdp/bat/issues/951#issuecomment-882098532

By my estimation, it will always be convenient to have the full SyntaxSet available, for things such as `--list-languages`. So regardless of the fate of the prototype in #951, I think this will end up being a sensible change. Because we only want to load the full syntax set when necessary (since it is slow).

### DIFF
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -18,7 +18,7 @@ use crate::syntax_mapping::{MappingTarget, SyntaxMapping};
 
 #[derive(Debug)]
 pub struct HighlightingAssets {
-    pub(crate) syntax_set: SyntaxSet,
+    syntax_set: SyntaxSet,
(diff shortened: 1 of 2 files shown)


keith-hall (approved):

Enselic:
Thanks for the code review! Let's merge this then :muscle: (more PR:s coming up)

sharkdp:
Thanks!

PR #1737, opened 2021-07-19
HighlightingAssets: Introduce private fn new() helper

Enselic:
It already now reduces code duplication slightly, but will become even more
useful in the future when we add more complicated logic such as lazy-loading. See e.g. https://github.com/sharkdp/bat/issues/951#issuecomment-882098532.

### DIFF
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -41,6 +41,14 @@ const IGNORED_SUFFIXES: [&str; 10] = [
 ];
 
 impl HighlightingAssets {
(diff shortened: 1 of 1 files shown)


keith-hall (approved):

Enselic:
Thanks again for reviewing. Merging.

Next PR will be about lazy-loading existing `syntaxes.bin`, which is actually enough to speed up startup in some cases. I have that working in my prototype, but it will take some time to get that code production ready.

PR #1747, opened 2021-07-22
Reduce startup time in loop-through mode by 80%-90%

Enselic:
`HighlightingAssets::get_syntax_set()` is never called when e.g. piping the `bat` output to a file (see `Config::loop_through`), so by loading the `SyntaxSet` only when needed, we radically improve startup time when it is not needed.

On my low-end machine, I get the following numbers when doing
`time bat tests/examples/multiline.txt > /tmp/output.txt`:

**git master**
```
0,09s user 0,00s system 93% cpu 0,103 total
```

**git master + this PR**
```
0,01s user 0,00s system 63% cpu 0,026 total
```

We can also benchmark with `hyperfine` , in which case we don't need to redirect the output manually, because the internal redirection of `hyperfine` is enough to trigger loop-through mode:
`hyperfine --export-markdown /dev/tty 'bat tests/examples/multiline.txt'`

**git master**
|:---|---:|---:|---:|---:|
| `bat tests/examples/multiline.txt` | 96.1 ± 2.2 | 93.3 | 104.3 | 1.00 |

**git master + this PR**
|:---|---:|---:|---:|---:|
| `bat tests/examples/multiline.txt` | 11.2 ± 1.1 | 9.7 | 15.9 | 1.00 |

So on my system, the speedup amounts to (1-11.2/96.1)*100 ~ 88%.

Lazy-loading was one of the key aspects of my [prototype](https://github.com/sharkdp/bat/issues/951#issuecomment-882098532) on how to improve the startup speed of bat, and this PR is one step on the journey to improve startup speed in general.

I have done sanity checking of this change, and all regression tests pass, so this PR should not be far off production-quality code. But don't hesitate to give criticism when doing code review :) And if you think my approach is completely off track, I would gladly like to know!

I also want to clarify that I think this is a good move long term, even if we start to load partial syntax sets for improved  performance. The ability to load the full syntax set is still useful, both for fallback purposes, and for implementing things like `—list-languages`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -101,6 +101,7 @@ dependencies = [
  "globset",
  "grep-cli",
  "lazy_static",
+ "lazycell",
  "nix",
  "path_abs",
  "predicates",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ ansi_term = "^0.12.1"
 ansi_colours = "^1.0"
 console = "0.14.1"
 lazy_static = { version = "1.4", optional = true }
+lazycell = "1.0"
 wild = { version = "2.0", optional = true }
 content_inspector = "0.2.4"
 encoding = "0.2"
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -54,7 +54,7 @@ impl HighlightingAssets {
 
     pub fn from_files(source_dir: &Path, include_integrated_assets: bool) -> Result<Self> {
         let mut theme_set = if include_integrated_assets {
-            Self::get_integrated_themeset()
+            get_integrated_themeset()
         } else {
             ThemeSet {
                 themes: BTreeMap::new(),
@@ -83,7 +83,7 @@ impl HighlightingAssets {
             builder.add_plain_text_syntax();
             builder
         } else {
-            Self::get_integrated_syntaxset().into_builder()
+            get_integrated_syntaxset().into_builder()
         };
 
         let syntax_dir = source_dir.join("syntaxes");
@@ -109,19 +109,8 @@ impl HighlightingAssets {
         ))
     }
 
-    fn get_integrated_syntaxset() -> SyntaxSet {
-        from_binary(include_bytes!("../assets/syntaxes.bin"))
-    }
-
-    fn get_integrated_themeset() -> ThemeSet {
-        from_binary(include_bytes!("../assets/themes.bin"))
-    }
-
     pub fn from_binary() -> Self {
-        HighlightingAssets::new(
-            Self::get_integrated_syntaxset(),
-            Self::get_integrated_themeset(),
-        )
+        HighlightingAssets::new(get_integrated_syntaxset(), get_integrated_themeset())
     }
 
     pub fn save_to_cache(&self, target_dir: &Path, current_version: &str) -> Result<()> {
@@ -325,6 +314,14 @@ impl HighlightingAssets {
             .ok()
             .and_then(|l| syntax_set.find_syntax_by_first_line(&l)))
     }
+}
+
+fn get_integrated_syntaxset() -> SyntaxSet {
+    from_binary(include_bytes!("../assets/syntaxes.bin"))
+}
+
+fn get_integrated_themeset() -> ThemeSet {
+    from_binary(include_bytes!("../assets/themes.bin"))
 }
 
 fn asset_to_cache<T: serde::Serialize>(asset: &T, path: &Path, description: &str) -> Result<()> {
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -1,7 +1,9 @@
 use std::collections::BTreeMap;
 use std::ffi::OsStr;
 use std::fs;
-use std::path::Path;
+use std::path::{Path, PathBuf};
+
+use lazycell::LazyCell;
 
 use syntect::dumps::{dump_to_file, from_binary, from_reader};
 use syntect::highlighting::{Theme, ThemeSet};
@@ -17,7 +19,8 @@ use crate::syntax_mapping::{MappingTarget, SyntaxMapping};
 
 #[derive(Debug)]
 pub struct HighlightingAssets {
-    syntax_set: SyntaxSet,
+    syntax_set_cell: LazyCell<SyntaxSet>,
+    serialized_syntax_set: Option<SerializedSyntaxSet>,
     pub(crate) theme_set: ThemeSet,
     fallback_theme: Option<&'static str>,
 }
@@ -40,9 +43,21 @@ const IGNORED_SUFFIXES: [&str; 10] = [
 ];
 
 impl HighlightingAssets {
-    fn new(syntax_set: SyntaxSet, theme_set: ThemeSet) -> Self {
+    fn new(
+        syntax_set: Option<SyntaxSet>,
+        serialized_syntax_set: Option<SerializedSyntaxSet>,
+        theme_set: ThemeSet,
+    ) -> Self {
+        assert!(syntax_set.is_some() || serialized_syntax_set.is_some());
+
+        let syntax_set_cell = LazyCell::new();
+        if let Some(syntax_set) = syntax_set {
+            syntax_set_cell.fill(syntax_set).expect("can never fail");
+        }
+
         HighlightingAssets {
-            syntax_set,
+            syntax_set_cell,
+            serialized_syntax_set,
             theme_set,
             fallback_theme: None,
         }
@@ -97,20 +112,30 @@ impl HighlightingAssets {
         }
 
         Ok(HighlightingAssets::new(
-            syntax_set_builder.build(),
+            Some(syntax_set_builder.build()),
+            None,
             theme_set,
         ))
     }
 
     pub fn from_cache(cache_path: &Path) -> Result<Self> {
         Ok(HighlightingAssets::new(
-            asset_from_cache(&cache_path.join("syntaxes.bin"), "syntax set")?,
+            None,
+            Some(SerializedSyntaxSet::FromFile(
+                cache_path.join("syntaxes.bin"),
+            )),
             asset_from_cache(&cache_path.join("themes.bin"), "theme set")?,
         ))
     }
 
     pub fn from_binary() -> Self {
-        HighlightingAssets::new(get_integrated_syntaxset(), get_integrated_themeset())
+        HighlightingAssets::new(
+            None,
+            Some(SerializedSyntaxSet::FromBinary(
+                get_serialized_integrated_syntaxset(),
+            )),
+            get_integrated_themeset(),
+        )
     }
 
     pub fn save_to_cache(&self, target_dir: &Path, current_version: &str) -> Result<()> {
@@ -137,7 +162,17 @@ impl HighlightingAssets {
     }
 
     pub(crate) fn get_syntax_set(&self) -> Result<&SyntaxSet> {
-        Ok(&self.syntax_set)
+        if !self.syntax_set_cell.filled() {
+            self.syntax_set_cell.fill(
+                self.serialized_syntax_set
+                .as_ref()
+                .expect("a dev forgot to setup serialized_syntax_set, please report to https://github.com/sharkdp/bat/issues")
+                .deserialize()?
+             ).unwrap();
+        }
+
+        // It is safe to .unwrap() because we just made sure it was .filled()
+        Ok(self.syntax_set_cell.borrow().unwrap())
     }
 
     /// Use [Self::get_syntaxes] instead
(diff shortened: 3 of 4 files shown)


Cargo.lock
```diff
@@ -101,6 +101,7 @@ dependencies = [
  "globset",
  "grep-cli",
  "lazy_static",
+ "lazycell",
```
Enselic, line 104: syntect already directly depends on this crate, so it is not a big deal for us to also start doing that

Cargo.toml
```diff
@@ -38,6 +38,7 @@ ansi_term = "^0.12.1"
 ansi_colours = "^1.0"
 console = "0.14.1"
 lazy_static = { version = "1.4", optional = true }
+lazycell = "1.0"
```
Enselic, line 41: Note that syntect already depends on this crate, so it is not a big deal for us to also start doing that.

src/assets.rs
```diff
@@ -284,6 +303,42 @@ impl HighlightingAssets {
...
+}
+
+impl SerializedSyntaxSet {
+    fn deserialize(&self) -> SyntaxSet {
+        match self {
+            SerializedSyntaxSet::FromBinary(data) => {
+                from_binary(data)
+            },
+            SerializedSyntaxSet::FromFile(ref path) => {
+                asset_from_cache(&path, "syntax set").expect("cache corrupt, consider rebuilding or clearing, see https://github.com/sharkdp/bat#adding-new-syntaxes--language-definitions on how")
```
Enselic, line 324: One downside of lazy loading is that we can't fail "nicely" early on. We could return a Result here and propagate it onwards, but it would cascade into very unpleasant code since then get_syntax_set() would also need to return a Result. It is enough of an edge for this to fail to justify .expect() here, IMHO.

Enselic:
Also starting to lazy-load `theme_set` should also improve startup speed, although not as dramatic since it is much smaller.

I plan on looking into that once this PR has landed.

keith-hall (approved):
Nice achievement, great job!

src/assets.rs
sharkdp: Sounds okay to me for now.

sharkdp:
This looks great - thank you very much!

Just for my understanding: The use of `LazyCell` (instead of a simple `Option`, for example) is needed in order to make the `get_syntax_set` method `&self` (and not `&mut self`), right?

I can confirm the benchmark results. I thought it would also be good to make sure that normal syntax set loading is not (significantly) slower, even if I wouldn't expect it to be. This can also be confirmed:

```
hyperfine \
  --warmup 5 \
  -L bat bat-master,bat-feature \
  --export-markdown results-loopthrough.md \
  './{bat} tests/examples/multiline.txt' 
```

|:---|---:|---:|---:|---:|
| `./bat-master tests/examples/multiline.txt` | 48.5 ± 1.2 | 47.1 | 52.6 | 7.61 ± 1.04 |
| `./bat-feature tests/examples/multiline.txt` | 6.4 ± 0.9 | 5.3 | 12.8 | 1.00 |


```
hyperfine \
  --warmup 5 \
  -L bat bat-master,bat-feature \
   --export-markdown results-normal.md \
  './{bat} -f src/less.rs'
```

|:---|---:|---:|---:|---:|
| `./bat-master -f src/less.rs` | 64.6 ± 0.9 | 63.4 | 67.9 | 1.00 |
| `./bat-feature -f src/less.rs` | 64.8 ± 1.0 | 63.5 | 69.0 | 1.00 ± 0.02 |

sharkdp (approved):

Enselic:
> Just for my understanding: The use of LazyCell (instead of a simple Option, for example) is needed in order to make the get_syntax_set method &self (and not &mut self), right?

Yes, the use of `LazyCell` is to enable `get_syntax_set()` on `&self`, which is a [recommended pattern](https://doc.rust-lang.org/std/cell/#implementation-details-of-logically-immutable-methods) (with some variation) to implement "hidden" caching, or lazy-loading.  But that is not the full story. I also wanted to avoid breaking the public bat-as-a-library API. In particular [`HighlightingAssets::syntaxes()`](https://docs.rs/bat/0.18.1/bat/assets/struct.HighlightingAssets.html#method.syntaxes). That method assume things can never go wrong, so there would be no sensible way to implement it if `get_syntax_set()` would return, say, an `Option`.

When it is time to look over the public API and make breaking API changes, we could perhaps make it necessary for clients to handle failure to load the cache. Not sure it's worth the complexity though... I suggest we wait and see how much a problem my `.expect("cache corrupt ...` code is in practice.

Thanks for the code reviews. I will try to merge this shortly. I just need some more time to explore how the code behaves when the cache is corrupt.

Enselic:
So, the PR in its current form introduces a behavioral change when `syntaxes.bin` is missing from the cache, and I can't make up my mind about if this is a problem or not. I would love to get a second opinion on it.

### Step-by-step
1. Build a cache. One easy way is to do this:
`bat cache --build --blank --source ~/src/bat/assets`
2. Remove only `syntaxes.bin` from the cache:
`rm ~/.cache/bat/syntaxes.bin`
3. Ask `bat` to highlight a file:
`bat examples/simple.rs`

### git master behavior:
Bat silently falls back to using assets (both themes and syntaxes) from the binary.

### PR behavior:
Bat panics:
```
% bat-pr examples/simple.rs 
thread 'main' panicked at 'cache corrupt, consider rebuilding or clearing, see https://github.com/sharkdp/bat#adding-new-syntaxes--language-definitions on how: Error(Msg("Could not load cached syntax set '/home/martin/.cache/bat/syntaxes.bin'"), State { next_error: Some(Os { code: 2, kind: NotFound, message: "No such file or directory" }), backtrace: InternalBacktrace })', src/assets.rs:324:55
```

## Alternative approach
It is straightforward to also make `bat` in this PR fallback to binary assets in the scenario above. We just need to (try to) read the file during startup, and only defer parsing (which is the slow part) of it. The code for that looks like this: https://github.com/Enselic/bat/commit/4197558a35

In terms of performance with a fully valid cache and with no cache, it doesn't matter much what route we take. Here are the numbers, where I compare "this PR" with "this PR + above commit":
```
% bat cache --build --blank --source ~/src/bat/assets
% hyperfine --export-markdown /dev/tty -L bat bat-pr,bat-pr-prime -L color never,always '{bat} --color {color} examples/simple.rs'
% mv ~/.cache/bat ~/.cache/bat-disabled
% hyperfine --export-markdown /dev/tty -L bat bat-pr,bat-pr-prime -L color never,always '{bat} --color {color} examples/simple.rs'
```

|:---|---:|---:|---:|---:|
| `bat-pr --color never examples/simple.rs` | 16.8 ± 1.1 | 15.0 | 21.8 | 1.00 |
| `bat-pr-prime --color never examples/simple.rs` | 17.5 ± 1.2 | 15.7 | 22.5 | 1.04 ± 0.10 |
| `bat-pr --color always examples/simple.rs` | 113.5 ± 1.7 | 109.1 | 117.7 | 6.74 ± 0.47 |
| `bat-pr-prime --color always examples/simple.rs` | 115.2 ± 2.3 | 111.4 | 122.2 | 6.84 ± 0.49 |

|:---|---:|---:|---:|---:|
| `bat-pr --color never examples/simple.rs` | 11.8 ± 1.4 | 10.1 | 21.8 | 1.00 ± 0.16 |
| `bat-pr-prime --color never examples/simple.rs` | 11.8 ± 1.2 | 10.2 | 17.1 | 1.00 |
| `bat-pr --color always examples/simple.rs` | 109.0 ± 2.5 | 105.5 | 117.3 | 9.24 ± 0.97 |
| `bat-pr-prime --color always examples/simple.rs` | 112.9 ± 2.4 | 109.1 | 121.1 | 9.57 ± 1.01 |

I'm starting to love hyperfine more and more btw, such a great tool! :D You really have a talent for coming up with useful tools, David :)

Anyway, the question is basically: Do you think we need to worry about people poking around in the cache like this?

sharkdp:
Hm. I usually try to invest a lot of time to get rid of possible `panic`s because I don't want users to see these messages. It would really turn me away from a tool. This sounds like a weird edge case where it might be okay for us to show a panic message to the user, but it does look rather scary for the average user ("panicked?", "corrupt?", "what the hell is a backtrace"?). If we can be sure that this is the only scenario where it will be shown, I'm okay with it. Otherwise, I would prefer your alternative solution.

sharkdp:
> I'm starting to love hyperfine more and more btw, such a great tool! :D You really have a talent for coming up with useful tools, David :)

Thank you for the feedback! If you are interested, I encourage you to learn about [Warmup runs and prepare commands](https://github.com/sharkdp/hyperfine#warmup-runs-and-preparation-commands) which can be really useful to compare warm disk cache with cold disk cache. In this scenario here, it could actually be really useful to compare the two, as we are actually loading additional files from disk.

Enselic:
As far as I know, the panic will only happen in the scenario I described. But I do understand your concerns, so I have decided to go the extra mile and incorporate the extra change into this PR. Thanks for pushing me in the right direction :+1: (That will also be a good opportunity for me to play around with `--prepare` etc of `hyperfine` to explore performance numbers with cold file caches.)

Btw, I have noticed you seem to prefer `hyperfine --export-markdown results.md ...`. I was curious if you picked up on my trick you use `hyperfine --export-markdown /dev/tty ...` instead? That way the results ends up directly in the terminal for immediate copy-paste. In your case, don't you need an extra step to access the contents of `results.md`?

sharkdp:
> Btw, I have noticed you seem to prefer `hyperfine --export-markdown results.md ...`. I was curious if you picked up on my trick you use `hyperfine --export-markdown /dev/tty ...` instead? That way the results ends up directly in the terminal for immediate copy-paste. In your case, don't you need an extra step to access the contents of `results.md`?

I have used this in the past as well. I didn't know `/dev/tty`, so I used `/dev/stdout`. I didn't use it here, as it doesn't work anymore with the latest master version of `hyperfine`(!), which I just noticed when I commented here. This is probably due to https://github.com/sharkdp/hyperfine/pull/339 which is also very useful. Will have to think about whether to restore that old export-to-stdout "feature" somehow :smile:

sharkdp:
For bonus points:
```
hyperfine 'sleep 0.1' 'sleep 0.2' --export-markdown >(xclip)
```

Enselic:
I have now rebased this PR on top of the merged PRs #1758, #1755 and #1755, and here is updated performance numbers and behavior.

First, let's compare startup time between current git master bat and bat from this PR:
`hyperfine --export-markdown /dev/tty -L bat bat-new,bat -L color never,always '{bat} --color {color} examples/simple.rs'`

|:---|---:|---:|---:|---:|
| `bat-new --color never examples/simple.rs` | 11.1 ± 0.9 | 9.8 | 14.7 | 1.00 |
| `bat --color never examples/simple.rs` | 95.9 ± 1.2 | 93.9 | 98.6 | 8.62 ± 0.74 |
| `bat-new --color always examples/simple.rs` | 107.6 ± 1.4 | 105.3 | 110.8 | 9.67 ± 0.83 |
| `bat --color always examples/simple.rs` | 108.0 ± 1.6 | 105.7 | 112.1 | 9.71 ± 0.84 |

And we can confirm that the PR bat is still significantly faster in loop-through mode, and same as before in normal mode. Now let's use the new bat in this PR to compare performance with and without custom assets (that are otherwise byte-by-byte identical):
`hyperfine --export-markdown /dev/tty -L bat bat-new -L assets --no-custom-assets, -L color never,always  '{bat} --color {color} examples/simple.rs {assets}'`

|:---|---:|---:|---:|---:|
| `bat-new --color never examples/simple.rs --no-custom-assets` | 11.7 ± 0.9 | 10.3 | 14.3 | 1.01 ± 0.11 |
| `bat-new --color never examples/simple.rs ` | 11.6 ± 0.8 | 10.3 | 14.1 | 1.00 |
| `bat-new --color always examples/simple.rs --no-custom-assets` | 110.5 ± 1.3 | 107.8 | 113.8 | 9.53 ± 0.68 |
| `bat-new --color always examples/simple.rs ` | 108.9 ± 1.4 | 106.5 | 112.2 | 9.39 ± 0.67 |

And as expected, it practically does not matter if custom or integrated assets are used.

Finally, let's now see what happens if the cache is corrupt. First, let's keep `syntaxes.bin` around but put garbage in it:
```
% echo garbage >! ~/.cache/bat/syntaxes.bin
% bat-new examples/simple.rs
[bat error]: Could not parse cached syntax set
```

Instead of panicing, bat now explains that the cached syntax set could not be parsed. And if we remove the file completely:
```
% rm ~/.cache/bat/syntaxes.bin
% bat-new examples/simple.rs
[bat error]: Could not load cached syntax set
'/home/martin/.cache/bat/syntaxes.bin'
```
bat now says that the file can not be found. Note that this is changed behavior compared to before, where we silently used integrated assets if the cached assets were corrupt. But I think the new behavior is better. If the cache is corrupt, the user probably wants to know.

If we remove the cache completely, then of course integrated assets are used:
```
% rm -rf ~/.cache/bat
% bat-new examples/simple.rs
───────┬─────────────────────────────
       │ File: examples/simple.rs
───────┼─────────────────────────────
[...]
```

Would be great if you could take a renewed look on this PR before I merge!

sharkdp (approved):
This looks great -thank you!

> Note that this is changed behavior compared to before, where we silently used integrated assets if the cached assets were corrupt. But I think the new behavior is better. If the cache is corrupt, the user probably wants to know.

I completely agree :+1:

Enselic:
CI problems are unrelated. Merged. (:taco: )

PR #1754, opened 2021-07-26
Make loading of cached assets closer in performance to integrated assets

### LINKED ISSUE #1753: Loading assets from the cache is significantly slower than from the binary
### Step-by-step:

1. Measure the time it takes bat to highlight `examples/simple.rs` without cached assets, i.e. when the assets integrated into the binary are used:
```
% rm -rf ~/.cache/bat
% hyperfine --export-markdown /dev/tty 'bat --color always examples/simple.rs'
```
|:---|---:|---:|---:|---:|
| `bat --color always examples/simple.rs` | 107.4 ± 2.1 | 104.1 | 112.5 | 1.00 |

2. Create cached assets that are identical to the integrated assets:
```
% mkdir -p ~/.cache/bat
% echo "---\nbat_version: 0.18.2" > ~/.cache/bat/metadata.yaml
% cp assets/syntaxes.bin ~/.cache/bat/syntaxes.bin
% cp assets/themes.bin ~/.cache/bat/themes.bin
```

3. Measure again
```
% hyperfine --export-markdown /dev/tty 'bat --color always examples/simple.rs'
```
|:---|---:|---:|---:|---:|
| `bat --color always examples/simple.rs` | 115.0 ± 2.1 | 111.9 | 120.1 | 1.00 |

### Expected result
There is no significant difference.

### Actual result
There is a significant difference.

I will soon create a PR with a proposed fix.

Enselic:
When [measuring performance](https://github.com/sharkdp/bat/pull/1747#issuecomment-886392521) of lazy loading, I noticed a somewhat surprising discrepancy for loop-through performance with and without cached assets. So I looked into possible causes for that and noticed we can make loading of cached assets faster. (Remember: we still eagerly load themes.bin in the above lazy-load PR.)

Using `BufReader` makes sense for large files, but I am convinced assets are never large enough (until proven otherwise) to require buffering. It is significantly faster to load the file contents in one go, so let's do that instead.

The numbers are identical when loading integrated assets. Here are the performance measurements when loading assets that are identical to the integrated assets. I am comparing git master with the code in this PR:
```
% mkdir -p ~/.cache/bat
% echo "---\nbat_version: 0.18.2" > ~/.cache/bat/metadata.yaml
% cp assets/syntaxes.bin ~/.cache/bat/syntaxes.bin
% cp assets/themes.bin ~/.cache/bat/themes.bin
% hyperfine --export-markdown /dev/tty -L bat bat,bat-pr '{bat} --color always examples/simple.rs'
```
|:---|---:|---:|---:|---:|
| `bat --color always examples/simple.rs` | 115.9 ± 1.6 | 113.1 | 121.0 | 1.06 ± 0.03 |
| `bat-pr --color always examples/simple.rs` | 109.3 ± 2.4 | 105.0 | 113.2 | 1.00 |

Closes #1753

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ## Other
 
+- Load cached assets as fast as integrated assets, see #1753 (@Enselic)
+
 
 ## Syntaxes
 
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -1,7 +1,6 @@
 use std::collections::BTreeMap;
 use std::ffi::OsStr;
-use std::fs::{self, File};
-use std::io::BufReader;
+use std::fs;
 use std::path::Path;
 
 use syntect::dumps::{dump_to_file, from_binary, from_reader};
@@ -298,15 +297,14 @@ fn asset_to_cache<T: serde::Serialize>(asset: &T, path: &Path, description: &str
 }
 
 fn asset_from_cache<T: serde::de::DeserializeOwned>(path: &Path, description: &str) -> Result<T> {
-    let asset_file = File::open(&path).chain_err(|| {
+    let contents = fs::read(path).chain_err(|| {
         format!(
             "Could not load cached {} '{}'",
             description,
             path.to_string_lossy()
         )
     })?;
-    from_reader(BufReader::new(asset_file))
-        .chain_err(|| format!("Could not parse cached {}", description))
+    from_reader(&contents[..]).chain_err(|| format!("Could not parse cached {}", description))
 }
 
 #[cfg(test)]
@@ -316,7 +314,7 @@ mod tests {
     use std::ffi::OsStr;
 
     use std::fs::File;
-    use std::io::Write;
+    use std::io::{BufReader, Write};
     use tempfile::TempDir;
 
     use crate::input::Input;


Enselic:
I almost forgot. Here are measurements for this PR, first with a cache:

|:---|---:|---:|---:|---:|
| `./target/release/bat --color always examples/simple.rs` | 106.9 ± 1.2 | 104.2 | 109.6 | 1.00 |

and without a cache:

|:---|---:|---:|---:|---:|
| `./target/release/bat --color always examples/simple.rs` | 105.2 ± 1.0 | 102.8 | 107.3 | 1.00 |

sharkdp:
> Using `BufReader` makes sense for large files, but I am convinced assets are never large enough (until proven otherwise) to require buffering. It is significantly faster to load the file contents in one go, so let's do that instead.

Wow, how did you find this? I wouldn't have expected `BufReader` to have such a large impact.

sharkdp:
I can confirm your results (I was mostly curious if I could set this up as a single hyperfine benchmark :smile:)

`master`:

|:---|---:|---:|---:|---:|
| `cache = true` | 64.6 ± 1.0 | 62.8 | 68.4 | 1.05 ± 0.03 |
| `cache = false` | 61.5 ± 1.2 | 59.9 | 64.7 | 1.00 |

`Enselic:improve-performance-of-loading-of-cached-assets`:

|:---|---:|---:|---:|---:|
| `cache = true` | 62.6 ± 1.3 | 61.0 | 65.8 | 1.02 ± 0.04 |
| `cache = false` | 61.2 ± 1.7 | 59.4 | 68.2 | 1.00 |


```bash
hyperfine \
    --parameter-list cache true,false \
    --warmup 3 \
    --min-runs 30 \
    --command-name 'cache = {cache}' \
    --export-markdown results.md \
    --prepare 'if {cache}; then bat cache --build; else bat cache --clear; fi' \
    'bat --color=always examples/simple.rs'
```
(using https://github.com/sharkdp/hyperfine/pull/391)

sharkdp (approved):

Enselic:
> Wow, how did you find this? I wouldn't have expected BufReader to have such a large impact.

It was when I benchmarked the lazy-load PR that I noticed a big difference between using cached assets vs integrated assets even in loop-through mode (when SyntaxSet is not loaded at all, only ThemeSet). And I knew that BufReader was the only significant difference between the two code paths. So it was a natural first experiment. And it turned out to make a (surprisingly, I agree) big difference.

I have confirmed now that the discrepancy I saw in the lazy-load PR benchmarking is removed with this change. I go from 17.7 ms to 12.3 ms with cached assets in loop-through mode. Using integrated assets is *slightly* better at 11.3 ms, but I'm going to leave it at that. :)

Merging!

PR #1755, opened 2021-07-27
HighlightingAssets: Make .syntaxes() and .syntax_for_file_name() failable

Enselic:
Or rather, introduce new versions of these methods and deprecate the old ones.

This is preparation to enable robust and user-friendly support for lazy-loading.
With lazy-loading, we don't know if the SyntaxSet is valid until after we try to
use it, so wherever we try to use it, we need to return a Result. See discussion
about panics in #1747.

### DIFF
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -128,7 +128,7 @@ impl HighlightingAssets {
         let _ = fs::create_dir_all(target_dir);
         asset_to_cache(&self.theme_set, &target_dir.join("themes.bin"), "theme set")?;
         asset_to_cache(
-            self.get_syntax_set(),
+            self.get_syntax_set()?,
             &target_dir.join("syntaxes.bin"),
             "syntax set",
         )?;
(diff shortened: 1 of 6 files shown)


src/assets.rs
```diff
@@ -280,7 +293,11 @@ impl HighlightingAssets {
     fn get_first_line_syntax(&self, reader: &mut InputReader) -> Option<&SyntaxReference> {
         String::from_utf8(reader.first_line.clone())
             .ok()
-            .and_then(|l| self.get_syntax_set().find_syntax_by_first_line(&l))
+            .and_then(|l| {
+                self.get_syntax_set()
+                    .ok()
```
sharkdp, line 298: This will conflate a missing syntax_set with the non-availability of a first line syntax. Maybe it's better to use an explicit `unwrap` here?

sharkdp: Similar for the other `.ok()` calls.

Enselic: Hmm, wouldn't it be better to change to Result<Option> instead? So that we have three cases:
* Failure to look up
* Nothing was found
* Something was found

sharkdp: It depends whether or not the `get_syntax_set` method can fail at this particular place. Based on the `.unwrap()` calls in similar methods below, I thought it could not fail here. In that case I would have preferred `.unwrap()` or `.expect()`. If it can indeed fail, `Result<Option<_>>` is definitely the way to go. You will probably need one of my [favorite helper functions](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.transpose)

Enselic: Aha, you probably missed that my `.unwrap()`s are in `mod tests { ...`, where I think it is (more) OK to take shortcuts. (And in `PrettyPrinter`, but there is can never fail, as motivated in the added comment.)

I decided to go full robustness, and that it was not enough to handle `~/.cache/bat/syntaxes.bin`. I also want to robustly (meaning: giving helpful message to the user, and not panicing) handle `echo garbage >! ~/.cache/bat/syntaxes.bin`. So my proposed quickfix in the lazy-load PR was not enough.

(I'm not going to bother protecting against flipped bits in the bat binary itself though...)

Anyway, I'll go through which code paths with `Option` that can never fail (because of them never being the first call to get_syntax_set()), and adjust accordingly. Thanks for the code review!

Enselic: I've cascaded the return of `Result` now to all necessary places, and all regression tests pass, so would be great if you could take a second look.

src/bin/bat/main.rs
```diff
@@ -101,7 +101,10 @@ pub fn get_languages(config: &Config) -> Result<String> {
                 true
             } else {
                 let test_file = Path::new("test").with_extension(extension);
-                match assets.syntax_for_file_name(test_file, &config.syntax_mapping) {
+                let syntax = assets
+                    .get_syntax_for_file_name(test_file, &config.syntax_mapping)
+                    .unwrap(); // safe since .get_syntaxes() above worked
```
sharkdp, line 106: definitely out of scope for this PR, but ideally we would have an API which would avoid this. Something like:
```
let syntaxes = assets.get_syntaxes()?;

// and then later:

syntaxes.get_syntax_for_file_name(…) // no unwrap here
```

sharkdp (approved):
Very nice! Thank you

src/assets.rs
sharkdp: > Aha, you probably missed that my `.unwrap()`s are in `mod tests { ...`? I think it is (more) OK to take shortcuts in tests. (I also do .unwrap() in `PrettyPrinter`, but there is can never fail, as motivated in the added comment.)

I saw the unwrap in PrettyPrinter and thought that it might apply to these other situations. Thanks for clarifying.

src/bin/bat/main.rs
Enselic: Good idea. Let's try to move in that direction long-term.

PR #1756, opened 2021-07-27
reference Rob Pike's classical USENIX presentation on cat(1)

ischwarze:
Almost all cat(1) manual pages reference Rob Pike's classical USENIX presentation on cat(1) below SEE ALSO, for example:

    https://man.openbsd.org/cat.1#SEE_ALSO
    https://www.freebsd.org/cgi/man.cgi?query=cat#end
    https://man.netbsd.org/cat.1#SEE%20ALSO
    https://leaf.dragonflybsd.org/cgi/web-man?command=cat
    https://man.bsd.lv/4.4BSD-Lite2/cat#SEE_ALSO
    https://en.wikipedia.org/wiki/Cat_(Unix)   [below "External links"]

I believe bat(1) users would benefit from being offered the same pointer below MORE INFORMATION.

### DIFF
--- a/assets/manual/bat.1.in
+++ b/assets/manual/bat.1.in
@@ -229,3 +229,6 @@ These can be installed to `\fB$({{PROJECT_EXECUTABLE}} --config-dir)/themes\fR`,
 For more information and up-to-date documentation, visit the {{PROJECT_EXECUTABLE}} repo:
 .br
 \fBhttps://github.com/sharkdp/bat\fR
+.PP
+Rob Pike, \(lqUNIX Style, or cat -v Considered Harmful\(rq,
+.I USENIX Summer Conference Proceedings, 1983.


Enselic:

I fail to see how that reference adds value to bat users, however. According to the reference, bat as a tool should not exist. So it's kind of weird to ask us to reference it.

ischwarze:
Hello Martin, thank you for taking the time to look at this pull request!

When publishing in the fields of science and engineering (including CS), it is customary to not only cite sources you agree with, but those that you directly build upon, that are fundamental to the topic you are concerned with, no matter whether you affirm, extend, or refute their conclusions.

If manuals of cat(1) implementations that should not exist according to Rob Pike's presentation would not reference it, then none of those i listed would mention it: they all provide -v.  According to Pike, they should indeed not exist and all those systems would be better off using a more faithful clone of

    https://man.bsd.lv/UNIX-7/cat.1

Yet all of them do include the reference because it is fundamental to their topic, and because users benefit from having access to more than one perspective when learning about a topic.

Yours, Ingo.

Enselic:
Thank you for providing a perspective I did not think about. And I encourage you and others to keep raising concerns like this. I hope you don't mind if I elaborate on my position a bit. I want to stress that they are purely personal, and do not represent an official stance by the bat project as a whole.

Citing sources that goes against your conclusions makes a lot of sense in academia and science. But bat is not a science project. It is a tool. And I stand by my initial judgement that users of this tool will not benefit from this reference.

I am happy to keep having an open-minded discussion about it though. And I encourage my co-maintainers to chime in.

ischwarze:
Hello Martin, that's a position i can understand to a certain degree.
I do a lot of work on a wide range of manual pages, probably on the order of 1000 to 3000 commits all told during the last 10 years, and a few thousand more on manual page formatting tools, and i freely admit that i don't follow full academic standards either.  Having footnotes throughout and a complete, long list of references at the end would be just too distracting in most manual pages.
That said, i do try to follow the spirit of academic citing, that is (1) in the HISTORY section, mention in which release of which operating system the tool was first provided, and when it was added to the operating system currently being documented; (2) in the AUTHORS section, mention the author(s) of the original implementation and the author(s) of the current implementation; (3) in the STANDARDS section, mention the latest revisions of the standard(s) the tool conforms to, if any, and which features are extensions to these standards; and (4) in the SEE ALSO section, mention publications - academic or otherwise - that are particularly noteworthy and particularly relevant to to the topic.
Then again, i certainly won't feel insulted if your policies and/or personal preferences happen to differ.
That's merely the working style i'm used to, so i thought the patch might be useful.
Yours, Ingo.

tubojetengine:
@Enselic @sharkdp please be advise this pr seems in bad faith based on public openbsd email threads [1][2] critizizing/mocking bat and rust. you can see op specific message too [3] where they said they made the pr

[1] https://marc.info/?t=155015237900005&r=1&w=2
[2] https://marc.info/?t=162739733000001&r=1&w=2
[3] https://marc.info/?l=openbsd-ports&m=162740235411770&w=2

sharkdp:
@ischwarze Great to see that you are all having a good laugh (https://marc.info/?l=openbsd-ports&m=162740602514134, https://marc.info/?l=openbsd-ports&m=162569536100530, https://marc.info/?l=openbsd-ports&m=162570185303872). I'll try to respond in a calm way.

Whenever bat is discussed somewhere on the internet, there are some people making fun of it. The http://harmful.cat-v.org/cat-v page has been mentioned multiple times. I'm okay with that.

But let me tell you my point of view.

Unfortunately, I'm not old enough to have experienced the golden age of Unix in the 80s, 90s myself. I really mean that. I would love to have that kind of experience. But I am someone who has been in love with Unix-style operating systems (and the command line in particular) for a long time ... around 20 years. Like many of you probably, I spend *a lot* of time in the terminal each day - at work and during my free time. I consider myself quite proficient in it's use. Contrary to what people seem to think of me (throwing the "`cat -v`" article at me), I am not only familiar with the Unix philosophy, but I also think that it makes some really good points.

I would really like to encourage everyone to be a bit more confident in our generation of developers. Why do some people seem to imply that truly great things have to come from 1980? Why are programs judged by their choice of programming language? What's wrong with some experimentation? What's wrong with providing alternatives?

I would like to talk about some of my own command-line tools. Not to advertise them, but to explain my viewpoint.

* [shell-functools](https://github.com/sharkdp/shell-functools) was an experiment to bring the expressive power of functional programming to the shell. I hadn't seen anything similar, so I had to implement it myself to see if it's a good idea or not. I think that this project follows the Unix guidelines very closely. This is a collection of small tools where each does one thing, and hopefully does it well. They also [compose nicely via Unix pipes](https://github.com/sharkdp/shell-functools#lazy-evaluation). I don't use these tools myself anymore. There are a few fundamental problems. But I still think it was worth to create this project and publish it.
* [`pastel`](https://github.com/sharkdp/pastel) is a collection of small command line tools that work with colors. Each of them does one job. And they compose with Unix pipes.
* [`fd`](https://github.com/sharkdp/fd) is my own alternative to `find`. In comparison to many other classic Unix tools, I think that `find` has a horrible "user interface". `fd` is my attempt to provide something which is more convenient. And much faster.

And finally there is `bat`. It also started as a small experiment. Yes, you can mock it for doing hundreds of different things. But you could also see it as a tool that does one thing (display source code on the terminal), and hopefully does it well.

sharkdp:
@tubojetengine Thanks. I had seen that as well.

tubojetengine:
> @tubojetengine Thanks. I had seen that as well.

no prob. thank you for bat and fd, they get alot of use by me and friends

PR #1757, opened 2021-07-27
HighlightingAssets::get_extension_syntax(): Split up into smaller methods

Enselic:
To make the code easier to understand and change.

### DIFF
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -251,29 +251,42 @@ impl HighlightingAssets {
     }
 
     fn get_extension_syntax(&self, file_name: &OsStr) -> Option<&SyntaxReference> {
+        self.find_syntax_by_file_name(file_name).or_else(|| {
+            self.find_syntax_by_file_name_extension(file_name)
+                .or_else(|| self.get_extension_syntax_with_stripped_suffix(file_name))
+        })
+    }
+
+    fn find_syntax_by_file_name(&self, file_name: &OsStr) -> Option<&SyntaxReference> {
         self.get_syntax_set()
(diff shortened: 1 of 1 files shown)


src/assets.rs
```diff
@@ -251,29 +251,42 @@ impl HighlightingAssets {
...
+            file_path
+                .extension()
+                .and_then(|x| x.to_str())
+                .unwrap_or_default(),
+        )
+    }
+
+    /// If we find an ignored suffix on the file name, e.g. '~', we strip it and
+    /// then try again to find a syntax without it. Note that we do this recursively.
+    fn get_extension_syntax_with_stripped_suffix(
```
keith-hall, line 277: I know we have syntax highlighting test files to cover this, but maybe it could make sense to add some unit tests too, now that it is split out and thus should be easier to test?

keith-hall (approved):

src/assets.rs
Enselic: Thanks for the code review and suggestion!

I'm a big fan of end-to-end tests and integration tests, but not a super-big fan of unit tests. They certainly have their place in certain scenario, but I find that broad use of unit tests take long time to write and maintain, and tends to lock one in to a certain implementation since they are highly implementation dependent. In other words, they make refactoring hard.

I tried to be careful not to change any logic; only split things up into smaller methods, so given that our existing regression test pass, I'm confident enough in that I did not break anything to merge this.

Enselic:
For the record, this was preparation for making the method return `Result` (see discussion in #1755), which became way too messy in the old code.

PR #1758, opened 2021-07-27
Support a hidden arg --no-custom-assets that skips loading assets from the cache

Enselic:
The main use-case I have in mind for this flag is to make it easy to parameterize hyperfine (`hyperfine --parameter-list assets --no-custom-assets, 'bat {assets} examples/simple.rs'`) to compare performance between custom vs integrated assets, as we work on #951. But I would be surprised if it would not come in handy for other uses as well in the future.

I choose to name the arg `--no-custom-assets` rather than `--no-cache` in order to be more future-proof with upcoming changes. Since it is a hidden arg, I figured it would be OK to do it like this. I do not mention it in CHANGELOG.md though, since I don't think we need to make it publicly visible at this point. But let me know if you disagree!

### DIFF
--- a/src/bin/bat/assets.rs
+++ b/src/bin/bat/assets.rs
@@ -23,7 +23,7 @@ pub fn clear_assets() {
     clear_asset("metadata.yaml", "metadata file");
 }
 
-pub fn assets_from_cache_or_binary() -> Result<HighlightingAssets> {
+pub fn assets_from_cache_or_binary(use_custom_assets: bool) -> Result<HighlightingAssets> {
     let cache_dir = PROJECT_DIRS.cache_dir();
     if let Some(metadata) = AssetsMetadata::load_from_folder(&cache_dir)? {
         if !metadata.is_compatible_with(crate_version!()) {
@@ -41,8 +41,12 @@ pub fn assets_from_cache_or_binary() -> Result<HighlightingAssets> {
         }
     }
 
-    Ok(HighlightingAssets::from_cache(&cache_dir)
-        .unwrap_or_else(|_| HighlightingAssets::from_binary()))
+    let custom_assets = if use_custom_assets {
+        HighlightingAssets::from_cache(&cache_dir).ok()
+    } else {
+        None
+    };
+    Ok(custom_assets.unwrap_or_else(HighlightingAssets::from_binary))
 }
 
 fn clear_asset(filename: &str, description: &str) {
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -796,6 +796,17 @@ fn does_not_print_unwanted_file_named_cache() {
 }
 
 #[test]
+fn accepts_no_custom_assets_arg() {
+    // Just make sure --no-custom-assets is considered a valid arg
+    // Don't bother to actually verify that it works
+    bat()
+        .arg("--no-custom-assets")
+        .arg("test.txt")
+        .assert()
+        .success();
+}
+
+#[test]
 fn unicode_wrap() {
     bat_with_config()
         .arg("unicode-wrap.txt")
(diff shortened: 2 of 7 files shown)


src/bin/bat/assets.rs
```diff
@@ -41,8 +41,14 @@ pub fn assets_from_cache_or_binary() -> Result<HighlightingAssets> {
         }
     }
 
-    Ok(HighlightingAssets::from_cache(&cache_dir)
-        .unwrap_or_else(|_| HighlightingAssets::from_binary()))
+    let no_custom_assets = wild::args_os().any(|arg| arg == "--no-custom-assets");
```
sharkdp, line 44: It feels wrong to me to parse the entire set of environment arguments again here. It also ignores the possibility that `--no-custom-assets` could be part of `bat`s config file (even if unlikely).

Couldn't we pass a `custom_assets: bool` into this function as a parameter? Make that part of the `Config`, like other settings?

src/bin/bat/assets.rs
```diff
@@ -41,8 +41,14 @@ pub fn assets_from_cache_or_binary() -> Result<HighlightingAssets> {
...
+    let from_cache = if no_custom_assets {
+        Err("Skipping custom assets due to --no-custom-assets arg".into())
```
sharkdp, line 46: Took me a while to understand this. Is this just a dummy error in order to take the "else" path in `unwrap_or_else` below? Isn't there another, simpler way to write this logic?

tests/integration_tests.rs
```diff
@@ -795,6 +795,19 @@ fn does_not_print_unwanted_file_named_cache() {
     bat_with_config().arg("cach").assert().failure();
 }
 
+#[test]
+fn accepts_no_custom_assets_arg() {
+    // Just make sure --no-custom-assets is considered a valid arg
+    // Don't bother to actualy verify that it works
```
sharkdp, line 801: typo: actually

tests/integration_tests.rs
```diff
@@ -795,6 +795,19 @@ fn does_not_print_unwanted_file_named_cache() {
...
+    bat()
+        .arg("--no-custom-assets")
+        .arg("test.txt")
+        .assert()
+        .success()
+        .stdout("hello world\n")
+        .stderr("");
```
sharkdp, line 808: Let's remove these two lines.

sharkdp (changes requested):
Thanks. Sounds like a good idea.

src/bin/bat/assets.rs
Enselic: I actually went in that direction first, but backed off when I noticed `--no-config` was not in `Config`. But hmm I just now I realized why: `--no-config` is a meta-config, i.e. a config to not use configs :bulb: 

I'll add my new arg to `Config`.

src/bin/bat/assets.rs
Enselic: I wasn't super happy with it either, but though it was good enough. But if you struggled with it on first try it definitely is not clear enough code. I'll see if I can come up with something clearer.

tests/integration_tests.rs
Enselic: will fix

tests/integration_tests.rs
Enselic: will fix

Enselic:
All comments should be addressed now, so feel free to take another look when you have time.

src/bin/bat/assets.rs
```diff
@@ -41,8 +41,12 @@ pub fn assets_from_cache_or_binary() -> Result<HighlightingAssets> {
...
     }
 
-    Ok(HighlightingAssets::from_cache(&cache_dir)
-        .unwrap_or_else(|_| HighlightingAssets::from_binary()))
+    let custom_assets = if use_custom_assets {
+        HighlightingAssets::from_cache(&cache_dir).ok()
+    } else {
+        None
+    };
+    Ok(custom_assets.unwrap_or_else(|| HighlightingAssets::from_binary()))
```
sharkdp, line 49: Could be `custom_assets.unwrap_or_else(HighlightingAssets::from_binary)` to fix the clippy suggestion.

sharkdp (approved):

src/bin/bat/assets.rs
Enselic: Dang it, I didn't see that one. Will fix before merge. Thanks for review!

PR #1759, opened 2021-07-28
Add HTTP Request/Response syntax as a git submodule

### LINKED ISSUE #1748: Support for multiple syntaxes in a single file
I have a bunch of files like this. They're incoming web requests that I've captured to a file:

```
POST /foo/bar HTTP/1.1
X-Forwarded-For: 54.187.216.72
Content-Length: 3077
Cache-Control: no-cache
X-Forwarded-Proto: https
Content-Type: application/json; charset=utf-8
Host: 7261363d4b2c.ngrok.io
User-Agent: Stripe/1.0 (+https://stripe.com/docs/webhooks)
Accept: */*; q=0.5, application/xml
Accept-Encoding: gzip

{
  "id": "blahblahblahblah",
  "object": "event",
  "api_version": "2020-03-02",
  "created": 1626790174,
  "data": {
  ...
```

and it would be Dead Handy to be able to highlight the two sections independently, both the HTTP headers and the JSON body. Obviously this would require a manually specified `-l` and probably also telling bat what the boundary between the two languages is, something like this perhaps:

`bat -l http+json --boundary='\r\n\r\n' filename`

keith-hall:
Fixes #1748


sharkdp (approved):
Nice!

Thank you for improving the docs as well.

PR #1760, opened 2021-07-29
HighlightingAssets: Encapsulate theme_set behind a getter

Enselic:
Mainly to prepare for potential lazy-loading in the future to improve startup performance. Like we did for `syntax_set`.

I currently don't plan on making `ThemeSet` lazy-loaded, because the impact of that is negligible compared to other things we can do. But in case we want to do it later, it's nice to have the code prepared like this.

### DIFF
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -21,7 +21,7 @@ use crate::syntax_mapping::{MappingTarget, SyntaxMapping};
 pub struct HighlightingAssets {
     syntax_set_cell: LazyCell<SyntaxSet>,
     serialized_syntax_set: Option<SerializedSyntaxSet>,
-    pub(crate) theme_set: ThemeSet,
+    theme_set: ThemeSet,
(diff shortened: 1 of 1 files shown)


src/assets.rs
```diff
@@ -229,7 +237,8 @@ impl HighlightingAssets {
                 if !theme.is_empty() {
                     bat_warning!("Unknown theme '{}', using default.", theme)
                 }
-                &self.theme_set.themes[self.fallback_theme.unwrap_or_else(|| Self::default_theme())]
+                &self.get_theme_set().themes
+                    [self.fallback_theme.unwrap_or_else(|| Self::default_theme())]
```
Enselic, line 241: kinda weird formating, but `cargo fmt` insists...

sharkdp: indeed :-)

sharkdp (approved):

PR #1761, opened 2021-07-29
Update git2 dependency

sharkdp:
..  to fix an incompatibility with Rust 1.54

See https://github.com/rust-lang/rust/issues/85574


Enselic:
For the curious: direct link to the fix in git2-rs: https://github.com/rust-lang/git2-rs/pull/712

Enselic (approved):

Issue #915, opened 2020-04-12
bat panics on simple Vue file


sharkdp:
When calling `bat` on this Vue file
``` vue
<style lang="stylus">
</style>
```
it panics with:
```
thread 'main' panicked at
  'Can only call resolve on linked references:
    ByScope { scope: <source.stylus>, sub_context: None }',
  …/syntect-4.1.0/src/parsing/syntax_definition.rs:188:18
```

found with the Python script in #913.

sharkdp:
This happens because we didn't have a syntax for "Stylo" source code, which is included from the Vue syntax when using `<style lang="stylus">`.

nemu626:
```
thread 'main' panicked at 'Can only call resolve on linked references: ByScope { scope: <text.pug>, sub_context: None }', /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/syntect-3.3.0/src/parsing/syntax_definition.rs:192:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
I meet same error with single Vue component file with `pug` template syntax.

sharkdp:
Thank you for reporting this. That's unfortunate.

It should be possible to somehow verify that syntaxes do not include broken references when building the binary assets. This might need to be fixed upstream (in `syntect`).

Enselic:
I don't understand how to reproduce the problem. Can someone please provide a Vue component file with `pug` template syntax that can be used to reproduce the panic? Maybe you @nemu626 ?

sharkdp:
I think you can take the one from the first post and replace `stylus` with one of the other languages supported by the syntax file, e.g. Pug

Enselic:
Turns out it was necessary to also change `style` to `template`. Here is how to reproduce with a one-liner: 

```bash
% echo "<template lang='pug'>\n</template>" > /tmp/crash.vue | bat /tmp/crash.vue
[...]
thread 'main' panicked at 'Can only call resolve on linked references: ByScope { scope: <text.pug>, sub_context: None }', [...]
```

keith-hall:
> It should be possible to somehow verify that syntaxes do not include broken references when building the binary assets. This might need to be fixed upstream (in `syntect`).

I've created a PR which will add a new API to syntect for this, but I'm sure it could be improved: https://github.com/trishume/syntect/pull/332

keith-hall:
[...]

PR #1771, opened 2021-08-01
Warn when building assets from files if some referenced contexts are missing

### LINKED ISSUE #1616: Warn at asset building time when a syntax definition context reference can't be resolved
It would be great to make use of the new syntect API to detect when a syntax definition has some context references (specifically push/set targets - missing includes are not such a problem and can be just treated as warnings) which couldn't be linked/resolved and fail the CI build. https://github.com/sharkdp/bat/issues/915#issuecomment-823581511

keith-hall:
Upgrade to syntect 4.6.0 and add a warning to stdout when building assets from files and some context references are unresolved. Fixes #1616 and makes it easier to work on #915 and prevent such cases occurring again in future :)

### DIFF
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -111,8 +111,17 @@ impl HighlightingAssets {
             );
         }
 
+        let syntax_set = syntax_set_builder.build();
+        let missing_contexts = syntax_set.find_unlinked_contexts();
+        if !missing_contexts.is_empty() {
+            println!("Some referenced contexts could not be found!");
+            for context in missing_contexts {
+                println!("- {}", context);
+            }
+        }
+
         Ok(HighlightingAssets::new(
-            Some(syntax_set_builder.build()),
+            Some(syntax_set),
             None,
             theme_set,
         ))
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -381,9 +381,9 @@ dependencies = [
 
 [[package]]
 name = "fancy-regex"
-version = "0.3.5"
+version = "0.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ae91abf6555234338687bb47913978d275539235fcb77ba9863b779090b42b14"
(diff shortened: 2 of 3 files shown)


src/assets.rs
```diff
@@ -111,8 +111,17 @@ impl HighlightingAssets {
             );
         }
 
+        let syntax_set = syntax_set_builder.build();
+        let missing_contexts = syntax_set.find_unlinked_contexts();
+        if missing_contexts.len() > 0 {
```
Enselic, line 116: Nice! Just fix this lint and this PR is golden:

```
warning: length comparison to zero
   --> src/assets.rs:116:12
    |
116 |         if missing_contexts.len() > 0 {
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^ help: using `!is_empty` is clearer and more explicit: `!missing_contexts.is_empty()`
    |
    = note: `#[warn(clippy::len_zero)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#len_zero
```

I did a test-run of the code and it seemed to work fine for me too :)

Enselic (approved):

sharkdp (approved):

Enselic (approved):

Enselic:
@keith-hall Oh, I just noticed, seems like you forgot to git add the changes to `Cargo.lock` ?

Enselic (changes requested):
Cargo.lock seems forgotten?

Enselic:
Our CI is supposed to detect this, but it appears as if us forgetting to pass `--locked` to early `cargo` commands fixed `Cargo.lock` for later commands.

I just created a [PR](https://github.com/sharkdp/bat/pull/1772) that fixes this. I verified that with this CI fix, this PR would cause [CI to fail](https://github.com/Enselic/bat/pull/19/checks):

Enselic (approved):
:+1:

PR #1776, opened 2021-08-03
When returning a SyntaxReference, also return the SyntaxSet that contains it

Enselic:
Here comes what I think will be the most controversial change on the #951 journey, because it has the highest impact on the bat-as-a-library API. Let me know what you think about it.

To improve startup performance, we will later load smaller `SyntaxSet`s instead of one giant one. However, the current API assumes only one `SyntaxSet` is ever used, and that that implicitly is the `SyntaxSet` from which returned `SyntaxReference`s comes.

This change changes the API to reflect that `SyntaxSet` and `SyntaxReference` are tightly coupled, and enables the use of several `SyntaxSet`.

I think it will make sense to add a helper method to reduce the boilerplate a bit, but I would like to get early feedback on the change itself before I do that.

### DIFF
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -163,7 +163,7 @@ impl<'a> InteractivePrinter<'a> {
             panel_width = 0;
         }
 
-        let highlighter_and_syntax_set = if input
+        let (highlighter, syntax_set) = if input
             .reader
             .content_type
             .map_or(false, |c| c.is_binary() && !config.show_nonprintable)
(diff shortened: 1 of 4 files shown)


src/printer.rs
```diff
@@ -163,23 +163,29 @@ impl<'a> InteractivePrinter<'a> {
             panel_width = 0;
         }
 
-        let highlighter = if input
+        let highlighter_and_syntax_set = if input
```
keith-hall, line 166: is it possible to deconstruct the tuple here into separate variables? Does it make sense to do it? My reasoning is just to make lines 200/201 prettier instead of referring to `.0` and `.1`.

keith-hall (approved):

sharkdp (approved):
The proposed API changes looks good to me. The code as well.

src/printer.rs
Enselic: Good idea! Will fix

Enselic: This is fixed now and is a clear improvement.

Enselic:
I fixed the destructuring comment. I also looked into ways to introduce a helper function, but couldn't come up with something I was happy with. So I'll go ahead and merge this now.

Issue #1777, opened 2021-08-03
LLVM


ioncodes:
**Syntax:**
[LLVM](https://llvm.org/)

**Guideline Criteria:**
Not on packagecontrol.io

**Reasoning:**
LLVM is a very common assembly-like (intermediate representation/IR) format used in various compilers. Due to its huge popularity, I think it should be included by default. Being able to navigate LLVM IR files on the command line with a nice syntax highlighting would greatly benefit my (and other's) workflow.

ioncodes:
As per https://github.com/sharkdp/bat/pull/1778 I'll be implementing this in my own repository and then submit a PR that links to it as a submodule.

sharkdp:
Thank you. I would love to see LLVM IR as a supported language.

Enselic:
Closed by #1780

PR #1778, opened 2021-08-03
Implemented LLVM syntax highlighting and regression tests

ioncodes:
This PR implemented LLVM IR syntax highlighting as per #1777. Let me know if this change would be beneficial and I'll add it to the CHANGELOG.md!


ioncodes:
Here's a screenshot of what it looks like: image

keith-hall:
Hi, nice that you worked on a syntax definition for a language you are interested in and shared it here - thanks! :)

I have a question - did you consider maintaining it in a separate repository? The reason I ask is, if/when it becomes popular, I can foresee that we may see lots of issues about it or PRs relating to improving it, and it would be nice to be able to just bump a git submodule rather than have lots of noise in this repo

ioncodes:
I could definitely do that! Would you like me to close this PR and make a new repo with the syntax file and then submit a PR with that submodule added to 02_Extras or just wait it out until it receives some traffic first?

keith-hall:
Up to you :) It may make sense to wait and see what others think, but the good news is we've added some syntax definitions recently which don't meet the 10k package control requirement (i.e. https://github.com/sharkdp/bat/pull/1685#issuecomment-886187561), and we are working on performance improvements so that having a large number of syntaxes should be less of a problem. But I certainly think we're more likely to agree to add another submodule than a sublime-syntax directly :)

vzaliva:
I am working with LLVM a lot and would like to be able to us this syntax highligting. How can I do this. I have bat version 0.18.2.
Can I install additional syntax locally?

ioncodes:
@vzaliva 

The way I did it was:

1. Pull my commit (make sure to fetch all submodules)
2. `bash assets/create.sh`
3. `cargo install --path . --locked --force`

This creates a new syntax blob and compiled bat with it. It will also install bat (and overwrite your current one I'd assume) globally.

If that doesn't work let me know. If it's urgent you can reach out to me on [Twitter](https://twitter.com/layle_ctf).

Issue #1726, opened 2021-07-13
Errors for `bat cache --build` when cache already exists and BAT_CACHE_PATH is specified


zachriggle:
**Describe the bug you encountered:**

I am attempting to distribute a custom `bat` theme with my tool to use.  I want to avoid the possibility of messing with the user's installed `bat` themes and theme cache.

Notably, the `bat cache --build --blank` command does not work when the destination path already exists, and it throws an unusual error message.

Note that it works the first time:

```sh
$ mkdir -p /tmp/bat-cache-bug

$ cd /tmp/bat-cache-bug

$ XDG_CONFIG_HOME=$PWD BAT_CACHE_PATH=$PWD/cache bat cache --build
No themes were found in '/tmp/bat-cache-bug/bat/themes', using the default set
No syntaxes were found in '/tmp/bat-cache-bug/bat/syntaxes', using the default set.
Writing theme set to /tmp/bat-cache-bug/cache/themes.bin ... okay
Writing syntax set to /tmp/bat-cache-bug/cache/syntaxes.bin ... okay
Writing metadata to folder /tmp/bat-cache-bug/cache ... okay
```

But fails the second time, with the *exact* same command:

```
$ XDG_CONFIG_HOME=$PWD BAT_CACHE_PATH=$PWD/cache bat cache --build
error: Found argument '--build' which wasn't expected, or isn't valid in this context

USAGE:
    bat [OPTIONS] [FILE]...
    bat <SUBCOMMAND>

For more information try --help
```

**What did you expect to happen instead?**

Rebuilding the cache would also be nice, as that's the expected behavior.

A better error message would be helpful.

**How did you install `bat`?**

`brew install bat`

---

**bat version and environment**

```sh
$ XDG_CONFIG_HOME=$PWD BAT_CACHE_PATH=$PWD/cache bat --diagnostic
```

#### Software version

bat 0.18.1 (<no git>)

#### Operating system

macOS 12.0 (Darwin 21.0.0)

#### Command-line

```bash
bat --diagnostic
```

#### Environment variables

```bash
SHELL=/bin/zsh
PAGER=less
LESS='-F -i -M -R -S -w -X -z-4'
BAT_PAGER=<not set>
BAT_CACHE_PATH=/tmp/bat-cache-bug/cache
BAT_CONFIG_PATH=<not set>
BAT_OPTS=<not set>
BAT_STYLE=plain
BAT_TABS=<not set>
BAT_THEME=<not set>
[...]

PR #1782, opened 2021-08-05
Draft: Replace cache subcommand with --custom-assets-* args

Enselic:
This PR goes over the code base and replaces the `cache` subcommand with new `--custom-assets-*` equivalents, to see what that would be like.

This is Proposal C in https://github.com/sharkdp/bat/issues/1726#issuecomment-892862794

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Features
 
+- TODO: Write about new CLI
+
 
 ## Bugfixes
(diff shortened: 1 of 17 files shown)


Enselic:
I'm closing this to keep PR inbox clean. We need some more time to discuss what a new CLI should look like.

If we end up choosing a path akin to the CLI in this PR, we can always reopen it again later.

PR #1791, opened 2021-08-10
Make 'build-assets' an optional capability for application

Enselic:
The main short-term benefit of this PR is to make building `minimal-application` (previously called `quick-build-application`) even faster.

But it also opens up a major long-term possibility:
* If we end up creating a separate `bat-assets` binary to build assets, we can still allow users to optionally keep the current capability in the main app

For the curious, here are some numbers, comparing `minimal-application` with `application` for stripped release builds. Note that I am including https://github.com/trishume/syntect/pull/345 in below numbers:

| variant               | stripped release binary size | dependencies to build |
| ----------------------|------------------------------|-----------------------|
| `minimal-application` | 3,5M                         | 121                   |
| `application`         | 4,7M                         | 166                   |

I have not yet done broad verification of this PR. Would be great to get your high-level feedback on it before I polish it.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -246,12 +246,12 @@ jobs:
         command: check
         args: --locked --target=${{ matrix.job.target }} --verbose --lib --no-default-features --features regex-onig,git,paging
 
-    - name: "Feature check: quick-build-application"
+    - name: "Feature check: minimal-application"
       uses: actions-rs/cargo@v1
       with:
         use-cross: ${{ matrix.job.use-cross }}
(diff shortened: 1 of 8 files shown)


sharkdp:
I think we should maybe be careful not to make the conditional compilation configuration *too* fine-grained to avoid combinatorial complexity, but the asset building seems like a very reasonable part to make optional :+1:

sharkdp (approved):

Enselic:
I've done some additional verification now, and everything seems to work as it should. Merging.

PR #1793, opened 2021-08-11
Syslog highlight improvements

scop:
Make FQDN's and literal IPv4 and IPv6 addresses work, ditto process names with periods in them.


keith-hall (approved):

keith-hall:
Thanks, looks good. Do you want to update the changelog so others will know what has been fixed/improved? :)

scop:
Why not :)

Issue #1727, opened 2021-07-13
BAT_CONFIG_DIR not respected, despite CHANGELOG.md


zachriggle:
**Describe the bug you encountered:**

Not sure if this is a bug or a feature request, but I'll go with bug since there's at least some documentation.

CHANGELOG.md states:

```
CHANGELOG.md-754-## Changes
CHANGELOG.md-755-
CHANGELOG.md:756:- New themes in `$BAT_CONFIG_DIR/themes` are now loaded *in addition* to
CHANGELOG.md-757-  the default themes (they may also override), see #172
```

This value `$BAT_CONFIG_DIR` is not respected in `directories.rs`, only `$XDG_CONFIG_HOME` and `$HOME` are.

There are no references to `BAT_CONFIG_DIR` anywhere in the source except the changelog.  It looks like it used to be documented, but was never implemented.  Documentation was removed in d3d795ffb2adad97d507192f21225f1382559d99 and other commits.

It would be nice to see support for `BAT_CONFIG_DIR` in `directories.rs`.  This makes it easier for tools to ship custom bat themes / configs / caches, without overriding `$XDG_CONFIG_HOME`.

**What did you expect to happen instead?**

I should be able to specify `$BAT_CONFIG_DIR`.

**How did you install `bat`?**

`brew install bat`

---

**bat version and environment**

```
$ bat --version
bat 0.18.1
```

sharkdp:
I think you are looking for `BAT_CONFIG_PATH`, which is also mentioned in the `README`. The commit you are referring to removed `BAT_CONFIG_DIR` from a README code snippet. It was never used in `bat` internally.

The CHANGELOG entry you are referring to is misleading though, I agree. It's a historical artifact that refers to the README entry at the time. We could change that to read `BAT_CONFIG_PATH`, I guess.

zachriggle:
As far as I can tell, `BAT_CONFIG_PATH` is a file, not a directory containing e.g. a `themes` subdirectory.

```
$ BAT_CONFIG_PATH=/tmp/bat-cache-bug bat --diagnostic
...
#### Config file

Could not read contents of '/tmp/bat-cache-bug': Is a directory (os error 21).
```

sharkdp:
[...]

PR #1797, opened 2021-08-12
Recognize $BAT_CONFIG_DIR and give it precedence over $XDG_CONFIG_HOME

billrisher:
This PR closes issue #1727

I added a nearly-identical test case to one that already exists, for verbosity's sake I left it there but I understand if we'd want to wrap the new test case into an existing one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Features
 
-- $BAT_CONFIG_DIR is now a recognized environment variable and has precedent over $XDG_CONFIG_HOME. (@billrisher)
+- `$BAT_CONFIG_DIR` is now a recognized environment variable. It has precedence over `$XDG_CONFIG_HOME`, see #1727 (@billrisher)
 
 ## Bugfixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Features
 
+- $BAT_CONFIG_DIR is now a recognized environment variable and has precedent over $XDG_CONFIG_HOME. (@billrisher)
 
 ## Bugfixes
 
--- a/src/bin/bat/directories.rs
+++ b/src/bin/bat/directories.rs
@@ -4,8 +4,9 @@ use std::path::{Path, PathBuf};
 use lazy_static::lazy_static;
 
 /// Wrapper for 'dirs' that treats MacOS more like Linux, by following the XDG specification.
-/// This means that the `XDG_CACHE_HOME` and `XDG_CONFIG_HOME` environment variables are
-/// checked first. The fallback directories are `~/.cache/bat` and `~/.config/bat`, respectively.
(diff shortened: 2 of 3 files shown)


src/bin/bat/directories.rs
```diff
@@ -4,8 +4,9 @@ use std::path::{Path, PathBuf};
 use lazy_static::lazy_static;
 
 /// Wrapper for 'dirs' that treats MacOS more like Linux, by following the XDG specification.
-/// This means that the `XDG_CACHE_HOME` and `XDG_CONFIG_HOME` environment variables are
-/// checked first. The fallback directories are `~/.cache/bat` and `~/.config/bat`, respectively.
+/// The `XDG_CACHE_HOME` environment variable is checked first. `BAT_CONFIG_DIR`
+///  is then checked before the `XDG_CONFIG_HOME` environment variable.
+///  The fallback directories are `~/.cache/bat` and `~/.config/bat`, respectively.
```
billrisher, line 9: There's probably a better way to word this if anyone has any ideas.

keith-hall (approved):

tests/integration_tests.rs
```diff
@@ -747,6 +747,25 @@ fn config_location_when_generating() {
     assert!(tmp_config_path.exists());
 }
 
+#[test]
```
sharkdp, line 750: Can you please add the new env. variable to `bat_raw_command_with_config` to make sure it is *removed from the environment* for all integration tests? This is important in order to make sure that tests pass, even in environments where a user has set some env. variables.

tests/integration_tests.rs
```diff
@@ -747,6 +747,25 @@ fn config_location_when_generating() {
...
+    // Create file at BAT_CONFIG_DIR
+    bat_with_config()
+        .env("BAT_CONFIG_DIR", bat_conf_dir.to_str().unwrap())
+        .arg("--generate-config-file")
+        .assert()
+        .success()
+        .stdout(
+            predicate::str::is_match("Success! Config file written to .*config\n")
+                .unwrap()
+        );
```
sharkdp, line 763: Another idea for a (maybe simpler) test would be similar to `config_read_arguments_from_file`. Wouldn't require a temp. directory. But I guess this is fine as well. Thank you for writing a test!

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Features
 
+- $BAT_CONFIG_DIR is now a recognized environment variable and has precedent over $XDG_CONFIG_HOME. (@billrisher)
```
sharkdp, line 5:
```suggestion
- `$BAT_CONFIG_DIR` is now a recognized environment variable. It has precedence over `$XDG_CONFIG_HOME`, see #1727 (@billrisher)
```

sharkdp:

sharkdp (changes requested):

sharkdp (approved):

PR #1800, opened 2021-08-16
Fix syslog syntax highlighting when no colon after "process"

### LINKED ISSUE #1794: syslog highlighting confused if there is no colon in message

Syslog highlighter gets confused if a message without colon occurs. Highlighting "leaks" to the next line, gets likely terminated by the colon in the next line's timestamp, and works thereafter.

Example (on Ubuntu 20.04, rsyslog):
```shellsession
$ logger --server localhost hello
$ tail -n 1 /var/log/syslog
Aug 11 21:40:31 uncleman scop hello
```
Screenshot from 2021-08-11 21-45-01

No confusion, lines with messages without colons would work.

Current git master.

keith-hall:
Fixes #1794 

I consider the existing changelog entry referring to syslog improvements to be enough, so I didn't update the changelog here.


sharkdp (approved):
Thank you!

PR #1803, opened 2021-08-18
Make Package Control reference a link

scop:


scop:
I think some clarification on acceptance criteria is also in order. For example #1777 adds a syntax that's not at Package Control at all, but the way the criteria are currently written, I suppose that would not have met it so the 10,000 downloads over there is apparently not the only way to meet them...?

keith-hall (approved):

keith-hall:
I believe we are going to re-address the criteria after our performance improvements for bat startup time (which is primarily related to the number & size of syntax definitions we ship) have landed. (See https://github.com/sharkdp/bat/issues/951)

sharkdp (approved):

PR #1811, opened 2021-08-21
Add release checklist

sharkdp:
This adds some documentation concerning the release of new `bat` versions. The first commit is my partially outdated version. Let's discuss which steps can be removed / should be changed / should be added.

### DIFF
--- a/doc/release-checklist.md
+++ b/doc/release-checklist.md
@@ -1,0 +1,27 @@
+Release checklist:
+
+- [ ] Check for outdated dependencies (`cargo outdated`)
+- [ ] Optional: update dependencies with `cargo update`.
+      See also https://deps.rs/repo/github/sharkdp/bat
+- [ ] Update syntaxes and themes (`cargo install -f --path .; assets/create.sh`).
+- [ ] Update README (features, usage, languages, ..).
+- [ ] Update man page
+
+
+- [ ] Update version in `Cargo.toml`. Run `cargo build` to update `Cargo.lock`
+- [ ] Update version in README and possibly update minimum Rust version
+- [ ] Run `cargo fmt`
+- [ ] Run `cargo test`
+- [ ] Run `cargo install --path . -f`
+- [ ] Test new features & command-line options
+- [ ] Check `-h` and `--help` texts
+
+
+- [ ] `cargo publish --dry-run --allow-dirty`.
+- [ ] write GitHub release notes
+- [ ] check if CI succeeds
+- [ ] `git tag vX.Y.Z; git push --tags`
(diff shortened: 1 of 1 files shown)


doc/release-checklist.md
```diff
@@ -0,0 +1,63 @@
...
+- [ ] Create a tag and push it to the remote `git tag vX.Y.Z; git push --tags`.
+      This will trigger the deployment via GitHub Actions.
+- [ ] Go to https://github.com/sharkdp/bat/releases/new to create the new
+      release. Select the new tag and also use it as the release title. For the
+      release notes, copy the corresponding section from `CHANGELOG.md` and
+      possibly add additional remarks for package maintainers.
+      Publish the release.
+- [ ] Check if the binary deployment works (archives and Debian packages should
+      appear when the CI run for the Git tag has finished).
+- [ ] Publish to crates.io by running `cargo publish` in a *clean* repository.
```
Enselic, line 57: I suppose you need to give me explicit permission to do this? This is me: https://crates.io/users/Enselic

For the record: Yes, My GitHub account (which is also used to login on crates.io) _do_ require 2fa :)

doc/release-checklist.md
```diff
@@ -0,0 +1,63 @@
...
+      The safest way to do this is to cloning a fresh copy.
+
+## Post release
+
+- [ ] Optional: Inform package maintainers about the update:
```
Enselic, line 62: I apologize in advance for my ignorance, but I would expect package maintainers to be subscribed to the project so that they get notified of releases? bat seems to be packaged widely, and informing all package maintainers would be a daunting task.

Enselic:
It all seems very reasonable. We could consider adding steps for what to do for hotfix releases (like 0.18.3 will be), but I don't think that would be worth it. I think the current list is totally sufficient as a baseline to make case-by-case adaptions for hotfix releases.

I'm setting Approved so that this is ready to be merged once we sort out the remaining open questions.

doc/release-checklist.md
sharkdp: Done

doc/release-checklist.md
sharkdp: I'm okay with removing this.

sharkdp:
> We could consider adding steps for what to do for hotfix releases (like 0.18.3 will be), but I don't think that would be worth it. I think the current list is totally sufficient as a baseline to make case-by-case adaptions for hotfix releases.

Right. There are definitely some things that can be skipped for hotfix releases. And sometimes also for normal releases. Please don't see the checklist as an fixed instruction that needs to be followed, but rather as a guideline.

PR #1813, opened 2021-08-22
Experiment: remove CARGO_TEST_OPTIONS

sharkdp:
In principle, this *could* work. `cargo cross` runs the tests via QEMU.
For integration tests, the only difficulty is that we run `bat` from
within the tests. But maybe this is handled by assert_cmd.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -194,21 +194,12 @@ jobs:
         echo ::set-output name=BIN_PATH::${BIN_PATH}
         echo ::set-output name=BIN_NAME::${BIN_NAME}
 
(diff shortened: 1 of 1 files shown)


sharkdp:
Apparently not :-/

Enselic:
Just in case you forgot: Since 9db9a38565 you can trigger a CI pipeline manually for any branch, without creating a temporary PR.

Just do like this:
1. Push the branch
2. Actions tab -> Click "CICD" -> Click "Run workflow" button to the right of "This workflow has a workflow_dispatch event trigger." and select what branch you want to run with

sharkdp:
I did indeed forget about this. Thanks.

PR #1815, opened 2021-08-22
doc/release-checklist.md: Recommend git push origin tag vX.Y.Z

Enselic:
Also recommend to just push the tag for the release, and not all tags.

### DIFF
--- a/doc/release-checklist.md
+++ b/doc/release-checklist.md
@@ -52,8 +52,6 @@ See this page for a good overview: https://deps.rs/repo/github/sharkdp/bat
       release notes, copy the corresponding section from `CHANGELOG.md` and
       possibly add additional remarks for package maintainers.
       Publish the release.
-      NOTE: If you are too slow, the CI job you triggered in the step above will create
-      the release for you, and the author of the release will become **github-actions**.
 - [ ] Check if the binary deployment works (archives and Debian packages should
       appear when the CI run for the Git tag has finished).
 - [ ] Publish to crates.io by running `cargo publish` in a *clean* repository.
--- a/doc/release-checklist.md
+++ b/doc/release-checklist.md
@@ -45,13 +45,15 @@ See this page for a good overview: https://deps.rs/repo/github/sharkdp/bat
 
 ## Release
 
-- [ ] Create a tag and push it to the remote `git tag vX.Y.Z; git push --tags`.
+- [ ] Create a tag and push it to the remote `git tag vX.Y.Z; git push origin vX.Y.Z`.
       This will trigger the deployment via GitHub Actions.
 - [ ] Go to https://github.com/sharkdp/bat/releases/new to create the new
       release. Select the new tag and also use it as the release title. For the
       release notes, copy the corresponding section from `CHANGELOG.md` and
       possibly add additional remarks for package maintainers.
       Publish the release.
+      NOTE: If you are too slow, the CI job you triggered in the step above will create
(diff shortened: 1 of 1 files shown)


doc/release-checklist.md
```diff
@@ -45,13 +45,15 @@ See this page for a good overview: https://deps.rs/repo/github/sharkdp/bat
 
 ## Release
 
-- [ ] Create a tag and push it to the remote `git tag vX.Y.Z; git push --tags`.
+- [ ] Create a tag and push it to the remote `git tag vX.Y.Z; git push origin vX.Y.Z`.
```
sharkdp, line 48: Maybe this is bad practice, but I do sometimes tend to name branches after upcoming releases, i.e. with the same name as the tag. In that case, I would need to `git push origin tag vX.Y.Z`.

doc/release-checklist.md
```diff
@@ -45,13 +45,15 @@ See this page for a good overview: https://deps.rs/repo/github/sharkdp/bat
...
       This will trigger the deployment via GitHub Actions.
 - [ ] Go to https://github.com/sharkdp/bat/releases/new to create the new
       release. Select the new tag and also use it as the release title. For the
       release notes, copy the corresponding section from `CHANGELOG.md` and
       possibly add additional remarks for package maintainers.
       Publish the release.
+      NOTE: If you are too slow, the CI job you triggered in the step above will create
+      the release for you, and the author of the release will become **github-actions**.
```
sharkdp, line 56: Note that release notes, title etc. can be updated after creating the release. So you can quickly create the release and then take your time to edit the release notes.

doc/release-checklist.md
Enselic: I would agree it is bad practice :)

But doesn't hurt to change the command to what works in even more cases, so I'll do that.

doc/release-checklist.md
Enselic: One downside of that is that the notification email that GitHub sends for new releases is without release notes. Edits to release notes are not distributed to watchers of the project.

So I wonder: Maybe we should change CI to trigger the `Publish archives and packages` job via the creation of a release, and not trigger via push of a tag? That way you can take your time to edit release notes, and there is no timing issues.

sharkdp: > One downside of that is that the notification email that GitHub sends for new releases is without release notes. Edits to release notes are not distributed to watchers of the project.

I didn't even know that releases trigger email notifications :-)



> So I wonder: Maybe we should change CI to trigger the `Publish archives and packages` job via the creation of a release, and not trigger via push of a tag? That way you can take your time to edit release notes, and there is no timing issues.

If that's possible, great :+1:. Otherwise, we can adapt the release notes such that maintainers are asked to prepare the release notes upfront. This was actually part of my original list, but with `CHANGELOG.md` it has become much easier, so I removed it at some point.

Enselic: I don't think you can create a release without first pushing a tag though? Since you have to select for what tag to make a release. But maybe you can select a temporary tag and change to the real tag later :thinking: 

Anyway, good to know you are open to triggering the CI IS_RELEASE step via creation of a release, instead of creation of a tag (note that the tag will still be needed and be pushed, it would just not trigger the IS_RELEASE part of CI). I can imagine myself experimenting with that at some point later. Maybe as part of the doing the next release when it is time.

I'll undo my changes here, and just keep the updated git push origin tag vX.Y.Z change.

sharkdp (approved):

PR #1825, opened 2021-08-27
Make asset compression optional at compile time

Enselic:
This will be needed to later support zero-copy deserialization of independent syntax sets (for which we will not use compression on the "outer" lookup data structure), but is interesting and useful on its own.

Instead of defering serialization and de-serialization to syntect, we implement it ourselves in the same way, but make compression optional.

I'd like to do a bit more verification before merge, but this is tested enough to be ready for detailed code review.


src/assets.rs
```diff
@@ -307,18 +314,41 @@ pub(crate) fn get_serialized_integrated_syntaxset() -> &'static [u8] {
 }
 
 pub(crate) fn get_integrated_themeset() -> ThemeSet {
-    from_binary(include_bytes!("../assets/themes.bin"))
+    from_binary(include_bytes!("../assets/themes.bin"), COMPRESS_THEMES)
+}
+
+pub(crate) fn from_binary<T: serde::de::DeserializeOwned>(v: &[u8], compressed: bool) -> T {
+    asset_from_contents(v, "n/a", compressed)
+        .expect("data integrated in binary is never faulty, but make sure `compressed` is in sync!")
```
sharkdp, line 322: I was shortly thinking about this before reading this piece of code. Maybe we could use the asset metadata file to store whether or not the binaries are compressed? This way, we could even make a runtime decision.

sharkdp (approved):
Looks good to me!

src/assets.rs
Enselic: That's certainly a tempting feature to introduce at some point!

Enselic:
With #1829 now being part of regressions tests, I am confident I did not break anything. Merging.

PR #1826, opened 2021-08-27
Add PowerShell completion file

rashil2000:


rashil2000:
@Enselic The file extension needs to be `.ps1`, otherwise PowerShell won't be able to source it.

Enselic:
@rashil2000 Sorry to mess up the PR, I thought I knew what I was doing, but now CI fails. Any chance you can take over, since you know what needs to be done? My goal was to package the completion file in the tar.gzs and .zips that our CI pipeline build. You can download CI artifacts here: https://github.com/sharkdp/bat/actions/runs/1174961314. But of course the link is different for each run.

rashil2000:
The CI was failing because the file `bat.pwsh` didn't exist. It's fixed now.

.github/workflows/CICD.yml
```diff
@@ -285,6 +285,7 @@ jobs:
         # Autocompletion files
         cp 'target/${{ matrix.job.target }}/release/build/${{ env.PROJECT_NAME }}'-*/out/assets/completions/bat.bash "$ARCHIVE_DIR/autocomplete/${{ env.PROJECT_NAME }}.bash"
         cp 'target/${{ matrix.job.target }}/release/build/${{ env.PROJECT_NAME }}'-*/out/assets/completions/bat.fish "$ARCHIVE_DIR/autocomplete/${{ env.PROJECT_NAME }}.fish"
+        cp 'target/${{ matrix.job.target }}/release/build/${{ env.PROJECT_NAME }}'-*/out/assets/completions/bat.ps1 "$ARCHIVE_DIR/autocomplete/_${{ env.PROJECT_NAME }}.ps1"
```
Enselic, line 288: Double-check: Should the `_` really be here? That makes the file be packaged as `_bat.ps1`

rashil2000: It is intentional. I think it's a convention for PowerShell scripts - for example see here https://github.com/sharkdp/fd/blob/3ba90dd7685a73923a3a872c4ffc8756a198d755/.github/workflows/CICD.yml#L189

rashil2000: I think I should add that underscore in build.rs itself, what do you think?

Enselic: Yes, if the convention is to prefix with `_`, I agree it makes sense to let that be the output of `build.rs` :+1:

Enselic: I would go as far as to say the input file should be called `_bat.ps1.in`

rashil2000: Done

Enselic:
Thanks! I can confirm that `_bat.ps1` is packaged in e.g. `bat-v0.18.3-i686-pc-windows-msvc.zip` now.

Can you also add an entry about this to `CHANGELOG.md` as per instructions in https://github.com/sharkdp/bat/blob/master/CONTRIBUTING.md#add-an-entry-to-the-changelog please? Thanks in advance!

(I'm setting Approved already now)

Enselic (approved):

sharkdp (approved):
Thank you

Hrxn:
Maybe it should also be mentioned how to use the PowerShell completion file?
I mean, I don't think that this is immediately obvious..

rashil2000:
Completion files are usually dot-sourced. For example you can put this line in your `$PROFILE`:
```pwsh
. /file/path/to/_bat.ps1
```

Hrxn:
I'm aware, I was just a bit surprised because this does not seem to be mentioned anywhere on the project site. Although it could arguably be considered part of the install instructions, in my opinion.

For what's it worth, it could be considered presumed knowledge for a PowerShell user, but still..

Speaking of presumed knowledge, why does the name of that file start with an underscore character (`_`)?
Is that some sort of new PowerShell convention as well?

rashil2000:
> Is that some sort of new PowerShell convention as well?

It does seem like a convention. I've seen it in many places.

PR #1829, opened 2021-08-30
Add regression testing for the custom assets functionality

Enselic:
Our current CI pipeline does not regression test the custom assets functionality at all. We get a [green](https://github.com/Enselic/bat/runs/3459232947) pipeline even if we replace `HighlitingAssets::from_cache` with this:

```rust
    pub fn from_cache(cache_path: &Path) -> Result<Self> {
        #[allow(unused_variables)]
        let res = Ok(HighlightingAssets::new(
            SerializedSyntaxSet::FromFile(cache_path.join("syntaxes.bin")),
            asset_from_cache(&cache_path.join("themes.bin"), "theme set")?,
        ));

        panic!("does this code run in CI?");

        #[allow(unreachable_code)]
        res
    }
```

This PR adds code to regression test the custom assets functionality. It can be seen as preparatory work for the verification of #1825, but is of course useful on its own.

### DIFF
--- a/tests/syntax-tests/test_custom_assets.sh
+++ b/tests/syntax-tests/test_custom_assets.sh
@@ -1,43 +1,59 @@
 #!/usr/bin/env bash
 set -o errexit -o nounset -o pipefail
 
-test_file="tests/syntax-tests/source/BatTestCustomAssets/NoColorsUnlessCustomAssetsAreUsed.battestcustomassets"
+### HELPER VARS
 
-# First make sure our test syntax is not part of integrated syntaxes
-echo "Testing that BatTestCustomAssets is an unknown syntax:"
-bat -f --language BatTestCustomAssets "${test_file}" &&
-    echo "should have failed because of unknown syntax" &&
-    exit 1
-
-# Now build custom assets and include our test syntax
-# We do as instructed to regular users at
-# https://github.com/sharkdp/bat/blob/master/README.md#adding-new-syntaxes--language-definitions
-
-# Create dir to put our test syntax
 custom_syntaxes_dir="$(bat --config-dir)/syntaxes"
+
+installed_custom_syntax="${custom_syntaxes_dir}/BatTestCustomAssets.sublime-syntax"
+
+custom_syntax_args=(
+    "--language=BatTestCustomAssets"
+    "tests/syntax-tests/source/BatTestCustomAssets/NoColorsUnlessCustomAssetsAreUsed.battestcustomassets"
+)
+
+integrated_syntax_args=(
+    "--language=Rust"
+    "examples/simple.rs"
+)
+
+### HELPER FUNCTIONS
+
+echo_step() {
+    echo -e "\n== $1 =="
+}
+
+fail_test() {
+    echo "FAIL: $1"
+    exit 1
+}
+
+### TEST STEPS
+
+echo_step "TEST: Make sure 'BatTestCustomAssets' is not part of integrated syntaxes"
+bat -f "${custom_syntax_args[@]}" &&
+    fail_test "EXPECTED: 'unknown syntax' error ACTUAL: no error occured
+               HINT: try manually doing 'bat cache --clear' before running tests"
+
+echo_step "PREPARE: Install custom syntax 'BatTestCustomAssets'"
 mkdir -p "${custom_syntaxes_dir}"
+cp -v "tests/syntax-tests/BatTestCustomAssets.sublime-syntax" "${installed_custom_syntax}"
 
-# Put the syntax in place
-cp tests/syntax-tests/BatTestCustomAssets.sublime-syntax "${custom_syntaxes_dir}/BatTestCustomAssets.sublime-syntax"
-
-# Build custom assets to include the above syntax
-echo "Building custom assets to include the BatTestCustomAssets syntax"
+echo_step "PREPARE: Build custom assets to enable 'BatTestCustomAssets' syntax"
 bat cache --build
 
-# Now bat shall not fail when the syntax is used. If it does not fail, we can be reasonably sure
-# that the custom assets functionality is working
-echo "Testing that BatTestCustomAssets is a KNOWN syntax:"
-bat -f --language BatTestCustomAssets "${test_file}" ||
-    (echo "command shall not have failed this time, but did" &&
-        exit 1)
+echo_step "TEST: 'BatTestCustomAssets' is a known syntax"
+bat -f "${custom_syntax_args[@]}" ||
+    fail_test "EXPECTED: syntax highlighting to work ACTUAL: there was an error"
 
-# While we're at it  and for extra safety in the test, we also
-# make sure that the --no-custom-assets flag work as intended
-echo "Testing that BatTestCustomAssets is an unknown syntax when using --no-custom-assets:"
-bat -f --no-custom-assets --language BatTestCustomAssets "${test_file}" &&
-    echo "should have failed because of unknown syntax via --no-custom-assets" &&
-    exit 1
+echo_step "TEST: The 'Rust' syntax is still available"
+bat -f "${integrated_syntax_args[@]}" ||
+    fail_test "EXPECTED: syntax highlighting to still work for integrated assets ACTUAL: there was an error"
 
-# We clean up after ourselves to reduce risk of problems later
+echo_step "TEST: 'BatTestCustomAssets' is an unknown syntax with --no-custom-assets"
+bat -f --no-custom-assets "${custom_syntax_args[@]}" &&
+    fail_test "EXPECTED: 'unknown syntax' error because of --no-custom-assets, but no error occured"
+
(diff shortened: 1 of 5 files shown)


tests/syntax-tests/test_custom_assets.sh
```diff
@@ -0,0 +1,43 @@
...
+echo "Building custom assets to include the BatTestCustomAssets syntax"
+bat cache --build
+
+# Now bat shall not fail when the syntax is used. If it does not fail, we can be reasonably sure
+# that the custom assets functionality is working
+echo "Testing that BatTestCustomAssets is a KNOWN syntax:"
+bat -f --language BatTestCustomAssets "${test_file}" ||
+    (echo "command shall not have failed this time, but did" &&
+        exit 1)
+
```
Enselic, line 33: I just realised it would also be a good idea to sanity check an existing syntax, to verify that the custom syntax is not the ONLY syntax that is now available (i.e. verify that `include_integrated_assets` work)

Enselic: @sharkdp Would you mind doing a high level code review of this PR before I do the above?

sharkdp: This looks great - thank you! In principle, we would like to run this on every supported platform, right? (not sure if we can run bash scripts on Windows)

keith-hall: Not sure about from CI, but "Git Bash for Windows" is a thing  :) (obviously the "git" part isn't so relevant here)

Enselic: Running bash on Windows is no issue, because we already do it a lot. The entire 'Create tarball' step in CI is written in bash for example, but runs just fine on Windows (`windows-2019` to be precise).

In principle, yes we probably should run this in the matrix. But I think the risk of windows-specific problems is low enough to postpone looking into that.

Enselic:
Thanks to both of you for taking a look! This PR is now ready for detailed code review.

tests/syntax-tests/test_custom_assets.sh
```diff
@@ -0,0 +1,59 @@
...
+fail_test() {
+    echo "FAIL: $1"
+    exit 1
+}
+
+### TEST STEPS
+
+echo_step "TEST: Make sure 'BatTestCustomAssets' is not part of integrated syntaxes"
+bat -f "${custom_syntax_args[@]}" &&
```
Enselic, line 34: If you are unfamiliar with the `"${custom_syntax_args[@]}"` syntax, you are in good company. But it is what is recommended by https://github.com/koalaman/shellcheck/wiki/SC2086#exceptions

tests/syntax-tests/test_custom_assets.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
+set -o errexit -o nounset -o pipefail
+
+### HELPER VARS
+
+custom_syntaxes_dir="$(bat --config-dir)/syntaxes"
+
+installed_custom_syntax="${custom_syntaxes_dir}/BatTestCustomAssets.sublime-syntax"
```
sharkdp, line 8: This shell script will modify the local environment. Even if it cleans up in the end, there might be changes to the environment (e.g. existing serialized custom assets will be lost).

I wonder if we should add a comment at the top or rename it slightly (`ci_pipeline_test_…`) to make sure that this is not intended to be ran locally.

Another way would be to perform everything in a temporary directory. By setting `BAT_CONFIG_DIR`/`BAT_CACHE_DIR` (or respective command line options) appropriately.

A third option could be to move everything inside a `script:` block in the CI configuration (maybe inside a new yaml file).

tests/syntax-tests/test_custom_assets.sh
```diff
@@ -0,0 +1,59 @@
...
+bat -f "${integrated_syntax_args[@]}" ||
+    fail_test "EXPECTED: syntax highlighting to still work for integrated assets ACTUAL: there was an error"
+
+echo_step "TEST: 'BatTestCustomAssets' is an unknown syntax with --no-custom-assets"
+bat -f --no-custom-assets "${custom_syntax_args[@]}" &&
+    fail_test "EXPECTED: 'unknown syntax' error because of --no-custom-assets, but no error occured"
+
+echo_step "CLEANING: To avoid unwanted side effects of running tests"
+bat cache --clear
+rm -rfv "${installed_custom_syntax}"
```
sharkdp, line 59: Why `-r`?

tests/syntax-tests/test_custom_assets.sh
Enselic: Making use of `BAT_CONFIG_DIR`/`BAT_CACHE_DIR` is a great idea. Will fix.

tests/syntax-tests/test_custom_assets.sh
Enselic: Habit :sweat_smile: Will fix!

Enselic: (we need it in the new code)

tests/syntax-tests/test_custom_assets.sh
Enselic: I think it is important to be able to run the script locally. That way it is easy to use it for local verification, but it is also easy to work on the script. So the idea to use the env vars was a great idea. It turned me into such a good mode that I also added another test that makes sure `bat cache --clean` removes all files.

Enselic:
All comments taken care of. Looking forward to another review round.

sharkdp:
Thank you for the update and the extra test.

If we want developers to run this script locally in their 'dirty' environment, there are some more things that need to be taken care of. We should probably pass `--no-config` and also remove the `BAT_*` environment variables that could be relevant. We might actually be able to skip the second part here, as none of them could interfere. See similar code in

https://github.com/sharkdp/bat/blob/960219591028984053a80fa6aeb3e506e9d791e7/tests/integration_tests.rs#L34-L51

and

https://github.com/sharkdp/bat/blob/960219591028984053a80fa6aeb3e506e9d791e7/tests/syntax-tests/create_highlighted_versions.py#L48-L56

Enselic:
Hmm. Not sure I agree it is a good idea to spend time on making the script run in any thinkable environment. There is a difference between "being able to run the script locally" and "making sure the script works, no matter in what environment it is run".

There are many facets to my opinion.
* I consider the CI environment to be the "reference" environment. If a developer wants to run tests, I do not think it is unreasonable to require the developer to make their environment comparable to CI. Or run test in a docker container.
* It will take considerable effort to make the script run in any thinkable environment. Effort that I think is better spent elsewhere.
* It will complicate the code and make it harder to understand and change.

So why did I think it was a good idea to make use of e.g. `BAT_CONFIG_DIR` and `BAT_CACHE_PATH` then? Because:
* As with so many things in software development, things are not black and white. I do think it makes sense to prevent casual explorers of the git repo from destructively clearing their cache.
* I think it adds clarity and predictability to the test, without overdoing it.

Admittedly, my opinion might be a bit tainted by my focus on #951. Spending more time on this script would to me be a distraction from #951 rather than help us get #951 fixed quicker. At this point, I just want to get this PR merged, so that I can merge #1825, so that I can update #1787.

sharkdp:
> I consider the CI environment to be the "reference" environment. If a developer wants to run tests, I do not think it is unreasonable to require the developer to make their environment comparable to CI. Or run test in a docker container.

Hm, yes. But as long as we don't provide a clean docker environment for testing out of the box, it's rather inconvenient for developers to modify their local environment to make it "CI-like". It's not unreasonable to have a `bat` config file as a developer and maybe to have a few `BAT_*` environment variables locally set.

I didn't add all of this clean up code just because I was thinking of hypothetical scenarios. I added it because tests *were* locally failing in some environments.

That being said, for this particular script, I don't think we need more clean-up work. And if we missed something, we can also add it in later. It *is* possible to break the tests (e.g. by setting `BAT_OPTS="--no-custom-assets"`), but that's a constructed example of course.

> It will take considerable effort to make the script run in any thinkable environment. Effort that I think is better spent elsewhere.

Partially agreed. I have spent time in the past figuring out why I couldn't reproduce a bug report. Or a test failure. Time that I could have saved by proactively making sure that the environment is clean/comparable.

For testing, it would be ideal to have a completely side-effect-free program. Somewhat unfortunately, `bat` has grown to be an application that is quite the opposite. It is very much dependent on its environment (env. variables, terminal properties, less version, OS flavor, config file, …).

Again - not speaking of this PR here, but more generally.

> Admittedly, my opinion might be a bit tainted by my focus on #951. Spending more time on this script would to me be a distraction from #951 rather than help us get #951 fixed quicker. At this point, I just want to get this PR merged, so that I can merge #1825, so that I can update #1787.

That's completely fair. I don't want to slow you down. This PR is a great addition for our tests.

sharkdp (approved):

Enselic:
Thanks for being pragmatic despite making valid counter-points 👍 

Merging.

PR #1831, opened 2021-08-31
CICD: Put documentation testing in its own job

Enselic:
The time varies quite a lot, but instead of 10 - 13 min, it will take 6 - 7 min. We achieve the speedup by doing the following:
* Stop buildling e.g. `git2` by using `--features minimal-application,build-assets`.
* Stop running `cargo test` in "Syntaxes and themes regression tests". We run tests several times in other jobs, and there is no need to run them again in "Syntaxes and themes regression tests".
* To be able to get rid of `cargo test` completely, we convert "All themes are present" from an `#[ignore]` cargo test to a script that we run in CI. (I think it is quite nice to not have any `#[ignore]`d tests left, even if its just semantics.)
* Move documentation testing in its own job, since it does not depend on things we do in the syntaxes and themes regression test. This speeds up the syntaxes and themes regression test by about another minute.

Bonus side effect: Instead of having three files in the repo named `assets.rs`, we now only have two :)


sharkdp:
> Stop running `cargo test` in "Syntaxes and themes regression tests". We run tests several times in other jobs, and there is no need to run them again in "Syntaxes and themes regression tests".

But there is a difference! Some tests do rely on the available syntaxes (e.g. `no_duplicate_extensions`). This is why I wanted to run `cargo test` again after building the new assets. This way, we would immediately see a CI error if someone adds a syntax with a duplicate extension - and not just once we create the release.

Enselic:
> But there is a difference!

Right, my bad. I wasn't familiar enough with the existence of these tests to get that insight. I should have looked through all tests beforehand. It is definitely a good idea to run e.g. `no_duplicate_extensions` for each commit with latest assets.

I've turned this PR into a PR to "CICD: Put documentation testing in its own job" instead, which I think is uncontroversial, and still a worthwhile thing to do, by my estimation.

sharkdp:
> > But there is a difference!
> 
> Right, my bad. I wasn't familiar enough with the existence of these tests to get that insight. I should have looked through all tests beforehand. It is definitely a good idea to run e.g. `no_duplicate_extensions` for each commit with latest assets.

Sorry, I wasn't intending to sound this "commanding" :-).

> I've turned this PR into a PR to "CICD: Put documentation testing in its own job" instead, which I think is uncontroversial, and still a worthwhile thing to do, by my estimation. It makes the existing job finish about one minute faster, as well as creating a job for later adding e.g. steps to make the docs easy to read by for example publish the docs as an artifact.

:+1:

sharkdp (approved):

PR #1832, opened 2021-08-31
README.md: Don't recommend --locked with cargo install bat

Enselic:
In our `README.md`, we currently say this:

--- START QUOTE ---

### From source

If you want to build `bat` from source, you need Rust 1.45 or higher. You can then use `cargo` to build everything:

```bash
cargo install --locked bat
```

--- END QUOTE ---

But passing `--locked` does not make sense when installing from a crate, since `Cargo.toml` and `Cargo.lock` can not be out of sync. (I remember being confused about `--locked` when I first started out with Rust, and I think the current advice risks confusing more than it helps.) So this PR removes `--locked` from all versions of `README.md`

To be extra sure, I have verified that `cargo install bat` is enough, by doing:

```
% docker run -it debian:stable
# apt update
# apt install build-essential curl
# curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh # with defaults
# source $HOME/.cargo/env
# cargo install bat
# bat ~/.cargo/registry/src/github.com-1ecc6299db9ec823/bat-0.18.3/examples/simple.rs # works!
```


Enselic:
I'm happy with this PR now. We already have a section on [development](https://github.com/sharkdp/bat/blob/master/README.md#development), so no need to have another section on "From source" I think. Instead call it "From crates.io" and adjust the text a bit.

Here is the commit message I will use:

```
README.md: Don't recommend --locked with cargo install bat

It does not make sense with `--locked` when installing from a crate, since
`Cargo.toml` and `Cargo.lock` can not be out of sync.

Also stop pointing towards a `build` folder since it is tricky to find when
installing bat from its crate. It just exists temporarily at a path looking
something like this:
/tmp/cargo-installYa2TPV/release/build/bat-367c2f59177968a0/output
```

sharkdp:
> But passing `--locked` does not make sense when installing from a crate, since `Cargo.toml` and `Cargo.lock` can not be out of sync. (I remember being confused about `--locked` when I first started out with Rust, and I think the current advice risks confusing more than it helps.)

Maybe I'm also confused about the meaning of `--locked`, but I understood it as: "install the *exact* version of each dependency that is specified in `Cargo.lock`. Do not attempt to install newer versions, even if they are compatible in terms of the ranges specified in `Cargo.toml`.

Quoting from https://doc.rust-lang.org/cargo/commands/cargo-install.html:

> By default, the Cargo.lock file that is included with the package will be ignored. This means that Cargo will recompute which versions of dependencies to use, possibly using newer versions that have been released since the package was published. The --locked flag can be used to force Cargo to use the packaged Cargo.lock file if it is available. This may be useful for ensuring reproducible builds, to use the exact same set of dependencies that were available when the package was published. It may also be useful if a newer version of a dependency is published that no longer builds on your system, or has other problems. The downside to using --locked is that you will not receive any fixes or updates to any dependency.

I would personally rather recommend the usage of `--locked` in order to be sure that users will definitely get a compiling and working version of `bat`. Yes, there might be upstream bugs that could be fixed by removing `--locked`, but breakage is more likely (even if it shouldn't happen if everyone is using semver correctly).

Enselic:
It is not you that's confused, it is me. I am starting to get over-confident in thinking I know by now how cargo works, and not double-checking assumptions. I'll just go ahead and close this PR. Thanks for taking a look and apologies for wasting everyone's time.

sharkdp:
No worries

Issue #1611, opened 2021-04-15
Support increments with `--highlight-line`


XPhyro:
For periodic data, such as in `git log` or say in a CSV file, being able to pass an increment to the `--highlight-line` option would be helpful. I suppose the syntax could be `--highlight-line 30:40:2`. This would highlight every second line from line 30 to line 40. As far as I am aware, this syntax is fully backwards compatible.

XPhyro:
I suppose `--line-range` could also benefit from this.

t-mangoe:
Hi. I tried implementing. I implemented codes so that you can pass the value of increment.
How about this?
![スクリーンショット 2021-08-21 18 00 46](https://user-images.githubusercontent.com/22494246/130316895-5bb2e039-6a53-48b1-bac9-086c3db4c42e.png)

XPhyro:
Hi. This is exactly what I had in mind. Thank you.

I suppose this is not yet in the upstream, right?

t-mangoe:
@XPhyro 
Thank you for your comment.
Yes, I have not sent a pull request yet, so this is not in the upstream.
I'm planning to add tests and then make a pull request.
Maybe it may take some time. I'm glad if you could wait patiently.

XPhyro:
I definitely could wait, no worries. Thanks again for the feature.

sharkdp:
Could you please explain your usecase? I'm not really sure what I would use this for.

XPhyro:
Sure. To start with, I cannot think of any uses for general code; but for text data such as system logs or a CSV file (which are already supported), this can be used to highlight or only show milestones, or to reduce the number of shown data points.

To be more concrete, at the time of my opening the issue, I wanted to quickly verify some voltage vs. wavelength vs. radius data that I had in CSV format. Since the data was sorted and mostly periodic, I could have used `bat --highlight-line 200:300:10` (or `--line-range`) to see the data points where the voltage was between X and Y, with I increments.

Since the data was in text, a general command-line tool would have been the easiest and fastest, with the added bonus of bat's highlighting.

sharkdp:
Thank you.

I'm not 100% sold, but also not strictly against adding this. I'll let the other maintainers decide. @Enselic expressed interest in #1841. CC @eth-p @keith-hall

Enselic:
I think the feature request is reasonable, especially since apparently at least two persons found interest in it.

[...]

PR #1841, opened 2021-09-04
Add increment option in `--highlight-line` and `--line-range`

t-mangoe:
This pull request adds increment option in `--highlight-line` and `--line-range`.
By this PR, you will become able to pass an increment value to these options, for example `--highlight-line 30:40:2`.

This PR solves this #1611 issue.

スクリーンショット 2021-08-21 18 00 46

### DIFF
--- a/src/bin/bat/app.rs
+++ b/src/bin/bat/app.rs
@@ -216,8 +216,12 @@ impl App {
                 _ => VisibleLines::Ranges(
                     self.matches
                         .values_of("line-range")
-                        .map(|vs| vs.map(LineRange::from).collect())
-                        .transpose()?
+                        .map(|vs| {
+                            vs.map(LineRange::from)
+                                .flatten()
+                                .flatten()
+                                .collect::<Vec<_>>()
+                        })
                         .map(LineRanges::from)
                         .unwrap_or_default(),
                 ),
@@ -229,8 +233,12 @@ impl App {
             highlighted_lines: self
                 .matches
                 .values_of("highlight-line")
-                .map(|ws| ws.map(LineRange::from).collect())
-                .transpose()?
+                .map(|ws| {
+                    ws.map(LineRange::from)
+                        .flatten()
+                        .flatten()
+                        .collect::<Vec<_>>()
+                })
                 .map(LineRanges::from)
                 .map(HighlightedLineRanges)
                 .unwrap_or_default(),
--- a/src/bin/bat/app.rs
+++ b/src/bin/bat/app.rs
@@ -216,13 +216,7 @@ impl App {
                 _ => VisibleLines::Ranges(
                     self.matches
                         .values_of("line-range")
-                        .map(|vs| {
-                            vs.map(LineRange::from)
-                                .flatten()
-                                .flatten()
-                                .collect::<Vec<_>>()
-                        })
-                        .map(LineRanges::from)
+                        .map(App::get_line_ranges)
                         .unwrap_or_default(),
                 ),
             },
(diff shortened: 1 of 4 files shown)


src/line_range.rs
```diff
@@ -59,15 +62,29 @@ impl LineRange {
...
             }
+            3 => {
+                new_range.lower = line_numbers[0].parse()?;
+                new_range.upper = line_numbers[1].parse()?;
+                new_range.increment = Some(line_numbers[2].parse()?);
+                Ok(new_range)
+            }
             _ => Err(
-                "Line range contained more than one ':' character. Expected format: 'N' or 'N:M'"
+                "Line range contained more than one ':' character. Expected format: 'N' or 'N:M' or 'N:M:L'"
```
Enselic, line 72: "Line range contained more than one ':' character." needs to be reworded

tests/integration_tests.rs
```diff
@@ -224,6 +224,16 @@ fn line_range_multiple() {
         .stdout("line 1\nline 2\nline 4\n");
 }
 
+#[test]
+fn line_range_increment() {
```
Enselic, line 228: I love that you added tests to the PR :+1:

src/line_range.rs
```diff
@@ -4,13 +4,15 @@ use crate::error::*;
 pub struct LineRange {
     lower: usize,
     upper: usize,
+    increment: Option<usize>,
```
Enselic, line 7: I saw there was a discussion in #1611 about the use case for this feature, but it seems to me like it can be useful.

Maybe you want to let that discussion settle before working on the code more.

Brainstorming: I wonder if the code wouldn't become nicer and more modular if you split up the range in several `LineRange`s, rather than adding an `increment` special case to `LineRange`.

tests/integration_tests.rs
t-mangoe: Thank you!!

src/line_range.rs
t-mangoe: Thank you for your comment.
I think it is good to split up the range in several `LineRange`s, too.
I'm considering fixing the code, but I'm not good at rust programming skill.
I'm happy for you to tell me advice or help comment.

src/line_range.rs
t-mangoe: I changed the syntax to `sed` style.
So, I reverted this error message.

src/line_range.rs
t-mangoe: I changed the code to split up the range in several `LineRange`s.

src/bin/bat/app.rs
```diff
@@ -216,8 +216,12 @@ impl App {
                 _ => VisibleLines::Ranges(
                     self.matches
                         .values_of("line-range")
-                        .map(|vs| vs.map(LineRange::from).collect())
```
Enselic, line 219: Would be great if you could de-duplicate the code for line-range and highlight-line

src/line_range.rs
```diff
@@ -74,34 +92,45 @@ impl LineRange {
 #[test]
 fn test_parse_full() {
     let range = LineRange::from("40:50").expect("Shouldn't fail on test!");
```
Enselic, line 94: I think we need think more about the interface here. I would expect `LineRange::from("40:50")` to only return a single line range.

Enselic (changes requested):
I took a quick look. Unfortunately I don't have time to deep dive into the code at this point, but occasionally I can take a look and give you my spontaneous thoughts and feedback.

There is a risk that this will lead to a bit back and forth before we land at a final solution. I hope that works for you.

Enselic:
Closing this for now. Feel free to reopen if you resume work on it.

PR #1842, opened 2021-09-06
Consolidate environment variable lists in tests

sharkdp:
We want to make sure that all of our test environments are clean from
possible outside modification. This consolidates the list of used
environment variables in Rust-based and Python-based integration tests.

Note that there is also a similar list in `src/bin/bat/main.rs` which
is even more exhaustive (for bug report collection). However, some
of these variables can not possibly have an effect on test environments.

### DIFF
--- a/tests/syntax-tests/create_highlighted_versions.py
+++ b/tests/syntax-tests/create_highlighted_versions.py
@@ -47,12 +47,16 @@ def create_highlighted_versions(output_basepath):
     ):
         try:
             env = os.environ.copy()
-            env.pop("PAGER", None)
-            env.pop("BAT_PAGER", None)
+            env.pop("BAT_CACHE_PATH", None)
+            env.pop("BAT_CONFIG_DIR", None)
             env.pop("BAT_CONFIG_PATH", None)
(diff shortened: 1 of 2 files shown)


tests/syntax-tests/create_highlighted_versions.py
```diff
@@ -47,12 +47,16 @@ def create_highlighted_versions(output_basepath):
...
+            env.pop("BAT_CACHE_PATH", None)
+            env.pop("BAT_CONFIG_DIR", None)
             env.pop("BAT_CONFIG_PATH", None)
+            env.pop("BAT_OPTS", None)
+            env.pop("BAT_PAGER", None)
             env.pop("BAT_STYLE", None)
-            env.pop("BAT_THEME", None)
             env.pop("BAT_TABS", None)
+            env.pop("BAT_THEME", None)
+            env.pop("NO_COLOR", None)
```
Enselic, line 58: `COLORTERM` is missing here, right?

Enselic: it's two lines down :sweat_smile:

Enselic (approved):

PR #1843, opened 2021-09-06
Parallelize syntax regression tests

sharkdp:
The syntax highlighting regression tests can be trivially parallelized.
On my notebook (8 core), this results in a 3.9x speedup.

### DIFF
--- a/tests/syntax-tests/create_highlighted_versions.py
+++ b/tests/syntax-tests/create_highlighted_versions.py
@@ -6,7 +6,7 @@ import sys
 import os.path as path
 import os
 import argparse
-
+from multiprocessing import Pool
 
 BAT_OPTIONS = [
     "--no-config",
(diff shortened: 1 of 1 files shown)


Enselic:
Hmm, no difference on my dual core *Intel(R) Celeron(R) J4005 CPU @ 2.00GHz*, even tough `os.cpu_count()` which `Pool` uses is `2` on my system. I'll try to figure out why and review the code in more detail later. Maybe it's simply because it is mostly IO bound. But it is interesting that you only saw a 4x speedup instead of 8x.

On a quick glance the code looks good, so feel free to merge if you don't want to wait for a detailed review.

sharkdp:
On the CI workers, time has improved from 10 seconds to 6 seconds. A huge win! :smile:

sharkdp:
> Hmm, no difference on my dual core _Intel(R) Celeron(R) J4005 CPU @ 2.00GHz_, even tough `os.cpu_count()` which `Pool` uses is `2` on my system. I'll try to figure out why

> But it is interesting that you only saw a 4x speedup instead of 8x.

Not sure, but I would guess it's just coincidence. Parts of `bat` are parallelized (e.g. during startup), so we can't expect it to be 8x faster.

> Maybe it's simply because it is mostly IO bound.

I don't think so. `cat` is I/O bound, `bat` isn't (yet :smile:).

> On a quick glance the code looks good, so feel free to merge if you don't want to wait for a detailed review.

There is no rush to merge it, but I'm not sure if it's worth spending too much time on the review. The diff looks large but I only changed a `for` loop to a `Pool.map`.

keith-hall (approved):

PR #1844, opened 2021-09-07
Add context to .ino configuration

KevinJohnMulligan:
This allows a user to search the README or the entire repository for "Arduino" to find the necessary configuration.


KevinJohnMulligan:
I realise this change is trivial and searching ".ino" will already give the correct result however I spent quite a lot of time trying to figure out how to add a new syntax. So this might save some people some time.

Enselic (approved):
Thanks for taking the time to make it easier for people to find this info in the future!

.ino indeed is an Arduino file format: https://fileinfo.com/extension/ino

Not sure if the grammar ends up correct in Korean, but I guess someone that knows Korean can correct it later.

This PR looks good to me.

sharkdp (approved):

PR #1850, opened 2021-09-13
Extract some private submodules from 'bat::assets'

Enselic:
As I continued working on #951, `src/assets.rs` grew uncomfortably large. So I figured it would be a good idea to first split it up into private submodules without changing the public API.

This is ready for detailed review, but as usual I'd like to do some final verification before merge.

(I use this hacky command to verify that the public API is not changed:

```
% rm -f /tmp/all-*.html && for commit in master introduce-internal-assets-submodules; do git ch $commit && cargo clean -p bat && cargo doc --no-deps --all-features && cp target/doc/bat/all.html /tmp/all-$commit.html; done && diff -u /tmp/all-*.html | diff-so-fancy
```
)


keith-hall (approved):

Enselic:
Sorry for the last minute addition,  but I got some extra time this evening to also throw in `ignored_suffixes.rs` in this PR. I know from continued prototyping on #951 that it will be useful to have this code generic, so it can be reused in `MinimalAssets`.

src/assets/ignored_suffixes.rs
```diff
@@ -0,0 +1,43 @@
...
+];
+
+/// If we find an ignored suffix on the file name, e.g. '~', we strip it and
+/// then try again without it. Note that we do this recursively.
+pub fn try_with_stripped_suffix<T, F>(file_name: &OsStr, func: F) -> Result<Option<T>>
+where
+    F: Fn(&OsStr) -> Result<Option<T>>,
+{
+    let file_path = Path::new(file_name);
+    let mut syntax = None;
```
keith-hall, line 33: is this `syntax` variable name still correct?

Enselic: Thanks, good catch. I'll push a fix as soon as I get a green CI on the current code, to make sure the refactoring itself did not break anything.

Enselic:
I'll go ahead and merge this now since:
* Keeping it open has quite high risk of conflicts
* I'm happy with the code
* Keith is happy with the code :)

PR #1851, opened 2021-09-13
Add some highlighting assets helpers

Enselic:
Reduce code duplication to ease solving #951.

(Expect to see more refactoring PRs soon)

### DIFF
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -306,6 +306,13 @@ impl HighlightingAssets {
             .map(|syntax| SyntaxReferenceInSet { syntax, syntax_set }))
     }
 
+    fn find_syntax_by_extension(&self, extension: &str) -> Result<Option<SyntaxReferenceInSet>> {
+        let syntax_set = self.get_syntax_set()?;
+        Ok(syntax_set
+            .find_syntax_by_extension(extension)
+            .map(|syntax| SyntaxReferenceInSet { syntax, syntax_set }))
+    }
+
     fn get_extension_syntax(&self, file_name: &OsStr) -> Result<Option<SyntaxReferenceInSet>> {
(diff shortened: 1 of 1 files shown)


src/assets.rs
```diff
@@ -308,6 +299,20 @@ impl HighlightingAssets {
...
+            .map(|syntax| SyntaxReferenceInSet { syntax, syntax_set }))
+    }
+
+    fn find_syntax_by_extension(&self, extension: &str) -> Result<Option<SyntaxReferenceInSet>> {
+        let syntax_set = self.get_syntax_set()?;
+        Ok(syntax_set
+            .find_syntax_by_extension(extension)
+            .map(|syntax| SyntaxReferenceInSet { syntax, syntax_set }))
+    }
+
```
Enselic, line 315: These two can also be de-duplicated with some generics, but I need to do some other preparatory work first to maximize the reach of such a refactoring.

Enselic: The generic function would look like this:
```
    fn find_syntax<'s, A, F>(&'s self, func: F, arg: A) -> Result<Option<SyntaxReferenceInSet>>
    where
        F: FnOnce(&'s SyntaxSet, A) -> Option<&'s SyntaxReference>,
    {
        let syntax_set = self.get_syntax_set()?;
        Ok(func(syntax_set, arg).map(|syntax| SyntaxReferenceInSet { syntax, syntax_set }))
    }
```

and you would invoke it like this:

```
self.find_syntax(SyntaxSet::find_syntax_by_name, syntax_name)
self.find_syntax(SyntaxSet::find_syntax_by_extension, file_name)
```

But
* It can't be maximally used in the current structure of the code
* I'm not sure it is an improvement

So I will wait with introducing it, if ever.

keith-hall (approved):

PR #1852, opened 2021-09-14
Simplify HighlightingAssets::get_syntax() first_line logic

Enselic:
The first commit makes the following work via first_line_match:

```
bat --map-syntax='*.invalid-syntax:InvalidSyntax' tests/examples/regression_tests/first_line_fallback.invalid-syntax
```
and adds a regression test for it.

The second commit simplifies the code, thanks to the fix in the first commit. It is funny how adding features can result in simpler code.

All regression tests pass, but we don't seem to have that many regression tests for `--map-syntax`, so I might add some more after code review of this.

(The reason I am bothering to simplify the code at all is of course to make #951 easier to implement.)

### DIFF
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -298,9 +298,9 @@ impl HighlightingAssets {
         match path_syntax {
             // If a path wasn't provided, or if path based syntax detection
             // above failed, we fall back to first-line syntax detection.
-            Err(Error::UndetectedSyntax(syntax_name)) => self
+            Err(Error::UndetectedSyntax(path)) => self
                 .get_first_line_syntax(&mut input.reader)?
-                .ok_or(Error::UndetectedSyntax(syntax_name)),
+                .ok_or(Error::UndetectedSyntax(path)),
             _ => path_syntax,
         }
     }
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -256,8 +256,6 @@ impl HighlightingAssets {
                 .ok_or_else(|| Error::UnknownSyntax(language.to_owned()));
         }
 
-        let line_syntax = self.get_first_line_syntax(&mut input.reader)?;
-
         // Get the path of the file:
         // If this was set by the metadata, that will take priority.
         // If it wasn't, it will use the real file path (if available).
@@ -270,7 +268,7 @@ impl HighlightingAssets {
                 _ => None,
             });
 
-        if let Some(path_str) = path_str {
+        let path_syntax = if let Some(path_str) = path_str {
             // If a path was provided, we try and detect the syntax based on extension mappings.
             let path = Path::new(path_str);
             let absolute_path = PathAbs::new(path)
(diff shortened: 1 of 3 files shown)


src/assets.rs
```diff
@@ -289,13 +288,17 @@ impl HighlightingAssets {
                 None => {
                     let file_name = path.file_name().unwrap_or_default();
                     self.get_extension_syntax(file_name)?
-                        .or(line_syntax)
                         .ok_or_else(|| Error::UndetectedSyntax(path.to_string_lossy().into()))
                 }
             }
         } else {
             // If a path wasn't provided, we fall back to the detect first-line syntax.
```
keith-hall, line 295: this comment feels a little out of place now - the fallback is done a few lines further down

src/assets.rs
```diff
@@ -289,13 +288,17 @@ impl HighlightingAssets {
...
-            line_syntax.ok_or_else(|| Error::UndetectedSyntax("[unknown]".into()))
+            Err(Error::UndetectedSyntax("[unknown]".into()))
+        };
+
+        match path_syntax {
+            Err(err) => self.get_first_line_syntax(&mut input.reader)?.ok_or(err),
```
keith-hall, line 300: maybe we should show some type of warning that the mapped syntax was invalid/ignored?

src/assets.rs
Enselic: agree, will fix

src/assets.rs
Enselic: Hmm yes maybe it is confusing to change the behavior as I did after all.

Rather than creating new problems, I will aim for not changing behavior from before, while still keeping the simplification of the code. I will soon push a commit with code that restores old behavior but keeps the refactoring, so we can discuss it.

Enselic:
Ok, code pushed. Let me know what you think. I think the new code makes more semantic sense. Only if we could not *detect* syntax (`Error::UndetectedSyntax`), we fall back to first_line. If we end up with an `UnkownSyntax`, we should let the user know. And we do.

The comment talks about "if path based syntax detection above failed", so it also talks about `UndetectedSyntax`.

keith-hall (approved):
Looks good to me!

src/assets.rs
```diff
@@ -289,13 +288,20 @@ impl HighlightingAssets {
...
         } else {
-            // If a path wasn't provided, we fall back to the detect first-line syntax.
-            line_syntax.ok_or_else(|| Error::UndetectedSyntax("[unknown]".into()))
+            Err(Error::UndetectedSyntax("[unknown]".into()))
+        };
+
+        match path_syntax {
+            // If a path wasn't provided, or if path based syntax detection
+            // above failed, we fall back to first-line syntax detection.
+            Err(Error::UndetectedSyntax(syntax_name)) => self
```
Enselic, line 301: Thanks for review! 

Note to self for tomorrow: it should be `UndetectedSyntax(path)`

Enselic:
It turned out we had good test coverage for all other code paths. Merging.

PR #1853, opened 2021-09-16
HighlightingAssets: Simplify absolute_path code

Enselic:
This is three commits. The last one is perhaps a matter of taste, but the problem with the `absolute_path` variable was that it did not necessarily contain an absolute path. Rather then renaming it to, say, `maybe_absolute_path`, I inlined it. That way one does not have to come up with a good name. It was only used once anyway.


keith-hall (approved):

Enselic:
All regression tests pass, and manual sanity checking also confirms the code works as it should. Merging.

PR #1858, opened 2021-09-20
Track dependents for each syntax

Enselic:
Progress on #951 continues, and I now have work-in-progress code where `syntaxes.bin` has been removed, and every syntax is present in `minimal_syntaxes.bin`. Notably, without duplicates, so the size of `minimal_syntaxes.bin` is the same as `syntaxes.bin`, differing only by some tens of kbs.  See https://github.com/Enselic/bat/pull/38.

While there still are problems to solve, I am confident in some parts of the code, and will ~~start~~ continue to create separate PRs to take the code base in that direction. (Actually, #1856 is also this kind of PR.)

This PR consists of three commits. The first commit just renames stuff in preparation for the third commit. The second commits just moves code around a bit to fit better with the third commit. The "real" code is in the third commit. It adds code to add information about what dependents a syntax has. For example, the **Rust** syntax is a dependent for the **Markdown** syntax, because **Markdown** depends on **Rust** via `embed: scope:source.rust`.

Knowing the dependents of a syntax is useful when constructing `SyntaxSet`s without duplicate `SyntaxReference`s (and is used in https://github.com/Enselic/bat/pull/38).

While this code adds info about syntax dependents, there is a temporary env var flag that makes it not used yet when building assets. The env var will be removed later, and the dependents info will start being used when building assets.

### DIFF
--- a/src/assets/build_assets.rs
+++ b/src/assets/build_assets.rs
@@ -11,18 +11,18 @@ use crate::assets::*;
 type SyntaxName = String;
 
 /// Used to look up what dependencies a given [SyntaxDefinition] has
-type SyntaxToDependencies = HashMap<SyntaxName, Vec<Dependency>>;
+type SyntaxToDependencies = HashMap<SyntaxName, Vec<OtherSyntax>>;
 
-/// Used to look up which [SyntaxDefinition] corresponds to a given [Dependency]
-type DependencyToSyntax<'a> = HashMap<Dependency, &'a SyntaxDefinition>;
+/// Used to look up which [SyntaxDefinition] corresponds to a given [OtherSyntax]
+type OtherSyntaxLookup<'a> = HashMap<OtherSyntax, &'a SyntaxDefinition>;
 
-/// Represents a dependency on an external `.sublime-syntax` file.
+/// Represents some other `*.sublime-syntax` file, i.e. another [SyntaxDefinition].
 #[derive(Debug, Eq, PartialEq, Clone, Hash)]
-enum Dependency {
-    /// By name. Example YAML: `include: C.sublime-syntax`
+enum OtherSyntax {
+    /// By name. Example YAML: `include: C.sublime-syntax` (name is `"C"`)
     ByName(String),
 
-    /// By scope. Example YAML: `embed: scope:source.c`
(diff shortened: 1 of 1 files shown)


src/assets/build_assets.rs
```diff
@@ -237,31 +246,57 @@ fn build_minimal_syntax_sets(
...
-/// what dependencies it has. This functions generates that data for each syntax.
-fn generate_maps(syntaxes: &[SyntaxDefinition]) -> (SyntaxToDependencies, DependencyToSyntax) {
+/// In order to analyze dependencies, we need three key pieces of data.
+///
+///  * When we have a [OtherSyntax], we need to know what [SyntaxDefinition]
+///    that corresponds to
+///  * When we have a [SyntaxDefinition], we need to know what dependencies it
+///    has
+///  * When we have a [SyntaxDefinition], we need to know what other syntaxes
+///    that depend on it
```
keith-hall, line 256:
```suggestion
///    depend on it
```

src/assets/build_assets.rs
```diff
@@ -10,19 +10,22 @@ use crate::assets::*;
...
+/// Used to look up which [SyntaxDefinition] corresponds to a given [OtherSyntax]
+type OtherSyntaxLookup<'a> = HashMap<OtherSyntax, &'a SyntaxDefinition>;
+
 /// Used to look up what dependencies a given [SyntaxDefinition] has
-type SyntaxToDependencies = HashMap<SyntaxName, Vec<Dependency>>;
+type SyntaxToDependencies = HashMap<SyntaxName, Vec<OtherSyntax>>;
 
-/// Used to look up which [SyntaxDefinition] corresponds to a given [Dependency]
-type DependencyToSyntax<'a> = HashMap<Dependency, &'a SyntaxDefinition>;
+/// Used to look up what other [SyntaxDefinition]s that depends on a given [SyntaxDefinition]
```
keith-hall, line 19:
```suggestion
/// Used to look up what other [SyntaxDefinition]s depend on a given [SyntaxDefinition]
```

keith-hall (approved):

Enselic:
Thanks for review!

(I will do a force-push to restore original commit structure before I merge)

PR #1860, opened 2021-09-22
Enable optional dump of syntax dependencies to Graphviz dot file

Enselic:
Also include a couple of fixes for dependency analysis (that does not affect the existing `minimal_syntaxes.bin` file) that I discovered by looking at the generated file. (There is still a problem left with dependencies being tracked on "self", that I might fix in some other PR.)

Here is a map of current syntax dependencies (click to enlarge):
bat-syntax-dependencies

Instructions on how to build a PNG yourself is included in the code as a comment, but I also duplicate it here for convenience:

```
% sudo apt install graphviz
% BAT_SYNTAX_DEPENDENCIES_TO_GRAPHVIZ_DOT_FILE=/tmp/bat-syntax-dependencies.dot cargo run -- cache --build --source assets --blank --target /tmp
% dot /tmp/bat-syntax-dependencies.dot -Tpng -o /tmp/bat-syntax-dependencies.png
% open /tmp/bat-syntax-dependencies.png
```

### DIFF
--- a/src/assets/build_assets.rs
+++ b/src/assets/build_assets.rs
@@ -20,8 +20,8 @@ type SyntaxToDependencies = HashMap<SyntaxName, Vec<OtherSyntax>>;
 type SyntaxToDependents<'a> = HashMap<SyntaxName, Vec<OtherSyntax>>;
 
 /// Represents some other `*.sublime-syntax` file, i.e. another [SyntaxDefinition].
-#[derive(Debug, Eq, PartialEq, Clone, Hash)]
-enum OtherSyntax {
+#[derive(Debug, Eq, PartialEq, PartialOrd, Ord, Clone, Hash)]
+pub(crate) enum OtherSyntax {
     /// By name. Example YAML: `include: C.sublime-syntax` (name is `"C"`)
     ByName(String),
 
@@ -304,6 +304,7 @@ fn dependencies_for_syntax(syntax: &SyntaxDefinition) -> Vec<OtherSyntax> {
         .collect();
 
     // No need to track a dependency more than once
+    dependencies.sort();
     dependencies.dedup();
 
     dependencies
--- a/src/assets/build_assets.rs
+++ b/src/assets/build_assets.rs
@@ -8,6 +8,8 @@ use syntect::parsing::{Scope, SyntaxSet, SyntaxSetBuilder};
 
 use crate::assets::*;
 
+mod graphviz_utils;
+
 type SyntaxName = String;
 
 /// Used to look up which [SyntaxDefinition] corresponds to a given [OtherSyntax]
@@ -221,6 +223,11 @@ fn build_minimal_syntax_sets(
     let (other_syntax_lookup, syntax_to_dependencies, syntax_to_dependents) =
         generate_maps(syntaxes);
 
+    maybe_write_syntax_dependencies_to_graphviz_dot_file(
+        &other_syntax_lookup,
+        &syntax_to_dependencies,
+    );
+
     // Create one minimal SyntaxSet from each (non-hidden) SyntaxDefinition
     syntaxes.iter().filter_map(move |syntax| {
         if syntax.hidden {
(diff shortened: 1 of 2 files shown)


src/assets/build_assets.rs
```diff
@@ -221,6 +223,20 @@ fn build_minimal_syntax_sets(
     let (other_syntax_lookup, syntax_to_dependencies, syntax_to_dependents) =
         generate_maps(syntaxes);
 
+    if let Ok(dot_file_path) = std::env::var("BAT_SYNTAX_DEPENDENCIES_TO_GRAPHVIZ_DOT_FILE") {
+        // To genarate a Graphviz dot file of syntax dependencies, do this:
```
Enselic, line 227: genarate -> generate (I will wait with updating PR until first review round is done)

Enselic: had to fix CI, so I fixed this typo too

src/assets/build_assets.rs
```diff
@@ -328,10 +345,31 @@ fn dependencies_from_pattern(pattern: &Pattern) -> Vec<OtherSyntax> {
     .collect()
 }
 
+/// Removes any subscope from the scope.
```
keith-hall, line 348: It may be clearer to use syntect terminology here. Maybe, "Removes any context name from the syntax reference"? And then to rename "subscope" elsewhere to "context" (or "subcontext" - without a `#something`, `#main` is implied)

keith-hall:
Wow, it's really awesome to see the syntax definition dependencies like this, nice one! I don't see anything wrong with the code, I just some minor comments about consistent naming.

src/assets/build_assets.rs
Enselic: That is definitely a good idea! Thanks for pointing that out. I have not internalized syntect terminology enough to fluently speak it yet :)

I pushed a commit that I think improves the terminology. Looking forward to your thoughts on the new code.

keith-hall (approved):

src/assets/build_assets.rs
keith-hall: Looks great - thanks!

sharkdp:
Nice - very cool idea. I was recently playing with graphviz. Note that there are [other layout engines](https://graphviz.org/docs/layouts/) that sometimes do a better job of untangling this mess. For example, if you simply replace `dot` with `fdp`, you get this:

out

Enselic:
> why do we have self-referential syntaxes (e.g. Swift)? And even cycles (Git Config <=> Git Common)?

Do you mean why we track it in our dependency analysis, or why the syntaxes are constructed the way they are?

The reason we track it is because it requires less code than not tracking it, and it (does not yet, at least) make a difference. As I noted in the PR description, it might be worthwhile someday to adjust that.

The reason they are constructed like that is probably better answered by @keith-hall. I can only note that they are. For example, Swift references its own syntax as a way to support syntax highlighting of String Interpolation expressions (see https://docs.swift.org/swift-book/LanguageGuide/StringsAndCharacters.html#ID292). It looks like this in `./assets/syntaxes/02_Extra/Swift.sublime-syntax`:

```
351   │             - include: scope:source.swift
```

Similarly, Git Common includes Git Config:
```
 341   │         - include: Git Config.sublime-syntax#color-value
```
and Git Config includes Git Common, e.g.
```
  20   │     - include: Git Common.sublime-syntax#comments
```

I haven't checked the other ones, but I assume they contain similar constructs.

sharkdp (approved):

Enselic:
Btw, thanks for reminding me about how easy it is to change Graphviz layout engine. I had forgot about that. Last time I messed around with Graphviz was 12 years ago :stuck_out_tongue: 

https://gitlab.gnome.org/GNOME/gegl/-/commit/a274f464277e0ca7f814b20042d3a8622a553307
```
commit a274f464277e0ca7f814b20042d3a8622a553307
Author: Martin Nordholts <martinn@src.gnome.org>
Date:   Sun Apr 12 11:40:02 2009 +0000

    gegl: Add gegl_dot_node_to_png()
    
    Add gegl_dot_node_to_png() which is a debug function supposed to be
    called directly from a debugger. E.g. in gdb it is used like this:
    
      p gegl_dot_node_to_png (node, "/tmp/foo.png")
    
    where 'node' is a node of interest. That will show the graphviz graph
    of 'node' and its dependencies.
    
    svn path=/trunk/; revision=3051
```

PR #1862, opened 2021-09-25
HighlightingAssets: Extract new public API method get_syntax_for_path()

Enselic:
We do this mainly to reduce duplication of logic. The move to adjust the public API is more of a side effect. It so happens that adjusting the public API also simplifies the code.

One of the reasons this refactoring works is that the `.file_name()` of a `file_name` is `file_name` unchanged:
```
    #[test]
    fn file_name_of_file_name_is_same() {
        let path = Path::new("test.rs");
        let expected = Path::new("test.rs");
        assert_eq!(path.file_name().unwrap_or_default(), expected);
    }
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@
 
 ## `bat` as a library
 
-- Deprecate `HighlightingAssets::syntaxes()` and `HighlightingAssets::syntax_for_file_name()`. Use `HighlightingAssets::get_syntaxes()` and `HighlightingAssets::get_syntax_for_file_name()` instead. They return a `Result` which is needed for upcoming lazy-loading work to improve startup performance. They also return what `SyntaxSet` the returned `SyntaxReference` belongs to. See #1747, #1755 and #1776 (@Enselic)
+- Deprecate `HighlightingAssets::syntaxes()` and `HighlightingAssets::syntax_for_file_name()`. Use `HighlightingAssets::get_syntaxes()` and `HighlightingAssets::get_syntax_for_path()` instead. They return a `Result` which is needed for upcoming lazy-loading work to improve startup performance. They also return which `SyntaxSet` the returned `SyntaxReference` belongs to. See #1747, #1755, #1776, #1862 (@Enselic)
 - Remove `HighlightingAssets::from_files` and `HighlightingAssets::save_to_cache`. Instead of calling the former and then the latter you now make a single call to `bat::assets::build`. See #1802 (@Enselic)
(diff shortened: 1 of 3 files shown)


src/assets.rs
```diff
@@ -159,12 +159,30 @@ impl HighlightingAssets {
...
-        let file_name = file_name.as_ref();
-        Ok(match mapping.get_syntax_for(file_name) {
-            Some(MappingTarget::MapToUnknown) => None,
-            Some(MappingTarget::MapTo(syntax_name)) => self.find_syntax_by_name(syntax_name)?,
-            None => self.get_extension_syntax(file_name.as_os_str())?,
-        })
+        Ok(self.get_syntax_for_path(file_name.as_ref(), mapping).ok())
+    }
+
+    fn get_syntax_for_path(
```
Enselic, line 165: Hmm it would probably be a good idea to change the pub API. We already deprecated one method. Need to think a bit more about this…

Enselic: I tried it and liked how it turned out. Let me know what you think about it.

CHANGELOG.md
```diff
@@ -33,7 +33,7 @@
 
 ## `bat` as a library
 
-- Deprecate `HighlightingAssets::syntaxes()` and `HighlightingAssets::syntax_for_file_name()`. Use `HighlightingAssets::get_syntaxes()` and `HighlightingAssets::get_syntax_for_file_name()` instead. They return a `Result` which is needed for upcoming lazy-loading work to improve startup performance. They also return what `SyntaxSet` the returned `SyntaxReference` belongs to. See #1747, #1755 and #1776 (@Enselic)
+- Deprecate `HighlightingAssets::syntaxes()` and `HighlightingAssets::syntax_for_file_name()`. Use `HighlightingAssets::get_syntaxes()` and `HighlightingAssets::get_syntax_for_path()` instead. They return a `Result` which is needed for upcoming lazy-loading work to improve startup performance. They also return what `SyntaxSet` the returned `SyntaxReference` belongs to. See #1747, #1755, #1776, #1862 (@Enselic)
```
keith-hall, line 36: > They also return what `SyntaxSet` the returned `SyntaxReference` belongs to.

I think this may be easier to read/understand if we replace `what` with `which`.

keith-hall (approved):

src/assets.rs
```diff
@@ -140,29 +140,39 @@ impl HighlightingAssets {
...
-    pub fn get_syntax_for_file_name(
-        &self,
-        file_name: impl AsRef<Path>,
-        mapping: &SyntaxMapping,
-    ) -> Result<Option<SyntaxReferenceInSet>> {
-        Ok(self.get_syntax_for_path(file_name.as_ref(), mapping).ok())
-    }
-
-    fn get_syntax_for_path(
+    /// Detect the syntax based on, in order:
```
keith-hall, line 155: Btw, I really like this docstring, nice one for adding it :)

CHANGELOG.md
Enselic: will fix before merge :+1:

src/assets.rs
Enselic: Thanks! I also like how it turned out :)

PR #1868, opened 2021-09-30
Update the doc for Ubuntu and Debian

RoneoOrg:
`bat` is available in the Ubuntu repositories for a while now and in Debian `stable` since August.
Let's update the doc accordingly.


Enselic (approved):
Thanks for updating! Looks like a clear improvement to my eyes.

PR #1875, opened 2021-10-01
chore: add funtoo linux instructions

invakid404:
Hello 👋🏻,

Since I maintain the bat package in [Funtoo Linux](https://www.funtoo.org), I thought I might as well add a few lines about it to the readme :^)


sharkdp:
Thank you

PR #1882, opened 2021-10-03
Add support for tmux

### LINKED ISSUE #1703: Why isn't this Tmux.sublime-syntax file working for me?
I wanted to add syntax highlighting for for .tmux.conf file, so I add this file to `bat/syntaxes/`

https://raw.githubusercontent.com/gerardroche/sublime-tmux/master/Tmux.sublime-syntax

I then `bat cache -b` and try viewing my `.tmux.conf` file, bit it is still plain with no syntax highlighting. Have I missed something or does this file not work with bat for some reason?

cbolgiano:
Adds support for Tmux.


cbolgiano:
Will run fmt shortly.

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@
 - Support for `x:+delta` syntax in line ranges (e.g. `20:+10`). See  #1810 (@bojan88)
 
 ## Bugfixes
-
+
```
SuperSandro2000, line 9:
```suggestion
```

src/syntax_mapping.rs
```diff
@@ -81,6 +81,16 @@ impl<'a> SyntaxMapping<'a> {
                 .unwrap();
         }
 
+        // Support syntax highlighting for Tmux, resolves issue #1703
+        for glob in &[
+            "tmux.conf",
+	    "tmux",
+	    ".tmux.conf",
+	    ".tmux",
```
SuperSandro2000, line 89:
```suggestion
            "tmux.conf",
            "tmux",
            ".tmux.conf",
            ".tmux",
```

Enselic: Thanks a lot for looking into this issue!

I think it might be possible to solve this in a generic way, i.e. not only for tmux.conf.

Our current algorithm to detect syntax is this:

1. See if there is a syntax mapping. If there is a syntax mapping to "unknown", use first-line fallback (this is the code path wrongly taken for tmux.conf). If there is a syntax mapping to a specific syntax, use that syntax
2. See if any syntax handles the file name in question, e.g. `.tmux.conf`. This is the code path we want for `.tmux.conf`.
3. See if any syntax handles the file name extension in question.
4. If all else fails, use first_line fallback

The code for this logic is here: https://github.com/sharkdp/bat/blob/master/src/assets.rs#L214

I wonder if it wouldn't be possible to introduce the concept of "map to unknown syntax, but only if there is no exact file name match". So that instead of

```
        mapping
            .insert("*.conf", MappingTarget::MapToUnknown)
            .unwrap();
```

we would have something like

```        mapping
            .insert("*.conf", MappingTarget::MapToUnknownUnlessExactFileNameMatch)
            .unwrap();
```

That would also allow us to get rid of this special case:

```
        // Re-insert a mapping for resolv.conf, see #1510
        mapping
            .insert("resolv.conf", MappingTarget::MapTo("resolv"))
            .unwrap();
```

cbolgiano: Thank you for the suggestion. I can look into making the change.

cbolgiano:
I almost ready for a push an update. I love this project. It is so cool.

cbolgiano:
Latest work has been pushed.

cbolgiano:
Looking at the raw log of the regression tests this is what I see:

2021-10-06T04:25:43.0189536Z === Error: Could not highlight source file '/home/runner/work/bat/bat/tests/syntax-tests/source/DotENV/.env
2021-10-06T04:25:43.0190260Z === bat stdout:
2021-10-06T04:25:43.0190512Z 
2021-10-06T04:25:43.0190809Z === bat stderr:
2021-10-06T04:25:46.9500990Z thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Ok(SyntaxReferenceInSet { syntax: SyntaxReference { name: "resolv", file_extensions: ["resolv.conf"], scope: <source.resolv>

^ I am new to rust but it seems like I need to wrap the Error in Err. Going to try that to fix the failing regression tests. If I am incorrect and there is something else I introduced please let me know.

Enselic:
Thanks for the update! I don’t have time to deep dive into the code right now, but I glanced it over, and have these spontaneous comments:

* would you be ok with splitting this PR up into two? One for adding tmux and one for adding MapToUnknownUnlessExactFileNameMatch? That will make code review and everything else easier.
* Rather than adding more calls to self.get_extension_syntax, try to split it up and rearrange. The goal is to only do each thing once. I think that can be achieved by different ordering.

cbolgiano:
Absolutely. I don't mind. I will work on that soon. Thank you!

cbolgiano:
Moved the implementation of MapToUnknownUnlessExactFileNameMatch to a different MR: https://github.com/sharkdp/bat/pull/1889

assets/syntaxes/02_Extra/Tmux/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+The color scheme in the screenshot is [MonokaiFree](https://github.com/gerardroche/sublime-monokai-free).
+
+## INSTALLATION
+
+### Manual installation
+
+Close Sublime Text then download or clone this repository to a directory named `Tmux` in the Sublime Text Packages directory for your platform:
+
+* Linux: `git clone https://github.com/gerardroche/sublime-tmux.git ~/.config/sublime-text-3/Packages/Tmux`
```
MichaelCurrin, line 15:
```suggestion
- Linux:
    ```bash
    git clone https://github.com/gerardroche/sublime-tmux.git ~/.config/sublime-text-3/Packages/Tmux
    ```
```

assets/syntaxes/02_Extra/Tmux/README.md
```diff
@@ -0,0 +1,21 @@
...
+* OSX: `git clone https://github.com/gerardroche/sublime-tmux.git ~/Library/Application\ Support/Sublime\ Text\ 3/Packages/Tmux`
```
MichaelCurrin, line 16: "OSX" is the older systems. Apple uses "macOS" for their newer systems.

Enselic (changes requested):
Looks like the sub module is not actually added like a sub module like the other syntaxes?

Also, how many downloads does this syntax have on package control?

Enselic:
Looks like tmux is quite far from 10k?

This PR was more of a stepping stone to #1889 I suppose.

Big thanks again for that already legendary PR btw 🙂

I think we can go ahead and close this PR though.

PR #1883, opened 2021-10-03
CICD: Make the 'cargo fmt' check a toplevel job

Enselic:
Mainly to make it easier to see what went wrong when CI fails.

If this ever gets of out sync with a particular Rust version, we can most likely
save the situation by introducing a `rustfmt.toml` file.

Note that I choose to make the config minimal (e.g. without `name`). I think it does not make things harder to understand. In fact, I think it makes things easier to understand and maintain. I also use the default Rust installation instead of installing our own. That's just unnecessary to do in this case.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -14,6 +14,13 @@ on:
       - '*'
 
 jobs:
+  ensure_cargo_fmt:
+    name: Ensure 'cargo fmt' has been run
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v2
+    - run: cargo fmt -- --check
+
   min_version:
     name: Minimum supported rust version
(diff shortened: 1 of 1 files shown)


sharkdp:
> Note that I choose to make the config minimal (e.g. without `name`).

You mean.. without a `Install rust toolchain` task?

Enselic:
Yep. Rust stable is installed by default: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#rust-tools

sharkdp:
To be honest, I'd rather spend a bit more time in CI than to have a *third* (undocumented and uncontrolled) Rust Version around.

Enselic:
Ok, will add a step for that shortly :+1:

sharkdp (approved):
Thank you for the update.

PR #1887, opened 2021-10-05
syntax-tests: Make CpuInfo test actually work

Enselic:
File extension matching is case-sensitive, so extension needs to be .cpuinfo for
the syntax to actually be used.

Fixing this means one problem less with bat tests when using syntect master. Syntect master use case insensitive matching, so this test will start to fail with syntect master. See https://github.com/trishume/syntect/pull/369.

With this fix, the test will work with both current syntect and syntect master.


sharkdp (approved):
Thanks!

Enselic:
Oops. Wanted to rewrite the “Squash and merge” commit messag, but am on mobile, and somehow managed to click merge while trying to edit the commit msg.

Oh well, no big harm, just mildly annoying.

PR #1888, opened 2021-10-06
Add systemwide config file support

### LINKED ISSUE #668: Separate system and user config files

patrickpichler:
There is now support for a systemwide config file. The location of the
system wide config file is `$(BAT_SYSTEM_CONFIG_PREFIX)/bat/config`.
`$(BAT_SYSTEM_CONFIG_PREFIX)` has to be provided at compile time as an
environment variable. If the environment variable is not set, a default
is used. This default is `C:\ProgramData` for windows and `/etc` for every other os.

fixes #668


patrickpichler:
@sharkdp any update on this?

tests/system_wide_config.rs
```diff
@@ -0,0 +1,47 @@
...
+    cmd.env_remove("BAT_THEME");
+    cmd.env_remove("COLORTERM");
+    cmd.env_remove("NO_COLOR");
+    cmd.env_remove("PAGER");
+    cmd
+}
+
+fn bat() -> assert_cmd::Command {
+    assert_cmd::Command::from_std(bat_raw_command())
+}
```
sharkdp, line 24: it would be great to move this duplicated(?) code to a common test-helper module

tests/system_wide_config.rs
```diff
@@ -0,0 +1,47 @@
...
+
+// This test is ignored, as it needs a special system wide config put into place
```
sharkdp, line 26: can we please add instruction on how to enable it?

sharkdp:
Thank you for the updates!

patrickpichler:
@sharkdp I've added the `CHANGELOG.md` entry, plus added a short description of the feature to the `README.md`

sharkdp (approved):
Thank you

sharkdp:
Tests are currently failing

tests/integration_tests.rs
```diff
@@ -27,41 +25,11 @@ mod unix {
 use unix::*;
 
 mod utils;
+use utils::command::{bat, bat_raw_command, bat_with_config};
```
sharkdp, line 28: Can we fix the unused import warning here?

tests/utils/command.rs
```diff
@@ -0,0 +1,33 @@
...
+    cmd.env_remove("BAT_STYLE");
+    cmd.env_remove("BAT_TABS");
+    cmd.env_remove("BAT_THEME");
+    cmd.env_remove("COLORTERM");
+    cmd.env_remove("NO_COLOR");
+    cmd.env_remove("PAGER");
+    cmd
+}
+
+pub fn bat_raw_command() -> Command {
```
sharkdp, line 21: I think this function (or the whole module) should be `#[cfg(test)]`, to silence the warnings. Sorry, just saw this now.

Similar for the other functions in this file.

tests/integration_tests.rs
patrickpichler: I'm a bit confused. Where do you see a warning for this here? (I'm still pretty new to rust)

tests/utils/command.rs
patrickpichler: I've tried putting `#[cfg(test)]` on there, but `cargo test` still emits the warning when building.
Or did you mean some other warnings?

tests/integration_tests.rs
sharkdp: I meant the warnings that are show in the PR diff view:

image

patrickpichler: Oooh :smile: 
Yeah I didn't look there :smile: 
Thanks for pointing this out!

tests/utils/command.rs
patrickpichler: After looking at it again, I guess rust still emits a warning, as the function is only
used in unix tests. I'm not 100% sure how to fix this though, as one cannot put multiple
`cfg` annotations on top. Do you maybe have an idea?

patrickpichler:
@sharkdp @Enselic any update on this?
I looked once again into the warnings on running the system wide tests and I guess this
is caused as code is shared between the different test types. 

The warnings also show up for other shared utility methods such as the one defined in 
`tests/utils/mocked_pagers.rs`. 

Do you have any plan what I could do here?

Enselic:
Looks like I got rid of the warnings now, and this looks like a nice
change to me. It has tests, CI is green, and @sharkdp even approved it
already once. Sorry for taking so long, @patrickpichler.

@sharkdp Any chance you can take a final look and then merge? I
haven't done a detailed review yet, but if you think it look good, I'm
sure it's good enough. And even if it is not perfect (what is,
really?) it certainly seems good enough to start off with and then
iteratively improve on over several versions of bat.

tests/utils/command.rs
Enselic: #[cfg(test)] is not enough (but does not hurt), because the mod is brought into scope by a test, but then never used. There are ways to restructure the code, but I don't think it is worth it, so let's just allow it, at least for now.

sharkdp:
I did have another look and fully agree. Thank you @patrickpichler and @Enselic for the update.

PR #1899, opened 2021-10-11
Improve message on cache file is not existing

rhysd:
As you know [`Result::ok`](https://doc.rust-lang.org/std/result/enum.Result.html#method.ok) is a method to convert `Result<T, E>` to `Option<T>`. Using it without substituting the result to other place has not effect.

I thought those misusage aimed to check  no error happened basically. In the case `Result::unwrap` should be used. This PR fixes the `Result::ok` with `Result::unwrap` to ensure no error happened.

In the case of removing assets cache, an error is expected (when a cache does not exist). I added a better output in the case (rather than just saying 'okay').

### DIFF
--- a/src/bin/bat/assets.rs
+++ b/src/bin/bat/assets.rs
@@ -1,5 +1,6 @@
 use std::borrow::Cow;
 use std::fs;
+use std::io;
 
 use clap::crate_version;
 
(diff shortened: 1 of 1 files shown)


sharkdp:
I do actually prefer `.ok()` in favor of `let _`, yes. And not just for stylistic reasons:

Compare:
```rust
foo().ok();
let _ = foo();
```
and imagine what happens if the return type of the `foo()` method changes from `Result<T>` to something else. In the first case, we will get a compile error (which is good!). In the second case, we will just discard whatever `foo` returns now. Might be a plain `T`, might be something else entirely.

rhysd:
I understood. I'll update this branch following your preference.

PR #1907, opened 2021-10-17
Make grep-cli optional dependency

rhysd:
`grep-cli` crate is used for resolving a path of pager command safely. So it is only necessary with `paging` feature. This PR changes it to optional dependency so that installing bat as library with `--no-default-features` can reduce number of dependencies (111 → 108).


sharkdp:
Thank you

PR #1910, opened 2021-10-20
Changelog enriched with `cargo changelog` (a binary of `cargo-smart-release`)

Byron:
This is the result of running `cargo changelog --write`, which regenerates the changelog to amend information from the git history while leaving user edits in tact. This works here as the changelog format is already close to the one suggested by keepachanglog.

* [rendered changelog](https://github.com/Byron/bat/blob/master/CHANGELOG.md)

There are two auto-generated sections, `Commit Details` and `Commit Statistics`. Both can be disabled by running `cargo changelog --without commit-statistics commit-details`. The added header can just be replaced with an empty line and won't be regenerated.

If `cargo smart-release` is used for release, a `cargo release` compatible tag will be generated which carries the relevant section from the changelog, as well as a GitHub release, if the `gh` executable is found. Note that it generates special versions for each target to get the best possible rendering.

Lastly, even without `cargo smart-release`, `cargo changelog` will none-destructively regenerate the changelog and respect edits.

Please feel free to close this PR as it's definitely unsolicited, but I thought I'd try in case the change is likable or can be made that way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,11 @@
-# unreleased
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+# Unreleased
 
 ## Features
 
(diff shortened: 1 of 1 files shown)


Enselic:

My personal take on this is that it is already quite labor intensive  to keep the current CHANGELOG.md up to date, so I am not that excited about making it even more complicated. Spare time is a precious resource.

Enselic:
This doesn't seem to spark great enthusiasm among other maintainers, so I'll go ahead and close this now.

Thanks anyway for taking the time to create a PR. It *could* have been a smash hit 🙂

PR #1920, opened 2021-10-23
Add `bat::PrettyPrinter::clear_highlights`

### LINKED ISSUE #1919: Cannot clear highlights in `bat::PrettyPrinter` instance

Let's say we have the following code:

```rust
use bat::PrettyPrinter;

let targets: (PathBuf, Vec<(usize, usize)>) = vec![];

let mut pp = PrettyPrinter::new();
for (path, ranges) in targets.iter() {
    pp.input_file(path);
    for (start, end) in ranges.iter() {
        pp.highlight_range(*start, *end);
    }

    // This clears inputs, but does not clear highlights
    pp.print().unwrap();
}
```

Since there are multiple files with different highlight ranges, I want to print each files with different ranges. However, currently there is no way to clear highlights in `bat::PrettyPrinter` instance.

Creating `bat::PrettyPrinter` instance for each file is an option. However it is much slower because `bat::PrettyPrinter::new` deserializes binary assets and it is time consuming.

I'd like `bat::PrettyPrinter` to provide some API to clear highlights.

```toml
[dependencies]
bat = { version = "0.18.3", default-features = false }
```

in my project's `Cargo.toml`.

---

- bat: 0.18.3
- OS: macOS 10.15
- Rust: 1.56

rhysd:
Fixes #1919

With this API, the code written in #1919 can be improved as follows:

```rust
use bat::PrettyPrinter;

let targets: (PathBuf, Vec<(usize, usize)>) = vec![];

let mut pp = PrettyPrinter::new();
for (path, ranges) in targets.iter() {
    pp.input_file(path);
    for (start, end) in ranges.iter() {
        pp.highlight_range(*start, *end);
    }

    pp.print().unwrap();
    // Clear highlights for next iteration
    pp.clear_highlights();
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@
 ## `bat` as a library
 
 - Make `bat::PrettyPrinter::syntaxes()` iterate over new `bat::Syntax` struct instead of `&syntect::parsing::SyntaxReference`. See #2222 (@Enselic)
+- Clear highlights after printing, see #1919 and #1920 (@rhysd)
 
 
 # v0.21.0
--- a/src/pretty_printer.rs
+++ b/src/pretty_printer.rs
@@ -229,6 +229,14 @@ impl<'a> PrettyPrinter<'a> {
         self
     }
 
+    /// Clear highlighted lines added by [`PrettyPrinter::highlight`] and
+    /// [`PrettyPrinter::highlight_range`]. This is useful when reusing a
+    /// `PrettyPrinter` instance for multiple different ranges.
+    pub fn clear_highlights(&mut self) -> &mut Self {
+        self.highlighted_lines.clear();
+        self
+    }
+
     /// Specify the highlighting theme
     pub fn theme(&mut self, theme: impl AsRef<str>) -> &mut Self {
         self.config.theme = theme.as_ref().to_owned();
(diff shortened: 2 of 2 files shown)


Eehsanahmadi (approved):

sharkdp:
> The point is that it is a breaking change in API. I avoided the breaking change but clearing the line ranges at `print()` is sufficient for me. Do you think the breaking change is OK?

Not sure I understand this correctly: wouldn't the breaking change only affect the cases where the behavior is currently broken/buggy anyway? (#1919)

sharkdp:
@rhysd Any update on here? Are you still interested in this change? Or has the situation improved due to syntax lazy loading (see accompanying ticket)

rhysd:
I'm sorry for the lack of response. I forgot this PR. I'll take a look tonight.

rhysd:
> wouldn't the breaking change only affect the cases where the behavior is currently broken/buggy anyway?

I'm not sure it is broken/buggy, but I agree that it is very edge case.

rhysd:
@sharkdp I updated this branch at 3d7817d. Would you review changes?

sharkdp:
Looks great - thank you. I like the implementation :+1:

sharkdp (approved):

PR #1923, opened 2021-10-27
Update alternatives.md

h3athen:
I have been using the bat for a while now saw that it's also registered for the Hacktober competition but due to my lack of knowledge I could only contribute a change to this document. :) Thanks for making this awesome tool


Enselic:
Thanks for taking the time to make a contribution! I skimmed over the change, but I'm not quite sure if the new version is an improvement over what we have TBH. Maybe you would like to elaborate?

Issue #1897, opened 2021-10-10
Clarify control chars behavior on `bat`


isinyaaa:
Edit:

As explained below by @Enselic, `bat` doesn't support paging files that already contain colors, which has led me (and may lead others) to think this is a bug.

It'd be nicer if this could be clarified in the [troubleshooting section](https://github.com/sharkdp/bat#troubleshooting) of the README, so that no one else mistakes it for a bug (i.e. `bat` behaving exactly the same as `less -R` with those files).

Original issue:

> **Describe the bug you encountered:**
> 
> Using `less -R [filename]` directly renders output as expected, but when doing it through `bat` colors codes show up in output.
> 
> FYI I'm looking at some colored `make` output. On `less` it looks like this:
> 
> ![image](https://user-images.githubusercontent.com/39812919/136694699-87257622-cb88-4aaa-b31b-adc1a2658de3.png)
> 
> Then on `bat`:
> 
> ![image](https://user-images.githubusercontent.com/39812919/136694720-ddd2142f-c184-4139-a0e4-5db3ed5587c3.png)
> 
> **What did you expect to happen instead?**
> 
> Consistent behavior when using `less -R` alone and doing it through `bat`
> 
> **How did you install `bat`?**
> 
> `pacman`, but also tried using `cargo`
> 
> **bat version and environment**
> 
> 
> 
> #### Software version
> 
> bat 0.18.3 (b146958)
> 
> #### Operating system
> 
> Linux 5.14.9-arch2-1
> 
> #### Command-line
> 
> ```bash
> bat --diagnostic amd_build/modules1.clean.log 
> ```
> 
> #### Environment variables
> 
> ```bash
> SHELL=/usr/bin/fish
> PAGER='less -R'
> LESS=<not set>
> BAT_PAGER='less -RF'
> BAT_CACHE_PATH=<not set>
> BAT_CONFIG_PATH=<not set>
> BAT_OPTS=<not set>
> BAT_STYLE=<not set>
> BAT_TABS=<not set>
> BAT_THEME=Coldark-Dark
> XDG_CONFIG_HOME=<not set>
> XDG_CACHE_HOME=<not set>
> COLORTERM=truecolor
> NO_COLOR=<not set>
> MANPAGER=<not set>
> ```
> 
> #### Config file
> 
> ```
> # This is `bat`s configuration file. Each line either contains a comment or
> # a command-line option that you want to pass to `bat` by default. You can
[...]

Issue #1823, opened 2021-08-26
Hyperlinks don't work with bat but work fine with cat


schneiderfelipe:
Hi! I just realized that `bat` couldn't handle [hyperlinks in the terminal](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda), even though in a compatible terminal emulator `cat` is fine with it:

![Captura de tela de 2021-08-26 14-44-43](https://user-images.githubusercontent.com/37125/131010720-f36f8dec-3001-480b-9f73-334d92ef9082.png)

(🦇 to the left, 🐱 to the right.)
See how `cat` properly underlines hyperlinks, while `bat` shows the raw escapes defining them.
It would be very nice to have this feature.

A way of reproducing the above:

```bash
$ # A test example from the GNOME terminal
$ wget -c https://gitlab.gnome.org/GNOME/vte/raw/master/perf/hyperlink-demo.txt
$ bat hyperlink-demo.txt
$ /bin/cat hyperlink-demo.txt # In case you have aliased cat like me 😄

Issue #1572, opened 2021-03-02
bat display a lot ^M（e.g. python tqdm output）


zhouyangnk:
I know I can just replace '^M' with some other character, but is there any good way to show it in bat?

Enselic:
Thanks for asking a question. I don't quite understand what you mean though. How do you get `bat` to display a lot of `^M`s? Maybe `bat --show-all` is what you want?

zhouyangnk:
> Thanks for asking a question. I don't quite understand what you mean though. How do you get `bat` to display a lot of `^M`s? Maybe `bat --show-all` is what you want?

Thank you for your reply, for exam:
![微信截图_20210303102533](https://user-images.githubusercontent.com/10472264/109743045-0547ad00-7c0b-11eb-9f97-ba321f0f7aff.png)
ps: test.out is produced by python's tqdm module

Enselic:
Thanks for providing more info. Here is the output file in question: [test.txt](https://github.com/sharkdp/bat/files/6073926/test.txt)

This is an interesting case, because `cat` and `bat` behaves differently. The file contains carriage-return characters, aka ^M, produced by the progress bar redrawing itself.

When displayed with `cat`, it looks correct, because `cat` handles ^M correctly, in the sense that it "erases" what was previously printed.

```
% cat test.txt
100%|██████████| 50/50 [00:00<00:00, 97.21it/s]
```

in the case of `bat` however, we interpret ^M like any other character, and just prints it:

```
% bat test.txt
───────┬────────────────────────────────────────────────────
       │ File: test.txt
───────┼────────────────────────────────────────────────────
   1   │ ^M  0%|          | 0/50 [00:00<?, ?it/s]^M 20%|██      
       │   | 10/50 [00:00<00:00, 98.02it/s]^M 40%|████      | 
       │ 20/50 [00:00<00:00, 97.72it/s]^M 60%|██████    | 30/5
       │ 0 [00:00<00:00, 97.65it/s]^M 80%|████████  | 40/50 [0
       │ 0:00<00:00, 97.00it/s]^M100%|██████████| 50/50 [00:00
       │ <00:00, 97.27it/s]^M100%|██████████| 50/50 [00:00<00:
       │ 00, 97.21it/s]
───────┴────────────────────────────────────────────────────
```

Maybe we should start to doing what `cat` does unless `--show-all` is passed? Not sure if it's a good idea though.

zhouyangnk:
[...]

Issue #403, opened 2018-11-05
'bat' gives different output than 'cat' when printing from IPython


kbd:
A "bat has different behavior than cat" bug report (I have `bat` aliased to `cat` now):

```shell
⚡~/proj/stderr$ ipython -c "print('hello')"
hello
⚡~/proj/stderr$ ipython -c "print('hello')" | cat
───────┬───────────────────────────────────────────────────────────
       │ STDIN
───────┼───────────────────────────────────────────────────────────
   1   │ 0;IPython: proj/stderr^Ghello
───────┴───────────────────────────────────────────────────────────
⚡~/proj/stderr$ ipython -c "print('hello')" | \cat
hello
⚡~/proj/stderr$ python3 -c "print('hello')" | cat
───────┬───────────────────────────────────────────────────────────
       │ STDIN
───────┼───────────────────────────────────────────────────────────
   1   │ hello
───────┴───────────────────────────────────────────────────────────
```

I dunno what the `0;...^G` syntax means or why IPython adds it, but figured I'd report.

Thanks again for `bat`!

sharkdp:
Thank you for reporting this.

Interestingly, we can use `bat`s new `-A`/`--show-all` option (#395) to figure out what's going on here. This option shows non-printable characters in a human-readable form:

```
/tmp                                                                             
▶ ipython -c "print('hello')" | bat -A
───────┬──────────────────────────────────────────────────────────────────────────
       │ STDIN
───────┼──────────────────────────────────────────────────────────────────────────
   1   │ ␛]0;IPython:•/tmp␇hello␊
───────┴──────────────────────────────────────────────────────────────────────────
```

Instead of just printing `hello␊` (hello + line feed / new line), IPython first outputs `␛]0;` (some form of escape sequence), then `IPython: /tmp` (I'm in the `/tmp` folder), then `␇` (BEL ASCII character). It turns out that these three pieces make up a Terminal control sequence to set the title of the terminal Window (see e.g. http://tldp.org/HOWTO/Xterm-Title-3.html). Indeed, if you just execute `ipython -c "print('hello')"`, you should be able to see your terminal title change (just as long as the process is running).

IPython has a `--no-term-title` option to disable this behavior, so this works:
```
▶ ipython --no-term-title -c "print('hello')" | bat       
───────┬───────────────────────────────────────────────────────────────
       │ STDIN
───────┼───────────────────────────────────────────────────────────────
   1   │ hello
[...]

PR #1937, opened 2021-11-07
README.md: Add 'Garbled output' to 'Troubleshooting'

isinyaaa:
As per #234, #403, #1572, #1823 and #1897, this PR clarifies that `bat` cannot properly handle escape sequences.

### DIFF
--- a/README.md
+++ b/README.md
@@ -674,6 +674,13 @@ bat() {
 
 ## Troubleshooting
 
+### Garbled output
+
+If an input file contains color codes or other ANSI escape sequences, `bat` will have problems
+performing syntax highlighting and text wrapping, and thus the output can become garbled.
+When displaying such files it is recommended to disable both syntax highlighting and wrapping by
+passing the `--color=never --wrap=never` options to `bat`.
+
 ### Terminals & colors
 
 `bat` handles terminals *with* and *without* truecolor support. However, the colors in most syntax


Enselic:
Thanks for taking the time to look into writing documentation for this.

As I was reviewing, I was trying to wrap my head around what the actual problem is that bat is having. I think bat can get away with not being able to perform syntax highlighting on files that contain color codes.

But when syntax highlighting is disabled (`--color=never`), I thought bat should be able to display colored files. So I experimented a bit. First I created a colored file to test with:

```
bat --color=always --style=plain examples/simple.rs > /tmp/simple.rs
```

Then I tried to show that file with bat **with syntax highlighting disabled**, but it was a mess:

```
% bat --color=never /tmp/simple.rs             
```
Screenshot_2021-11-16_21-46-25

However, when I disabled wrapping, it worked perfectly:
```
% bat --wrap=never --color=never /tmp/simple.rs
```
Screenshot_2021-11-16_21-47-53

So that got me thinking. If `--wrap=auto` (the default), it would perhaps be a good idea to disable wrapping if we encounter ANSI escape codes in the input. Would that have helped you in #1897 @isinyaaa ? Does the output look good for you too if you pass `--wrap=never` to `bat` (and `--color=never` to disable syntax highlighting)?

It would say it is still a good idea to document the troubles bat have with files with ANSI escape code, but we could perhaps make bat behave a bit better by default still.

isinyaaa:
> Thanks for taking the time to look into writing documentation for this.

No problem! Glad to help.

> So that got me thinking. If `--wrap=auto` (the default), it would perhaps be a good idea to disable wrapping if we encounter ANSI escape codes in the input. Would that have helped you in #1897 @isinyaaa ? Does the output look good for you too if you pass `--wrap=never` to `bat` (and `--color=never` to disable syntax highlighting)?

Can confirm it does work pretty nicely :)

image
 
> It would say it is still a good idea to document the troubles bat have with files with ANSI escape code, but we could perhaps make bat behave a bit better by default still.

Maybe disabling wrapping and colors whenever bat encounters an escape sequence would work? I'd like to help out more if I can but I've little to no experience with rust so I don't want to keep holding this back in case anyone else wants to do it.

Enselic:
After thinking about it some more I'm not sure how we could implement that. We would have to analyze the whole file before wrapping the first line, unless we can tolerate turning off wrapping half-way down a file.

So I think we are better of simply documenting the current behavior.

I will review your proposal again with this in mind.

README.md
```diff
@@ -674,6 +674,26 @@ bat() {
...
+   uses its own decorations, so it's not strictly compliant with output that
+   would work on `less -R` alone.
+
+2. That needs reprinting (eg. a progress bar).
+
+   `bat` can only work as a drop-in `cat` replacement whenever it detects a
+   non-interactive terminal session, so you might be able to make it work by
+   using `bat -pp`, if that doesn't suit your needs, please check out the man
+   pages for more options.
+
```
Enselic, line 696: In light of what we talked about recently, what do you think about shortening this section and just having it say the following?

```suggestion
### Garbled output

If an input file contains color codes or other ANSI escape sequences, `bat` will
have troubles to perform syntax highlighting and text wrapping. The output can
become garbled. When displaying such files it is recommended to disable both
syntax highlighting and wrapping by passing `--color=never --wrap=never` to
`bat`.

```

isinyaaa: Sure, that's a lot cleaner and does seem to work :)

I'll only change a couple words for clarity.

README.md
```diff
@@ -674,6 +674,13 @@ bat() {
 
 ## Troubleshooting
 
+### Garbled output
+
+If an input file contains color codes or other ANSI escape sequences, `bat` will have problems
+performing syntax highlighting and text wrapping and, thus the output can become garbled.
```
Enselic, line 680: Comma placements seems a bit weird? Maybe change to

```suggestion
performing syntax highlighting and text wrapping, and thus the output can become garbled.
```

Otherwise it looks fine to me.

isinyaaa: done :)

Enselic (approved):
Thanks again for taking action on improving the documentation :) This looks good to me now. We can always iteratively polish this section over time.

sharkdp:
It's pretty easy to remove ANSI sequences from a given input (https://superuser.com/a/380778). Maybe we could suggest to use this instead, if someone is interested in re-highlighting some output?
```bash
program_with_colorized_output | sed 's/\x1b\[[0-9;]*m//g' | bat
```

Enselic:
That's a good thing to discuss. Thanks for bringing that up.

But I actually don't think that technique is applicable to the use cases at hand. None of the linked issues are about re-highlighting. Instead, they are alll about making bat preserve ANSI escape codes in the output. So in that sense, it was silly of me to use `bat examples/simple.rs` as example input. If a file is highlighted already, there is typically no need to highlight it again.

An additional complication is that it seems as if there is no default tool on macOS to strip ANSI escape sequences easily.

So all in all, I think we are better off leaving that technique out of this section in the troubleshooting guide.

If you insist on incorporating it, that is fine with me though of course.

PR #1938, opened 2021-11-08
Strip BOM from output in interactive mode

### LINKED ISSUE #1922: A space appears at the beginning of the file (Byte order mark)

If you use bat on C# source files (.cs, .xaml and others), a space appears in the first line. This is due to byte order mark (BOM)
Maybe reproduced for others files on Windows systems 
https://en.wikipedia.org/wiki/Byte_order_mark#Byte_order_marks_by_encoding
**Sample file with BOM:**
[Program.cs.txt](https://github.com/sharkdp/bat/files/7420061/Program.cs.txt)

IWpnW3OyGgg

Screenshot_57

If I delete these bytes: 
Screenshot_58

bat works correctly :

Screenshot_59

GitHub release:
**bat-v0.18.3-x86_64-pc-windows-gnu.zip
bat-v0.18.3-x86_64-pc-windows-msvc.zip**

---

#### Software version

bat 0.18.3 (b146958)

#### Operating system

Windows 6.2.9200

#### Command-line

```bash
C:\WINDOWS\system32\bat.exe --diagnostic
```

#### Environment variables

```bash
SHELL=<not set>
PAGER=<not set>
LESS=<not set>
BAT_PAGER=<not set>
BAT_CACHE_PATH=<not set>
BAT_CONFIG_PATH=<not set>
BAT_OPTS=<not set>
BAT_STYLE=<not set>
BAT_TABS=<not set>
BAT_THEME=<not set>
XDG_CONFIG_HOME=<not set>
XDG_CACHE_HOME=<not set>
COLORTERM=<not set>
NO_COLOR=<not set>
MANPAGER=<not set>
```

#### Config file

Could not read contents of 'C:\Users\timoxa\AppData\Roaming\bat\config': Системе не удается найти указанный путь. (os error 3).

#### Compile time information

- Profile: release
- Target triple: x86_64-pc-windows-gnu
- Family: windows
- OS: windows
- Architecture: x86_64
- Pointer width: 64
- Endian: little
- CPU features: fxsr,sse,sse2
- Host: x86_64-pc-windows-msvc

dag-h:
Fixes #1922 

This is my first contribution to an open source project so I apologize in advance for any misunderstandings!

I tried finding somewhere where the output is known to be shown in an interactive mode, and `InteractivePrinter` looked right on. Removed the BOM if present on the first line. Also made sure the BOM is not stripped if output is piped to a file/other program.

Contents of first line copied from `bat Program.cs.txt` (see #1922 for example file):

Screenshot from 2021-11-08 09-54-57

Contents of file obtained from `bat Program.cs.txt > test.txt`:

Screenshot from 2021-11-08 09-58-06

The `utf16` test then fails since it expects the output to contain the BOM. As far as I can tell the test runs `bat` in interactive mode, so I assumed this to be incorrect now. I changed the expected test output to not contain the BOM anymore. 

I have probably missed some important things to consider, and there are probably better ways to do this than `std::str::replace`.


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Python syntax highlighting no longer suffers from abysmal performance in specific scenarios. See #1688 (@keith-hall)
 - First line not shown in diff context. See #1891 (@divagant-martian)
 - Do not ignore syntaxes that handle file names with a `*.conf` extension. See #1703 (@cbolgiano)
+- Strips byte order mark from output in interactive mode. See #1922 (@dag-h)
```
Enselic, line 13: It's actually not dependent on interactive mode (atty). Because even when piping to a file, we want to strip the BOM if syntax highlighting is enabled.

I think the cases are:
 * In `config.loop_through` mode, never strip anything
 * When `config.colored_output` (I think it is that one) is enabled, always strip BOM, no matter if interactive or not. Otherwise regexes for syntaxes seems to break
 * If not loop through and if not colored output, I suppose we should keep the BOM, but maybe we can get away with stripping it if it makes the implementation much simpler

It would be fantastic if you could also ensure that there is a regression tests for each case above, although I would not go as far as saying that it is absolutely required for the PR.

Enselic (changes requested):
Thanks a lot for working on this! I took a high level look and had a high level comment.

CHANGELOG.md
Enselic: > If not loop through and if not colored output, I suppose we should keep the BOM

Actually, we should probably strip it also in this case, since some terminals have problems displaying it, and it doesn't do anything useful, especially not if it's in the "middle" of the output when bat prints its decorations.

dag-h: Sorry for the late response! I will have a look at this.

dag-h:
BOM is now stripped from all non-`config.loop_through` output. I'm not entirely confident in how robust the regression tests are, but I had a go at them as well.

src/printer.rs
```diff
@@ -393,6 +393,7 @@ impl<'a> Printer for InteractivePrinter<'a> {
                     .map_err(|_| "Invalid UTF-16BE")?,
                 _ => String::from_utf8_lossy(line_buffer).to_string(),
             }
+            .replace('\u{feff}', "") // Removing BOM from all non-loop-through output
```
sharkdp, line 396: This will call `.replace` on every single line in the output, right?
It will also replace the "zero width no-break space" character everywhere in the output, not just at the beginning.

Do we really need/want that? Doesn't this add a runtime overhead?

I would have expected that it is enough to call `.strip_prefix('\u{feff}')` on the very first line of the output.

https://doc.rust-lang.org/stable/std/string/struct.String.html#method.strip_prefix

dag-h: If it's expected to only appear at the first line, and always at the beginning of a line (without any other strange unicode characters before it perhaps), `strip_prefix` should be enough. The only issue I can think of then would be if we have two concatenated files, both with a BOM at the beginning. Will the BOM be stripped from them individually if we were to strip it in `print_line`?

I agree that calling `.replace` on every line might have too big of a hit on performance. I just wasn't sure if it was enough to only strip from the first line. I'm also not familiar enough with file encodings to know if the BOM always appears as the very first character.

dag-h: BOM is now only stripped when `line_number` is 1 and only from the very start of the line using `strip_prefix`.

sharkdp:
Sorry for leaving this hanging for so long. I think this actually looks quite good! I have one minor question: Is there any chance that this would interfere with any of the other `bat` command line options? For example, does everything still work in `bat -A` mode? I.e.: is the BOM still shown?

Or does this interfere/interact somehow with `--line-range` or similar options?

src/printer.rs
sharkdp: > The only issue I can think of then would be if we have two concatenated files, both with a BOM at the beginning

I'd say we can safely ignore this use case for now. I *think* that would be valid in terms of the Unicode standard. But it's certainly not the recommended way of designing a Unicode text file.

tests/integration_tests.rs
```diff
@@ -1225,3 +1228,48 @@ fn grid_for_file_without_newline() {
...
+        .stdout(
+            "\
+───────┬────────────────────────────────────────────────────────────────────────
+       │ File: test_BOM.txt
+───────┼────────────────────────────────────────────────────────────────────────
+   1   │ hello world
+───────┴────────────────────────────────────────────────────────────────────────
+",
+        );
+}
```
sharkdp, line 1275: :heart:

src/printer.rs
```diff
@@ -430,6 +430,15 @@ impl<'a> Printer for InteractivePrinter<'a> {
                     .decode(line_buffer, DecoderTrap::Replace)
                     .map_err(|_| "Invalid UTF-16BE")?,
                 _ => String::from_utf8_lossy(line_buffer).to_string(),
+            };
+            // Remove byte order mark from the first line if it exists
+            if line_number == 1 {
+                match line.strip_prefix('\u{feff}') {
+                    Some(stripped) => stripped.to_string(),
```
sharkdp, line 437: I wonder if there is an alternative implementation where we do not have to allocate a new string? by operating on `&str` instead of `String`.

But .. this is only relevant for files starting with a BOM. So maybe it's not the most urgent question.

sharkdp:
Fixed the merge conflicts.

sharkdp (approved):

CHANGELOG.md
sharkdp: Looks like this has been resolved with the three new regression tests.

Enselic:
Looks like tests are failing in Mac and Windows, but I will be to review those once that has been sorted out (I did a quick try at fixing the failures myself but it did not seem trivial to fix)

sharkdp:
> I did a quick try at fixing the failures myself but it did not seem trivial to fix

Indeed. Two distinct tests were failing on macOS and Windows, for two different reasons:
- macOS has a different color scheme by default (presumably due to the dark mode changes) and one test relied on explicit ANSI color codes => fixed by matching with a regex
- Windows test was failing because the terminal width was not explicitly set, and the Windows runner (presumably) has a different terminal width.

In addition, the "Run tests with updated syntaxes and themes" run was failing because we have one syntax test with a UTF-16 BOM (PowerShell). Previously, `bat`s highlighted output looked like this:

image

The `ef bb bf` part is the UTF-8-encoded version of the UTF-16 BOM, which is surrounded by some ANSI highlighting (for whatever reason). This PR improves the output and removes that whole first section (enclosed in red).

Enselic (approved):
Phenomenal! Let's merge this then! Code looks good to me 👍

PR #1945, opened 2021-11-14
Blend alpha value of gutter foreground color

rhysd:
Currently bat ignores an alpha value of gutter foreground color. This breaks gutter colors of some color themes. Here is the example with `Nord` theme:

- Current master branch

before

- This PR

after

As you see, gutter color of `Nord` theme is very light on current master branch. This is because the gutter color has actually an alpha value smaller than 255, which makes the color darker.

With this PR, the alpha value is correctly considered by blending the foreground color and the background color on initializing a printer instance.

### DIFF
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -29,7 +29,7 @@ use crate::error::*;
 use crate::input::OpenedInput;
 use crate::line_range::RangeCheckResult;
 use crate::preprocessor::{expand_tabs, replace_nonprintable};
-use crate::terminal::{as_terminal_escaped, to_ansi_color};
+use crate::terminal::{as_terminal_escaped, blend_foreground_alpha, to_ansi_color};
 use crate::wrapping::WrappingMode;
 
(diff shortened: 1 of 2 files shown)


src/printer.rs
```diff
@@ -647,7 +647,12 @@ impl Colors {
                 // Note: It might be the special value #00000001, in which case
                 // to_ansi_color returns None and we use an empty Style
                 // (resulting in the terminal's default foreground color).
-                Some(c) => to_ansi_color(c, true_color),
+                Some(mut fg) => {
+                    if let Some(bg) = theme.settings.background {
```
Enselic, line 651: Wouldn't it be more appropriate to use `ThemeSettings::gutter` as the background color for `gutter_foreground` here?

```
    /// Background color of the gutter.
    pub gutter: Option<Color>,
```

Enselic:
Closing due to inactivity to keep he PR inbox clean, but of course feel free to re-open if you resume work on this!

PR #1951, opened 2021-11-22
Remove minimal syntaxes

Enselic:
Now that [my syntect PR to lazy-load syntaxes](https://github.com/trishume/syntect/pull/393) has landed in syntect git master (:tada:), we can start to prepare for that in bat.

The first step is to remove all **minimal_syntaxes.bin** related code. It's not needed any longer since lazy-loading of syntaxes in syntect works much better.

|:---|---:|---:|---:|---:|
| `bat-v0.18.3 --color=always --language Rust examples/simple.rs` | 106.1 ± 2.0 | 103.5 | 113.1 | 4.58 ± 0.23 |
| `bat-lazy --color=always --language Rust examples/simple.rs` | 25.6 ± 1.0 | 23.9 | 29.2 | 1.10 ± 0.07 |
| `bat-minimal-syntaxes --color=always --language Rust examples/simple.rs` | 23.1 ± 1.1 | 21.4 | 27.7 | 1.00 |

* **bat-lazy** is this PR + syntect git master (https://github.com/Enselic/bat/pull/62)
* **bat-minimal-syntaxes** is current bat git master that uses minimal_syntaxes.bin.

As we can see, syntect lazy-loading is almost as fast. And I'm pretty confident in that if we were to take `minimal_syntaxes.bin "all the way", it would also add the extra milliseconds that the syntax lazy-loading has, if not more.

Much more importantly, lazy-loading syntect is faster in *all* cases, not just for half of our syntaxes when using `--language`. There are a lot more to say and do about syntect lazy-loading. More PRs and benchmarks will come in the near future :)

A bit sad to see the nifty graphviz analyzis code go of course, but it's not needed any longer. And it will always be available in git log. And in our hearts :)


sharkdp (approved):
Thank you very much! I have been closely following your PRs on the syntect repo. Awesome work.

> A bit sad to see the nifty graphviz analyzis code go of course, but it's not needed any longer. And it will always be available in git log. And in our hearts :)

Absolutely :heart_eyes:.

PR #1952, opened 2021-11-22
Replace `lazycell` and `lazy_static` with `once_cell`

Enselic:
We originally started to use lazycell because syntect was already using lazycell.

Syntect recently started to use once_cell instead to fix a race condition, see https://github.com/trishume/syntect/pull/391.

Once we start using the upcoming version of syntect, we will also want to stop using lazycell. So we might as well stop doing that already now.

once_cell is maintained by matklad (https://github.com/matklad/once_cell) and there is an open RFC to add its API to the standard library: https://github.com/rust-lang/rfcs/pull/2788. So starting to use once_cell should be a safe move.

Another nice thing with once_cell is that it supports lazy_static use cases without any macro magic (see https://docs.rs/once_cell/1.8.0/once_cell/#lazy-initialized-global-data). So this PR also removes our use of lazy_static and replaces it with once_cell.


sharkdp (approved):
Nice!

PR #1959, opened 2021-11-28
Better syntax highlighting for Dart

Ersikan:
Closes  #1958

This PR changes the dart highlighting source package to  [elMuso/Dartlight](https://github.com/elMuso/Dartlight).

This fixes the bug mentioned in the issue. I added a syntax test to check that comments in function calls are highlighted correctly.

However, the highlighting changes quite a bit. The two screenshots below show the difference :

Old:
Old Dart highlighting

New:
New Dart highlighting

PR #1961, opened 2021-11-28
CICD: Add check for accidental inclusion of GPL:ed code

Enselic:
I have confirmed that the script catches the problem that #1960 fixes.


sharkdp (approved):
Thank you!

Enselic:
I have also confirmed now that the CI check would have failed for the initial version of #1959 (and also that the fix makes the check pass)

PR #1968, opened 2021-12-01
release-checklist.md: Update `cargo outdated` section with link and arg

Enselic:
I don't think we can do much about transitive dependencies being out of date? So I think we want to run `cargo outdated --root-deps-only`?


keith-hall (approved):

sharkdp:
Ah, nice. Didn't know that option.

PR #1969, opened 2021-12-01
Lazy load themes

Enselic:
To further improve upon #951: It is quite easy to start lazy-loading themes, and it can be done without any changes to syntect.

It makes `bat` start ~25% faster when syntaxes are also lazy loaded. 

This might feel like overkill, but it is necessary if we want to be able to add new themes without feeling guilty about impacting startup time negatively.

Running `hyperfine` on `bat --color=always tests/benchmarks/test-src/small-CpuInfo-file.cpuinfo`:

|:---|---:|---:|---:|---:|
| git master | 99.6 ± 1.6 | 97.0 | 103.0 | 6.74 ± 0.35 |
| git master + lazy themes | 95.8 ± 1.0 | 94.5 | 98.6 | 6.49 ± 0.32 |
| git master + lazy syntaxes | 19.4 ± 0.9 | 18.0 | 22.0 | 1.31 ± 0.09 |
| git master + lazy themes + lazy syntaxes | 14.8 ± 0.7 | 13.5 | 16.5 | 1.00 |


Enselic:
Oops, forgot CHANGELOG.md. Will fix that after the first review round.

keith-hall (approved):

sharkdp (approved):
Very cool! :heart:

PR #1970, opened 2021-12-04
add a README-zh.md as the Chinese version of README.md

ac682:
I checked the Chinese one and found it out of dated for three years. So I wrote my own version from the original README.md.


Enselic (approved):
Thank you. Looks good to me. I did some Google Translate samples of the new text and it seemed reasonable.

The old fork we linked to does not seem maintained. Better to maintain everything in our repo. So this PR in a step in the right direction IMHO.

I will give other maintainers a chance to chime in before I merge this.

sharkdp (approved):

PR #1975, opened 2021-12-08
run-benchmarks.sh: Don't benchmark startup time syntaxes twice

Enselic:
Currently the files for startup time is benchmarked twice, which is a bit annoying. So separate highlight speed and startup time files to not mix them up in the benchmark code.

Left: After PR. Right: Before PR.

no-dups


Enselic:
I think this change is uncontroversial, and I would like to create a new PR that depends on this PR, so I'll just go ahead and merge this one.

PR #1977, opened 2021-12-08
pipe separated values

patatetom:
hi,
may adding the pipe character in fields separator ?
`column --table --separator $'\t' --output-separator ' | ' myfile.tsv | bat --plain --language tsv` will give a nice aligned and colored table ;-)
regards, lacsaP.

### DIFF
--- a/assets/syntaxes/02_Extra/CSV.sublime-syntax
+++ b/assets/syntaxes/02_Extra/CSV.sublime-syntax
@@ -7,7 +7,7 @@ file_extensions:
   - tsv
 scope: text.csv
 variables:
-  field_separator: (?:[,;\t])
+  field_separator: (?:[|,;\t])
   record_separator: (?:$\n?)
 contexts:
   prototype:


Enselic:
Thank you for making a contribution!

Since this syntax is our own and not from a submodule, we are quite free to do what we want.

I don't quite understand the usecase though, and your example command does not work for me unfortunately.

Can you provide the exact input you use please? So just do `column --table --separator $'\t' --output-separator ' | ' myfile.tsv > /tmp/output.txt` and then copy-paste a few lines of the contents of that file here.

patatetom:
hi,
I regularly use `bat` (thanks to him/you) with Tab Separated Data (tsv) : the coloring is well done but the alignment, like a table, is not (`bat` has nothing to do with it).
here is a command that should work and allow you to better understand my pull request :
```console
$ find ./bat-master/ -maxdepth 1 -type f -printf '%p\t%s\t%m\t%i\t%Tc\n' | bat -fpl tsv | column -t -s $'\t' -o ' | '
```
_this being said, I realize with this example that it is not necessary to modify `bat` to achieve my goals..._
regards, lacsaP.

keith-hall:
fwiw, the example didn't work for me:
```
column: invalid option -- 'o'
usage: column [-txne] [-c columns] [-s sep] [file ...]
```

but pipe separated csv files are common enough that I believe this PR has merit. It would be nice to add a syntax regression test for it, and a changelog entry.

patatetom:
a problem of version on `column` can be...
```console
$ LANG= column -V
column from util-linux 2.37.2
```

patatetom:
the `csv` regex may need to be revised as the coloring may be altered by the presence of characters that are not separators but considered as such.

for example this row whose fields are separated by the pipe character will not be colored correctly because of the presence of several characters considered as separators :

`field 1 | field 2, with a comma | field 3; with a semicolon | field 4`

keith-hall:
I have a potential solution at https://github.com/forkeith/bat/commit/c1a4e981897844cb145b6224f2fc4be42b1fdcce, but it gets hard to maintain while syntect doesn't support `extends`, `pop: 2` etc. I ran out of time to test more right now - to see if the regression tests pass etc. and to add new ones.

patatetom:
I'm not comfortable with that.
another solution - less technical - would be to have one file (regex) per separator : `csv` for comma only, `tsv` for tabulation only and `psv` for pipe only...

Enselic:
One syntax per separator like Keith experimented with in https://github.com/sharkdp/bat/pull/1977#issuecomment-992931431 seems to me like the most robust way to add support for pipe-separates values. 

In other words, I think we can close this particular PR, since its approach is too fragile, albeit simple.

Let me know if you disagree.

PR #1985, opened 2021-12-23
Underline highlighted lines in ANSI theme

### LINKED ISSUE #1730: Add lineHighlight for ansi theme
The ansi theme doesn't have `lineHighlight` defined, so the highlighted line is not apparent.

mdibaiee:
Hey 👋 
I have attempted to fix this in two ways: the [first commit](https://github.com/sharkdp/bat/pull/1985/commits/129210e9fa181c06409d5ad085a0b0744cc5fb1d) is a simple, ANSI-specific condition.
The [second commit](https://github.com/sharkdp/bat/pull/1985/commits/31404f6a172c74936481189fec58e0ec05975efd) is using a custom scope to allow each theme to customise how the highlighted line's font style is set.

Let me know what you think and I can squash to a single final commit.

Fixes #1730

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Features
 
 - New style component `header-filesize` to show size of the displayed file in the header. See #1988 (@mdibaiee)
+- Use underline for line highlighting on ANSI, see #1730 (@mdibaiee)
 
 ## Bugfixes
 
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -463,6 +463,10 @@ impl<'a> Printer for InteractivePrinter<'a> {
         let highlight_this_line =
             self.config.highlighted_lines.0.check(line_number) == RangeCheckResult::InRange;
 
+        if highlight_this_line && self.config.theme == "ansi" {
+            self.ansi_style.update("^[4m");
+        }
+
         let background_color = self
             .background_color_highlight
             .filter(|_| highlight_this_line);
@@ -647,6 +651,10 @@ impl<'a> Printer for InteractivePrinter<'a> {
                 )?;
             }
             writeln!(handle)?;
+        }
+
+        if highlight_this_line && self.config.theme == "ansi" {
+            self.ansi_style.update("^[24m");
         }
 
         Ok(())
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -655,6 +655,7 @@ impl<'a> Printer for InteractivePrinter<'a> {
 
         if highlight_this_line && self.config.theme == "ansi" {
             self.ansi_style.update("^[24m");
+            write!(handle, "\x1B[24m")?;
         }
 
         Ok(())
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -1290,6 +1290,25 @@ fn grid_for_file_without_newline() {
         .stderr("");
 }
 
+// For ANSI theme, use underscore as a highlighter
+#[test]
+fn ansi_highlight_underline() {
+    bat()
+        .arg("--paging=never")
+        .arg("--color=never")
+        .arg("--terminal-width=80")
+        .arg("--wrap=never")
+        .arg("--decorations=always")
+        .arg("--theme=ansi")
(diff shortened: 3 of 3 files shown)


assets/themes/ansi.tmTheme
```diff
@@ -423,6 +423,17 @@
                     <string>#03000000</string>
                 </dict>
             </dict>
+            <dict>
+                <key>name</key>
+                <string>Line highlight style</string>
+                <key>scope</key>
+                <string>line_highlight</string>
```
keith-hall, line 430: I'm not sure how I feel about this being a scope selector with specific meaning. It's a very clever solution, for sure :) but I wonder if it would be better to re-use/re-purpose a global color scheme "setting" which already exists, like `tags_options` (http://www.sublimetext.com/docs/color_schemes.html#global_settings-tags) which otherwise wouldn't have any use in a tool like `bat`. To me this would seem more portable and unsurprising.

keith-hall (commented):
Thanks for your contribution! I like the first commit - despite it being specific for the ANSI theme and hard-coded, it is very simple. I have a few concerns about the second commit's approach, namely it seems weird to use a "custom" scope selector as a global setting, so I'd be interested to hear what the other maintainers think :)

assets/themes/ansi.tmTheme
mdibaiee: @keith-hall wouldn't re-purposing another scope cause more surprise and confusion for people? Are there actually _standard_ scopes at all? My sense is that they are kind of arbitrary, and different editors can have their own, but I might be mistaken.

mdibaiee: From the [docs](http://www.sublimetext.com/docs/color_schemes.html#global_settings-tags):
> Author of syntaxes can assign whatever scopes they want to a given token. This combined with the fact that there are hundreds of community-maintained syntaxes means that is can be hard to know what scopes to target. The official Scope Naming guidelines were established to help syntax and color scheme authors use a common set, for better interoperability. The Usage in Color Schemes section provides a baseline set of scopes that color scheme authors should strive to handle. 

My understanding is that themes should try to handle the "baseline" scopes, but they are still free to add their own. But again, I'm new to this area and I'm sure you know better about the best practices on this subject.

keith-hall: It's a good point - as I understand, syntax definitions define the scopes for the tokens, and color schemes can choose which of those to refer to / target. It would be bad practice to introduce a scope which is not part of the guidelines as only a single color scheme would use it.

mdibaiee: Okay... I don't mind going with the simple, ansi-specific approach of the first commit either. Let's see what other maintainers think.

mdibaiee: @sharkdp @Enselic any ideas on this one? we could use some input here. Thanks!

Enselic: I honestly don't know much about how to best and most properly structure themes and syntaxes. I fully trust Keiths judgement in these matters. He is one of the world's most prominent experts in this area :)

mdibaiee: @keith-hall I'm happy to go with the first commit approach. I also trust you know better about the usage of the themes and their scopes, etc., and if that doesn't look right to you, let's avoid it.

sharkdp:
This PR still has a change in `syntaxes.bin`. Should that be removed?

mdibaiee:
@sharkdp I had overlooked that! removed that change 👍

Enselic:
Would you mind adding a simple regression test for this change? It is useful not only preventing regressions, but also to make it clearer what the use case is and how it works.

It would be great if you could set Approved on this when you think it looks good @keith-hall (when you have time, of course).

sharkdp (approved):
Approving. But I agree with @Enselic that a regression test would be great.

mdibaiee:
Yes definitely, good call. I just added a test, and while doing so realised it actually had a bug with `--style=plain` (it worked with the grid style) where it wouldn't clean up the underline. 🐛 All good now.

PR #1987, opened 2021-12-24
Add note regarding options passed to less by default in manpage 

### LINKED ISSUE #1913: Bat displays escape codes when run with sudo and less -X or -F

```sh
❯ sudo bat .
[bat error]: '.' is a directory.
❯ sudo bat . --pager=less
[bat error]: '.' is a directory.
❯ sudo bat . --pager='less -F'
ESC[31m[bat error]ESC[0m: '.' is a directory.
❯ bat . --pager=less
[bat error]: '.' is a directory.
❯ bat . --pager='less -F'
[bat error]: '.' is a directory.
❯ bat .
[bat error]: '.' is a directory.
❯ sudo bat . --pager='less -X'
ESC[31m[bat error]ESC[0m: '.' is a directory.
```
image

Adding options to the pager does not break coloring.

Arch

---

#### Software version

bat 0.18.3 (b146958)

#### Operating system

Linux 5.14.9-arch2-1

#### Command-line

```bash
bat . '--pager=less -F' --diagnostic 
```

#### Environment variables

```bash
SHELL=/bin/zsh
PAGER=less
LESS=<not set>
BAT_PAGER=<not set>
BAT_CACHE_PATH=<not set>
BAT_CONFIG_PATH=<not set>
BAT_OPTS=<not set>
BAT_STYLE=<not set>
BAT_TABS=<not set>
BAT_THEME=<not set>
XDG_CONFIG_HOME=<not set>
XDG_CACHE_HOME=<not set>
COLORTERM=truecolor
NO_COLOR=<not set>
MANPAGER=<not set>
```

#### Config file

Could not read contents of '/root/.config/bat/config': No such file or directory (os error 2).

#### Compile time information

- Profile: release
- Family: unix
- OS: linux
- Architecture: x86_64
- Pointer width: 64
- Endian: little
- CPU features: fxsr,sse,sse2

#### Less version

```
> less --version 
less 590 (PCRE2 regular expressions)
Copyright (C) 1984-2021  Mark Nudelman

```

mdibaiee:
Fixes #1913

Signed-off-by: Mahdi Dibaiee <mdibaiee@pm.me>

### DIFF
--- a/assets/manual/bat.1.in
+++ b/assets/manual/bat.1.in
@@ -118,6 +118,8 @@ which pager is used, see the '\-\-pager' option. Possible values: *auto*, never,
 Determine which pager is used. This option will override the PAGER and BAT_PAGER
 environment variables. The default pager is 'less'. To control when the pager is used, see
 the '\-\-paging' option. Example: '\-\-pager "less \fB\-RF\fR"'.
+
+Note: By default, if the pager is set to 'less' (and no command-line options are specified), 'bat' will pass the following command line options to the pager: '-R'/'--RAW-CONTROL-CHARS', '-F'/'--quit-if-one-screen' and '-X'/'--no-init'. The last option ('-X') is only used for 'less' versions older than 530. The '-R' option is needed to interpret ANSI colors correctly. The second option ('-F') instructs less to exit immediately if the output size is smaller than the vertical size of the terminal. This is convenient for small files because you do not have to press 'q' to quit the pager. The third option ('-X') is needed to fix a bug with the '--quit-if-one-screen' feature in old versions of 'less'. Unfortunately, it also breaks mouse-wheel support in 'less'. If you want to enable mouse-wheel scrolling on older versions of 'less', you can pass just '-R' (as in the example above, this will disable the quit-if-one-screen feature). For less 530 or newer, it should work out of the box.
 .HP
 \fB\-m\fR, \fB\-\-map\-syntax\fR <glob-pattern:syntax-name>...
 .IP


Enselic (changes requested):
Thanks for taking the time to make a contribution, but I think you missed the paragraph 4 paragraphs down that also talks about `--RAW-CONTROL-CHARS`?

mdibaiee:
@Enselic ah, you are right, then I guess the original issue #1913 can be closed as this is already documented?

Enselic:
The issue talks about the man page. I suspect we don't mention it in the man page. Maybe we should?

mdibaiee:
@Enselic oh I see, that's my misunderstanding then. I'll see if I can add it to the manpage 👍

sharkdp:
@mdibaiee Thank you very much. Did you verify that this "renders" correctly in `man` by building `bat` and checking the resulting `bat.1` file via `bat -l bat.1`?

mdibaiee:
Yes, this is the resulting manpage section:

```
man target/debug/build/bat-62ebb303f306a280/out/assets/manual/bat.1
```
```
       --pager <command>

              Determine which pager is used. This option will override the PAGER and BAT_PAGER environment variables. The default pager is 'less'. To control when the pager is used,
              see the '--paging' option. Example: '--pager "less -RF"'.

              Note: By default, if the pager is set to 'less' (and no command-line options are specified), versions older than 530.  The '-R' option is needed to interpret ANSI col-
              ors  correctly.  The  second option ('-F') instructs less to exit immediately if the output size is smaller than the vertical size of the terminal.  This is convenient
              for small files because you do not have to press 'q' to quit the pager. The third option ('-X') is needed to fix a bug with the '--quit-if-one-screen' feature  in  old
              versions of 'less'. Unfortunately, it also breaks mouse-wheel support in 'less'.  If you want to enable mouse-wheel scrolling on older versions of 'less', you can pass
              just '-R' (as in the example above, this will disable the quit-if-one-screen feature). For less 530 or newer, it should work out of the box.
```

image

mdibaiee:
Okay now it's rendering correctly:

```
              Determine which pager is used. This option will override the PAGER and BAT_PAGER environment variables. The default pager is 'less'. To control when the pager is used,
              see the '--paging' option. Example: '--pager "less -RF"'.

              Note: By default, if the pager is set to 'less' (and no command-line options are specified), 'bat'  will  pass  the  following  command  line  options  to  the  pager:
              '-R'/'--RAW-CONTROL-CHARS',  '-F'/'--quit-if-one-screen'  and '-X'/'--no-init'. The last option ('-X') is only used for 'less' versions older than 530. The '-R' option
              is needed to interpret ANSI colors correctly. The second option ('-F') instructs less to exit immediately if the output size is smaller than the vertical size  of  the
              terminal.  This  is convenient for small files because you do not have to press 'q' to quit the pager. The third option ('-X') is needed to fix a bug with the '--quit-
              if-one-screen' feature in old versions of 'less'. Unfortunately, it also breaks mouse-wheel support in 'less'. If you want to enable  mouse-wheel  scrolling  on  older
              versions  of  'less',  you can pass just '-R' (as in the example above, this will disable the quit-if-one-screen feature). For less 530 or newer, it should work out of
              the box.
```

image

sharkdp (approved):
Thank you!

Enselic (approved):
Am out in the forest so my abilities are limited, but it looks good to me

Issue #1743, opened 2021-07-21
Overlap between --style=auto and --decorations=auto is confusing


lilyball:
**Describe the bug you encountered:**

There's two different ways of specifying that decorations should not be displayed when the output is piped, and they both independently suppress the decorations. In particular, this is `--style=auto` and `--decorations=auto`. Both are documented as being the default (though it appears that `--style=auto` is not in fact the default, see #1742, possibly because of this issue). This is confusing as it means `bat --style=full foo | cat` omits decorations, and `bat --style=auto --decorations=always foo | cat` also omits decorations (this is more confusing if `--style=auto` is set in config, and so `bat -f foo | cat` will still omit decorations).

This is also frustrating if you want to turn off the auto style but aren't sure if the inherited style is actually auto or not. For example, `bat --style=full -f foo | cat` will ensure all decorations are on, but if my config includes `--number` then this will override that and turn on all decorations. Which is to say, there's no way to ensure decorations are visible without either knowing the current style or risking changing the style.

**What did you expect to happen instead?**

At the very least, `bat` needs to stop claiming `--style=auto` is the default, and it really should make it clear what the relationship is between `--style` and `--decorations`, and that saying `--style=auto` means `--decorations=always` isn't sufficient to print decorations when piped. This doesn't solve the problem of forcing auto decorations on though.

Also it looks like `bat` can actually mark some style components as auto and others as always, by writing this like `--style=auto,numbers`, such that `bat --style=auto,numbers -f foo | cat` will print numbers but printing to the terminal will display all decorations. This is undocumented, but is an argument against changing the actual meaning of `--style=auto`.

But I think we can also do something intelligent here by providing per-style-component control over auto behavior. We could allow for writing something like `--style=numbers:auto,header` (and then `--style=auto` just becomes the same thing as `--style=full:auto`), or perhaps a prefix sigil like `--style=*numbers,header` (or maybe some other sigil that doesn't need escaping from the shell). This doesn't solve the problem of "both style and decorations have independent auto toggles that stack", but it may make the fact that they stack a bit easier to understand (i.e. "`--decorations=auto` controls decorations when piping normally, but you can flag independent style components as auto if you want them to still be omitted when piping with `-f`").

[...]

PR #1988, opened 2021-12-25
Add new `--style` called `header-filesize` and display it by default

### LINKED ISSUE #1701: Extra file information in the header
Sometimes, you want to quickly glance at more information, in the header, than just the file name, e.g.: mime type, or file size, or relevant timestamps/dates, or full path, or permissions, etc.

I've managed to somewhat accomplish this with just the `--file-name` option, but (A) I don't feel it's really appropriate, and (B) syntax detection fails, so you don't get any highlighting which sucks, and (C) the grid lines aren't printed:

```bash
❯ f=LICENSE.txt && bat \
--file-name "$f
       Canonical: $(greadlink -f $f)
       Filetype:$(file $f | cut -d':' -f2)
$(stat -x $f | tail -n3 | sed -E 's/^(.+)/       \1/')" \
$f
```

image

---

So yeah I thought this might be an interesting feature, that perhaps others might find useful as well. Also this would make the `fzf` & `bat` integration really quite nice and even more useful.

---

To be a bit more precise on the actual feature, without getting into implementation: a feature that allows inserting a variety of other useful information on the file, in the header. Either you could provide us with a list of possible additional informations, kinda like `--style`'s possible values, and we list them out in an option, e.g.:

```
bat --header-info=fullpath,filetype,access,modify,change[,...]
```

AND-additionally/OR provide the ability to specify arbitrary scripts that produce some text output, or some formatted (e.g. JSON, XML, HTML, or your custom designed format) output, something that you can easily read and print in the header; also you could limit the output to 1 line, or maybe not, I don't know. This could be kind of like `fzf`'s `--bind '...:execute(<script>)'` option, where the user provides a script, i.e some chain of commands and not a script file, of what they want. E.g.:

```
bat\
--header-info=fullpath \
--header-info=execute(file {} | cut -d':' -f2)
```

`{}` being a placeholder for the file.

The output would be printed in the header, 1 line under the other, in the order specified.

---

Also, Thank you for your wonderful work with `bat` and all the other command line tools you've created, they're really useful !

mdibaiee:
Fixes #1701


src/input.rs
```diff
@@ -84,9 +87,17 @@ impl<'a> InputKind<'a> {
     }
 }
 
+#[derive(Clone)]
+pub(crate) struct InputPermissions {
+    pub(crate) mode: u32,
```
mdibaiee, line 92: Apparently getting user and group is not supported in the std library and we would need to develop further for that, so I left that as a later improvement to avoid complicating this PR further.

src/printer.rs
```diff
@@ -315,17 +310,73 @@ impl<'a> Printer for InteractivePrinter<'a> {
...
+                        metadata
+                            .permissions
+                            .clone()
+                            .map(|perm| perm.mode)
+                            .unwrap_or(0)
+                    ),
+
+                    HeaderComponent::LastModified => {
+                        let format = format_description::parse(
+                            "[day] [month repr:short] [year] [hour]:[minute]:[second]",
```
mdibaiee, line 363: I'm following `ls -lh` format here, with the difference that I'm including seconds.

src/printer.rs
```diff
@@ -315,17 +310,73 @@ impl<'a> Printer for InteractivePrinter<'a> {
...
+                        self.colors.filename.paint(description.title()),
+                        mode
+                    ),
+
+                    HeaderComponent::Size => {
+                        let bsize = metadata
+                            .size
+                            .map(|s| format!("{}", ByteSize(s)))
+                            .unwrap_or("".into());
+                        writeln!(handle, "Size: {}", bsize)
```
mdibaiee, line 348: Not sure if we should completely omit the line if the value is `None`, or just print an empty value?

src/header.rs
```diff
@@ -0,0 +1,88 @@
...
+            "permissions" => Ok(HeaderComponent::Permissions),
+            "last-modified" => Ok(HeaderComponent::LastModified),
+            "full" => Ok(HeaderComponent::Full),
+            _ => Err(format!("Unknown header-info '{}'", s).into()),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct HeaderComponents(pub Vec<HeaderComponent>);
```
mdibaiee, line 66: This is a `Vec` instead of `HashSet`, because unlike `style`, the ordering of these options matters. `HashSet` creates an unpredictable output ordering (unless we sort it in the code somehow).

src/bin/bat/app.rs
```diff
@@ -338,4 +341,38 @@ impl App {
...
+            matches
+                .values_of("header-info")
+                .map(|header| {
+                    header
+                        .map(|header| header.parse::<HeaderComponent>())
+                        .filter_map(|header| header.ok())
+                        .collect::<Vec<_>>()
+                })
+                .or(env_header_components)
+                .unwrap_or_else(|| vec![HeaderComponent::Full])
```
mdibaiee, line 367: The default I've set to `Full` now, but I'm not sure if we should have the default as `Filename` alone, to be backward compatible?

Cargo.toml
```diff
@@ -65,6 +65,8 @@ dirs-next = { version = "2.0.0", optional = true }
 grep-cli = { version = "0.1.6", optional = true }
 regex = { version = "1.0", optional = true }
 walkdir = { version = "2.0", optional = true }
+time = { version = "0.3.5", features = ["formatting"] }
```
mdibaiee, line 68: I just noticed this breaks our minimum rust version support, as the minimum supported version of this library is 1.51. If necessary, I can switch to `chrono` or some other library for formatting the timestamp.

mdibaiee:
I have not written tests yet, waiting for some feedback on my approach and implementation, I will make necessary changes and then write tests in the final steps

Cargo.toml
Enselic: Hi and thanks for making a contribution! I don't have time to review at the moment, but me or someone else can probably look ath these in max a few weeks

however, I can already now say that bat will soon bump to 1.51 as well, because syntect already has, and bat will soon depend on the soon to be made syntect 4.7.1 release

Enselic: In fact, feel free to create a PR that bumps MSRV to 1.51, I would set Approved on such a PR

mdibaiee: Hello! Sounds good, will create a MR for that as well. No rush for reviewing the MR either, but thanks for letting me know in advance.

mdibaiee: https://github.com/sharkdp/bat/pull/1994

src/bin/bat/app.rs
mdibaiee: I actually changed this to be `Filename` since I noticed a lot of tests were failing as well.

Cargo.toml
```diff
@@ -65,6 +65,8 @@ dirs-next = { version = "2.0.0", optional = true }
...
+bytesize = {version = "1.1.0", features = ["serde"]}
```
Enselic, line 69: Again, thanks a lot for working on this.

I wrote a proposal in #1701 that maybe we should keep using `--style`.

Another note: I'm a big believer in small and incremental steps when developing software. I think a good first minimal scope would be to only add support for showing the file size in bytes (so we don't need to add a dependency on `bytesize` in the first iteration). That way we get a clear picture of what the infrastructure would look like for supporting more kinds of info in the header.

Once that basic PR has landed, we can open new PRs that adds more header info types and e.g. adds a dependency on `bytesize`.

In my experience an incremental approach to develop software is much more pleasant for everyone. It's pleasant for maintainers since CR becomes easier. It's pleasant for contributors since they don't end up with a giant branch that gets conflicts over and over. And it's pleasant for users who gets new features quicker.

I want to stress that this is just a proposal. If you strongly prefer to do it all in one go in one PR then we can make that work too :)

mdibaiee: @Enselic I'll experiment with your proposal on #1701 and we'll see how it turns out, I think it's a good idea!

Regarding small incremental steps, it all comes down the trade-off between overhead of multiple PRs vs. overhead of a larger PR: it's subjective and case-by-case can differ whether a PR is large enough to justify multiple PRs, or if it's okay to stay as one as adding more PRs might cause even more overhead: for maintainers to keep track and review multiple pieces, for the contributor to manage multiple branches, and for users to have their features changed in quick succession (e.g. output is byte now, and next week it's human formatted, that's a surprise).

In this case, I personally think the addition of human formatting, last modified at and permissions doesn't add much more complexity to the PR, the majority of the code is the infrastructure, and these 3 are some basic usages of that infrastructure, on the other hand, adding file owner and group would add a lot more complexity, hence I dropped it from this PR and said we can do it later. All in all though, I'm happy to split this into smaller PRs, or continue as it is, not a big difference for me.

mdibaiee:
Okay, so I moved the header fields into `--style`, and deprecated `header` value of `--style` in favour of all the others. If any of the header fields are enabled, `StyleComponents.header()` returns true, so it allows us to know whether we need to draw a header at all.

Example usage:

```
target/debug/bat --style grid,filename,filesize,permissions,lastmodified CONTRIBUTING.md

────────────────────────────────────────────────────────
File: CONTRIBUTING.md
Size: 1.6 KB
Permissions: 644
Last Modified At: 23 Dec 2021 21:16:17
────────────────────────────────────────────────────────
```

Note that I also added a deprecation warning for `header` value:
```
[bat warning]: Style 'header' is deprecated, use 'filename' instead.
```

This has broken a lot of tests, I will address those once I know we agree on the approach.

src/style.rs
```diff
@@ -11,6 +11,10 @@ pub enum StyleComponent {
     Grid,
     Rule,
     Header,
+    Filename,
```
Enselic, line 14: Just so I don't forget: We should consider marking this enum as `#[non_exhaustive]`, similar to #2012

sharkdp:
> Note that I also added a deprecation warning for header value:
> [bat warning]: Style 'header' is deprecated, use 'filename' instead.

I did not have time to review this at all, but dare to ask: would it maybe make sense to keep `header` around and change it's meaning to a default set of enabled header infos? Also, should we maybe prefix the other new options with `header-`? Something like:
```
header-filename
header-filesize
header-permissions
header-lastmodified
header = header-filename + header-filesize
header-full = header-filename + header-filesize + header-permissions + header-lastmodified
```

Just a suggestion - I haven't thought about it in detail.

mdibaiee:
@sharkdp I wouldn't mind that either, that might help by being:

1. backward compatible (somewhat... since we might add more than just filename to the default)
2. namespaces header components for clarity / ease of search

Let's hear what others have to say.

Enselic:
I actually had `header-filename` first before I submitted the comment, but was afraid it was too verbose, so I simplified to `filename`. Since our initial instinct was the same, maybe we should go with `header-filename` etc after all. I am open to either way.

mdibaiee:
Okay, I will update the code according to @sharkdp's comment (hopefully tomorrow). Thanks!

src/bin/bat/clap_app.rs
```diff
@@ -384,7 +384,7 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
                 .validator(|val| {
                     let mut invalid_vals = val.split(',').filter(|style| {
                         !&[
-                            "auto", "full", "plain", "header", "grid", "rule", "numbers", "snip",
+                            "auto", "full", "plain", "header", "header-full", "header-filename", "header-filesize", "header-permissions", "header-lastmodified", "grid", "rule", "numbers", "snip",
```
Enselic, line 387: Sorry to nit-pick about formatting (I try to avoid nit-picking), but can you split this up into many lines please? `cargo fmt` seems to be OK with having it like this instead, which is more manageable:

```
                            "auto",
                            "full",
                            "plain",
                            "header",
                            "header-full",
                            "header-filename",
                            "header-filesize",
                            "header-permissions",
                            "header-lastmodified",
                            "grid",
                            "rule",
                            "numbers",
                            "snip",
                            #[cfg(feature = "git")]
                            "changes",
```

Enselic:
I think this is starting to look good on a high level, so it's soon time to start detailed review.

One high-level thing to discuss is what header fields to show by default. Right now all new header fields are shown. I think Size is a reasonable default. Maybe even last modification time. But I'm not sure Permissions is worth showing by default. Especially since bat is read-only. What do you think of skipping showing Permissions by default?

I also think we should simplify `Last Modified At` to just `Modified`.

It would also be neat to align the header field values vertically. But I think it is fine to skip that in the first version of this feature.

One thing that might be worth doing in this initial PR is to also mark the other header fields values in bold, and not only the filename.

For context, here is current default output of `cargo run examples/simple.rs` in the PR.
```
───────┬──────────────────────────────────────────────────────────────────────────────────
       │ File: examples/simple.rs
       │ Size: 177 B
       │ Permissions: 644
       │ Last Modified At: 23 Jul 2021 18:57:19
───────┼──────────────────────────────────────────────────────────────────────────────────
   1   │ /// A simple program that prints its own source code using the bat library
   2   │ use bat::PrettyPrinter;
   3   │ 
   4   │ fn main() {
   5   │     PrettyPrinter::new().input_file(file!()).print().unwrap();
   6   │ }
───────┴──────────────────────────────────────────────────────────────────────────────────
```

I also took a look at what impact this PR has on the public API, using a new tool I wrote:
```
% cargo install cargo-public-items
% git checkout origin/master
% cargo public-items > /tmp/bat-master-api
% gh pr checkout 1988
% cargo public-items > /tmp/bat-header-info-api
% diff -u /tmp/bat-master-api /tmp/bat-header-info-api
--- /tmp/bat-master-api	2022-01-09 16:13:25.000000000 +0100
+++ /tmp/bat-header-info-api	2022-01-09 16:15:16.000000000 +0100
@@ -208,6 +208,11 @@
 bat::style::StyleComponent::Full
 bat::style::StyleComponent::Grid
 bat::style::StyleComponent::Header
+bat::style::StyleComponent::HeaderFilename
+bat::style::StyleComponent::HeaderFilesize
+bat::style::StyleComponent::HeaderFull
+bat::style::StyleComponent::HeaderLastModified
+bat::style::StyleComponent::HeaderPermissions
 bat::style::StyleComponent::LineNumbers
 bat::style::StyleComponent::Plain
 bat::style::StyleComponent::Rule
@@ -226,6 +231,10 @@
 bat::style::StyleComponents::fmt
 bat::style::StyleComponents::grid
 bat::style::StyleComponents::header
+bat::style::StyleComponents::header_filename
+bat::style::StyleComponents::header_filesize
+bat::style::StyleComponents::header_last_modified
+bat::style::StyleComponents::header_permissions
 bat::style::StyleComponents::new
 bat::style::StyleComponents::numbers
 bat::style::StyleComponents::plain
```

which looks like a reasonable diff.

mdibaiee:
@Enselic regarding the default, I prefer Filename and Filesize, leaving the rest to be specified by users if necessary (I rarely look up or use last modified date of files, but I can't say that for everyone).

I updated the code to make other component values bold as well.

So far then, I only need to write tests, but open to more feedback, too.

src/bin/bat/app.rs
```diff
@@ -322,7 +322,7 @@ impl App {
                             .collect::<Vec<_>>()
                     })
                     .or(env_style_components)
-                    .unwrap_or_else(|| vec![StyleComponent::Full])
+                    .unwrap_or_else(|| vec![StyleComponent::Auto])
```
Enselic, line 325: Not showing all header fields by default has surfaced a somewhat thorny problem.

First of all, we don't want to change this to `Auto`. Wether or not to show style components is and should remain to be `--decorations` which defaults to `auto`.

It might very well make sense to remove `StyleComponent::Auto`. At least from the CLI, to remain semver compatible with `bat` as a library. See https://github.com/sharkdp/bat/issues/1743 and linked issues.

src/bin/bat/clap_app.rs
```diff
@@ -422,11 +433,17 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
                      '--style=\"..\"' option to the configuration file or export the \
                      BAT_STYLE environment variable (e.g.: export BAT_STYLE=\"..\").\n\n\
                      Possible values:\n\n  \
-                     * full: enables all available components (default).\n  \
```
Enselic, line 425: Here we have another high level problem. We don't want to make this `auto` by default. But  we don't want `full` as default either since we don't want to show all header fields by default.

So we have two choices now:
1. We only add `header-filename` and `header-filesize` in this PR so we can keep `full` as default, which in turn allows us to postpone this problem to a follow-up PR (this is the option I prefer)
2. We deal with this problem in this PR. Maybe we should add a new variant `recommended` that is default? Or maybe just call the default `default`?

I would like to ask @sharkdp and @keith-hall: Are you OK with starting to show both `header-filename` and `header-filesize` by default? If not, then we might as well go with option 2 right away and skip a potential follow-up PR.

src/bin/bat/clap_app.rs
```diff
@@ -422,11 +433,17 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
...
-                     * auto: same as 'full', unless the output is piped.\n  \
+                     * auto: enables all components and 'header', unless the output is piped.
+                     * (default)\n  \
+                     * full: enables all available components.\n  \
                      * plain: disables all available components.\n  \
                      * changes: show Git modification markers.\n  \
-                     * header: show filenames before the content.\n  \
```
Enselic, line 429: I would prefer if we let `header` remain an alias for just `header-filename` rather than change its meaning.

src/bin/bat/clap_app.rs
```diff
@@ -422,11 +433,17 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
...
+                     * header: displays the filename and filesize.\n \
+                     * header-full: displays all header-* fields.\n \
```
Enselic, line 442: If we go with option 1 above I think we can also postpone adding `header-full` but maybe add it in a follow-up PR.

src/input.rs
```diff
@@ -84,9 +88,17 @@ impl<'a> InputKind<'a> {
...
+}
+
 #[derive(Clone, Default)]
 pub(crate) struct InputMetadata {
     pub(crate) user_provided_name: Option<PathBuf>,
+    pub(crate) size: Option<u64>,
+    pub(crate) permissions: Option<InputPermissions>,
```
Enselic, line 100: This might be an irrelevant point if we go with option 1, but why not store the `mode` directly? I suspect you introduced `InputPermissions` as a form of future-proofed abstraction, but since this is internal to `bat` and not part of the API I would prefer if we kept it as simple as possible.

src/printer.rs
```diff
@@ -315,17 +310,102 @@ impl<'a> Printer for InteractivePrinter<'a> {
         };
 
         let description = &input.description;
+        let metadata = &input.metadata;
+
+        // We use this iterator to have a deterministic order for
```
Enselic, line 315: If we go with option 1 (which I prefer we don't need a Vec, since the filename shall always go before the filesize if present.

I'm actually not sure we need to allow a custom order even if we go with option 2. But I would prefer if we went with option 1 so that we can postpone this problem to a follow-up PR.

src/printer.rs
```diff
@@ -315,17 +310,102 @@ impl<'a> Printer for InteractivePrinter<'a> {
...
-            "{}{}{}",
-            description
-                .kind()
-                .map(|kind| format!("{}: ", kind))
-                .unwrap_or_else(|| "".into()),
-            self.colors.filename.paint(description.title()),
-            mode
-        )?;
+            match component {
+                StyleComponent::HeaderFilename => writeln!(
```
Enselic, line 353: It would be nice it we could de-duplicate the code in the match arms a bit

Enselic (changes requested):
Here comes another round of comments.

May I also ask you to keep the CI green from now on, and before the next round of review? Let me know if you have troubles reproducing the failures in CI locally, and I'll help you out.

And thanks a lot for working on this PR!

src/bin/bat/clap_app.rs
keith-hall: I'm okay with it :)

src/printer.rs
mdibaiee: After having removed permissions and lastmodified, I don't see much to de-duplicate at this point. Let me know if you have ideas about it.

mdibaiee:
Thanks @Enselic for the review. I went with Option 1, removed permissions and lastmodified, so now it's only filename and size. Let me know if your thoughts.

src/bin/bat/clap_app.rs
```diff
@@ -426,7 +434,9 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
                      * auto: same as 'full', unless the output is piped.\n  \
                      * plain: disables all available components.\n  \
                      * changes: show Git modification markers.\n  \
-                     * header: show filenames before the content.\n  \
+                     * header: show filenames before the content.\n \
+                     * header-filename: displays the file name.\n  \
+                     * header-filesize: displays the size of the file in human-readable format.\n  \
```
Enselic, line 439: What do you think about reusing previous description in the new one for filename, refer to the new one in the old one, and make header-filesize description more similar in style to header-filesize? Like this:

```suggestion
                     * header: legacy alias for 'header-filename'.\n  \
                     * header-filename: show filenames before the content.\n  \
                     * header-filesize: show file sizes before the content.\n  \
```

tests/integration_tests.rs
```diff
@@ -966,7 +966,7 @@ fn filename_stdin_binary() {
     let vec = vec![0; 1];
     bat_with_config()
         .arg("--decorations=always")
-        .arg("--style=header")
+        .arg("--style=header-filename")
```
Enselic, line 969: In general I would prefer if we kept existing regression tests unchanged, because that is the safest way to make sure we don't have any regressions.

Sometimes it makes sense to change integration tests, but e.g. changing `header` to `header-filename` everywhere is not something we should do IMHO.

So I would prefer if you added new tests for the new style components.

I want to stress again: Please use your good judgement to decide when to side-step that general guideline.

src/bin/bat/clap_app.rs
```diff
@@ -396,11 +396,19 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
                 .validator(|val| {
```
Enselic, line 396: Don't forget to undo the change of commits for the submodules above

src/style.rs
```diff
@@ -11,6 +11,8 @@ pub enum StyleComponent {
     Grid,
```
Enselic, line 11: Add `#[non_exhaustive]` at the top of this enum please as previously discussed

src/printer.rs
```diff
@@ -289,21 +292,11 @@ impl<'a> Printer for InteractivePrinter<'a> {
 
         if self.config.style_components.grid() {
             self.print_horizontal_line(handle, '┬')?;
-
-            write!(
```
Enselic, line 293: I'm a bit short on time, so please accept my apology in advance in asking this question rather than figuring it out for myself:

Can we somehow refactor the code so that we don't need to check for `if self.config.style_components.grid()` twice? That would simplify the code which is always nice.

Eventually I will get time to figure it out for myself, but that can take a while.

src/printer.rs
```diff
@@ -315,17 +308,61 @@ impl<'a> Printer for InteractivePrinter<'a> {
...
+        header_components.iter().try_for_each(|component| {
+            if self.config.style_components.grid() {
+                write!(
+                    handle,
+                    "{}{}",
+                    " ".repeat(self.panel_width),
+                    self.colors
+                        .grid
+                        .paint(if self.panel_width > 0 { "│ " } else { "" }),
+                )?;
```
Enselic, line 339: To make this code easier to understand, can you introduce some helper functions please? I _think_ this block can be moved into a helper function for example.

Enselic (commented):
Here is another round of comments.

Again, big thanks for working on this 👍

src/input.rs
```diff
@@ -130,9 +132,21 @@ impl<'a> Input<'a> {
...
+                let size = meta.len();
+
+                InputMetadata {
+                    size: Some(size),
+                    ..InputMetadata::default()
+                }
+            }
+            Err(_) => InputMetadata::default(),
+        };
+
```
Enselic, line 146: Hmm this comment got lost so I'm making it again. You can simplify this code like this:
```suggestion
        let metadata = InputMetadata {
            size: fs::metadata(path).map(|m| m.len()).ok(),
            ..InputMetadata::default()
        };

```

src/printer.rs
mdibaiee: I tried to do that initially, but didn't find a straightforward way to do that. I can put comments on what each one of those if blocks does and why they exist separately. Basically the first one is only required before the first grid line to insert the cornering grid, or else insert a new-line break between files.

I moved the outside check closer to the loop as well which I think makes reasoning about them easier.

tests/integration_tests.rs
mdibaiee: I tried to avoid these renames as muchas possible and in cases where testing the header was the goal (and not just filename), I copied the tests over and added `header-filename,header-filesize` to test those two together.

Enselic (commented):
Really good job with the integration tests 👍 

I had one lint comment, but otherwise this is starting to look good to me. I am close to setting Approved on this, I just need to find the time for a final look before I set Approved.

src/printer.rs
```diff
@@ -315,17 +314,60 @@ impl<'a> Printer for InteractivePrinter<'a> {
...
+                        .unwrap_or_else(|| "".into()),
+                    self.colors.header_value.paint(description.title()),
+                    mode
+                ),
+
+                StyleComponent::HeaderFilesize => {
+                    let bsize = metadata
+                        .size
+                        .map(|s| format!("{}", ByteSize(s)))
+                        .unwrap_or("".into());
```
sharkdp, line 365: It's quite possible that `metadata.size` is empty, right? For example, if the input comes from STDIN. I think we should also add a test case for this. Maybe we want to skip the `Size: ` line completely in this case? Or at least print something like `Size: -`.

src/input.rs
```diff
@@ -130,9 +132,14 @@ impl<'a> Input<'a> {
 
     fn _ordinary_file(path: &Path) -> Self {
         let kind = InputKind::OrdinaryFile(path.to_path_buf());
+        let metadata = InputMetadata {
+            size: fs::metadata(path).map(|m| m.len()).ok(),
```
sharkdp, line 136: It's probably irrelevant since other things (like actually reading the file) are much slower, but this is an additional (stat) syscall for every file that we print - *even if we do not print the filesize*. It might make sense to run a quick benchmark in a scenario where we print a lot of small files.

If it has a significant performance impact, we could delay loading the metadata until we actually need it.

sharkdp (changes requested):
Thank you very much for your contribution! I added a few comments.

sharkdp:
One additional thing that might be useful in a future version that just comes to my mind: if someone runs `bat some-symlink`, we could display
```
Filename: some-symlink -> symlink-target
```
in the header.

src/input.rs
mdibaiee: I ran a benchmark against 100 small files.

master branch
```
|:---|---:|---:|---:|---:|
| `bat … --language=txt *.txt` | 46.6 ± 2.2 | 43.6 | 55.2 | 1.00 |

```

My branch:
```
|:---|---:|---:|---:|---:|
| `bat … --language=txt *.txt` | 47.3 ± 2.0 | 44.3 | 57.0 | 1.00 |
```

But while running these benchmarks, I noticed a large regression in performance between 0.18.3 and master branch on "Syntax highlighting speed jquery" (wrap=character and wrap=never). From milliseconds to 14s, it's definitely worth investigating:

master branch
```
       │ File: bat-master/syntax-highlighting-speed-jquery.js.md
───────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ | Command | Mean [s] | Min [s] | Max [s] | Relative |
   2   │ |:---|---:|---:|---:|---:|
   3   │ | `bat … jquery.js` | 13.258 ± 0.070 | 13.174 | 13.392 | 1.00 |

```

0.18.3
```
       │ File: bat-system/syntax-highlighting-speed-jquery.js.md
───────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ | Command | Mean [s] | Min [s] | Max [s] | Relative |
   2   │ |:---|---:|---:|---:|---:|
   3   │ | `bat … jquery.js` | 1.052 ± 0.008 | 1.039 | 1.062 | 1.00 |
```

Enselic: I'm not able to reproduce that big regression. Are you sure you used release builds in both cases?

`hyperfine --export-markdown=/dev/stdout -L bat bat-master,bat-v0.18.3 '{bat} --no-config --style=full --color=always --wrap=never --terminal-width=80 tests/benchmarks/highlighting-speed-src/jquery.js'`

|:---|---:|---:|---:|---:|
| `bat-master --no-config --style=full --color=always --wrap=never --terminal-width=80 tests/benchmarks/highlighting-speed-src/jquery.js` | 366.2 ± 1.5 | 364.8 | 369.5 | 1.02 ± 0.00 |
| `bat-v0.18.3 --no-config --style=full --color=always --wrap=never --terminal-width=80 tests/benchmarks/highlighting-speed-src/jquery.js` | 357.9 ± 0.7 | 357.2 | 359.1 | 1.00 |

The difference is even less pronounced with `--wrap=character`.

mdibaiee: @Enselic ah you are right, I was using debug builds instead of release builds. 🤦 😁 

New runs with release builds for many small files:

my branch:
|:---|---:|---:|---:|---:|
| `bat … --language=txt *.txt` | 12.8 ± 1.4 | 10.8 | 20.7 | 1.00 |


1.8.3
|:---|---:|---:|---:|---:|
| `bat … --language=txt *.txt` | 13.1 ± 1.4 | 11.2 | 24.2 | 1.00 |

And I don't see any significant difference between syntax highlighting benchmarks either, so all good 👍

Enselic: Thanks for adding a new performance benchmark!

I've also compared master vs this PR using all existing and the new performance benchmarks, and the difference is negligible (meaning: I don't really see any difference). I also used a low-end desktop for this, so if we had a problem, I should have seen it.

So this is a non-issue in practice as far as I can tell. It was good to make sure that was the case though!

Besides, if this turns out to be a problem later for whatever reason, it should be no big deal to change it to be lazy-loaded. It's an implementation detail that should not require changing public API.

src/style.rs
```diff
@@ -31,14 +34,17 @@ impl StyleComponent {
...
+            StyleComponent::HeaderFilesize => &[StyleComponent::HeaderFilesize],
             StyleComponent::LineNumbers => &[StyleComponent::LineNumbers],
             StyleComponent::Snip => &[StyleComponent::Snip],
             StyleComponent::Full => &[
                 #[cfg(feature = "git")]
                 StyleComponent::Changes,
                 StyleComponent::Grid,
-                StyleComponent::Header,
+                StyleComponent::HeaderFilename,
+                StyleComponent::HeaderFilesize,
```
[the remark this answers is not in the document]
Enselic: This is what we say in `--help` about `full` and `auto`:

```
                     * full: enables all available components (default).\n  \
                     * auto: same as 'full', unless the output is piped.\n  \
```

Let us now ask if we are doing a breaking change with regards to the behavior of `full` and this PR. I would argue that it is to be expected, given the documentation, that the exact set of styles in `full` can change over time. I would expect users that want a very particular style to have specified components individually, and for those users, the output should remain the same as before.

Let us now ask if we are doing a breaking change with regards to the behavior of `auto`. I would say the documentation is still correct. So I don't think there is any breaking change here, given the reasoning about for `full` above.

Besides, if people complain about that `full` changed meaning, the workaround we can tell them to use is easy: specify components explicitly.

I think the alternative, that we may never more add components to `full`, is more harmful than being liberal in interpreting what it means to use `full`.

Enselic (approved):
I had one more nit-pick comment, but still feel comfortable setting Approved on this now.

I think the important bits:
* public API
* backwards compatibility
* regression tests

look good by now. Maybe the implementation could be polished and simplified a bit more, but that can always be done iteratively later.

mdibaiee:
Thanks a lot for your time reviewing the MR @Enselic @sharkdp ❤️

sharkdp (approved):
Thank you very much @mdibaiee.

I'm going to approve this, but I'm not yet sure if I am personally in favor of activating `header-filesize` by default (via `full`/`auto` by default). I mainly use `bat` to show text files on the console. And I am usually not interested in file sizes of text files.

Don't get me wrong. I love that we have this new feature. I just tend to think that it should be an opt-in thing, rather than a new default. Sure, the "downside" is small... but it's one additional line in the output, that can eventually also decide whether or not we need a pager, etc.

mdibaiee:
@sharkdp I don't have a strong opinion on this, I don't mind it not being part of the default either, but I guess it just introduces a new category of behaviour: non-default style components that need to be enabled explicitly. (and hence new code to support this new category)

Enselic:
I created a ticket where we can discuss this further. If we introduce a new style called `default`, we don't need to make `header-filesize` default yet. But we still have the option to make it default later. See #2061.

I hope we can take care of this in a way that does not cause annoying conflicts for you @mdibaiee. The easiest way to avoid conflicts is probably to create a PR that solves #2061 that is based on this PR. And when that PR is ready, we first merge this PR, and then directly after the other PR.

Since we want to have the bat master branch always releasable, it would be wise to wait a bit with merging this, just to play it safe.

sharkdp:
I would be in favor of this (implementing #2061 and keeping `header-filesize` out of `default`). But I'm also okay with being overruled, if everyone else thinks we should `header-filesize` by default.

mdibaiee:
Okay, so should I just wait then?

Enselic:
Since sharkdp has been clear on that he is OK with this being merged, I think we should merge it actually. I think we have enough of a plan on how to remove `header-filesize` from the default style again if there is a backlash from users.

Personally I would have found it useful to have that info in the header by default. E.g. when looking around in rustdoc JSON files that can easily be  7 MB big, it takes several seconds to highlight the entire file. In cases like that, it is useful to be informed of that the file is huge. It makes bat/syntect seem less slow. And when it is not interesting to know the size of a file, it only takes up one line.

Another way to look at it is to see it as a proxy for the number of lines in a file, which can also occasionally be useful to know. When navigating in a new repo, it gives a quick indication of a file is interesting to explore further. We can't really print the actual number of lines in a file, since that would requires us to read the entire file up-front.

But I am also perfectly fine with if we change our mind and remove `header-filesize` by default. But I think we can try to have it as default.

Waiting with making it the default is for sure the most low-risk approach. But many of us write code for bat for fun, and it seems a lot more fun to me to merge this PR than to wait. And crucially, it is not prohibitively fun, but reasonably fun.

@mdibaiee I just realised we forgot to ask you to add a CHANGELOG.md entry for this change as per https://github.com/sharkdp/bat/blob/master/CONTRIBUTING.md. Could you do that please?

mdibaiee:
@Enselic that sounds more fun indeed, thanks!

sharkdp:
I think you accidentally committed changes to submodules in assets/syntaxes/02_Extra

mdibaiee:
@sharkdp fixed that, that always confuses me 😁

Enselic:
Big thanks for your contribution @mdibaiee! And thank you for your patience with us :)

I think we can add more info in CHANGELOG.md, e.g. mentioning that `header-filename` is an alias to `header` now, but I couldn't figure out a good exact wording, so I say we can postpone that. We always clean up CHANGELOG.md right before a release anyway.

IMHO we need to fix #2061 before we can add timestamp and file permission info like you originally had in this PR. But I look forward to any future PR that you will create :)

Now I will try to give some attention to #1985 that has also been sitting around for a while (sorry about that)

Issue #1978, opened 2021-12-09
default `--theme` should be neutral


ngam:
Hi, thank you for the good work --- I like this utility. I found the default color theme unusable (it's [supposed to be](https://github.com/sharkdp/bat#highlighting-theme) for dark mode only) and so I recommend you set a neutral default theme, for example `ansi`.

milespossing:
Alright I've closed the above PR, as some pretty good arguments were made against this change, but I would recommend looking at the [config file documentation](https://github.com/sharkdp/bat#configuration-file), specifically [the formatting section](https://github.com/sharkdp/bat#format)

ngam:
> Alright I've closed the above PR, as some pretty good arguments were made against this change, but I would recommend looking at the [config file documentation](https://github.com/sharkdp/bat#configuration-file), specifically [the formatting section](https://github.com/sharkdp/bat#format)

Hey thanks for looking into this. I am well aware of the formatting and it is relatively easy --- however, you should keep that PR open to gather more opinions. I will offer a counter point shortly.

PR #1997, opened 2021-12-31
Updating default theme

milespossing:
A potential solution to #1978, if desired.

Updating theme to be `ansi` by default. The current default theme (Monokai Extended) does not work well in light mode:

Light vs Dark w/ Monokai:

image

Light vs Dark w/ ansi:

image

- Changing theme to ansi by default
- Updating integration tests
- Updating CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 - Add PowerShell completion, see #1826 (@rashil2000)
 - Minimum supported Rust version (MSRV) bumped to 1.51, see #1994 (@mdibaiee)
 - Include git hash in `bat -V` and `bat --version` output if present. See #1921 (@Enselic)
+- Updated default theme to be `ansi` by default (@milespossing)
 
 ## Syntaxes
 
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -69,7 +69,7 @@ impl HighlightingAssets {
     }
 
     pub fn default_theme() -> &'static str {
-        "Monokai Extended"
+        "ansi"
     }
 
     pub fn from_cache(cache_path: &Path) -> Result<Self> {
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -1254,7 +1254,7 @@ fn ignored_suffix_arg() {
         .arg("test.json~")
         .assert()
         .success()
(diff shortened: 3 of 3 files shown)


Enselic:
Thank you for looking into this in a serious manner. Especially thank you for the before and after screenshots.

Looking at the screenshots, ansi vs Monokai Extended boils down to: Do we want
1. great highlighting for users of default terminal backgrounds (AFAIK dark on all three major OSes), or 
2. mediocre highlighting on both dark and light terminal backgrounds.

Of these, I think 1 is the best choice for the vast majority of users. If someone deviates from the default, by definition they are willing to go the extra mile and tweak their system theming. So it does not seem like too much to ask to also ask such users to change default bat theme.

That said, there are some additional points that can be made:
* Somewhere in our issue tracker I think someone posted sample code that ostensibly allowed detecting terminal theme/background colour in a multi-platform manner. That could be one route to explore further.
* Perhaps it would be worth making it easier to change default theme in bat. Like `bat --write-to-config-default-theme ansi` or something like that.

milespossing:
@Enselic that makes sense; I'm going to close this PR for now.

I think that ultimately the best solution would be the dark/light term detection + something like the `--write-to-config...` option you just mentioned. Maybe I'll hack around the dark/light detection and also look into how local configuration works in this project.

Thanks so much for the response!

ngam:
> 1. AFAIK dark on all three major OSes

Wrong.

ngam:
@milespossing please keep this open.

ngam:
Sure, it is a good idea to have auto detection, if possible, but for now a straightforward solution is having a a theme that works okay for all possible themes. The problem is that currently, whatever the default theme is, it is unusable in light mode as I reported in #1978. You're essentially saying it is okay to have an absolutely unusable theme out-of-the-box. I don't understand how this makes sense at all.

Also, glancing quickly on the screenshots, I am not sure `monokai` is color-blind friendly --- is it? The other option, `ansi`, looks like it is. I will share a screen shot of how monokai looks like on a standard light terminal in MacOS shortly.

ngam:
Screen Shot 2021-12-31 at 18 19 38
Screen Shot 2021-12-31 at 18 20 03

ngam:
Screen Shot 2021-12-31 at 18 22 14
Screen Shot 2021-12-31 at 18 22 35

ngam:
In the readme here, https://github.com/sharkdp/bat#highlighting-theme, you say: 

> `bat` looks good on a dark background by default. However, if your terminal uses a light background, some themes like `GitHub` or `OneHalfLight` will work better for you. You can also use a custom theme by following the 'Adding new themes' section below

But you have a much simpler solution: Set a neutral theme by default! Your equivalency about one being great in dark mode, but the other being mediocre in both is also debatable. I don't know about you, but `ansi` looks great in dark mode too! By the way, fwiw, the `GitHub ` theme you have isn't that good and I wouldn't be advertising it as an option. Your best theme --- by far --- is `ansi`.

ngam:
Screen Shot 2021-12-31 at 19 07 52
Screen Shot 2021-12-31 at 19 08 12

ngam:
Screen Shot 2021-12-31 at 19 10 56
Screen Shot 2021-12-31 at 19 11 16

PR #2010, opened 2022-01-02
most likely a bug/typo

ka7:
most likely this typo is a bug.


Enselic:
Thank you for the contribution! In this case however, the typo comes from upstream. While typos can be annoying, I don't think it is worth deviating from upstream in this case.

I suggest you open a PR against the upstream repo instead. One example of the typo is this file and line: https://github.com/microsoft/TypeScript-Sublime-Plugin/blob/master/TypeScript.tmLanguage#L1634

Does that sound OK to you?

PR #2011, opened 2022-01-02
small typos, should not interrupt anything

ka7:
small fixes.


Enselic:
Since the change is only in comments and the readme it should indeed be close to zero risk to merge this. Thank you for your contribution.

PR #2015, opened 2022-01-04
Highlight nimble files with nim syntax

Mrfiregem:
Nimble is Nim's package manager and the syntax for its package configuration is the same as nimscript (`nims`). This is a patch similar to #1683.


keith-hall:
Hi, thanks for your contribution. Maybe you have a sample nimble file you can add as a regression test, please? :) https://github.com/sharkdp/bat/blob/master/doc/assets.md#syntax-tests

Mrfiregem:
I've added a fairly complete test file that looks to be highlighted correctly.

keith-hall (approved):

sharkdp:

Issue #2026, opened 2022-01-15
Obtain syntect Theme structs from theme name when using bat as a library


dandavison:
Hi bat developers! Delta has, since the beginning, made use of bat's very helpful collection of language syntax and highlighting theme definitions. Until now, that's involved bat code vendored in the delta repo. I'd love to use the bat library for this. However, from a quick look, it appears that while the syntect syntax structs are exposed, the syntect themes are not (only theme name strings are exposed). Would it be possible (and fit with your aims) to provide an API to look up a syntect theme struct by theme name?

Thanks for all the work.

Enselic:
Hi and thanks for the request!

I think the first step here would be to replace the vendored code in the delta repo with a normal bat Cargo dependency and see how far that takes you.

I'm thinking you would leave in placeholders in the code like

```
// Here we would like to use any theme, not just the default
let theme = default_theme();
```

but otherwise get the code to compile and run. Once that is functional, we can more easily prototype what kind of API that is missing.

Does that sound reasonable?

sharkdp:
> Would it be possible (and fit with your aims) to provide an API to look up a syntect theme struct by theme name?

Not that I disagree with the approach suggested by @Enselic, but to give a more direct answer: yes, I think we generally want to provide this functionality to enable `delta` (and others) to use `bat`-as-a-library.

Note that we currently have this notice in the documentation regarding the usage of the "expert" API (everything except for the things that are directly accessible via `bat::Xyz`):

> If you need more control, you can also use the structs in the submodules […], but note that the API of these internal modules is much more likely to change. Some or all of these modules might be removed in the future.

Meaning that the API of (for example) `bat::assets::HighlightingAssets` might not (yet) be well designed for public usage. So in this sense, an approach as described above might be very helpful. Not just for new public-facing functionality, but also for the existing things inside submodules.

dandavison:
Thanks @Enselic & @sharkdp! I've made a proof-of-principle branch in delta: https://github.com/dandavison/delta/pull/903

However, it isn't correct yet and  in local development I'm hitting the `malloc: can't allocate region` errors due -- I believe -- to the recent syntect bug. Would you be able to tell me which syntect version / revision I should be using in delta and bat, and whether I need to change my local workflows? I've run `cargo clean` (and `rustup update`).

dandavison:
> in local development I'm hitting the `malloc: can't allocate region` errors

Please ignore that -- this problem has gone away (I'm not sure why it was occurring). So,

[...]

PR #2030, opened 2022-01-15
Expose syntect theme and syntax sets

dandavison:
Proof-of-principle draft PR allowing delta to replace vendored bat code with bat library usage.

See 

https://github.com/sharkdp/bat/issues/2026
https://github.com/dandavison/delta/pull/903 (companion delta PR)


src/assets.rs
```diff
@@ -15,7 +15,7 @@ use crate::syntax_mapping::ignored_suffixes::IgnoredSuffixes;
 use crate::syntax_mapping::MappingTarget;
 use crate::{bat_warning, SyntaxMapping};
 
-use lazy_theme_set::LazyThemeSet;
+pub use lazy_theme_set::LazyThemeSet;
```
Enselic, line 18: Spontaneously I would like to have a public API that hides wether or not themes are lazy-loaded. So exporting the name `LazyThemeSet` should be avoided if possible.

I'm a bit short on time, so maybe you could experiment a bit and see if that can be achieved?

dandavison:
> exporting the name `LazyThemeSet` should be avoided if possible.

Hi @Enselic, thanks for looking. How about something like 2de7e6e9a5be9fe1b718e3a527f120ce2305b8c3? That adds a new public enum to bat named `ThemeSet` that currently has a single variant only (the `LazyThemeSet`). The new enum implements the same API as `LazyThemeSet`.

src/assets/theme_set.rs
```diff
@@ -0,0 +1,26 @@
...
+impl ThemeSet {
+    /// Look up a theme by name
+    pub fn get(&self, name: &str) -> Option<&syntect::highlighting::Theme> {
+        match self {
+            ThemeSet::LazyThemeSet(lazy_theme_set) => lazy_theme_set.get(name),
+        }
+    }
+
+    /// An iterator over all theme names
+    pub fn themes(&self) -> impl Iterator<Item = &str> {
```
dandavison, line 21: I think perhaps this method would be better named `theme_names`

dandavison:
Since the enum has one variant only currently, I've switched to use a newtype struct in 3d019910e9a2b0696e9ff84be684921ed82689d3. I think that's preferable, but we can go back to the enum in 9c2b7379b550073d7784fb3306e8c3c635a7a24e if not.

dandavison:
Hi @Enselic, if you do get time to look at this again that would be great but no worries there are lots of other issues to attend to currently. It will be nice for delta if it turns out to be possible to go in this direction because I'll be able to delete some vendored code and update highlighting assets simply by bumping the bat library version.

Enselic:
@dandavison Sorry for letting you wait on this. I've had my plate full lately. 

I haven't had time to investigate this myself yet, so maybe I should just go ahead and ask instead:

I don't fully understand why you can't just the existing API based around theme names, i.e. `Config::theme`? Would be great if you could elaborate on that.

dandavison:
Hi @Enselic, no worries!

> I don't fully understand why you can't use the existing API based around theme names, i.e. [Config::theme](https://docs.rs/bat/0.19.0/bat/config/struct.Config.html). Would be great if you could elaborate on that.

Sure, that's easy to explain. My motivation here is to use the bat library in [delta](https://github.com/dandavison/delta). Delta is, like bat, a Rust application that uses syntect for highlighting. So I need the bat library to give me the syntect theme structs themselves, not just the names.

The connection between delta and bat is:

1. They both use syntect
2. Delta uses bat's collection of binary syntect highlighting assets (and also copies bat's pager-spawning code).

The purpose of this PR is to alter the bat library so that a client such as delta can use the bat library for its collection of binary highlighting assets.

Enselic:
Thanks for the explanation! That makes sense. If that is the case, isn't it enough to simply change

```
pub(crate) fn get_theme(&self, theme: &str) -> &Theme {
```

to

```
pub fn get_theme(&self, theme: &str) -> &Theme {
```

?

That would enable you to get the syntect `Theme` for a given theme name.

dandavison:
Yes, thank you @Enselic! You're right -- I think because delta had always used syntect's `ThemeSet`, I had overlooked the possibility of using bat's `get_theme` and was making this change more complicated than it needed to be. Updated this PR and the companion delta PR.

sharkdp:
I'm in favor of this (now minimal) change. Especially since we warn in the documentation that *the API of these internal modules is much more likely to change. Some or all of these modules might be removed in the future.*

Enselic:
I am also fine with this change as long as the API is only considered semi-stable, but I agree the disclaimer in the docs makes that clear.

I think it would be good to repeat that disclaimer in the docs of the individual changed methods just to make sure people see them though. But it is not a super strong wish.

@dandavison Can you also add an entry in CHANGELOG.md in the "bat as a library" section for this please? More info in https://github.com/sharkdp/bat/blob/master/CONTRIBUTING.md. Thanks!

Apart from the above I consider this PR Approved 🎉 

As a side-note, I have played around with the thought of having a `bat` crate and a `bat-impl` crate, similarly to the `thiserror` crate that depends on a `thiserror-impl` crate.

The `bat` crate would contain the `PrettyPrinter` stable API and the `bat` binary, and `PrettyPrinter` and the `bat` binary would depend on `bat-impl`. The API this PR adds would live in `bat-impl`. So `delta` would depend on `bat-impl` and not `bat`. Just an idea. Not sure if it is a good idea. Certainly out of scope for this PR.

dandavison:
Thanks @sharkdp and @Enselic, I've add the Changelog entry.

> As a side-note, I have played around with the thought of having a `bat` crate and a `bat-impl` crate,

What do you think about having a crate dedicated specifically to the collection of highlighting assets and the wrapping API for working with them? It's a very useful collection. (Are there syntect apps other than delta using them?) As a side effect I'm guessing you might appreciate not having all those submodules in your main repo.

sharkdp:
> What do you think about having a crate dedicated specifically to the collection of highlighting assets and the wrapping API for working with them?

This has been requested a few times, and I agree that this would certainly be very useful. I have written down my thoughts on this here: #919

sharkdp (approved):

Enselic:
@dandavison We are currently experimenting with time based releases, and if everything goes as planned, the next release will be in April. Just to give you a sense of when the API will be available to delta.

Enselic:
@dandavison @michaelblyons Heads-up that we will make a bat v0.19.1 release with this new API in the coming days.

We want to keep projects that depend on bat happy. Waiting until April to make a release seems way too long in this case. :)

dandavison:
Excellent, thanks a lot @Enselic. I'll make a delta release shortly afterwards.

Enselic:
Released in https://crates.io/crates/bat/0.20.0 🎉

PR #2038, opened 2022-01-20
Correctly render tab stops in `--show-all` 

Synthetica9:
New behaviour:

```
───────┬────────────────────────────────────────────────────────────
       │ STDIN
───────┼────────────────────────────────────────────────────────────
   1   │ ··├┤␊
   2   │ ·├─┤␊
   3   │ ├──┤␊
   4   │ 12345678␊
───────┴────────────────────────────────────────────────────────────
```

Old behaviour

```
───────┬────────────────────────────────────────────────────────────
       │ STDIN
───────┼────────────────────────────────────────────────────────────
   1   │ ··├──┤␊
   2   │ ·├──┤␊
   3   │ ├──┤␊
   4   │ 12345678␊
───────┴────────────────────────────────────────────────────────────
```

Does not work with multi-width characters, yet, is this a requirement? If so, how should this work with unicode escaped characters?


Enselic:
Nice 👍 I skimmed the code and it looks reasonable. Before a detailed review, can you please add at least one test (preferably more as long as it makes sense) for this in https://github.com/sharkdp/bat/blob/master/tests/integration_tests.rs? I.e. invoke the binary and assert on stdout.

That way we can be sure that this contribution and use case will never stop working.

Synthetica9:
Okay, added two integration tests. Not sure about the names, feel free to suggest better ones.

sharkdp:
Very nice! I like it.  

Could you please add an entry to the CHANGELOG? See instructions here: https://github.com/sharkdp/bat/blob/master/CONTRIBUTING.md#add-an-entry-to-the-changelog

tests/integration_tests.rs
```diff
@@ -1153,6 +1153,52 @@ fn show_all_mode() {
...
+55555├─┤?␊
+666666├┤?␊
+7777777↹?␊
+88888888├──┤?␊
+",
+        );
+}
+
+#[test]
+fn show_all_tabstops_long() {
```
sharkdp, line 1180: Feel free to choose very long test names (I have to admit that some other test names could be improved as well). Maybe:

`show_all_extends_tab_markers_to_next_tabstop`

Issue #1709, opened 2021-06-28
Add way to fail silently when no specific syntax has been found


scop:
Having ability to fail silently with no output and a non-zero exit status would be useful when configuring `bat` as an [input preprocessor](https://manpages.debian.org/buster/less/less.1.en.html#INPUT_PREPROCESSOR) for `less`. This would accomplish falling back to other preprocessors in case `bat` doesn't have a specific output available for the given file, such as ones displaying metadata for images, archives, etc.

I implemented this to `chroma` some time ago, not sure how easily something like this would fit to `bat` but based on a quick look, it might not be entirely trivial:
- https://github.com/alecthomas/chroma/issues/501
- https://github.com/alecthomas/chroma/pull/502
- https://github.com/alecthomas/chroma/commit/e87e5b2208ed79c5e371a7fb86c5e0185cbc6888#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

sharkdp:
Thank you for your feature request. I like this idea.

How would you design the CLI for this? Add a new `--fail-if-syntax-unsupported` flag (or similar name)?

Slightly related? #1597

PR #2041, opened 2022-01-23
Failing silently if syntax is unsupported

willful759:
This is my first commit to a open source project, I decided to tackle this specific issue: #1709 

Added the flag `fail-if-unsupported`, allows `bat` to fail without output and with a non-zero status code, so that it can be used as an input preprocessor

I'm new to rust and to contributing to open source so please give any feedback you have

### DIFF
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -1312,3 +1312,14 @@ fn acknowledgements() {
         )
         .stderr("");
 }
+
+// Regression test for https://github.com/sharkdp/bat/issues/1709
+#[test]
+fn fail_if_unsupported_supported() {
+    bat()
+        .arg("--fail-if-unsupported")
+        .arg("test.txt")
+        .assert()
+        .success()
+        .stdout("hello world\n");
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## Features
 
-- Flag `--fail-if-unsupported` added, allows `bat` to fail silently with a non-zero exit code, see #1709
+- Flag `--fail-if-unsupported` added, allows `bat` to fail silently with a non-zero exit code if
+  no specific syntax highlighting was found for the file, see #1709
 
 ## Bugfixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Features
 
+- Flag `--fail-if-unsupported` added, allows `bat` to fail silently with a non-zero exit code, see #1709
+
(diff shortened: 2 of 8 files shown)


keith-hall (commented):
Thanks for your contribution! Looks good to me :)

The only feedback I have is maybe to add some [regression tests]( https://github.com/sharkdp/bat/blob/master/tests/integration_tests.rs) if possible. (And to think about how we want it to work if there are multiple input files and some have a syntax mapping and some don't.)

It may also be nice to make the changelog entry a bit more descriptive about what "unsupported" means - basically just like the help entry.
Something we always forget is to update the [`man` page](https://github.com/sharkdp/bat/blob/master/assets/manual/bat.1.in) when adding/updating command line args - it would be great if you could add it there too please.

keith-hall:
Also, I would say, don't be afraid to force-push to tidy up the commit history (and fix typos) - at least before anyone has reviewed your changes :) Don't worry about it now, though - I think it may make sense to squash merge, but we'll see what the other maintainers say :)

willful759:
Thanks for the feedback! iḿ also quite inexperienced with git, I've only used personally, but I will look forward into your suggestions

In the case we have multiple input files and only some of them don't have specific syntax we can use, I would personally just let it fail with a different error code, but I would really appreciate community suggestions

tests/integration_tests.rs
```diff
@@ -1312,3 +1312,14 @@ fn acknowledgements() {
         )
         .stderr("");
 }
+
+// Regression test for https://github.com/sharkdp/bat/issues/1709
+#[test]
+fn fail_if_unsupported_supported() {
```
Enselic, line 1318: Thank you for taking the time to make a contribution!

Before a detailed review, it will be very useful to have a broad set of regression tests in place. The test you added is only for the positive case. We would need tests also for the negative case, i.e. when the exit code is not 0.

I am also a bit skeptical about the name `--fail-if-unsupported` because it does not make much sense in and of itself. Since the use case specifically is usage as a less input preprocessor, I think we can simply call the flag `--as-input-processor-for-less`. I think it is perfectly fine to have long names for "experimental" feature. If it turns out that the flag can by used more broadly, we can also broaden the flag to e.g. just `--as-input-preprocessor`.

It is better to start narrow and iteratively broaden, than to start broad and iteratively narrow. Because the latter is hard to do in a backwards compatible way.

Enselic:
Closing due to inactivity to keep he PR inbox clean, but of course feel free to re-open if you resume work on this!

mnp:
Came here looking for this feature and was sad it didn't get completed.

PR #2067, opened 2022-02-09
git global config - lookup $XDG_CONFIG_HOME faithfully

cyqsimon:
Related issue: #1191 

When #1191 was fixed, the implementation was incomplete. More specifically, it did not cover the case when `$XDG_CONFIG_HOME` is not set or empty, as specified in [`git-config` documentation](https://git-scm.com/docs/git-config#FILES).

This PR fixes that.


src/syntax_mapping.rs
```diff
@@ -122,9 +122,18 @@ impl<'a> SyntaxMapping<'a> {
...
-        if let Some(xdg_config_home) = std::env::var_os("XDG_CONFIG_HOME") {
-            let git_config_path = Path::new(&xdg_config_home).join("git");
-
+        // global git config files
+        // `$XDG_CONFIG_HOME/git`, or `$HOME/.config/git` if `$XDG_CONFIG_HOME` is not set or empty
+        // see https://git-scm.com/docs/git-config#FILES
+        if let Some(git_config_path) = std::env::var_os("XDG_CONFIG_HOME")
+            .and_then(|val| (!val.is_empty()).then(|| val))
+            .map(PathBuf::from)
+            .or_else(|| std::env::var_os("HOME").map(|home| Path::new(&home).join(".config")))
```
Enselic, line 131: Thanks you for taking the time to open a PR!

I think we want to highlight `$HOME/.config/git/config` regardless of what `XDG_CONFIG_HOME` is set to though, right? I think it is perfectly reasonable to assume that the former is a git file even if the latter is set.

So it seems to me like you should move the code inside the if to a helper function and then apply it to both the possible paths to `.../git/config`.

cyqsimon: Yep, sounds reasonable. Will make that happen in a bit.

src/syntax_mapping.rs
```diff
@@ -122,29 +122,44 @@ impl<'a> SyntaxMapping<'a> {
...
+        fn insert_git_config_global(mapping: &mut SyntaxMapping, config_home: impl AsRef<Path>) {
+            let git_config_path = config_home.as_ref().join("git");
 
             mapping
                 .insert(
                     &git_config_path.join("config").to_string_lossy(),
                     MappingTarget::MapTo("Git Config"),
                 )
-                .ok();
+                .unwrap();
```
Enselic, line 148: Why change to unwrap here?

Enselic:
Would be great if you could add a couple of integration tests for this change, see tests/integration_test.rs

src/syntax_mapping.rs
cyqsimon: I believe `Result::ok` here is semantically inaccurate. If I understand correctly, `SyntaxMapping::insert` is infallible in this specific context, and therefore `unwrap` (or `expect`) should be used. If `SyntaxMapping::insert` is fallible here then I stand corrected.

In the original PR (#1195), @sharkdp stated "`XDG_CONFIG_HOME` could contain invalid UTF8" as the reason for removing `unwrap`, but that applies to `git_config_path.join("config").into_os_string().to_str().unwrap()`, not to this `insert` call. So I suspect they might have made this inaccuracy in a rush.

cyqsimon:
I wrote a test, but I'm not quite sure why it's failing. Probably some funky stuff going on with the ANSI escape codes. Help needed.

tests/integration_tests.rs
```diff
@@ -1117,7 +1117,9 @@ Single Line
 ────────────────────────────────────────────────────────────────────────────────
 ",
         )
-        .stderr("\x1b[33m[bat warning]\x1b[0m: Style 'rule' is a subset of style 'grid', 'rule' will not be visible.\n");
+        .stderr(
+            "\x1b[33m[bat warning]\x1b[0m: Style 'rule' is a subset of style 'grid', 'rule' will not be visible.\n",
+        );
```
cyqsimon, line 1122: rustfmt did this. Don't ask me why 🤷

src/syntax_mapping.rs
Enselic: I appreciate you wanting get the code semantically accurate. I tried to convince myself of that insert() can never fail, but it was not obvious to me. While it wasn't obvious that it can fail either, I would prefer if we could play it safe and keep `ok()` for now.

Enselic (approved):
Big thanks for adding tests. Automated tests are so great to have. With automated tests in place I consider this Approved (assuming you handle remaining comments). We could argue about if we want to define functions in functions, but I don't think it matters much with automated tests in place. Because then we can easily change that later.

It would be great if you could add an entry in the CHANGELOG.md before we merge. See https://github.com/sharkdp/bat/blob/master/CONTRIBUTING.md

src/syntax_mapping.rs
cyqsimon: Personally I prefer to fail fast if `insert` fails here, hence `unwrap`. But I'm happy to respect your preference if you insist.

cyqsimon:
I did a rebase against the newest `master` branch to resolve merge conflicts of `changelog.md`.

src/syntax_mapping.rs
Enselic: Failing fast is good for developers, but not for users. Users will prefer to be shown the file they requested bat to show, albeit without syntax highlighting, over bat crashing on them. So I would say I insist on reverting back to `.ok()`.

cyqsimon: Sure no problem.

src/syntax_mapping.rs
```diff
@@ -122,8 +122,23 @@ impl<'a> SyntaxMapping<'a> {
             .insert("*.hook", MappingTarget::MapTo("INI"))
             .unwrap();
 
-        if let Some(xdg_config_home) = std::env::var_os("XDG_CONFIG_HOME") {
-            let git_config_path = Path::new(&xdg_config_home).join("git");
+        // global git config files
+        // `$XDG_CONFIG_HOME/git`, or `$HOME/.config/git` if `$XDG_CONFIG_HOME` is not set or empty
+        // see https://git-scm.com/docs/git-config#FILES
+        // we cover both cases regardless of the state of `$XDG_CONFIG_HOME`
```
Enselic, line 128:
```suggestion
        // Global git config files rooted in `$XDG_CONFIG_HOME/git/` or `$HOME/.config/git/`
        // See e.g. https://git-scm.com/docs/git-config#FILES
```

src/syntax_mapping.rs
```diff
@@ -122,8 +122,23 @@ impl<'a> SyntaxMapping<'a> {
...
+        if let Some(xdg_config_home) =
+            std::env::var_os("XDG_CONFIG_HOME").and_then(|val| (!val.is_empty()).then(|| val))
+        {
+            insert_git_config_global(&mut mapping, &xdg_config_home);
+        }
+        if let Some(default_config_home) =
+            std::env::var_os("HOME").map(|home| Path::new(&home).join(".config"))
```
Enselic, line 135: Shouldn't you check `!val.is_empty()` also here? Maybe you can move that check into the helper function?

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 - Associate `poetry.lock` files with the `TOML` syntax. See #2049
 - Associate `.mesh`, `.task`, `.rgen`, `.rint`, `.rahit`, `.rchit`, `.rmiss`, and `.rcall` with the `GLSL` syntax. See #2050
 - Added support for `JQ` syntax, see #2072
+- `Git`: Associate global config files in all valid locations. See #2067 (@cyqsimon)
```
Enselic, line 19:
```suggestion
- Properly associate global git config files rooted in `$XDG_CONFIG_HOME/git/` or `$HOME/.config/git/`. See #2067 (@cyqsimon)
```

src/syntax_mapping.rs
cyqsimon: I mean, if `$HOME == ""` I think we can safely call the environment broken. But eh, why not 🤷.

I kept the check outside of the helper function because it's rather inconvenient to make it accept an `OsStr`. Also I realised `Option::filter` exists and therefore is best.

Enselic (approved):

sharkdp (approved):
Thank you very much

PR #2080, opened 2022-02-18
changed `MANPAGER` in `README.md`

ronkup12:
changed MANPAGER to behave more like the default man.
- always use pager, even for small man pages
- use less with colors, in order to remove man's output when exiting

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,7 +194,7 @@ bat main.cpp | xclip
 `MANPAGER` environment variable:
 
 ```bash
-export MANPAGER="sh -c 'col -bx | bat -l man -p'"
+export MANPAGER="sh -c 'col -bx | bat -l man -p --paging=always --color=always | less -r'"
 man 2 select
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -194,7 +194,7 @@ bat main.cpp | xclip
 `MANPAGER` environment variable:
 
 ```bash
-export MANPAGER="sh -c 'col -bx | bat -l man -p'"
+export MANPAGER="sh -c 'col -bx | bat -l man -p --paging=always --color=always | less -r'"
```
SuperSandro2000, line 197: Do we need both `--paging=always` and `less -r`?

ronkup12: you are right, it does seems to work without `--paging=always`. 
Although, I do wonder if both of them are necessary when the set pager isn't less...

sharkdp:
> use less with colors, in order to remove man's output when exiting

could you please go into a bit more detail?

sharkdp:
I'm closing this due to inactivity. Please feel free to comment in case it should be re-opened.

PR #2083, opened 2022-02-19
Add requirements.txt syntax

scop:
requirementstxt.sublime-syntax converted from `.tmLanguage`, patched to add `file_extensions`, and additional version specifiers from https://www.python.org/dev/peps/pep-0440/#version-specifiers

61k downloads at the moment at https://packagecontrol.io/packages/requirementstxt

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -239,3 +239,6 @@
 [submodule "assets/syntaxes/02_Extra/SublimeJQ"]
 	path = assets/syntaxes/02_Extra/SublimeJQ
 	url = https://github.com/zogwarg/SublimeJQ.git
+[submodule "assets/syntaxes/02_Extra/requirementstxt"]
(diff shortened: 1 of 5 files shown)


Enselic:
See https://github.com/sharkdp/bat/pull/1960 🙂

scop:
Sigh :|

Enselic:
I feel your pain. It's no fun to put effort into making a good PR only be told that it can't be merged :(

sharkdp:
The `requirements.txt` file format seems simple enough. Maybe we can write our own syntax definition from scratch? We've done so for some other formats as well.

scop:
That would be cool, I can help review/try out if someone submits one. References:
- https://pip.pypa.io/en/stable/reference/requirements-file-format/
- https://www.python.org/dev/peps/pep-0440/#version-specifiers

PR #2088, opened 2022-02-25
release-checklist.md: Remove '## Dependencies'

Enselic:
I would like us to discuss removing the section in the release checklist that encourages (by merely being in the checklist) bumping of dependencies.

Bumping dependencies right before a release might not be the best strategy, because

* Bumping deps comes with a risk of introducing regressions (that are not caught by our regression tests). Letting bumps sit in git master for a few weeks or so increases the chances of that kind of regressions being found before we make a release.

* Dependabot regularly bumps dependencies anyway (which is also pointed out in the removed text)

* Ideally, it should be pleasant and easy to make a bat release. The fewer steps, the easier. Even if we mark steps as "optional", it is easy to get a bad conscious if you skip the steps, and thus even optional steps have some amount of "slow down".

### DIFF
--- a/doc/release-checklist.md
+++ b/doc/release-checklist.md
@@ -1,16 +1,5 @@
 # Release checklist
 
-## Dependencies
-
-See this page for a good overview: https://deps.rs/repo/github/sharkdp/bat
-
-- [ ] Optional: update dependencies with `cargo update`. This is also done by
-      dependabot, so it is not strictly necessary.
-- [ ] Install [cargo-outdated](https://crates.io/crates/cargo-outdated). Check
-      for outdated dependencies with `cargo outdated --root-deps-only` and
-      decide for each of them whether we want to (manually) upgrade. This will
-      require changes to `Cargo.toml`.
-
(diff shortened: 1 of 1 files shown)


sharkdp:
I'm okay with removing it. I agree that it doesn't make a lot of sense with dependabot and regular releases. In the past, I wanted to avoid publishing a new version that depended on outdated packages because it sometimes led to conflicts for library users. Not sure if that was with bat or another project. But even with cargo supporting multiple versions of the same dependency, there are situations like
```
user_project -> bat -> other_dependency (v1.0, outdated)
user_project -> other_dependency (v2.0)
```
where a `user_project` depends on a dependency of `bat` directly. This is still fine, until a object/type from version 1.0 is passed into a function of version 2.0 (or vice versa).

sharkdp (approved):

Enselic:
Thanks for a lot for being willing to try this out. I really appreciate that.

I'm thinking that if suddenly get a problem with conflicts for library users, we fix by making another release 💯

Issue #2085, opened 2022-02-23
Automatically rebuild custom assets when necessary


Enselic:
Currently, if a user makes use of custom assets and then installs a new 0.x.0 version of bat, bat will stop working. That is not an ideal user experience. It would be good if we could make this use case smoother, e.g. by automatically doing the equivalent of running `bat cache --build` for the user.

# Step-by-step
Here is a step by step against the git repo, but regular users will get this problem too via regular releases from us and e.g. package manager updates of the bat app.

1.  `cargo run -- cache --build --blank --source assets`
2. Bump 0.x.0 version in `Cargo.toml`, e.g. from 0.19.0 to 0.20.0
3. `cargo run -- examples/simple.rs`

# Expected result
bat prints `examples/simple.rs`

# Actual result
```
[bat error]: The binary caches for the user-customized syntaxes and themes in '/Users/martin/.cache/bat' are not compatible with this version of bat (0.20.0). To solve this, either rebuild the cache (bat cache --build) or remove the custom syntaxes/themes (bat cache --clear).
For more information, see:

  https://github.com/sharkdp/bat#adding-new-syntaxes--language-definitions
```

sharkdp:
I did consider this when first implementing the above semver-incompatibility detection and error message. My only concern is that we can not know if the *binary* user assets and the *source* user assets are still "compatible". This might be a contrived case, but what I was thinking about is the following scenario:

- user puts custom assets in `$(bat --config-dir)/syntax`
- user runs `bat cache --build`
- user *somehow* modifies files in `$(bat --config-dir)/syntax` (e.g.: removes source files)
- new version of `bat` comes along that simply rebuilds the binary assets and changes the user behavior.

I therefore opted for the explicit version rather than the implicit/automatic one.

But I'm happy to reconsider this.

awerlang:
>     * user _somehow_ modifies files in `$(bat --config-dir)/syntax` (e.g.: removes source files, or adds some source files but then decides to not run `bat cache --build`)
> 
>     * new version of `bat` comes along that simply rebuilds the binary assets from source and changes the user behavior.

What if bat manages its caches transparently. Users should only have control over the plain configuration files. User behavior is governed by its config files.

Enselic:
You mean bat should check the timestamp of each custom theme and custom syntax, and then compare that with the timestamp of the built custom assets (a.k.a. "the cache"), and if any asset is newer than the cache, then the cache is automatically rebuilt?

[...]

PR #2107, opened 2022-03-03
Include contents of custom assets `metadata.yaml` in `--diagnostics`

Enselic:
# Output without custom assets:

#### Custom assets metadata

Could not read contents of '/Users/martin/.cache/bat/metadata.yaml': No such file or directory (os error 2).

# Output with custom assets:

#### Custom assets metadata

```
---
bat_version: 0.20.0
creation_time:
  secs_since_epoch: 1646283165
  nanos_since_epoch: 208676000
```

# Remark

When we have access to https://github.com/sharkdp/bugreport/pull/11 we will be able to get even more info for custom assets, but the metadata is a significant giver of valuable information by itself that we can add already now to `bat`, so let's do that.

This will help slightly with https://github.com/sharkdp/bat/issues/2085

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ## Other
 
+- Include contents of custom assets `metadata.yaml` in `--diagnostics`. See #2107 (@Enselic)
+
 ## Syntaxes
 
 ## Themes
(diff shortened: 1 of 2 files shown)


sharkdp (approved):
Thanks!

PR #2119, opened 2022-03-11
Add a --style=default option

### LINKED ISSUE #2061: Introduce new `--style` called `default` that is not necessarily same as `full`
There is a discussion in #1988 whether or not to include `header-filesize` in the default `--style`. The safest way forward is to be conservative and not include `header-filesize` by default. At least not yet. We can maybe make it default later.

To achieve that, we need a new style. I propose we simply call it `default`.

So we would have:
```
                     * default: enables recommended components (can change across versions) (default).\n  \
                     * full: enables all available components.\n  \
                     * auto: same as 'default', unless the output is piped.\n  \
                     * [... same as before ...]
```

Having `(can change across versions)` in the description is not very elegant. For now I just put it there for clarity. I think the implementation will be quite easy if we can agree on e.g. a name for the new style. We don't necessarily need to call it `default`.

It would be good to also take the opportunity to get rid of `auto` while we are at it. See #1743.

IsaacHorvath:
Hello, this is my first contribution, meant to resolve #2061. I added a new --style=default option that is effectively a carbon copy of --style=full, only I've taken out the file size in the header. I expect there to be more differences going forward, e.g. after implementing some of #1701.

I added a simple integration test that seems to pass okay, and updated the shell completions (though I only use bash so I haven't tested them).

I am brand spanking new to Rust, so apologies if I'm way out of line here. I'm hoping to try my hand at contributing a bit here and there to bat help practice.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Features
 
 - Correctly render tab stops in --show-all, see #2038 (@Synthetica9)
+- Add a --style=default option, less verbose than =full, see #2061 (@IsaacHorvath)
 
 ## Bugfixes
 
--- a/src/bin/bat/clap_app.rs
+++ b/src/bin/bat/clap_app.rs
@@ -398,6 +398,7 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
                         !&[
                             "auto",
                             "full",
+                            "default",
                             "plain",
                             "header",
                             "header-filename",
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -975,6 +975,30 @@ fn header_full_binary() {
 }
 
 #[test]
+fn header_default() {
+    bat()
+        .arg("--paging=never")
+        .arg("--color=never")
+        .arg("--terminal-width=80")
+        .arg("--wrap=never")
+        .arg("--decorations=always")
+        .arg("--style=default")
+        .arg("single-line.txt")
+        .assert()
+        .success()
+        .stdout(
+            "\
+───────┬────────────────────────────────────────────────────────────────────────
+       │ File: single-line.txt
+───────┼────────────────────────────────────────────────────────────────────────
+   1   │ Single Line
+───────┴────────────────────────────────────────────────────────────────────────
+",
+        )
+        .stderr("");
+}
+
+#[test]
 fn filename_stdin() {
     bat()
         .arg("--decorations=always")
--- a/assets/completions/_bat.ps1.in
+++ b/assets/completions/_bat.ps1.in
@@ -37,7 +37,7 @@ Register-ArgumentCompleter -Native -CommandName '{{PROJECT_EXECUTABLE}}' -Script
             [CompletionResult]::new('-m', 'm', [CompletionResultType]::ParameterName, 'Use the specified syntax for files matching the glob pattern (''*.cpp:C++'').')
             [CompletionResult]::new('--map-syntax', 'map-syntax', [CompletionResultType]::ParameterName, 'Use the specified syntax for files matching the glob pattern (''*.cpp:C++'').')
             [CompletionResult]::new('--theme', 'theme', [CompletionResultType]::ParameterName, 'Set the color theme for syntax highlighting.')
-            [CompletionResult]::new('--style', 'style', [CompletionResultType]::ParameterName, 'Comma-separated list of style elements to display (*auto*, full, plain, changes, header, header-filename, header-filesize, grid, rule, numbers, snip).')
+            [CompletionResult]::new('--style', 'style', [CompletionResultType]::ParameterName, 'Comma-separated list of style elements to display (*default*, auto, full, plain, changes, header, header-filename, header-filesize, grid, rule, numbers, snip).')
             [CompletionResult]::new('-r', 'r', [CompletionResultType]::ParameterName, 'Only print the lines from N to M.')
             [CompletionResult]::new('--line-range', 'line-range', [CompletionResultType]::ParameterName, 'Only print the lines from N to M.')
             [CompletionResult]::new('-A', 'A', [CompletionResultType]::ParameterName, 'Show non-printable characters (space, tab, newline, ..).')
--- a/assets/completions/bat.fish.in
+++ b/assets/completions/bat.fish.in
@@ -56,7 +56,7 @@ complete -c {{PROJECT_EXECUTABLE}} -s P -d "Disable paging. Alias for '--paging=
 
 complete -c {{PROJECT_EXECUTABLE}} -s A -l show-all -d "Show non-printable characters like space/tab/newline" -n "not __fish_seen_subcommand_from cache"
 
-complete -c {{PROJECT_EXECUTABLE}} -l style -xka "auto full plain changes header header-filename header-filesize grid rule numbers snip" -d "Comma-separated list of style elements or presets to display with file contents" -n "not __fish_seen_subcommand_from cache"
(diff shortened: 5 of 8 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Features
 
 - Correctly render tab stops in --show-all, see #2038 (@Synthetica9)
+- Add a --style=default option, less verbose than =full, see #2061 (@IsaacHorvath)
```
Enselic, line 6:
```suggestion
- Add a `--style=default` option, less verbose than `full`, see #2061 (@IsaacHorvath)
```

Enselic (commented):
Thanks a lot for working on this! I took a look and couldn't find anything worth commenting on (except some nitpick in CHANGELOG). This looks great on first sight. Especially thanks for adding an integration test. Those are super great to have.

Before I set Approved I think I would like to find the time to play around a bit with the implementation too, but code-wise I see no problems so far. So I think I will be able to set this as Approved within max a week or so.

src/bin/bat/clap_app.rs
```diff
@@ -398,6 +398,7 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
                         !&[
                             "auto",
                             "full",
+                            "default",
```
Enselic, line 401: Oh I see one thing that is missing now; updating the --help text in src/bin/bat/clap_app.rs line 425-ish

IsaacHorvath: I've added those, and also header-filename and header-filesize in the short help description.

IsaacHorvath:
Thanks for your feedback! I polished it up a bit, and I have a little more courage to snoop around the issues list now.

Enselic:
Sorry for the delay, but I got some other things I need to take care of, and I got less spare time than I anticipated, so it might take some time still until I can take a second look.

Enselic:
Shouldn't the file size be omitted by default with this PR? It does not seem to be the case when I try:

```
/Users/martin/src/bat
1166d6c Update help text for '--style'  2119  
% cargo run -- --no-config examples/simple.rs 
    Finished dev [unoptimized + debuginfo] target(s) in 0.20s
     Running `target/debug/bat --no-config examples/simple.rs`
───────┬───────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: examples/simple.rs
       │ Size: 177 B
───────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ /// A simple program that prints its own source code using the bat library
   2   │ use bat::PrettyPrinter;
   3   │ 
   4   │ fn main() {
   5   │     PrettyPrinter::new().input_file(file!()).print().unwrap();
   6   │ }
───────┴───────────────────────────────────────────────────────────────────────────────────────────────────────────
```

Again, big thanks for working on this, and sorry for not taking a second look sooner, but my last two weeks has been unusually busy.

src/bin/bat/clap_app.rs
```diff
@@ -430,7 +431,8 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
                      '--style=\"..\"' option to the configuration file or export the \
                      BAT_STYLE environment variable (e.g.: export BAT_STYLE=\"..\").\n\n\
                      Possible values:\n\n  \
-                     * full: enables all available components (default).\n  \
+                     * default: enables basic style components (default).\n  \
```
Enselic, line 434: This is just a proposal, so I am mostly curious to hear what you think:

Maybe we should use the word "recommended" instead of "basic"?

```suggestion
                     * default: enables recommended style components (default).\n  \
```

tests/integration_tests.rs
```diff
@@ -974,6 +974,30 @@ fn header_full_binary() {
...
 
+#[test]
+fn header_default() {
+    bat()
+        .arg("--paging=never")
+        .arg("--color=never")
+        .arg("--terminal-width=80")
+        .arg("--wrap=never")
+        .arg("--decorations=always")
+        .arg("--style=default")
```
Enselic, line 985: Any chance you can add a second test that is a regression test for https://github.com/sharkdp/bat/pull/2119/commits/1b475fb02e92e38ecf4fb53f13bedfa100d7a92e ?

I'm thinking you would simply copy-paste this test but leave out the explicit `.arg("--style=default")`. If you get problems with the default not being right when running tests you don't need to bother with adding a test for that. But if it is an easy test to add, it would be nice to have.

Enselic (approved):
This now seems to work as it should:

```
% cargo run -- --no-config examples/simple.rs       
   Compiling bat v0.20.0 (/Users/martin/src/bat)
    Finished dev [unoptimized + debuginfo] target(s) in 3.45s
     Running `target/debug/bat --no-config examples/simple.rs`
───────┬───────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: examples/simple.rs
───────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ /// A simple program that prints its own source code using the bat library
   2   │ use bat::PrettyPrinter;
   3   │ 
   4   │ fn main() {
   5   │     PrettyPrinter::new().input_file(file!()).print().unwrap();
   6   │ }
───────┴───────────────────────────────────────────────────────────────────────────────────────────────────────────
% cargo run -- --no-config examples/simple.rs --style=full
    Finished dev [unoptimized + debuginfo] target(s) in 0.30s
     Running `target/debug/bat --no-config examples/simple.rs --style=full`
───────┬───────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: examples/simple.rs
       │ Size: 177 B
───────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ /// A simple program that prints its own source code using the bat library
   2   │ use bat::PrettyPrinter;
   3   │ 
   4   │ fn main() {
   5   │     PrettyPrinter::new().input_file(file!()).print().unwrap();
   6   │ }
───────┴───────────────────────────────────────────────────────────────────────────────────────────────────────────
% cargo run -- --no-config examples/simple.rs --style=numbers,grid
    Finished dev [unoptimized + debuginfo] target(s) in 0.20s
     Running `target/debug/bat --no-config examples/simple.rs --style=numbers,grid`
─────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1 │ /// A simple program that prints its own source code using the bat library
   2 │ use bat::PrettyPrinter;
   3 │ 
   4 │ fn main() {
   5 │     PrettyPrinter::new().input_file(file!()).print().unwrap();
   6 │ }
─────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

I left some additional comments that would be good to respond to, but apart from that I consider this Approved now. Good job and thanks again!

src/bin/bat/clap_app.rs
IsaacHorvath: Sounds good, I remember that didn't quite feel right when I threw it in.

Enselic (approved):
Approved without reservation. Thank you for the updates.

sharkdp:
Thank you for your contribution. I think this part of  help text needs updating:
```
 * auto: same as 'full', unless the output is piped.
```
I guess this is now: "same as 'default', unless …".

Also, it would be great if the man page (template) in `assets/manual/bat.1.in` could be updated accordingly.

IsaacHorvath:
I did already update the --style section of the manpage with:  
`Possible values: *default*, full, auto, plain, changes, header, header-filename, header-filesize, grid, rule, numbers, snip.`  
Is there something I missed?

I'll add the help text change now.

Enselic:
Looks to me like @sharkdp's comments have been taken care of. Let's merge this now. Thanks again.

PR #2124, opened 2022-03-16
Add Catppuccin theme

WitherCubes:
[Catppuccin theme](https://github.com/catppuccin/catppuccin)

image

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -239,3 +239,6 @@
 [submodule "assets/syntaxes/02_Extra/SublimeJQ"]
 	path = assets/syntaxes/02_Extra/SublimeJQ
 	url = https://github.com/zogwarg/SublimeJQ.git
+[submodule "assets/themes/Catppuccin"]
+	path = assets/themes/Catppuccin
+	url = https://github.com/catppuccin/sublime-text
--- a/tests/assets.rs
+++ b/tests/assets.rs
@@ -31,6 +31,7 @@ fn all_themes_are_present() {
             "Sublime Snazzy",
             "TwoDark",
             "Visual Studio Dark+",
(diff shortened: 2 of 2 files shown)


Enselic:
I can't find this theme on https://packagecontrol.io/. Is it not there?

latipun7:
> I can't find this theme on https://packagecontrol.io/. Is it not there?

It's not there yet, I guess. Since the upstream installation guide still manual: https://github.com/catppuccin/sublime-text

WitherCubes:
> I can't find this theme on https://packagecontrol.io/. Is it not there?

Is it necessarily needed to be packaged?

Enselic:
Normally we require added syntaxes to have 10k downloads to be added. I was mostly curious to know how many downloads this theme had. I've been thinking about this PR some more, and here is how I personally see it.

There are lots of interesting and beautiful themes out there. But we can't add them all to bat, because then it becomes too hard for users to select a theme from the default set of themes.

So I think we should not add this theme to the default set of themes. Instead, we should ensure that it is easy to use custom themes with bat. For example by fixing https://github.com/sharkdp/bat/issues/2085.

WitherCubes:
If it was easy to add custom themes to bat without the hassle of rebuilding the cache, I would be :)

WitherCubes:
> Would you mind if I tested an idea I have on you? Would you like to be able to install themes in the proper place by simply doing `bat --install-theme ~/Downloads/some-theme.tmTheme`? That would simply copy the theme to `$(bat --config-dir)/themes`.

I think that would work for me.

Enselic:
Thank you for the input.

Let's go ahead and close this PR now.

PR #2144, opened 2022-04-02
Print info about custom assets in --diagnostics

Enselic:
It will look like this:

```
#### Custom assets

- metadata.yaml, 100 bytes
- syntaxes.bin, 726649 bytes
- themes.bin, 40475 bytes
```

or this:

```
#### Custom assets

'/Users/martin/.cache/bat' is empty
```

or this:

```
#### Custom assets

'/Users/martin/.cache/bat' not found
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ## Other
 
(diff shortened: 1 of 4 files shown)


sharkdp (approved):
Thank you!

Issue #2125, opened 2022-03-18
CVE-2022-24713


jubalh:
SUSE Security has requested an update to address [CVE-2022-24713](https://blog.rust-lang.org/2022/03/08/cve-2022-24713.html). Upstream [commit](https://github.com/rust-lang/regex/commit/ae70b41d4f46641dbc45c7a4f87954aea356283e).
Please take the time to adjust Cargo.lock and upgrade at least regex.

Thanks.

Enselic:
Closed by #2139

PR #2145, opened 2022-04-02
Lock regex crate at min 1.5.5 for CVE-2022-24713

Enselic:
Cargo.lock already specifies 1.5.5, but we should also do it in Cargo.toml.

I forgot that in #2139.

For #2125

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ## Bugfixes
 
+- Bump `regex` dependency from 1.5.4 to 1.5.5 to fix [CVE-2022-24713](https://blog.rust-lang.org/2022/03/08/cve-2022-24713.html), see #2145, #2139 (@Enselic)
+
 ## Other
 
 - Include contents of custom assets `metadata.yaml` in `--diagnostics`. See #2107 (@Enselic)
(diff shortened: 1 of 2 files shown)


sharkdp:
I don't know if bat is really "vulnerable" to this CVE, but anyway: to resolve that completely, would we also have to make sure that we do not *transitively* depend on that older regex version?

It looks good
```
▶ cargo tree | rg regex
│   ├── regex v1.5.5
│   │   └── regex-syntax v0.6.25
│   │   └── regex-automata v0.1.10
│   └── regex v1.5.5 (*)
│   ├── regex v1.5.5 (*)
├── regex v1.5.5 (*)
│   ├── regex-syntax v0.6.25
│   │   └── regex v1.5.5 (*)
```
but I wonder if that is guaranteed by having 1.5.5 specified in the lock file?

sharkdp (approved):

Enselic:
After skimming through https://doc.rust-lang.org/cargo/reference/resolver.html it seems as if different versions of the same crate is only allowed if the versions are within different semver domains. So if some dependency would have depended on say 1.3.0 then the resolver will still make sure that at least 1.5.5 is used for both, since they are in the same semver compatibility domain.

But if some crate transitively depended on 0.2.0 of regex, then we would build against both 0.2.0 and 1.5.5.

Anyway, thanks for checking with `cargo tree`, it seems like we are good to go.

I'm not sure either if we are vulnerable, but it seems easier to me to just do the bump than to dig deeper :)

sharkdp:
> I'm not sure either if we are vulnerable, but it seems easier to me to just do the bump than to dig deeper :)

Agreed

Issue #2130, opened 2022-03-29
Interest in adding a command --help syntax?


victor-gp:
**Syntax:**

As mentioned in #2115, I'm making [a syntax to scope command help messages](https://github.com/victor-gp/cmd-help-sublime-syntax). You can find some examples [here](https://github.com/victor-gp/cmd-help-sublime-syntax/tree/demo/docs/examples).

**Guideline Criteria:**

I don't plan on publishing it in packagecontrol.io because I can't think of an use case for this within Sublime. Considering this is made with bat in mind, I wanted to know if you'd consider including it.

It would be a nice alternative to `man.sublime-syntax`, the [currently recommended](https://github.com/sharkdp/bat/issues/1430) syntax to use with help messages.

`cmd-help` is more selective about what it scopes (only definitions, not descriptions) and it also covers some syntax patterns that `man` doesn't need to. Help messages are way more free-form than man pages.

It already gets many commands just right, scoping most things as intended. More importantly, its rate of false positives is negligible (on [a sample of 40+ help messages](https://github.com/victor-gp/cmd-help-sublime-syntax/tree/main/tests/highlighted)).

I'm still actively working on this, but it's definitely ready for prime time. Can I PR this syntax?

keith-hall:
> I don't plan on publishing it in packagecontrol.io because I can't think of an use case for this within Sublime. Considering this is made with bat in mind, I wanted to know if you'd consider including it.

I personally vote yes :) it feels like a good fit to me - nice work :)

Enselic:
Sounds like a good idea to me, and I can see you've already made good progress. I also vote yes. I would probably start using it myself 😇

victor-gp:
Thanks for your nice words ^^

@Enselic: help yourself! (Pun intended.) Please do report anything that doesn't quite work so I can fix it. Also, if I plan to make major changes in the future (like re-scoping things), can I @mention you there for feedback? 

@keith-hall: other than this, I have 2-3 ideas to improve Manpage.sublime-syntax. Nothing big (definitely not multi-scoping! XD), just tweaks on how the syntax targets some man elements and also what I did on cmd-help to avoid scoping normal text as headings (only useful for help messages, I think, but shouldn't break manpages). Maybe we can discuss them in another channel, or I just open another issue?

keith-hall:
As the Manpage syntax definition is part of this repo, I'm fine with you just opening another issue (or PR) to discuss improvements :)

victor-gp:
[...]

PR #2148, opened 2022-04-06
Syntax: add cmd-help

victor-gp:
As discussed in #2130.

I tried to add a syntax test but it's not highlighted because this is a syntax without auto-detection and the syntax tests scripts don't run `bat --language=X`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,9 +214,9 @@ The [`prettybat`](https://github.com/eth-p/bat-extras/blob/master/doc/prettybat.
 
 #### Highlighting `--help` messages
 
-You can use bat to colorize help text: `$ cp --help | bat -plhelp`
+You can use `bat` to colorize help text: `$ cp --help | bat -plhelp`
 
-But you may want to use a wrapper around this:
+You can also use a wrapper around this:
 
 ```bash
 # in your .bashrc/.zshrc/*rc
@@ -227,6 +227,8 @@ help() {
 ```
 
 Then you can do `$ help cp` or `$ help git commit`.
+
+Please report any issues with the help syntax in [this repository](https://github.com/victor-gp/cmd-help-sublime-syntax).
 
 
 ## Installation


sharkdp:
Thank you for this contribution!

You should be able to use this: https://github.com/victor-gp/cmd-help-sublime-syntax/blob/2cf2ecc0e183dcc5989cd184ef9d9cbed3e4dfb5/syntaxes/cmd-help.sublime-syntax#L4-L5, right?

Otherwise, you can also place a `bat_options` file in the `source` folder (see e.g. the Plaintext syntax). This could probably just contain `--language=cmd-help`.

By the way, for usage with the `-l`/`--language`, it would be nice to have a short name for this syntax I suppose? Like `help` instead of `cmd-help` maybe? This would allow using this like
```
git --help | bat -lhelp
```

Or do you have other ideas how this would be used? With a shell function a la
```sh
bat_help() {
  cmd="$1"
  $cmd --help | bat --language=cmd-help
}
```

victor-gp:
You are both right. I had overlooked that syntaxes are applied according to file extension, and cmd-help already supports that. :sweat_smile:

I added a test based on `bat -h`, which hits most syntax elements. I also included `help` as a file extension related to the syntax, and bumped the submodule here.

As for usage, I suggest in the readme to use [this](https://github.com/victor-gp/cmd-help-sublime-syntax/blob/main/docs/bathelp.sh), which covers most UX concerns (ask if you want me to elaborate). Surely it's not a very visible place, much less to the average bat user, so I'm open to suggestions with regards to this.

sharkdp:
> As for usage, I suggest in the readme to use [this](https://github.com/victor-gp/cmd-help-sublime-syntax/blob/main/docs/bathelp.sh), which covers most UX concerns (ask if you want me to elaborate). Surely it's not a very visible place, much less to the average bat user, so I'm open to suggestions with regards to this.

I think we could add a new "*Colorization of `--help` texts*" subsection here: https://github.com/sharkdp/bat#integration-with-other-tools (or similar title). I would first mention how to do it without any shell function and alias, and add the two commands that you have in your README.

victor-gp:
Done, see the latest commit. Open to edits :D

README.md
```diff
@@ -212,6 +212,22 @@ Also, note that this will [not work](https://github.com/sharkdp/bat/issues/1145)
 
 The [`prettybat`](https://github.com/eth-p/bat-extras/blob/master/doc/prettybat.md) script is a wrapper that will format code and print it with `bat`.
 
+#### Highlighting `--help` messages
+
+You can use bat to colorize help text: `$ cp --help | bat -plhelp`
```
keith-hall, line 217: In other places in the readme, we put `bat` in backticks.
```suggestion
You can use `bat` to colorize help text: `$ cp --help | bat -plhelp`
```

victor-gp: My bad! Fixed.

I also included a line to (try to) channel issues my way.

keith-hall (approved):

Enselic (approved):
I took a quick look and it looks OK to me too. Maybe we should be a bit more terse in the README, because I think our README currently is a bit too long, but we can do that later I think.

Since @sharkdp expressed support for this in the issue itself, I think we can merge this now.

Enselic:
If you have an idea for how to reduce the size of the readme AND make it easier to read as a whole, then: yes, I think you should do that :)

Issue #2152, opened 2022-04-08
colorize lines in logs differently depending on importance


miguelbarao:
Thanks for the work you put into `bat`.
I was considering using `tail -f` and `bat` to colorize the log of a running program.
In many logs there are keywords indicating the importance of a line. For example `INFO`, `WARNING`, `ERROR`, `CRITICAL`, `DEBUG`.
It would be useful to have the lines colored differently so that errors could be easily spotted.
Would you be interested in supporting this feature?

keith-hall:
Thanks for the feedback. We had a short discussion on this at https://github.com/sharkdp/bat/issues/1605#issuecomment-813601215, but it should be possible to cater for both styles - log level word vs full line - with some color scheme customization if we tweak the log syntax to support it.

MahdiNazemi:
Vim has a very nice syntax highlighting plugin with the following rules:
[https://github.com/MTDL9/vim-log-highlighting/blob/master/syntax/log.vim](https://github.com/MTDL9/vim-log-highlighting/blob/master/syntax/log.vim)

Is it possible to integrate similar rules in bat?

keith-hall:
You can customize/patch the themes to give a color to scopes like [`markup.error.log`](https://github.com/sharkdp/bat/blob/b13c4d5f8de932a7593271f74b7b3815a1e9ee26/assets/syntaxes/02_Extra/log.sublime-syntax#L58C5-L65C29). Something similar was done for various Manpage constructs in https://github.com/sharkdp/bat/pull/2994/files.
To use custom themes, see the section in the readme: https://github.com/sharkdp/bat?tab=readme-ov-file#adding-new-themes

PR #2153, opened 2022-04-09
Syntax: [Log] scope whole line based on log level

keith-hall:
Update the log syntax definition to allow color schemes to color entire lines differently based on log level.
Addresses #2152

### DIFF
--- a/assets/syntaxes/02_Extra/log.sublime-syntax
+++ b/assets/syntaxes/02_Extra/log.sublime-syntax
@@ -7,8 +7,15 @@ scope: text.log
 variables:
   ipv4_part: (?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)
   hours_minutes_seconds: (?:[01]\d|2[0-3]):(?:[0-5]\d):(?:[0-5]\d)
+  error: \b(?i:fail(?:ure|ed)?|error|exception|fatal|critical)\b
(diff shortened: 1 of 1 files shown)


keith-hall:
I was debating whether to put details of how to customise the color scheme in the changelog or readme, but seems too specific to belong in either place (its a shame Sublime doesn't support color scheme settings - "highlight_whole_log_line" would be a neat way to tackle this), so I'll likely drop a comment on the issue where it was requested. Not sure if worth mentioning this change in the changelog at all (as it won't affect anything without customization, unless we decide to change the scopes - I'm considering `markup.error.line` for example to cover the whole line instead of something like `meta.log-level.error`, and the `fatal` text could be `markup.error.word` or something, and color schemes can differentiate based on that)?

Enselic:
Nice! I fully trust your judgement on whether or not to mention this in CHANGELOG.md. Wouldn't it be good to also add a regression test though?

sharkdp:
How do we proceed here? Merge without an additional test?

keith-hall:
I think there's no way to add a regression test without updating one of the color schemes - this PR *enables* color schemes to color entire lines differently based on log level, but has no effect unless the color scheme targets the new scopes. So I'm thinking to merge without an additional test.

sharkdp:
Thanks

PR #2156, opened 2022-04-12
Add mapping for Unix mail spool

cyqsimon:
`/var/spool/mail` and `/var/mail` are often symlinked, but not necessarily so (https://serverfault.com/a/351783/549727). So I mapped both.


Enselic:
Thank you. Looks reasonable to me. Can you add an entry in CHANGELOG.md please?

Setting up syntax mapping has a measurable impact on startup performance once #951 is fixed, but I don't think we have yet reached a point where it is prohibitively expensive to add more.

cyqsimon:
Done.

Enselic (approved):

sharkdp (approved):

Issue #2143, opened 2022-04-02
Failure to use less from BusyBox


hungpham3112:
**Describe the bug you encountered:**
Step to reproduce:
1. scoop install busybox, bat
2. bat .\yourfile.

Here is the result in my machine:

![image](https://user-images.githubusercontent.com/75968004/161364026-bae24516-5efa-4650-985c-4b284c02cace.png)

**What did you expect to happen instead?**

No error

**How did you install `bat`?**

scoop install bat

---

**bat version and environment**

version: bat 0.20.0 (0655ecf)
OS: Window 11 lastest

Enselic:
Thank you for the taking the time to report a bug. But there is not much the bat project can do about this. Please try to reach out to the scoop packagers of bat.

hungpham3112:
> Thank you for the taking the time to report a bug. But there is not much the bat project can do about this. Please try to reach out to the scoop packagers of bat.

I don't know why but without busybox, it runs normal.

Enselic:
Probably busybox packages a bin called bat for some reason. In any case, let's close this particular issue in this tracker.

keith-hall:
Try with `--paging=never` or setting `BAT_PAGER` to override the command line args passed to `less` - it looks likely that busybox includes some dodgy version of `less`

Enselic:
Great save @keith-hall, that is most likely what is going on here 👍

Enselic:
Actually, we already do quite extensive version checks on less when bat starts. Perhaps we could try to catch this case too and print a warning or something. I suggest we re-open this. Might also be suitable for a new contributor could take a look at.

nfisher1226:
I'd like to take this one on since nobody else has claimed it.

keith-hall:
@nfisher1226 sounds great, thanks. Please let us know if you need any pointers :)

nfisher1226:
So looking at busybox less as it gets built with the default config, there's no color support, but it can handle bat's other formatting decorations (in somewhat reduced capacity). Example:
```
───────┬──────────────────────────────────────────────────────────────────────��
�─
       │ File: applets/applets.c
       │ Size: 350 B
───────┼──────────────────────────────────────────────────────────────────────��
�─
   1   │ /* vi: set sw=4 ts=4: */
   2   │ /*
   3   │  * Stub for linking busybox binary against libbusybox.
   4   │  *
   5   │  * Copyright (C) 2007 Denys Vlasenko <vda.linux@googlemail.com>
[...]

PR #2162, opened 2022-04-22
Add support for BusyBox less as pager

nfisher1226:
See issue [#2143](https://github.com/sharkdp/bat/issues/2143)

This is an initial working draft. I haven't updated the CHANGELOG, but will before merge if accepted.


keith-hall:
Looks good to me, but my Rust code reviews are lackluster compared to those of our other maintainers, so let's see what they say :)

I think to fix CI after unmarking the PR as a draft, you just need to execute `tests/syntax-tests/update.sh` and commit the file which will change :)

src/less.rs
```diff
@@ -3,21 +3,32 @@
...
+    } else {
+        parse_less_version(&cmd.stderr)
+    }
 }
 
-fn parse_less_version(output: &[u8]) -> Option<usize> {
-    if !output.starts_with(b"less ") {
-        return None;
+fn parse_less_version(output: &[u8]) -> Option<LessVersion> {
+    match std::str::from_utf8(output) {
```
Enselic, line 24: It would be good if we could continue being cautious about encoding, and check if this is less or BusyBox before assuming the output is UTF-8. Otherwise we are more or less certain to crash in scenarios where people use odd pagers (for whatever reason) that does not output UTF-8.

Enselic, line 24: It would be good if we could continue being cautious about encoding, and check if this is less or BusyBox before assuming the output is UTF-8. Otherwise we are more or less certain to crash in scenarios where people use odd pagers (for whatever reason) that does not output UTF-8.

src/less.rs
```diff
@@ -3,21 +3,32 @@
...
+            Some(LessVersion::Less(version[5..end].parse::<usize>().ok()?))
+        }
+        Ok(version) if version.contains("BusyBox ") => Some(LessVersion::BusyBox),
+        _ => None,
     }
-
-    let version = std::str::from_utf8(&output[5..]).ok()?;
-    let end = version.find(|c: char| !c.is_ascii_digit())?;
-    version[..end].parse::<usize>().ok()
 }
```
Enselic, line 32: Bonus points if you move this out to a helper function so that the diff is minimized. Something like this:

```rust
fn parse_vanilla_less_version(output: &[u8]) -> Option<usize> { 
    let version = std::str::from_utf8(&output[5..]).ok()?;
    let end = version.find(|c: char| !c.is_ascii_digit())?;
    version[..end].parse::<usize>().ok()
}

File: src/output.rs [hunk 3]
```diff
@@ -83,31 +83,39 @@ impl OutputType {
...
                 //
                 // See: http://www.greenwoodsoftware.com/less/news.530.html
                 //
                 // For newer versions (530 or 558 on Windows), we omit '--no-init' as it
                 // is not needed anymore.
-                match retrieve_less_version(&pager.bin) {
+                let less_version = retrieve_less_version(&pager.bin);
+                match less_version {
                     None => {
                         p.arg("--no-init");
```
Enselic, line 96: Do we want to pass this to BusyBox too? Isn't it better to not touch arguments at all if BuxyBox is detected? Right now the diff makes it a bit hard to quickly see if we behave the same when `less` is detected. Don't be afraid of introducing helper functions to improve structure and readability.

Enselic (commented):
Thanks a lot for working on this! I added some comments. For the record, this PR does not change the public API of bat-as-a-library, so we don't need to be super picky about naming of things:

```
% pr=2162; git fetch -f upstream pull/$pr/head:$pr && git checkout $pr
% cargo install cargo-public-api
% cargo public-api --diff-git-checkouts origin/master 2162                
(none)

(none)

Added items to the public API
=============================
(none)
```

Also please add an entry to CHANGELOG.md 🙏

src/less.rs
Enselic: Wait, looks like you do handle non-UTF-8 output. My bad. Great that you added a test case already. Maybe also add a test case with like 10 bytes of invalid UTF-8 to make sure we cover that case too? I.e. a test with `assert_eq!(None, parse_less_version(output));`.

nfisher1226: I can do that, sounds like a good idea.

src/less.rs
nfisher1226: Or how about put the original function back with the same name and the BusyBox specific code into a separate function? Since BusyBox doesn't accept long options it has to be detected from the stderr output anyway, which was the reason for the if / else block in `retrieve_less_version`. I was going for as concise as possible, but I can see that's probably not a great habit of mine.

```Rust
pub fn retrieve_less_version(less_path: &dyn AsRef<OsStr>) -> Option<LessVersion> {
    let resolved_path = grep_cli::resolve_binary(less_path.as_ref()).ok()?;

    let cmd = Command::new(resolved_path).arg("--version").output().ok()?;
    if cmd.status.success() {
        parse_less_version(&cmd.stdout)  // this function goes back to it's original form
    } else {
        parse_less_busybox(&cmd.stderr)  // diverge here to a new function
    }
}

fn parse_less_busybox(output: &[u8]) -> Option<LessVersion> {
```

src/output.rs
nfisher1226: Ok so there were two ways to go with this. BusBox doesn't support colored output, but it can filter out the escape sequences using the `-R` flag. So one can either turn off the color formatting in bat, or just pass a flag to have BusyBox ignore it. My thinking was that the latter was the less invasive to bat's codebase, but it's admittedly not going to be as efficient when using bat with busybox less (although probably not noticably so). My personal preference is for simpler code here, since BusyBox less being the only pager is admittedly an edge case.

As an aside, BusyBox doesn't support long options at all, so all of the other flags break it, which is what caused the bug report in the first place. But giving it a second look, it might also be handy to pass `-F` so that it quits if the output fits on one screen, and there is also an option to suppress the tilde characters it normally displays after the end of the output.

Anyway, that was my reasoning. I'll put more time in if the maintainers think it should be done differently. Or I can just add to the comments above this block to explain why we're passing only `-R` to BusyBox.

src/less.rs
Enselic: The smaller diff you can make, the better, because that is easier to review, plus reduces risk of surprising regressions. So keeping `parse_less_version()` unchanged sounds good to me. I think you should call the new function `parse_busybox_less_version()` instead of `parse_less_busybox()` though since "parse less busybox" doesn't fully make semantic sense on its own.

src/output.rs
```diff
@@ -83,31 +83,39 @@ impl OutputType {
             let replace_arguments_to_less = pager.source == PagerSource::EnvVarPager;
 
             if args.is_empty() || replace_arguments_to_less {
-                p.arg("--RAW-CONTROL-CHARS");
```
Enselic, line 86: If busybox less understands `-R` but not `--RAW-CONTROL-CHARS`, it would IMO make sense to change

```
p.arg("--RAW-CONTROL-CHARS");
```

into 

```
p.arg("-R"); // Short version of --RAW-CONTROL-CHARS to maximize compatibility
```

Same for `--quit-if-one-screen` and `-F`.

Does busybox have a short variant of `--chop-long-lines`? If so that would be great, because then we could avoid moving this block of code down, and instead reuse the logic for both buxybox less and less proper. The old code is like bomb waiting to explode with regressions, so it would be great if we could try to stick to "obviously safe" changes, with a minimal "structural" diff.

tests/syntax-tests/highlighted/Rust/output.rs
```diff
@@ -93,11 +93,14 @@
 [38;2;248;248;242m                            [0m[3;38;2;166;226;46mNone[0m[38;2;248;248;242m [0m[38;2;249;38;114m=>[0m[38;2;248;248;242m [0m[38;2;248;248;242m{[0m
 [38;2;248;248;242m                                p[0m[38;2;248;248;242m.[0m[38;2;102;217;239marg[0m[38;2;248;248;242m([0m[38;2;230;219;116m"[0m[38;2;230;219;116m--no-init[0m[38;2;230;219;116m"[0m[38;2;248;248;242m)[0m[38;2;248;248;242m;[0m
 [38;2;248;248;242m                            [0m[38;2;248;248;242m}[0m
-[38;2;248;248;242m                            [0m[3;38;2;166;226;46mSome[0m[38;2;248;248;242m([0m[38;2;248;248;242mversion[0m[38;2;248;248;242m)[0m
```
Enselic, line 96: IMHO it is better to leave the test files (both source and highlighted) unchanged, because that is the safest way to ensure there are no regressions. Plus, it is unnecessary effort to keep the test source file in sync with the "real" source file. The only purpose of the test file is to make sure highlight keeps working.

src/output.rs
Enselic: It's a bit hard to evaluate beforehand what approach would work best, so if it is OK with you, we maybe can iterate a bit on this? I added another comment with a proposal on how we can maybe reuse code for both less and busybox.

The guiding light should be:
* minimize diff and only make "obviously correct" changes so that we don't accidentally introduce a regression with how we modify regular less arguments. I consider changing from "less -RF" to "less -FR" as unnecessary risk, to give a concrete example (that does not apply to your change, but I think you get the point).

src/output.rs
nfisher1226: > Does busybox have a short variant of --chop-long-lines?
Yes, and I just checked that it is the same short option as regular less. So that's three options that could be passed the same way to both. Definitely good points.

src/less.rs
nfisher1226: Sounds like a good plan. I'm definitely seeing a theme in your comments, but it's a sensible one. I'll get this iterated in the next few days when I have time.

src/less.rs
```diff
@@ -78,3 +95,28 @@ fn test_parse_less_version_wrong_program() {
 
     assert_eq!(None, parse_less_version(output));
 }
+
```
Enselic, line 98: Can you add this test too please? After making the code compile:

```
#[test]
fn test_parse_less_version_invalid_utf_8() {
    let output = b"\xff";

    assert_eq!(None, parse_less_version(output));
    assert_eq!(None, parse_less_version_busybox(output));
}
```

Also please add an assert for `parse_less_version_busybox()` to `test_parse_less_version_wrong_program()`.

src/output.rs
```diff
@@ -83,13 +83,15 @@ impl OutputType {
...
+                if args.is_empty() || replace_arguments_to_less {
```
Enselic, line 86: Looks like you accidentally duplicated this if condition?

src/output.rs
```diff
@@ -83,13 +83,15 @@ impl OutputType {
...
+                    p.arg("-R");
```
Enselic, line 87: Let's save others the trouble of looking up what `-R` means by changing to this:

```suggestion
                    p.arg("-R"); // Short version of --RAW-CONTROL-CHARS for maximum compatibility
```

Same for the other short options.

Enselic (commented):
The diff was much smaller and easier to review now. Thanks!  I have 3 more comments.

src/output.rs
nfisher1226: Wow, that's embarrassing. Not sure how I missed it.

Enselic (approved):
This comes across to me as a low-risk diff now that should solve the issue. Thanks a lot for the help.

nfisher1226:
Many thanks for working with me on this. It was a good learning experience.

keith-hall (approved):

Enselic:
Let's go ahead and merge this now since both me and Keith has Approved

Issue #2164, opened 2022-04-23
bat stalls when it encounters extremely long strings (bug / request?)


Adamkadaban:
I have a bash script with a ~120000 character b64 string. When bat encounters it, it stalls for a about 3 minutes. It takes just a second if outputting something closer to ~4000 characters.

Default install of bat. Installed using `dpkg -i bat*.deb`

#### Software version

bat 0.20.0 (0655ecf)

#### Operating system

Linux 5.16.0-kali3-amd64

#### Command-line

```bash
/bin/batcat linpeas.sh --paging=never --diagnostic 
```

#### Environment variables

```bash
SHELL=/bin/bash
PAGER=<not set>
LESS=<not set>
LANG=en_US.UTF-8
LC_ALL=<not set>
BAT_PAGER=<not set>
BAT_CACHE_PATH=<not set>
BAT_CONFIG_PATH=<not set>
BAT_OPTS=<not set>
BAT_STYLE=<not set>
BAT_TABS=<not set>
BAT_THEME=<not set>
XDG_CONFIG_HOME=<not set>
XDG_CACHE_HOME=<not set>
COLORTERM=truecolor
NO_COLOR=<not set>
MANPAGER=<not set>
```

#### Config file

Could not read contents of '/root/.config/bat/config': No such file or directory (os error 2).

#### Compile time information

- Profile: release
- Target triple: x86_64-unknown-linux-gnu
- Family: unix
- OS: linux
- Architecture: x86_64
- Pointer width: 64
- Endian: little
- CPU features: fxsr,sse,sse2
- Host: x86_64-unknown-linux-gnu

#### Less version

```
> less --version 
less 590 (GNU regular expressions)
Copyright (C) 1984-2021  Mark Nudelman

less comes with NO WARRANTY, to the extent permitted by law.
For information about the terms of redistribution,
see the file named README in the less distribution.
Home page: https://greenwoodsoftware.com/less
```

keith-hall:
Thanks for reporting. I believe we are tracking this under https://github.com/sharkdp/bat/issues/1658

Enselic:
Duplicate of #1658

Issue #1751, opened 2021-07-25
bat release focus: performance


sharkdp:
For the next major release of `bat`, we decided to focus on **performance**.

This issue serves as a forum for discussion and as a summary of tasks that can be worked on:

- [x] Get #1596 by @eth-p merged. You can help by reviewing the code.
- [x] Work on #951 
  - [x] Get #1747 by @Enselic merged
  - [x] Fix #1753
  - [x] Discuss / review the prototype suggested by @Enselic in [this comment](https://github.com/sharkdp/bat/issues/951#issuecomment-882098532)
- [ ] Work on `bat` loop-through speed, see #304
- [ ] Set up some tooling to improve `bat` benchmarking. See `tests/benchmarks` for a start.

Other performance-related tickets:

- [x] #1658
- [x] #1630
- [ ] #1147
- [x] #1688

federicotorrielli:
Any updates on bat using the splice?

sharkdp:
> Any updates on bat using the splice?

I'm assuming you are referring to https://endler.dev/2018/fastcat. It would have been nice to at least put a reference. If that's indeed your question, there is a ticket for that already: https://github.com/sharkdp/bat/issues/1112 (and https://github.com/sharkdp/bat/issues/304).

federicotorrielli:
Yes. I was referring to a future splice call implementation. Since those issue did not really have news, issue [#194](https://github.com/sharkdp/bat/issues/194) was negative (but old) and the person who created #1112 did not make any comments on the issue #304 (that did not mention splice, btw), I was here to ask if we are seeing any new update on this or not. Thanks anyway!

Enselic:
Not sure this issue is worth keeping open any longer? I'll go ahead and close and unpin it.

We could pin the remaining issues individually, but I'm not sure they are important enough to fix to justify getting such a prominent position.

For example #2085 would by my estimation provide more value to users to fix, so it would be higher up in my list of issues that deserves pinning.

PR #2165, opened 2022-04-23
Skip syntax highlighting for long lines

keith-hall:
If a line is longer than 16384 characters, don't run it through the `syntect` parser to syntax highlight it - just output it using the theme's plain text coloring / coloring left over from the previous line. Should help solve slowdowns on long lines like #2164 as part of #1751


Enselic (approved):
Awesome to fix this 👍 
I especially like your integration test 🤩 

Maybe the we could improve the implementation, like introducing a helper function and try to reduce code duplication a bit?

But I don't think that needs to be done in this PR, since the current code is much better than no code. Since you had the good taste of adding a test it is easy to tweak the implementation later if needed.

tests/integration_tests.rs
```diff
@@ -1407,6 +1407,23 @@ fn ignored_suffix_arg() {
...
...dn_url":"https://developer.mozilla.org/docs/Web/API/AbstractRange","spec_url":"https://dom.spec.whatwg.org/#interface-abstractrange","support":{"chrome":{"version_added":"90"},"chrome_android":{"version_added":"90"},"edge":[{"version_added":"90"},{"version_added":"18","version_removed":"79"}],"firefox":{"version_added":"69"},"firefox_android":{"version_added":false},"ie":{"version_added":false},"opera":{"version_added":false},"opera_android":{"version_added":false},"safari":{"version_added":"14.1"},"safari_ios":{"version_added":"14.5"},"samsunginternet_android":{"version_added":false},"webview_android":{"version_added":"90"}},"status":{"experimental":false,"standard_track":true,"deprecated":false}},"collapsed":{"__compat":{"mdn_url":"https://developer.mozilla.org/docs/Web/API/AbstractRange/collapsed","spec_url":"https://dom.spec.whatwg.org/#ref-for-dom-range-collapsed①","support":{"chrome":{"version_added":"90"},"chrome_android":{"version_added":"90"},"edge":[{"version_added":"90"},{"version_added":"18","version_removed":"79"}],"firefox":{"version_added":"69"},"firefox_android":{"version_added":false},"ie":{"version_added":false},"opera":{"version_added":false},"opera_android":"# +
```
sharkdp, line 1414: We could also move this to a file and simply `include_str` it here. But I'm also fine with what we have. Adding this test incidentally makes this Rust file also a candidate for "skip syntax highlighting", I guess.

sharkdp (approved):
Nice, thank you!

dns2utf8:
Nice, this is good enough to view huge files. Looking forward to the next release :)

PR #2168, opened 2022-04-26
Run `cargo clippy --fix --all-targets --all-features` with Rust 1.60

Enselic:
Clippy in the newly released Rust 1.60 found some new lints.
Conveniently, all of them were fixable with `--fix`.

By fixing these lints it becomes easier for us and others to see when
new lints are introduced.


Enselic:
I'll spare you the overhead of reviewing this and just go ahead and merge this PR which I categorize as "obvoiusly correct" :)

Issue #1640, opened 2021-05-11
README.md - batdiff function


suchAdemon:
Hello, short question about the section [git-diff](https://github.com/sharkdp/bat#git-diff) from the readme.md file.

In there you descript how a function for showing a diff in git repos could be shown through bat.
The command you have used in your documentation is  `git diff --name-only --diff-filter=d | xargs bat --diff`

Is it on purpose that this diff function should only be used then from the root path of the repository?

Because if you are not inside of the root directory, git will still return the full path to the changed file, which means a false path would be handed over to bat:
```bash
$ cd ./templates

$ git diff --name-only --diff-filter=d | xargs bat --diff
[bat error]: 'templates/app.ini.erb': No such file or directory (os error 2)

$ git diff --name-only --diff-filter=d
templates/app.ini
```

To get that working the parameter `--relative` would be needed to get it working:
`git diff --name-only --diff-filter=d  --relative | xargs bat --diff`

sharkdp:
Thanks. I didn't know about the `--relative` option. Note that it has an additional effect though ("exclude changes outside the directory"):
```
       --relative[=<path>], --no-relative
           When run from a subdirectory of the project, it can be told to
           exclude changes outside the directory and show pathnames relative
           to it with this option. When you are not in a subdirectory (e.g.
           in a bare repository), you can name which subdirectory to make the
           output relative to by giving a <path> as an argument.
           --no-relative can be used to countermand both diff.relative config
           option and previous --relative.
```

Still, I think we should change this in the README.

PR #2172, opened 2022-04-28
Add `--relative` to `batdiff` in README.md

torqu3e:
The change emits the absolute path from root for changed files so one does not need to be in the repo root for `batdiff` to work, anywhere within the repo is fine.

```
$ git diff --name-only --diff-filter=d
gcp/terraform/apps/platform/sum/main.tf
gcp/terraform/apps/platform/sum/prd.tfvars
```
vs
```
$ git diff --name-only --relative --diff-filter=d 
main.tf
uat.tfvars
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,7 +171,7 @@ You can combine `bat` with `git diff` to view lines around code changes with pro
 highlighting:
 ```bash
 batdiff() {
-    git diff --name-only --line-prefix=`git rev-parse --show-toplevel`/ --diff-filter=d | xargs bat --diff
+    git diff --name-only --relative --diff-filter=d | xargs bat --diff
 }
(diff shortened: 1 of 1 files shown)


Enselic:
Thank you for taking the time to make a contribution. In https://github.com/sharkdp/bat/issues/1640#issue-886782263 there is a proposal to use a simpler solution, namely `--relative`. Would be great if you could see if that works for you, and if so, update this PR 🙏

torqu3e:
`--relative` is certainly cleaner than my originally proposed solution and I was not aware of it, TIL. Updated the PR to reflect the change.

Enselic:
Thank you for the help, now this PR

Closes #1640

PR #2181, opened 2022-05-04
Bump to syntect 5.0.0 to e.g. start lazy-loading syntaxes

### LINKED ISSUE #915: bat panics on simple Vue file

### LINKED ISSUE #951: Improve bat startup speed

### LINKED ISSUE #1846: Jenkins file autodetection
Jenkins is a Ci/CD pipeline platform, to get it working on a, ex github/bitbucket/gitlab, repository a Jenkinsfile needs to be placed on it with groovy syntax.

By default this file is simply named  `Jenkinsfile` or, in some cases, with the extension `.jenkinsfile`, either cases to get it nicely highlighted  one of the bellow options need to be done:
* bat -l groovy Jenkinsfile
* bat -l groovy something.jenkinsfile

so that would be a good addition to bat whether it could auto detect a jenkinsfile at set it up to groovy language

### LINKED ISSUE #1854: File extensions should be matched without case sensitivity
Perhaps there are some example use cases to the contrary, but in general it seems like it would preferable if file extensions were matched ignoring case. I recently experienced this with a file name that had the extension `SQL` instead of `sql` and `bat` did not automatically apply the desired highlighting.

Enselic:
This PR upgrades to syntect 5.0.0 ([release notes](https://github.com/trishume/syntect/releases/tag/v5.0.0)).

Closes #915
Closes #951
Closes #1846
Closes #1854


src/printer.rs
```diff
@@ -447,7 +447,7 @@ impl<'a> Printer for InteractivePrinter<'a> {
             };
             highlighter_from_set
                 .highlighter
-                .highlight(&line, highlighter_from_set.syntax_set)
+                .highlight_line(&line, highlighter_from_set.syntax_set)?
```
sharkdp, line 450: Do we really want to bubble up errors here? That would make bat exit.. and we would not show the end of the file, right?

Don't get me wrong. It's obviously better than panicking. I'm just trying to understand if there are reasonable error cases here that we would want to handle locally.

tests/syntax-tests/source/Vue/example.vue
```diff
@@ -8,6 +8,11 @@
   </div>
 </template>
 
+<template lang='pug'>
+  #container.col
+    p This shall be formated as Plain Text as long as a Pug syntax definition is missing
```
sharkdp, line 13: Small typo
```suggestion
    p This shall be formatted as Plain Text as long as a Pug syntax definition is missing
```

sharkdp (approved):
Thank you very much. Looking forward to having this integrated!

src/printer.rs
Enselic: @sharkdp Good point. Not sure yet exactly what to do other than bubble up though, and in what cases.

Since this PR is a pure improvement in this regard (erroring instead of panicing) I propose that we merge and release this as-is, and then incrementally improve the situation over time, once we figure out what to do instead. Sounds OK?

sharkdp: Yes - thank you

tests/syntax-tests/source/Vue/example.vue
Enselic: Thanks, fixed. I wrote that text before I installed a type-checker in VS Code 🙃

PR #2183, opened 2022-05-05
Simplify release-checklist.md by printing `-h`, `--help` and `man` in CI

Enselic:
Note that we run `find . -name bat.1` before we run anything with
`--all-features`, because `--all-features` is considered a separate
target, and so we will get a different (but identical) `bat.1`. For
example, we might have these:

    ./target/debug/build/bat-218e9538b4996215/out/assets/manual/bat.1
    ./target/debug/build/bat-89d6f56802af023f/out/assets/manual/bat.1

By showing the man page earlier, there is only one `bat.1` to pick from.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -118,6 +118,18 @@ jobs:
         toolchain: stable
         default: true
         profile: minimal
+    - name: Print -h
+      uses: actions-rs/cargo@v1
+      with:
+        command: run
+        args: --locked -- -h
+    - name: Print --help
+      uses: actions-rs/cargo@v1
+      with:
+        command: run
+        args: --locked -- --help
+    - name: Show man page
+      run: man $(find . -name bat.1) | tee
     - name: Check documentation
       env:
         RUSTDOCFLAGS: -D warnings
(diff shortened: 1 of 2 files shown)


.github/workflows/CICD.yml
```diff
@@ -118,6 +118,18 @@ jobs:
...
+      with:
+        command: run
+        args: --locked -- -h
+    - name: Print --help
+      uses: actions-rs/cargo@v1
+      with:
+        command: run
+        args: --locked -- --help
+    - name: Show man page
+      run: man $(find . -name bat.1) | tee
```
sharkdp, line 132: Just curious: why `man … | tee` instead of `man … | cat`? You can also do `man -l …` if you directly use it on files.

Enselic: Simply because I thought my current one-liner was sufficient since it did what I wanted and was not horrible :)

I mean, does it really matter if we pipe to `tee` or `cat`? I would argue it doesn't.

Thanks for the tip with `man -l` though, but it turns out that does not work on macOS, which I think it nice if possible.

The reason I added `| tee` in the first place was because I tested the command locally on my mac, and without it, you get into a pager, which I did not want.

Inside of CI we can simply do plain `man` though since the stdout is not a terminal, so I changed to that 👍

sharkdp: > I mean, does it really matter if we pipe to `tee` or `cat`? I would argue it doesn't.

No, as I said: just curious. Thanks for the explanation.

PR #2187, opened 2022-05-09
Tweak release-checklist.md

Enselic:
Using `--allow-dirty` increases the risk of the published code not being versioned properly in git, so stop recommending it.

Recommend `--locked` for `cargo install` so that the command will not fail if you are using the MSRV toolchain and the latest version of a dependency happens to have a higher MSRV than we do. Contrary to e.g. `cargo build`, `cargo install` will not default to versions in `Cargo.lock`, but instead try to use the latest compatible versions of everything, unless `--locked` is passed.

Recommend a single `git grep` command for MSRV that filters out irrelevant hits in `CHANGELOG.md` and `tests/`.

This is in preparation of the v0.21.0 release so that the diff becomes smaller when it is time to make the release for real.

### DIFF
--- a/doc/release-checklist.md
+++ b/doc/release-checklist.md
@@ -31,7 +31,7 @@
 - [ ] Optional: manually test the new features and command-line options. To do
       this, install the latest `bat` version again (to include the new syntaxes
       and themes).
-- [ ] Run `cargo publish --dry-run --allow-dirty` to make sure that it will
+- [ ] Run `cargo publish --dry-run` to make sure that it will
       succeed later (after creating the GitHub release).
 
 ## Release
--- a/doc/release-checklist.md
+++ b/doc/release-checklist.md
@@ -13,7 +13,7 @@
 
 ## Update syntaxes and themes (build assets)
 
-- [ ] Install the latest master version (`cargo clean && cargo install -f --path .`) and make
+- [ ] Install the latest master version (`cargo clean && cargo install --locked -f --path .`) and make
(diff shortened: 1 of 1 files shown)


sharkdp:
> Using `--allow-dirty` increases the risk of the published code not being versioned properly in git, so stop recommending it.

I used to have that as I ran the dry-run from my main development folder. Later, I used a clean folder for the release. But removing it completely is also fine for me.

doc/release-checklist.md
```diff
@@ -22,6 +22,7 @@
 
 - [ ] Review `-h`, `--help`, and the `man` page. All of these are shown in
       the output of the CI job called *Documentation*, so look there.
+      The CI workflow corresponding to the tip of the master branch is a good place to look.
```
sharkdp, line 25: Maybe add a link to https://github.com/sharkdp/bat/actions/workflows/CICD.yml?query=branch%3Amaster?

sharkdp (approved):

doc/release-checklist.md
Enselic: Nice! I will add a link to that before merging

Enselic: Hmm that's a bit risky though because that query sometimes returns master branches in forks, that might be out of date:

Skärmavbild 2022-05-09 kl  09 08 17

Personally I just go to https://github.com/sharkdp/bat and click the green checkmark next to the commit at the top.

So I think I'll go ahead and merge this PR as is.

Issue #2185, opened 2022-05-06
Issue handling text with escape sequences already embedded


eliminmax:
**What steps will reproduce the bug?**

1. Create a file with multiple embedded escape sequences in one line:
```sh
# shell one-liner to create a simple demo file for the bug
printf '\033[1;32mgreen,bold \033[39mcolorless,bold \033[35mpurple,bold \033[39mcolorless,bold \033[0mcolorless\n' > bat-bug-test
```
2. `bat -pp` the new file

**What happens?**

The first escape sequence overrides all following escape sequences before the final '[0m' (clear formatting) escape sequence

**What did you expect to happen instead?**

The escape sequences would render properly, or not at all

For context, both `cat` from coreutils version 8.32-4ubuntu2 and `busybox cat` from busybox 1:1.30.1-6ubuntu3.1 on Pop!_OS 21.04 result in the following:

![cat-proper](https://user-images.githubusercontent.com/33302691/167203333-98a7f016-a987-46a9-96f3-f7ecc02ff419.png)

Meanwhile, `bat -pp` from bat version 0.20.0 results in the following:

![bat-improper](https://user-images.githubusercontent.com/33302691/167203519-fdb35b01-e3e8-4e28-b6c4-3da0caeed082.png)

For context, I have a "shell collection" consisting of 17 command-line shells, with different colors for the prompts, depending on the shell, and I have an executable that prints out the different prompts, which can be seen [here](https://gist.github.com/eliminmax/22a4c9019dbee9c6442c6eb76088f346). I first ran into this issue when running `bat -pp` on that file.

**How did you install `bat`?**

From the GitHub releases `.deb` file

---

**bat version and environment**

#### Software version

bat 0.20.0 (0655ecf)

#### Operating system

Linux 5.16.19-76051619-generic

#### Command-line

```bash
bat --diagnostic -pp bat-bug-test 
```

#### Environment variables

```bash
SHELL=/usr/bin/bash
PAGER=<not set>
LESS=-Q
LANG=en_US.UTF-8
LC_ALL=<not set>
BAT_PAGER=<not set>
BAT_CACHE_PATH=<not set>
BAT_CONFIG_PATH=<not set>
BAT_OPTS=<not set>
BAT_STYLE=<not set>
BAT_TABS=<not set>
BAT_THEME=<not set>
XDG_CONFIG_HOME=/home/eliminmax/.config
XDG_CACHE_HOME=<not set>
COLORTERM=truecolor
[...]

PR #2189, opened 2022-05-10
Remove code that tries to handle ANSI escape inputs

Enselic:
Proof of concept to aid discussion in #2185. I have not made up my mind yet if this change in sensible, but I think it helps to see it in person. Make sure to ignore whitespace changes when diffing.

IMO this problem is not a blocker to get the v0.21.0 release out since this is not a change in behaviour compared to v0.20.0.

The main use case for the old code seems to be that colored input across newlines should be handled correctly. If you try this:

```
echo "\x1B[33mColor\nColor \x1B[m\nPlain\n" | ./target/debug/bat
```

it looks like this with `cat`:

Skärmavbild 2022-05-10 kl  08 42 04

and the same with the latest release of `bat`, but with the code in this PR it looks like this instead:

Skärmavbild 2022-05-10 kl  08 43 00

since we reset colors when we encounter `\n` it seems. But IMHO we should just consider ANSI escape inputs as out-of-scope for bat, since syntax highlighting is broken when input contains ANSI escape characters anyway.

I have also confirmed that this change makes `bat` show `bat-bug-test` in #2185 the same way as `cat`.


sharkdp:
So what are the things that *couldn't* be done anymore when removing the special ANSI sequence handling? CC @eth-p

Enselic:
Apart from the use case I described from the start, I know of no additional use of the ANSI escape code. And worth emphasising: the current code breaks the use case in #2185. I suppose it is possible to fix it. But is it worth it?

I would also greatly appreciate any input you might have here @eth-p.

sharkdp:
> Apart from the use case I described from the start, I know of no additional use of the ANSI escape handling code.

Oh, sorry. I misread your post. I understand now.

Enselic:
No objections have been raised, so I think we can go ahead and merge this. We can always change our mind later and bring back the code, if we want to.

I will wait another week or so, and then rebase and merge this unless any objections are raised.

3 months has passed since last release, and there a bunch of nice fixes waiting to be released, so it would be nice to get a release out in the coming weeks, IMHO.

CC @sharkdp @keith-hall @eth-p

sharkdp (approved):
Sounds good to me, thanks

PR #2190, opened 2022-05-10
Make `INI` syntax register as handler of `.inf` files

Enselic:
I discovered this while debugging an unexpected diff of `assets/syntaxes.bin` for the upcoming v0.21.0 release.

We need to type `inf` and `INF` as strings in `INI.sublime-syntax`,
otherwise `yaml-rust` interprets them as real numbers ("infinity") and
they do not get registered as file extensions:

    /Users/martin/src/yaml-rust # https://github.com/chyh1990/yaml-rust
    % cargo run --example dump_yaml ~/src/bat/assets/syntaxes/02_Extra/INI.sublime-syntax
    ---
    String("name"):
        String("INI")
    String("file_extensions"):
            String("ini")
            String("INI")
            Real("inf")        <--- NOTE: Interpreted as real number
            Real("INF")
    ...

Also add a regression test.

### DIFF
--- a/assets/syntaxes/02_Extra/INI.sublime-syntax
+++ b/assets/syntaxes/02_Extra/INI.sublime-syntax
@@ -5,8 +5,8 @@ name: INI
 file_extensions:
   - ini
   - INI
-  - inf
-  - INF
+  - "inf"
+  - "INF"
   - reg
   - REG
   - lng
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@
 - Associate `/var/spool/mail/*` and `/var/mail/*` with the `Email` syntax. See #2156 (@cyqsimon)
 - Added cmd-help syntax to scope --help messages. See #2148 (@victor-gp)
 - Slightly adjust Zig syntax. See #2136 (@Enselic)
+- Associate `.inf` files with the `INI` syntax. See #2190 (@Enselic)
 
 ## Themes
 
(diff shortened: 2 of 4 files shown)


assets/syntaxes/02_Extra/INI.sublime-syntax
```diff
@@ -5,8 +5,8 @@ name: INI
 file_extensions:
   - ini
   - INI
-  - inf
-  - INF
+  - "inf"
```
keith-hall, line 8: If I understood correctly, this `sublime-syntax` is manually converted from the `tmLanguage` in the INI submodule. Would it make sense to maintain this as a patch as opposed to directly editing the sublime-syntax?

keith-hall (approved):

assets/syntaxes/02_Extra/INI.sublime-syntax
Enselic: Thank you for code review! IMHO patches are a bit annoying and should be used sparingly, most appropriately when the file to be patched is not under our control. In this case the file happens to be versioned with us, so I think the most straightforward is to patch it directly, and then add a regression test so we detect if there is a regression. But if you strongly prefer a patch that is fine with me, I can change to that :)

keith-hall: Your reasoning is sound, let's leave it how it is :)

PR #2197, opened 2022-05-21
Make the default macOS theme depend on Dark Mode

Enselic:
We frequently get complaints from macOS users that bat does not work on
their default macOS terminal background, which is white.

Pay the price of slightly increased startup time to get a better default
on macOS. To avoid the slightly increased startup time, simply
specify a theme explicitly via `--theme`, `BAT_THEME`, or
`~/.config/bat`.

I did some quick testing, and it seems to work. With Dark Mode:

Skärmavbild 2022-05-21 kl  18 22 52

With default system theme:

Skärmavbild 2022-05-21 kl  18 22 33

Startup time is increased slightly due to this check:

|:---|---:|---:|---:|---:|
| `./target/release/bat -f examples/simple.rs` | 12.9 ± 1.3 | 12.1 | 23.2 | 1.50 ± 0.16 |
| `bat-v0.21.0 -f examples/simple.rs` | 8.6 ± 0.2 | 8.3 | 9.3 | 1.00 |

But there is no difference if an explicit theme is selected:

|:---|---:|---:|---:|---:|
| `./target/release/bat -f --theme "Monokai Extended" examples/simple.rs` | 8.7 ± 0.2 | 8.3 | 9.5 | 1.00 |
| `bat-v0.21.0 -f --theme "Monokai Extended" examples/simple.rs` | 8.7 ± 0.6 | 8.3 | 19.0 | 1.01 ± 0.08 |

Note that if there is an error when we check if Dark Mode is enabled, we
behave the same as on Windows and Linux; assume that the terminal
background is dark. This harmonizes behavior across platforms, and makes
bat behave the same as before, when Dark Mode was always assumed to be
enabled.


Enselic:
Some tests are failing. Will investigate some other day.

keith-hall (approved):

Enselic:
Thanks for review @keith-hall 🙏 

The tests were failing because it seems like the CI runner uses a light system theme by default on macOS. Making the tests explicitly set a theme fixed the failures.

I have contemplated to add automated tests for this, but since the test must also pass on systems that uses Dark Mode, and since I don't want our testing code to change macOS system theme, I don't currently plan on adding automated tests for this. If we get problems with regressions going ahead, I will reconsider. We could make use of dependency injection here, for example.

But as it stands, this PR is fundamentally ready to be merged. I want to keep it open for at least a few days though, maybe a week, to give time for more feedback.

Enselic:
Just realised I also need to add an entry to CHANGELOG.md ;)

sharkdp (approved):
Nice, thank you!

Enselic:
Thanks for review! Seems unlikely that more feedback will come in, so I will merge this now.

sh471:
The command you provided appears to be running a program written in the Rust programming language. The program uses the "bat" library, which is a library for syntax highlighting and pretty-printing of source code.

The program opens the file named "examples/simple.rs" which is it own source code and then uses the PrettyPrinter from the bat library to print the file. The print method will format the text with syntax highlighting and other pretty formatting options that can make the code easier to read.

It seems like the program is running without errors and has printed the source code of the program in the terminal window, as indicated by the message Running 'target/debug/bat examples/simple.rs'.

The file name given examples/simple.rs is in the relative path, if you want to run the file in other location, you might have to change the path accordingly.

Let me know if you have any questions about this code or any other Rust-related topic.

Issue #2191, opened 2022-05-11
Allow for preprocessing or pretty printing


salva:
I find myself repeatedly combining `bat` with a pretty printer, as in: `jq . file.json|bat -l JSON`

IMO, it would be very handy to have the preprocessing/pretty printing integrated into `bat`, so that one could just run it like `bat -P file.json`.

(And I know about `prettybat` from the `bat-extras` repository, but still, I think having the feature natively supported would be much better)

sharkdp:
Potential solution: #2203

eth-p:
Another potential solution: #2444.

Support for `LESSOPEN`/`LESSCLOSE` is not yet included in release bilds by default, but if/when it is, it would be possible to use it to run pretty printers as an input preprocessor.

Issue #2188, opened 2022-05-09
Support of plain file in .gz format - zless equivalent


Gu3nol:
Hi,
Are you support an equivalent feature to zless (opening .gz file in plain text)?
I tried using -A but the file is not human readable.
Best,

sharkdp:
Please see #748, #642 and especially #237.

Issue #2070, opened 2022-02-10
Feature request: Download files from the internet


laurinneff:
I often use bat to syntax highlight files from the internet, for example:

```sh
curl https://example.com | bat --language html
```

It would be nice if I could just do this instead:

```sh
bat https://example.com
```

and it detected the language from the `Content-Type` header. This might cause issues with filenames that look like URLs, but that could be solved by also adding support for the file scheme: `file:///path/to/file` and maybe adding a warning if there's a file with the same name

keith-hall:
Thanks for the feedback/suggestion. As `bat` is primarily a `cat` replacement, I personally don't feel this feature would be in scope. Sounds perfect for a wrapper script though! Maybe suggest it on https://github.com/eth-p/bat-extras? :)

Issue #1940, opened 2021-11-09
Provide useful output when `bat` is given a directory instead of a file


jasonkarns:
It's pretty common (for me) to run `bat thing` when `thing` is a directory. It's usually because I want to see what's _in_ `thing` but (initially) I may not know it's a directory.

Since bat's purpose (in my words) is to be a more useful `cat`, I think it would be helpful to print more useful output when given –accidentally or otherwise– a directory. My initial suggestion would be to print the directory's contents as if via `ls`. 

Granted, it still should be considered an error to invoke either `cat` and therefore also `bat` with a directory. So I would still consider the exit code to be `1`; and it should continue to print `[bat error]: 'thing' is a directory.`

But bat could emit a file listing to STDOUT (or STDERR) along with the existing error message.

Thoughts?

Enselic:
Personally I would consider this to be out-of-scope for bat.

It would be easy to write a wrapper script that invokes `ls` if the argument is a dir and `bat` if the argument is a file. With some trickery you can even call that script `bat` and have it on your `PATH`.

sharkdp:
This has been requested before and I previously had a similar reply like @Enselic. On the other hand, I do have to admit that it also happens to me from time to time :smile:. Maybe we could indeed print a simple directory listing. Similar to what `vim <directory>` shows, but without any of the controls. We could also call out to `ls`, but that comes with a lot of portability concerns (I do have some experience from `fd --list-details`).

But I'm also okay with closing this as "won't do". It's not really in the scope of `bat`.

Enselic:
This has indeed been requested before. I could easily find three examples that has been closed: #1156, #971, #427.

Let's keep this one open then and see what happens :)

keith-hall:
The tool mentioned at https://github.com/sharkdp/bat/issues/800#issuecomment-987855856 also can provide this functionality.

sharkdp:
Potential solution: #2203

kohane27:
I was having this exact issue and thanks to @keith-hall for linking to the awesome program [rsop](https://github.com/desbma/rsop) created by @desbma. Thank you to you two awesome people!

Here's my little guide for using `rsop` along with `fzf`.

In the `fzf` preview window, if it's a directory, it'll use `lsd` to preview the directory; if it's a file, it'll use `bat` to preview the file.

1. Install [rsop](https://github.com/desbma/rsop)
[...]

PR #2203, opened 2022-05-29
Proof of concept: Lua plugins

sharkdp:
This is a proof-of-concept implementation of something that we discussed amongst the maintainers recently: a **plugin mechanism for bat**. The idea of this PR is to get the discussion started. And to answer a few questions. Most importantly: do we really want/need this?

This whole idea came up because there are a lot of feature requests that keep popping up on `bat`s issue tracker that we previously always declined, mostly because we wanted to keep `bat` focused on its core functionality. And because of the additional maintenance overhead:
- **gzip support** (#237, #642, #748, #2188)
- **CURL support** (#4, #2070)
- **`bat <directory>` support** (#427, #971, #1156, #1940)
- **View binary files as hex dump** (#619, #800, #1554)
- **Pretty printing** (#565, #1592, #2191)

On the other hand, most of these functionalities *are* actually quite useful! A lot of workarounds have been suggested. Lots of people wrote their own wrapper functions. And with [bat-extras](https://github.com/eth-p/bat-extras), we even have a dedicated project working on extending `bat`s functionality. The "problem" with all of these solutions is that they are not directly included in `bat`, so you need to remember the name of the wrapper scripts. And you need to install them separately.

Here, we try to follow a different approach: instead of forcing users to wrap `bat`, we enable users to *add functionality to bat* by providing a plugin mechanism. We could still maintain those plugins here in this repository. But (1) we could easily argue that those plugins are not part of `bat`s core functionality, especially if all of them would be opt-in (2) it would be easier to maintain this functionality... or to throw it out again. And (3), users could easily customize those plugins or write their own.

In this PoC, plugins can be enabled by adding lines like the following to your `bat` config file:
```bash
--load-plugin "uncompress.lua"
--load-plugin "curl.lua"
--load-plugin "directories.lua"
```
Plugins are written in Lua. For now, there is just one type of plugin: a "preprocess" plugin that can modify the input to `bat`. These plugins have to provide a `preprocess` function that will be called for every input path to `bat`. The plugins then have the option to *return a different path* instead. This can be used to hand back paths to temporary files with the output of the preprocessing. For example, `directories.lua` looks like this:
```lua
function preprocess(path)
    if is_dir(path) then
        tmpfile = os.tmpname()
        os.execute("ls -alh --color=always '" .. path .. "' > '" .. tmpfile .. "'")
        return tmpfile
    else
        return path
    end
end
```
Obviously, *this is not a great plugin architecture*. It's extremely restrictive. And the interface between `bat` and the plugins is very limited. I could easily imagine a lot of different things that we would like to expose to the plugins. Or functionality that we would like to provide (e.g. to add new command-line options to `bat`). But it's somewhat surprising that this simple interface is (up to a few limitations) powerful enough to handle four of the five feature requests listed above:

## Showcase

**gzip support**:
image

**CURL support**:
image

**`bat <directory>` support**
image

**Preview binary files**

image

## Executable size, Benchmarks

I chose Lua because I heard that it is cheap to embed it into existing applications. And also fast. I have never programmed in Lua until today, so I would have probably have preferred another scripting language personally. But I don't think we want to ship a Python interpreter with `bat`.

But obviously, this still comes with some overhead in terms of binary size:
```
bat-master: 4.69 MB (4,685,824 bytes)
bat-2203:   5.05 MB (5,050,368 bytes)
```

Concerning startup time: the overhead here seems to be pretty much negligible. With three plugins activated and executed (but in the trivial "don't touch this path" mode) and 22 input files passed to `bat` (i.e. 3 x 22 = 66 Lua function invocations), we get a slowdown of less than a millisecond:

|:---|---:|---:|---:|---:|
| `./bat-master src/*.rs` | 5.2 ± 0.4 | 4.5 | 7.4 | 1.00 |
| `./bat-2203 src/*.rs` | 5.6 ± 0.4 | 4.9 | 7.9 | 1.08 ± 0.12 |

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -98,6 +98,7 @@ dependencies = [
  "path_abs",
  "predicates",
  "regex",
+ "rlua",
  "semver",
  "serde",
  "serde_yaml",
@@ -914,6 +915,30 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9a374af9a0e5fdcdd98c1c7b64f05004f9ea2555b6c75f211daa81268a3c50f1"
 dependencies = [
  "bytemuck",
+]
+
+[[package]]
+name = "rlua"
+version = "0.19.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "627ae78424400009e889c43b0c188d0b7af3fe7301b68c03d9cfacb29115408a"
+dependencies = [
+ "bitflags",
+ "bstr",
+ "libc",
+ "num-traits",
+ "rlua-lua54-sys",
+]
+
+[[package]]
+name = "rlua-lua54-sys"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b4e1fdfc6a5f7acfa1b1fe26c5076b54f5ebd6818b5982460c39844c8b859370"
+dependencies = [
+ "cc",
+ "libc",
+ "pkg-config",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ minimal-application = [
     "paging",
     "regex-onig",
     "wild",
+    "rlua"
 ]
 git = ["git2"] # Support indicating git modifications
 paging = ["shell-words", "grep-cli"] # Support applying a pager on the output
@@ -65,6 +66,7 @@ grep-cli = { version = "0.1.6", optional = true }
 regex = { version = "1.5.5", optional = true }
 walkdir = { version = "2.0", optional = true }
 bytesize = { version = "1.1.0" }
+rlua = { version = "0.19", optional = true }
 
 [dependencies.git2]
 version = "0.14"
--- a/plugins/curl.lua
+++ b/plugins/curl.lua
@@ -1,0 +1,21 @@
+function tempdir()
+    local stream = assert(io.popen('mktemp --directory'))
+    local output = stream:read('*all')
+    stream:close()
+    return string.gsub(output, "\n", "")
+end
+
+function preprocess(path_or_url)
+    filename_from_url = string.match(path_or_url, '^https?://.*/(.*)$')
+    if filename_from_url then
+        local temp_directory = tempdir()
+        local new_path = temp_directory .. "/" .. filename_from_url
+
+        -- TODO: how to prevent shell injection bugs?
+        os.execute("curl --silent '" .. path_or_url .. "' --output '" .. new_path .. "'")
+
+        return new_path
+    else
+        return path_or_url
+    end
+end
--- a/plugins/directories.lua
+++ b/plugins/directories.lua
@@ -1,0 +1,17 @@
+-- https://stackoverflow.com/a/3254007/704831
+function is_dir(path)
+    local f = io.open(path, "r")
+    local ok, err, code = f:read(1)
+    f:close()
+    return code == 21
+end
+
+function preprocess(path)
+    if is_dir(path) then
+        tmpfile = os.tmpname()
+        os.execute("ls -alh --color=always '" .. path .. "' > '" .. tmpfile .. "'")
+        return tmpfile
+    else
+        return path
+    end
+end
--- a/plugins/hexyl.lua
+++ b/plugins/hexyl.lua
@@ -1,0 +1,21 @@
+-- Note: this plugin depends on the existence of 'inspect' [1] and 'hexyl' [2]
+--
+-- [1] https://github.com/sharkdp/content_inspector
+-- [2] https://github.com/sharkdp/hexyl
+
+function is_binary(path)
+    local stream = assert(io.popen("inspect '" .. path .. "'"))
+    local output = stream:read('*all')
+    stream:close()
+    return string.find(output, ": binary\n")
+end
+
+function preprocess(path)
+    if is_binary(path) then
+        tmpfile = os.tmpname()
+        os.execute("hexyl --length 1024 --no-position --border=none --no-squeezing '" .. path .. "' > '" .. tmpfile .. "'")
+        return tmpfile
+    else
+        return path
+    end
+end
--- a/plugins/uncompress.lua
+++ b/plugins/uncompress.lua
@@ -1,0 +1,21 @@
+function tempdir()
+    local stream = assert(io.popen('mktemp --directory'))
+    local output = stream:read('*all')
+    stream:close()
+    return string.gsub(output, "\n", "")
+end
+
+function preprocess(path)
+    prefix = string.match(path, '^(.*)%.gz$')
+    if prefix then
+        local temp_directory = tempdir()
+        local new_path = temp_directory .. "/" .. prefix
+
+        -- TODO: how to prevent shell injection bugs?
+        os.execute("gunzip < '" .. path .. "' > '" .. new_path .. "'")
+
+        return new_path
+    else
+        return path
+    end
+end
--- a/src/bin/bat/app.rs
+++ b/src/bin/bat/app.rs
@@ -241,6 +241,11 @@ impl App {
                 .map(HighlightedLineRanges)
                 .unwrap_or_default(),
             use_custom_assets: !self.matches.is_present("no-custom-assets"),
+            plugins: self
+                .matches
+                .values_of_os("load-plugin")
+                .unwrap_or_default()
+                .collect(),
         })
     }
 
--- a/src/bin/bat/clap_app.rs
+++ b/src/bin/bat/clap_app.rs
@@ -475,6 +475,14 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
                 .long_help("Display a list of supported languages for syntax highlighting."),
         )
         .arg(
+            Arg::with_name("load-plugin")
+            .long("load-plugin")
+            .takes_value(true)
+            .value_name("name")
+            .help("Load plugin with specified name.")
+            .hidden_short_help(true)
+        )
+        .arg(
             Arg::with_name("unbuffered")
                 .short("u")
                 .long("unbuffered")
--- a/src/bin/bat/clap_app.rs
+++ b/src/bin/bat/clap_app.rs
@@ -477,7 +477,9 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
         .arg(
             Arg::with_name("load-plugin")
             .long("load-plugin")
+            .multiple(true)
             .takes_value(true)
+            .number_of_values(1)
             .value_name("name")
             .help("Load plugin with specified name.")
             .hidden_short_help(true)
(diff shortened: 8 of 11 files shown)


Enselic:
Very interesting! I think you are onto something very powerful and useful here. Some high level remarks/questions:

* A traditional plug-in system does not require explicitly enabling plug-ins, but instead uses all plugins in say `~/.config/bat/plugins`. Maybe we should also behave like that? 
* I think Lua is a perfectly reasonable choice as a scripting language, but I think we should consider a "binary" based interface, so that *any* language and runtime could be used to write plugins. For example: plug-ins get the path in argv and print a new path to stdout. Or maybe something even more sophisticated.
* It can be tricky to make `cargo install bat` also install plug-ins, which I think is necessary to handle some use cases out-of-the-box. We could maybe make some plug-ins internal/intrinsic, but then a binary architecture is probably out of the question.

Enselic:
Very interesting! I think you are onto something very powerful and useful here. Some high level remarks/questions:

* A traditional plug-in system does not require explicitly enabling plug-ins, but instead uses all plugins in say `~/.config/bat/plugins`. Maybe we should also behave like that? 
* I think Lua is a perfectly reasonable choice as a scripting language, but I think we should consider a "binary" based interface, so that *any* language and runtime could be used to write plugins. For example: plug-ins get the path in argv and print a new path to stdout. Or maybe something even more sophisticated.
* It can be tricky to make `cargo install bat` also install plug-ins, which I think is necessary to handle some use cases out-of-the-box. We could maybe make some plug-ins internal/intrinsic, but then a binary architecture is probably out of the question.

keith-hall:
I had similar thoughts. I like the idea in general, and have some concerns which I'd like us to discuss before we get too deep:

- You mention Lua is good for embedding, and it means we don't have to ship a Python interpreter. I'd argue that none of us know Lua very well, so maintaining plugins and reviewing PRs is going to be hard for us. I like Martin's direction of not being tied to a particular scripting language, or even scripting. If we were to use Python plugins by default, as an example, and the user has no Python interpreter, `bat` could just skip processing the plugins. But I do like how startup time is barely affected in this PoC! :)
- How will this affect our diagnostics and ability to debug issues? Would --diagnostics also append all plugin code? Seems a bit noisy? Maybe just a diff from the released plugins somehow?
- Is it really necessary to use temporary files instead of stdout? I realize it may be in some cases, but IMO would be nice to avoid when we can.
- I didn't quite grok why it was necessary to move from `pub(crate)` to `pub` for some struct fields. But it could be useful to pass the syntax name to the plugin, or have a plugin resolve which syntax to highlight with.

sharkdp:
> A traditional plug-in system does not require explicitly enabling plug-ins, but instead uses all plugins in say `~/.config/bat/plugins`. Maybe we should also behave like that?

Hm. I was thinking that plugins would be similar to syntaxes or themes. There could be builtin plugins. And in addition, there could be user plugins in `~/.config/bat/plugins`. It might make sense to enable everything in `~/.config/bat/plugins` automatically (since the user controls that folder). But I thought it would be good (at least in the beginning), to require opt-in for all builtin plugins. Doing the same for user-plugins would then be more consistent maybe.

> I think Lua is a perfectly reasonable choice as a scripting language, but I think we should consider a "binary" based interface, so that _any_ language and runtime could be used to write plugins. For example: plug-ins get the path in argv and print a new path to stdout. Or maybe something even more sophisticated.

> You mention Lua is good for embedding, and it means we don't have to ship a Python interpreter. I'd argue that none of us know Lua very well, so maintaining plugins and reviewing PRs is going to be hard for us. I like Martin's direction of not being tied to a particular scripting language, or even scripting. If we were to use Python plugins by default, as an example, and the user has no Python interpreter, `bat` could just skip processing the plugins. But I do like how startup time is barely affected in this PoC! :)

So the main reason why I went for a deep integration with a scripting language as opposed to a "binary interface" is this: I believe that we would quickly want more interaction between `bat` and the plugins. As described in the initial post, this PR sketches just a very basic interface (path in, path out), that could very easily be implemented with an IO-based approach where we simply communicate with an arbitrary process via Unix pipes. But I could imagine that "plugin authors" would quickly be in need for a much tighter integration. For example:
- Ways for plugins to query `bat`s internal list of syntaxes, syntax mappings, themes, etc.
- Ways for plugins to access `bat`s command-line arguments
- Allow plugins to add their own command-line arguments
- Allow plugins to directly generate output (instead of writing to temporary files)
- Allow plugins to modify `bat`s appearance (e.g. set a particular `--style` for certain file types)
- Allow plugins to control things like `--line-range` or `--highlight-line`. And all other settings.
- …
All of this could also be implemented with a generic IO-based interface (think JSON-RPC), but then we would basically need a full support library for each plugin language.

> It can be tricky to make `cargo install bat` also install plug-ins, which I think is necessary to handle some use cases out-of-the-box. We could maybe make some plug-ins internal/intrinsic, but then a binary architecture is probably out of the question.

Good point. I didn't think of that. My feeling is that `cargo`-based installations are a tiny minority, but it would be nice to further support it, of course. I mean… we already include a lot of assets inside the `bat` binary. It might seem strange, but we could, in theory, also embed (compressed) Lua source code for the builtin plugins.

sharkdp:
@Enselic @keith-hall Do you think this is worth exploring further?

Enselic:
My 100% honest opinion is that I don't want to increase the maintenance burden of this project with Lua scripts, or anything else that is not Rust :)

I do think that a good plugin-infrastructure has the potential to _decrease_ the long-term maintenance burden of this project, which would be great, but I don't think it should be Lua based.

Sorry for being somewhat discouraging, but I interpreted your request for honesty as a sincere request, so I wanted to be honest :)

sharkdp:
> Sorry for being somewhat discouraging, but I interpreted your request for honesty as a sincere request, so I wanted to be honest :)

Yes, of course. Thank you.

> My 100% honest opinion is that I don't want to increase the maintenance burden of this project with Lua scripts, or anything else that is not Rust :)
> I do think that a good plugin-infrastructure has the potential to decrease the long-term maintenance burden of this project, which would be great, but I don't think it should be Lua based.

I don't think that adding plugins will *ever* decrease the maintenance burden. Plugins will - hopefully - allow us to improve the functionality of the project. But maintenance-wise, they will almost certainly increase the effort. If easier maintenance is the only goal of this project, we should not accept any new features (which is also a valid approach). And certainly not a plugin-architecture that allows for a manifold of new features (and bugs).

So I'm not sure if it's fair to evaluate this *purely* from a maintenance-related perspective. First and foremost, I would like to evaluate this from a user-perspective:

* Is this new functionality something that users want? I certainly think so, given the amount of linked tickets in the top post
* Is a plugin architecture something that users want? Well, I'm not so sure. Probably, some of them would like to see all of those features directly included in the main application. On the other hand, some users might like the ability to create new plugins or modify existing ones. A plugin architecture certainly also allows us so prototype new features quicker, which should be a positive thing from a user-perspective.
* Is a scripting-language plugin architecture something that users want? Or would they prefer a Rust-based architecture? Again, not sure. But *I think* that most users would probably prefer a scripting language (even one they don't know) compared to a much harder-to-learn system programming language. I think that users would also prefer to have an easy way to create/modify plugins (simply edit a text file), without having to install an entire Rust toolchain. And without having to re-compile their plugins (into shared libraries?) on each change.
* Assuming we want a scripting-language-based plugin architecture, if Lua the best choice? From a user perspective: I don't think so. There are certainly much more people familiar with Python or JavaScript than with Lua. However, the big downside is the execution time (as detailed above). A large delay in `bat`s startup time is something that I personally - as a *user* - would not tolerate.

Coming back to the maintenance concern: would it help if we move all plugins to separate repositories (similar to how it works for Sublime syntaxes) and let them be maintained independently, by the respective authors? Instead of directly integrating them in our project, we could simply provide a list/registry of plugins somewhere... and users would simply install them by `git clone`ing the respective plugin repos into their `~/.config/bat/plugins` folder.

Enselic:
What you say make a lot of sense. What it boils down to for me personally is that I don't want to read or write Lua to solve problems in bat. I sloppily framed this as "increased maintenance". But I know very well that no one expects me to read or write Lua, so I would say my concern should not have too much weight here. It was just some spontaneous feedback. Merging this will not prevent me from continue to contribute to other parts of bat, so if that's what we end up doing, that's fine.

Thank you for reaching out for feedback and for elaborating on your position.

keith-hall:
> * A large delay in `bat`s startup time is something that I myself - as a _user_ - would not tolerate.

Agreed. Is there a way to avoid starting the plugin host on each `bat` invocation - at least for the (common?) case where no plugins would need to do anything with the input? Some type of configuration of when the plugin would apply? 

> would it help if we move all plugins to separate repositories (similar to how it works for Sublime syntaxes) and let them be maintained independently, by the respective authors? Instead of directly integrating them in our project, we could simply provide a list/registry of plugins somewhere... and users would simply install them by `git clone`ing the respective plugin repos into their `~/.config/bat/plugins` folder.

I personally worry about then having no control over the "base" set of plugins, which could make it hard to triage bug reports. Maybe having the diagnostic list all the installed plugins, but that could ad a lot of noise.

Those concerns aside, I like the idea in general, it definitely helps solve a lot of user requests. But at the same time, I feel those could also be solved with something external to `bat`, like [`rsop`](https://github.com/desbma/rsop).

So my "final verdict" : I'm undecided, but I'm happy to go with the flow. Adding plugins will no doubt be an interesting challenge / experience if we go for it.

sharkdp:
> > * A large delay in `bat`s startup time is something that I myself - as a _user_ - would not tolerate.
> 
> Agreed. Is there a way to avoid starting the plugin host on each `bat` invocation - at least for the (common?) case where no plugins would need to do anything with the input? Some type of configuration of when the plugin would apply?

I thought about that as well. But that would probably make plugins (a bit?) less powerful. Because the functionality of *when* to call the plugin would have to be done in the core `bat` application (e.g. via something like `--map-extension-to-plugin '*.gz:uncompress`). Or, with a plugin-specific config file, similar to:
```toml
[bat-plugin]
name = "uncompress"
description = "…"
filetype_filter = "*.gz"
```
That could work quite well for some plugins. But the "directories" and "curl" plugins above, for example, would not fit this pattern. And immediately call for an extension of the config format.

> I personally worry about then having no control over the "base" set of plugins, which could make it hard to triage bug reports. Maybe having the diagnostic list all the installed plugins, but that could ad a lot of noise.

Agreed. We will probably receive lots of falsely targeted bug reports in this repo if we go down that route, similar to bugs in the syntaxes.

> Those concerns aside, I like the idea in general, it definitely helps solve a lot of user requests. But at the same time, I feel those could also be solved with something external to `bat`, like [`rsop`](https://github.com/desbma/rsop).

True. But requires users to know about two tools. And gives them less control over the specific options passed to `bat`.

> So my "final verdict" : I'm undecided, but I'm happy to go with the flow. Adding plugins will no doubt be an interesting challenge / experience if we go for it.

Ok, thank you both for voicing your concerns. I suggest we close this topic for now. Focus on `bat`s core strengths. And we can always come back to this (or a better) idea, if we believe that we really need a plugin system in `bat`.

xeruf:
Just a note: If this ever gets implemented, the plugin should be able to return multiple file paths for one, such that one can view the contents of files in a compressed directory for example, this would be amazing!

I already wrote a script that does lots of this kind of preprocessing, but the pagination often ends up suboptimal because `bat` then receives the output as a stream rather than individual files.

Issue #2221, opened 2022-06-17
Tracking issue for bat v1.0.0


Enselic:
At one point it would be nice to release bat v1.0.0.

This issue tracks various initiatives that needs to be taken or that has been taken to reach that goal.

An incomplete list of things to do before before releasing bat v1.0.0:

* [ ] Split out the unstable parts of the `bat` library API into a new crate `bat-impl`. Draft: https://github.com/Enselic/bat/pull/125
* [ ] Go through all (or at least the most used) [crates that depend on bat](https://crates.io/crates/bat/reverse_dependencies) to make sure the bat v1.0.0 API is sufficient. I already looked at the most popular bat dependent, namely cargo-expand, and after https://github.com/dtolnay/cargo-expand/pull/148 that particular crate should have no troubles to upgrade to bat v1.0.0.
* [ ] Clean up the parts of what can become the bat v1.0 API, which will basically be just `bat::PrettyPrinter`, to avoid unwanted things outside of our control. For example, the API to list all syntaxes exposes a `syntect` data structure. We should abstract that away.
* [ ] I think we need to solve https://github.com/sharkdp/bat/issues/2085 to make bat version upgrades smooth. Maybe by having separate versioning of the binary format of assets.

PR #2222, opened 2022-06-19
Make `bat::PrettyPrinter::syntaxes()` iterate over new `bat::Syntax` struct

Enselic:
We can't keep `syntect::parsing::SyntaxReference` as part of the public API, because that might prevent us from bumping to syntect 6.0.0 without also bumping bat to v2.0.0.

So introduce a new stripped down struct `Syntax` and return that instead. Let it be fully owned to make the API simple. It is not going to be in a hot code path anyway.

I have looked at all code of our 27 dependents but I can't find a single instance of this method being used, so this change should be safe for v1.0.0. It is only used in our own example code as far as I can tell.

For reference (mostly to my future self), here is the script I used to download the code of all dependents. Then one can grep around in the code of dependents:

```bash
#/usr/bin/env bash

# curl -L 'https://crates.io/api/v1/crates/bat/reverse_dependencies' | jq --raw-output '.versions | map(.crate) | flatten[]'
# curl -L 'https://crates.io/api/v1/crates/bat/reverse_dependencies?page=2' | jq --raw-output '.versions | map(.crate) | flatten[]'
# curl -L 'https://crates.io/api/v1/crates/bat/reverse_dependencies?page=3' | jq --raw-output '.versions | map(.crate) | flatten[]'
bat_dependents="
barberousse
borrowing_exerci
bropages
cargo-expand
download_caretaker
dsconv
dts
etrade
fix-hidden-lifetime-bug-proc_macros
gistit-cli
gistit
git-delta-lib
git-delta
gooseberry
hgrep
longboard
mdn
next-gen-proc_macros
nu_plugin_textview
partiql-rs
piqel
pmis
qcow-cli
raen
riffle
trrs
with_locals-proc_macros
"

for crate in $bat_dependents; do
    echo "Processing $crate"
    mkdir -p "$crate"
    cd "$crate"
    newest_version=$(curl -L https://crates.io/api/v1/crates/$crate | jq --raw-output .crate.newest_version)
    echo "$newest_version"
    curl -L "https://crates.io/api/v1/crates/$crate/$newest_version/download" | tar -zxf -
    cd ..

    sleep 2 # avoid throttling
done
```

Also see #2221

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@
 
 ## `bat` as a library
 
+- Make `bat::PrettyPrinter::syntaxes()` iterate over new `bat::Syntax` struct instead of `&syntect::parsing::SyntaxReference`. See #2222 (@Enselic)
+
 
 # v0.21.0
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -48,7 +48,7 @@ pub(crate) mod syntax_mapping;
 mod terminal;
 pub(crate) mod wrapping;
 
-pub use pretty_printer::{Input, PrettyPrinter};
+pub use pretty_printer::{Input, PrettyPrinter, Syntax};
 pub use syntax_mapping::{MappingTarget, SyntaxMapping};
 pub use wrapping::WrappingMode;
 
--- a/src/pretty_printer.rs
+++ b/src/pretty_printer.rs
@@ -2,7 +2,6 @@ use std::io::Read;
 use std::path::Path;
 
 use console::Term;
-use syntect::parsing::SyntaxReference;
 
 use crate::{
     assets::HighlightingAssets,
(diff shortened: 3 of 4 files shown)


Enselic:
Just to be clear
* We should release this (and other 1.0.0 RC changes) as the usual bat 0.x.0 bump, i.e. we should not go directly to 1.0.0
* I will update CHANGELOG.md before merge

sharkdp (approved):
Thank you very much. This looks good to me!

Enselic:
For the record, here is the API diff from this PR:

```
% cargo public-api --diff-git-checkouts origin/master stabilize-syntaxes
(none)

-pub fn bat::PrettyPrinter::syntaxes(&self) -> impl Iterator<Item = &SyntaxReference>
+pub fn bat::PrettyPrinter::syntaxes(&self) -> impl Iterator<Item = Syntax> + '_

Added items to the public API
=============================
+#[non_exhaustive] pub struct bat::Syntax
+pub struct field bat::Syntax::file_extensions: Vec<String>
+pub struct field bat::Syntax::name: String
```

Issue #2023, opened 2022-01-12
PAGER=batcat turns into a fork nightmare


rsamuelklatchko:
I decided to make batcat my PAGER and when I tried setting:

    PAGER=batcat

it didn't work. After some troubleshooting I realized that batcat was forking/execing itself and was able to fix it by also adding:

    BAT_PAGER=less

while this solved the problem it was not obvious. One suggestion is when reading PAGER, if the value is the binary name (or ends with /{binary_name}) the value is ignored and the default is used instead.

Enselic:
Thanks for reporting!

There is code to prevent that from happening, but it does not take into account that the binary can be called something other than `bat`.

See https://github.com/sharkdp/bat/blob/9287cf6b0ba98be2611b5a682dffcc1a8583b151/src/pager.rs#L105

I would be happy to review PRs that solves this

rsamuelklatchko:
My thought would be to modify PagerKind::from_bin to support other names. I have a couple of ideas:

1. Just add "batcat" (I'm on Debian which is what the binary is named) and possibly any other common alternate names if you know of any.
2. Find the name the binary is being run as using `env::args`.

For #2, I'm thinking to reuse https://github.com/sharkdp/bat/blob/9287cf6b0ba98be2611b5a682dffcc1a8583b151/src/pager.rs#L43-L46 to exctract the name from env::args[0].

Enselic:
My first choice would be 2., but the code to get the program name is more something like `std::env::args().nth(0)`

Or even better is `std::env::args_os().nth(0)` but that might require a bigger change

rsamuelklatchko:
I have a code fix but having not used GitHub I'm not sure if exactly how to send it to you. What I'm suggesting is this:

    impl PagerKind {
        fn from_bin(bin: &str) -> PagerKind {
            use std::path::Path;
    
            let file_stem = Path::new(bin).file_stem();
    
            match file_stem.map(|s| s.to_string_lossy()).as_deref() {
                Some("bat") => PagerKind::Bat,
                Some("less") => PagerKind::Less,
                Some("more") => PagerKind::More,
                Some("most") => PagerKind::Most,
                _ => {
                    if env::args_os()
                        .nth(0)
                        .map(|arg0| Path::new(&arg0).file_stem() == file_stem)
                        .unwrap_or(false)
                    {
                        PagerKind::Bat
                    } else {
                        PagerKind::Unknown
                    }
                }
            }
        }
    }

[...]

PR #2235, opened 2022-07-01
Prevent fork nightmare with PAGER=batcat

johnmatthiggins:
Used rsamuelklatchko's suggested changes to fix recursive fork issue. Fixes issue #2023.

### DIFF
--- a/src/pager.rs
+++ b/src/pager.rs
@@ -40,16 +40,24 @@ impl PagerKind {
     fn from_bin(bin: &str) -> PagerKind {
         use std::path::Path;
 
-        match Path::new(bin)
-            .file_stem()
-            .map(|s| s.to_string_lossy())
-            .as_deref()
-        {
+        let file_stem = Path::new(bin).file_stem();
+
+        match file_stem.map(|s| s.to_string_lossy()).as_deref() {
             Some("bat") => PagerKind::Bat,
             Some("less") => PagerKind::Less,
             Some("more") => PagerKind::More,
             Some("most") => PagerKind::Most,
-            _ => PagerKind::Unknown,
+            _ => {
+                if env::args_os()
+                    .nth(0)
+                    .map(|arg0| Path::new(&arg0).file_stem() == file_stem)
+                    .unwrap_or(false)
+                {
+                    PagerKind::Bat
+                } else {
+                    PagerKind::Unknown
+                }
+            }
         }
     }
 }
--- a/src/pager.rs
+++ b/src/pager.rs
@@ -39,29 +39,27 @@ pub(crate) enum PagerKind {
 impl PagerKind {
     fn from_bin(bin: &str) -> PagerKind {
         use std::path::Path;
-        
+
         // Set to `less` by default on most Linux distros.
         let pager_bin = Path::new(bin).file_stem();
-        
+
         // The name of the current running binary. Normally `bat` but sometimes
         // `batcat` for compatibility reasons.
-        let current_bin = env::args_os()
-            .next();
-        
+        let current_bin = env::args_os().next();
+
         // Check if the current running binary is set to be our pager.
         let is_current_bin_pager = current_bin
             .map(|arg0| Path::new(&arg0).file_stem() == pager_bin)
             .unwrap_or(false);
-        
+
         if is_current_bin_pager {
             PagerKind::Bat
-        }
-        else {
+        } else {
             match pager_bin.map(|s| s.to_string_lossy()).as_deref() {
                 Some("less") => PagerKind::Less,
                 Some("more") => PagerKind::More,
                 Some("most") => PagerKind::Most,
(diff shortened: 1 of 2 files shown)


src/pager.rs
```diff
@@ -40,16 +40,24 @@ impl PagerKind {
...
 
-        match Path::new(bin)
-            .file_stem()
-            .map(|s| s.to_string_lossy())
-            .as_deref()
-        {
+        let file_stem = Path::new(bin).file_stem();
+
+        match file_stem.map(|s| s.to_string_lossy()).as_deref() {
             Some("bat") => PagerKind::Bat,
```
Enselic, line 46: I think we can remove this case then, and handle it generically? That would also allow us to piggy-back on existing tests so that we don't have to write new ones.

src/pager.rs
```diff
@@ -40,16 +40,24 @@ impl PagerKind {
...
             Some("less") => PagerKind::Less,
             Some("more") => PagerKind::More,
             Some("most") => PagerKind::Most,
-            _ => PagerKind::Unknown,
+            _ => {
+                if env::args_os()
```
Enselic, line 51: Sorry to nit-pick but I think this deserves its own variable with a clarifying comment. Something like:

```rust
// The bin name we run under. Normally `bat`, but can also be `batcat` for compatibility reasons.
let our_bin_name = env::args_os().next().
```

src/pager.rs
```diff
@@ -40,16 +40,24 @@ impl PagerKind {
...
+                    .nth(0)
+                    .map(|arg0| Path::new(&arg0).file_stem() == file_stem)
+                    .unwrap_or(false)
+                {
+                    PagerKind::Bat
```
Enselic, line 56: Just an idea, not sure if it works: Maybe we can match on `our_bin_name` earlier and keep the ` _ => PagerKind::Unknown,` line unchanged?

Enselic (changes requested):
Thanks for working on this! I added some comments.

src/pager.rs
johnmatthiggins: Totally agree, that `bat` case should already be covered by checking the binary name against the pager binary.

src/pager.rs
johnmatthiggins: Sounds like a good idea.

src/pager.rs
johnmatthiggins: Added it.

johnmatthiggins:
Made commits with suggested changes, ready for further review.

src/pager.rs
```diff
@@ -40,16 +40,27 @@ impl PagerKind {
...
+
+        if is_current_bin_pager {
+            PagerKind::Bat
+        } else {
+            match pager_bin.map(|s| s.to_string_lossy()).as_deref() {
+                Some("less") => PagerKind::Less,
+                Some("more") => PagerKind::More,
+                Some("most") => PagerKind::Most,
+                _ => PagerKind::Unknown,
+            }
```
Enselic, line 63: Probably worth simplifying to the following?

```suggestion
        match pager_bin.map(|s| s.to_string_lossy()).as_deref() {
            Some("less") => PagerKind::Less,
            Some("more") => PagerKind::More,
            Some("most") => PagerKind::Most,
            _ if is_current_bin_pager => PagerKind::Bat,
            _ => PagerKind::Unknown,
        }
```

Enselic (commented):
Thanks! Much clearer code. I proposed one possible further simplification. Can you also please add an entry to CHANGELOG.md? Like this:

```
## Bugfixes

- Prevent fork nightmare with `PAGER=batcat`. See #2235 (@johnmatthiggins)
```

When that is done I think I can set this PR as Approved!

johnmatthiggins:
Added to change log and altered match expression.

Enselic (approved):
I forgot to comment on the `arg0` name, so I pushed a commit for that. I hope you didn't mind.

This looks good to me now! Thanks a lot.

As far as I can tell our current regression tests cover this PR. For example, if we hardcode `is_current_bin_pager` to either `true` or `false`, tests start failing.

I didn't expect it to be, but just to make sure I ensured that startup time is not in practice affected, and it wasn't:

```
% hyperfine --export-markdown /dev/tty -L bat ./target/release/bat,bat-master-c8b1187 '{bat} -f --paging=always --theme="Monokai Extended" examples/simple.rs'
```

|:---|---:|---:|---:|---:|
| `./target/release/bat -f --paging=always --theme="Monokai Extended" examples/simple.rs` | 11.0 ± 1.1 | 10.1 | 21.2 | 1.00 |
| `bat-master-c8b1187 -f --paging=always --theme="Monokai Extended" examples/simple.rs` | 11.1 ± 1.1 | 10.1 | 21.9 | 1.01 ± 0.14 |

Enselic:
About time to merge this now I think.

PR #2236, opened 2022-07-04
Add `.mts` and `.cts` to TypeScript file extensions

kidonng:
These are [TypeScript counterparts](https://www.typescriptlang.org/docs/handbook/esm-node.html#new-file-extensions
) of Node.js `.mjs` and `.cjs` file extensions.

See also:

- https://github.com/github/linguist/pull/5905
- https://github.com/starship/starship/pull/3734


Enselic (changes requested):
Thank you for the contribution! Do you think we can upstream this change?

The file you changed is converted from a .tmLanguage, which makes maintenance a bit of a headache. I think we need to at least add some basic regression tests if we are to patch this file ourselves. Not sure what other maintainers think.

See https://github.com/sharkdp/bat/blob/master/doc/assets.md for some general pointers on adding tests.

sharkdp:
@kidonng Any update on this?

sharkdp:
The upstream repository seems to be unmaintained (https://github.com/microsoft/TypeScript-Sublime-Plugin/issues/786). So let's just patch this file and put a simply regression test in place.

sharkdp (approved):

Enselic (approved):

Issue #2246, opened 2022-07-18
Slight bug in fstab syntax definition?


H4CKY54CK:
At the risk of doing everything completely incorrectly, I have replaced the bug report template with something more concise and helpful. ~~I also did a quick search through these issues for "fstab" before making this bug report.~~ Correction, I looked through *open* issues for `fstab`. I just checked the closed ones, and it looks like this was actually noticed in the initial addition of the syntax definition, but may have just flew under the radar by accident. https://github.com/sharkdp/bat/pull/706#discussion_r339938877

In the syntax definition for `fstab`, it should also accept a `2` as a valid option for the 6th field. I only noticed because I was trying to use the syntax file in Sublime Text and saw the big red block indicating a syntax error. In the output of `bat /etc/fstab`, that `2` is just white, which *looked* correct. Here's a permalink to the line in question.

https://github.com/sharkdp/bat/blob/4583826a5bf456c305695ca85aff4716da872da6/assets/syntaxes/02_Extra/Fstab.sublime-syntax#L98

EDIT: I also meant to link to that original mention.

yuvalmo:
I would love to fix this issue. I noticed the same fix is needed for the `fstab_pass` field as well:
https://github.com/sharkdp/bat/blob/4583826a5bf456c305695ca85aff4716da872da6/assets/syntaxes/02_Extra/Fstab.sublime-syntax#L110
I also edited the fstab regression test so it covers all the possible range of numbers for these fields, and made sure it works correctly.

You can review the fix here: https://github.com/sharkdp/bat/pull/2250.
Please let me know if I missed anything.

keith-hall:
Thanks for your contribution @yuvalmo, looks good to me! :)

PR #2250, opened 2022-07-23
Color `fstab` dump and pass fields correctly (#2246)

yuvalmo:
Was missing the number 2 as a valid option in those fields.


keith-hall (approved):

sharkdp:
Thank you very much!

PR #2267, opened 2022-08-12
Bump MSRV to 1.56.1

Enselic:
This is what `hashbrown` did recently: https://github.com/rust-lang/hashbrown/pull/298

And it causes CI failures for us. Examples:
* https://github.com/sharkdp/bat/pull/2266
* https://github.com/sharkdp/bat/pull/2265

The path of least resistance for us is to also bump MSRV, so let's do that. 1.56.1 was released [9 months ago](https://blog.rust-lang.org/2021/11/01/Rust-1.56.1.html).

There were some discussion regarding the bump which is an interesting read, see https://github.com/rust-lang/hashbrown/issues/310.

I see no significant problems with `bat` bumping to 1.56.1. Does anyone else?


Enselic:
Although, this bump is only triggered by dependabot bumping dependencies. If we don't bother with bumping dependencies eagerly, we don't need to bump MSRV now.

Instead, we could start bumping dependencies on a as-needed basis, i.e. pull rather than push. If someone wants something from a newer dependency, they can simply create a PR for it.

Although for syntaxes I do think it makes more sense to get regular pings by dependabot when there are updates. But we could manually turn them off for dependencies if we wanted.

Just some thoughts.

Enselic:
Another option could be to change our MSRV strategy to "last 3 stable releases", and adjust CI accordingly. Then we would not have to manually bump MSRV because of CI failures every now and then. Assuming all of our deps supports the last three stable versions of course, but I'm pretty confident that they do.

sharkdp:
> I see no significant problems with `bat` bumping to 1.56.1. Does anyone else?

No, let's go ahead :+1:

sharkdp:
> Although, this bump is only triggered by dependabot bumping dependencies. If we don't bother with bumping dependencies eagerly, we don't need to bump MSRV now.

True. But then outdated dependencies are also often causing problems. Not just because of missing bugfixes, but also with distributions that are packaging dependencies independently, for example.

> Instead, we could start bumping dependencies on a as-needed basis, i.e. pull rather than push. If someone wants something from a newer dependency, they can simply create a PR for it.

As suggested elsewhere, another option would be to only bump quarterly? or twice a year?

> Although for syntaxes I do think it makes more sense to get regular pings by dependabot when there are updates. But we could manually turn them off for library dependencies if we wanted.

We should be able to control that independently: https://github.com/sharkdp/bat/blob/3a93798e8b5189c10058c98a2ceab32bbb734045/.github/dependabot.yml#

sharkdp (approved):

Issue #651, opened 2019-09-04
bat.fish: There should be only one


adiabatic:
A while back, during  0.11.x, I thought `bat` didn't have any `fish` completion file, so I wrote one. When 0.12.0 came out, I updated the fish completion file to support the freshly-added `-L` option. This file is [in the `fish` repository](https://github.com/fish-shell/fish-shell/blob/master/share/completions/bat.fish), and will be released to the public whenever the fish maintainers get around to releasing a new version (in all likelihood, not too soon).

Today, when watching 0.12.1 come in via Homebrew, I noticed that Homebrew installed a `bat` completion of its own into `/usr/local/share/fish/vendor_completions.d`. Seems like I've been scooped.

I figure that the completion in the `fish` repo should be deleted in favor of the one provided with `bat` itself. I think [the one I wrote](https://github.com/fish-shell/fish-shell/blob/master/share/completions/bat.fish) has a number of improvements over the one that's bundled with `bat`:

- It's in nice-looking columns and is very easy to read, assuming you have a wide-enough monitor. (Multi-cursor editing is a godsend.)
- It moves the "which is the default" verbiage into the option-completion part of the completion scripts, saving horizontal space and making sure it doesn't get collapsed to a "…".
- It supports all the `bat cache` options, not just `build` and `clear`.

There are also a couple other differences that I feel I should point out:

- My options are sorted by the order seen in the `--help` text. The `bat.fish` that ships with `bat` sorts everything by `-l` value (long option).
- My `__bat_complete_languages_and_extensions` assumes that one can specify a language by typing out its fancy name and not just an extension. (Maybe this assumption is wrong.)
- Mine omits `-u`/`--unbuffered`, since it's just a POSIX-compliance thing that gets ignored by `bat` anyway.

Would `bat` be better if the completion currently sitting in the `fish` repository were incorporated into the `bat` distribution? Regardless of whether the answer is "yes" or "no", I ought to remove the one in the `fish` repository.

eth-p:
This is likely going to need to be discussed in depth. There's a few things that need to be taken into consideration here.

In regards to how the completions are distributed, I think having them distributed in the installed bat packages would definitely be ideal. This would allow for them to be relevant to user's installed version of bat rather than set in stone until the next fish release. My only lingering concern with this method is how it should be handled when bat and fish have different install prefixes (e.g. if bat is installed to `/usr/local` while fish is installed to `/usr`).

[...]

PR #2275, opened 2022-08-16
better fish completions

zgracem:
Back in 2020, the command-line completions for bat that had been shipping with fish were removed in favour of the completions file that currently ships with bat (see discussion in #651). Unfortunately, the [deprecated fish completions](https://github.com/fish-shell/fish-shell/blob/ea8a2b2/share/completions/bat.fish) had some nifty features that weren't carried over. This PR restores those features and expands upon them.

Improvements vs. the current completions:

- Less verbose descriptions → better columned layout.
  - Notably, default options (e.g. `always`, `auto`) are indicated in the option-completion text, not the description of the option itself.
- Supports all documented options, including those for `bat cache`.
- Fully (!) supports completing `bat --map-syntax <glob pattern>:<language syntax>`.
- Supports multiple comma-separated arguments to `bat --style`.
- Regular output options (e.g. `--style`) won't be suggested if a previous option (e.g. `--help`) overrides regular output/behaviour.
- The `{{PROJECT_EXECUTABLE}}` placeholders have been place-held by the locally-scoped `$bat`, for ease of reading and editing.
  - The names of the support functions (e.g. `__bat_complete_map_syntax`) are no longer parameterized, but "the other bat" [doesn't include shell completions](https://packages.debian.org/sid/x32/bacula-console-qt/filelist) as far as I can tell, so there shouldn't be any collisions.

Please let me know if you have any questions!

***

Before:

bat-before

After:

bat-after

`bat --italic-text=` before:

bat-italic-before

`bat --italic-text=` after:

bat-italic-after

`bat --style=` before:

bat-style-before

`bat --style=` after:

bat-style-after

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 ## Other
 
 - Relaxed glibc requirements on amd64, see #2106 and #2194 (@sharkdp)
+- Improved fish completions. See #2275 (@zgracem)
 
 ## Syntaxes
 
--- a/assets/completions/bat.fish.in
+++ b/assets/completions/bat.fish.in
@@ -100,6 +100,14 @@ set -l color_opts '
 set -l decorations_opts $color_opts
 set -l paging_opts $color_opts
 
+# Include some examples so we can indicate the default.
+set -l pager_opts '
+    less\tdefault
+    less\ -FR\t
+    more\t
+    vimpager\t
+'
+
 set -l italic_text_opts '
     always\t
     never\tdefault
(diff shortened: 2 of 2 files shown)


Enselic:
Looks reasonable, so I am tempted to blindly merge this.

Ideally, all important aspects of fish completions should have automated tests, because manually testing stuff is not manageable in the long term. Until we have tests, I think blindly merging stuff that looks reasonable is not unreasonable.

I'll leave this comment here for a while to give other maintainers a chance to object if they want.

@keith-hall You seemed to enjoy the change, so what do you think about blindly merging?

We should add an entry to CHANGELOG.md though.

keith-hall:
I had the same thoughts regarding automated tests, but this change is described well enough (and looks manually tested enough by the author) that I'd be okay with blindly merging it. Worst case outcome, if something breaks, it's only fish completions ;)

Enselic:
Sounds good, thanks for the input!

@zgracem Can you add an entry to CHANGELOG.md please? After that we can merge this PR

sharkdp (approved):

zgracem:
> Can you add an entry to CHANGELOG.md please?

All done!

Enselic:
Thank you!

Issue #2253, opened 2022-07-31
Add clp to alternatives.md


jpe90:
Hello! I wrote a similar program called [clp](https://github.com/jpe90/clp). Would it be possible to add to alternatives.md? The diff below fills out a new column for clp but leaves placeholders for execution time.

I can re-run the benchmarks on my machine and submit a PR, or a maintainer for this project could run it on theirs. 

If you prefer the latter, could you build clp with luajit? ([build instructions available here](https://github.com/jpe90/clp#building-from-source))

```Diff
diff --git a/doc/alternatives.md b/doc/alternatives.md
index 665398f..956d849 100644
--- a/doc/alternatives.md
+++ b/doc/alternatives.md
@@ -4,17 +4,17 @@ The following table tries to give an overview *from `bat`s perspective*, i.e. we
 categories which are relevant for `bat`. Some of these projects have completely different goals and
 if you are not looking for a program like `bat`, this comparison might not be for you.
 
-|                                              | bat                                                                 | [pygments](http://pygments.org/) | [highlight](http://www.andre-simon.de/doku/highlight/highlight.php) | [ccat](https://github.com/jingweno/ccat) | [source-highlight](https://www.gnu.org/software/src-highlite/) | [hicat](https://github.com/rstacruz/hicat)          | [coderay](https://github.com/rubychan/coderay)      | [rouge](https://github.com/jneen/rouge)             |
-|----------------------------------------------|---------------------------------------------------------------------|----------------------------------|---------------------------------------------------------------------|------------------------------------------|----------------------------------------------------------------|-----------------------------------------------------|-----------------------------------------------------|-----------------------------------------------------|
-| Drop-in `cat` replacement                    | :heavy_check_mark: [*](https://github.com/sharkdp/bat/issues/134)   | :x:                              | :x:                                                                 | (:heavy_check_mark:)                     | :x:                                                            | :x: [*](https://github.com/rstacruz/hicat/issues/6) | :x:                                                 | :x:                                                 |
-| Git integration                              | :heavy_check_mark:                                                  | :x:                              | :x:                                                                 | :x:                                      | :x:                                                            | :x:                                                 | :x:                                                 | :x:                                                 |
-| Automatic paging                             | :heavy_check_mark:                                                  | :x:                              | :x:                                                                 | :x:                                      | :x:                                                            | :heavy_check_mark:                                  | :x:                                                 | :x:                                                 |
[...]

PR #2276, opened 2022-08-17
update alternatives.md

jpe90:
Adds [clp](https://github.com/jpe90/clp) to the comparison table.

Full script results that I used to fill in the table available here: https://paste.sr.ht/~eskin/2229ccc7d39e5d6128a3c014a30f0572cadc375a

I wasn't sure which 370KB XML file was used, so I used a 957KB file I found from Wikipedia.

Addresses  #2253


sharkdp:
Thank you for the contribution and congratulations on your project.

I've been looking into parsing myself recently, and this looks like a very cool approach. It certainly looks like this has huge runtime advantages. In principle, it should also be more powerful in terms of *truly* supporting the syntax of languages.

If you find the time, I'd be interested in your opinion:
- Do you think the current highlighting in `clp` is on-par with what `bat` provides (based on Sublime text syntaxes)?
- Are there any downsides to your approach (as compared to using regex-based syntaxes, like bat does)?
- Are there any alternatives to LPEG if one chooses to go down the route of a full parsing approach? Maybe tree-sitter?
- Do you think it would be a route worth exploring for `bat`? Hypothetically, we could even go with a hybrid approach and use LPEG grammars for some languages and regex-based syntaxes as a fallback.

jpe90:
clp doesn't always match bat's syntax highlighting. There's some cases where highlight groups aren't identified correctly. I think it just comes down to the quality and completeness of the LPEG lexers. 

I don't think I'm the right person to speak intelligibly on the downsides of LPEG or whether it makes sense to adopt PEG lexers for bat; I'm not very knowledgeable about formal language theory. clp was just a lucky break- I saw how simple LPEG/Scintillua lexers were, I knew they could perform fairly accurate syntax highlighting since they're used in a few text editors, so I ran some benchmarks and built out clp when they showed promise. 

Tree-sitter, in my experience, tends to be slow on initial parses and fast to process iterative changes. I think that makes sense for text editors, but it probably doesn't help much for bat or clp, where you mostly care about the speed of the initial parse.

PR #2279, opened 2022-08-18
CI: Build MSRV without git feature enabled

Enselic:
git2-rs MSRV policy is to only support latest stable rust (see https://github.com/rust-lang/git2-rs#rust-version-requirements), so it does not make sense to run our MSRV tests with git enabled. Also see discussion in https://github.com/sharkdp/bat/pull/2258

Also note the following separate commit:
```
tests/tester: Do not build as tests

mod tester does not contains any tests, so do not build the module as
containing tests. Instead use the mod.rs approach described in
https://doc.rust-lang.org/book/ch11-03-test-organization.html#submodules-in-integration-tests.
```


Enselic:
I hear no objections, so let's merge! 🎉

PR #2299, opened 2022-09-02
refactor: Migrate to clap 3

### LINKED ISSUE #2056: Upgrade to clap 3
I think everyone agrees we need/should upgrade to clap 3 at some point. This ticket tracks that work.

I mostly create this ticket as a reminder for us to turn Dependabot updates back on for clap 3.x when we have done that work. I will turn off Dependabot updates for clap 3.x versions now.

epage:
This updates to clap v3 and resolves all deprecations as of right now.

I mostly relied on the tests.

Help output has two main changes that I noticed
- The sections are no "Args", "Options", "Subcommands"
- The text for `--version` is slightly different

I was confused as to why external subcommand support was enabled but I left it.

Some potential further changes
- Preserve parsed values (like numbers) rather than using strings
- Switch lists of possible values to implementing or deriving the `ValueEnum` trait on enums
- Switching to the derive

Fixes #2056

### DIFF
--- a/src/bin/bat/app.rs
+++ b/src/bin/bat/app.rs
@@ -84,7 +84,7 @@ impl App {
             Some("never") => PagingMode::Never,
             Some("auto") | None => {
                 // If we have -pp as an option when in auto mode, the pager should be disabled.
-                let extra_plain = self.matches.occurrences_of("plain") > 1;
+                let extra_plain = self.matches.get_count("plain") > 1;
                 if extra_plain || self.matches.is_present("no-paging") {
                     PagingMode::Never
                 } else if inputs.iter().any(Input::is_stdin) {
@@ -297,7 +297,7 @@ impl App {
                 HashSet::new()
             } else if matches.is_present("number") {
                 [StyleComponent::LineNumbers].iter().cloned().collect()
-            } else if matches.is_present("plain") {
+            } else if 0 < matches.get_count("plain") {
                 [StyleComponent::Plain].iter().cloned().collect()
             } else {
                 let env_style_components: Option<Vec<StyleComponent>> = env::var("BAT_STYLE")
--- a/src/bin/bat/app.rs
+++ b/src/bin/bat/app.rs
@@ -85,7 +85,7 @@ impl App {
             Some("auto") | None => {
                 // If we have -pp as an option when in auto mode, the pager should be disabled.
                 let extra_plain = self.matches.get_count("plain") > 1;
-                if extra_plain || self.matches.is_present("no-paging") {
+                if extra_plain || self.matches.get_flag("no-paging") {
                     PagingMode::Never
                 } else if inputs.iter().any(Input::is_stdin) {
                     // If we are reading from stdin, only enable paging if we write to an
@@ -153,13 +153,13 @@ impl App {
                 .get_one::<String>("language")
                 .map(|s| s.as_str())
                 .or_else(|| {
-                    if self.matches.is_present("show-all") {
+                    if self.matches.get_flag("show-all") {
                         Some("show-nonprintable")
                     } else {
                         None
                     }
                 }),
-            show_nonprintable: self.matches.is_present("show-all"),
+            show_nonprintable: self.matches.get_flag("show-all"),
             wrapping_mode: if self.interactive_output || maybe_term_width.is_some() {
                 match self.matches.get_one::<String>("wrap").map(|s| s.as_str()) {
                     Some("character") => WrappingMode::Character,
@@ -178,7 +178,7 @@ impl App {
                 // There's no point in wrapping when this is the case.
                 WrappingMode::NoWrapping(false)
             },
-            colored_output: self.matches.is_present("force-colorization")
+            colored_output: self.matches.get_flag("force-colorization")
                 || match self.matches.get_one::<String>("color").map(|s| s.as_str()) {
                     Some("always") => true,
                     Some("never") => false,
@@ -194,7 +194,7 @@ impl App {
                     .get_one::<String>("decorations")
                     .map(|s| s.as_str())
                     == Some("always")
-                || self.matches.is_present("force-colorization")),
+                || self.matches.get_flag("force-colorization")),
             tab_width: self
                 .matches
                 .get_one::<String>("tabs")
@@ -221,7 +221,7 @@ impl App {
                     }
                 })
                 .unwrap_or_else(|| String::from(HighlightingAssets::default_theme())),
-            visible_lines: match self.matches.is_present("diff") {
+            visible_lines: match self.matches.contains_id("diff") && self.matches.get_flag("diff") {
                 #[cfg(feature = "git")]
                 true => VisibleLines::DiffContext(
                     self.matches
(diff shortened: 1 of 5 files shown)


src/bin/bat/clap_app.rs
```diff
@@ -133,47 +135,48 @@ pub fn build_app(interactive_output: bool) -> ClapApp<'static, 'static> {
     {
         app = app
                 .arg(
-                    Arg::with_name("diff")
+                    Arg::new("diff")
```
epage, line 138: CI is failing because this arg is conditionally present but is referenced by args that are unconditionally present

epage:
Looks like its now ready

Enselic (approved):
This is great, thank you so much.

I skimmed over the code, and it looks very reasonable. Since you managed to make all regression tests pass, I honestly think we should just go ahead and merge this. I would be very surprised if there is not a single regression that has sneaked in, but finding it via CR and manual testing is simply too much work. It is better to take a chance. Then once we find the regressions, we fix them and add tests for them so they don't occur again.

It would be great if you could add an entry to CHANGELOG.md before we merge though. I think this is worth mentioning in CHANGELOG.

epage:
> It would be great if you could add an entry to CHANGELOG.md before we merge though. I think this is worth mentioning in CHANGELOG.

As this is mostly internal, how would you like that expressed in the changelog?

Enselic:
(ping: @sharkdp @keith-hall, see below)

Hmm, you're right, maybe this is mostly making bat developers happy, and regular users don't care very much :) I'm fine with not mentioning it in CHANGELOG.md, on a second thought.

I will wait for other maintainers to chime in if they want though, so it might take a while before we actually merge this. But at the same time it is not good to wait too long, because the risk for merge conflicts is high if we wait...

@sharkdp @keith-hall Maybe would be good if you could put this PR high in your prio list?

src/bin/bat/app.rs
```diff
@@ -125,36 +125,43 @@ impl App {
...
-                    Some("show-nonprintable")
-                } else {
-                    None
-                }
-            }),
-            show_nonprintable: self.matches.is_present("show-all"),
+            language: self
+                .matches
+                .get_one::<String>("language")
+                .map(|s| s.as_str())
```
sharkdp, line 154: Just curious: this seems quite a bit more verbose than what we had before. Also, it looks like it now requires some dynamically allocated String, whereas before we directly got access to a `&str` or `&OsStr`. What were the reasons for this API change?

sharkdp (approved):
Thank you very much for the support! (and as mentioned before: thank you for your work on `clap`!).

The code changes look good, except for the `self.matches.get_one::<String>(…).map(|s| s.as_str())` dance that we now have to go through multiple times (see inline comment)

src/bin/bat/app.rs
epage: > What were the reasons for this API change?

Clap now allows you to store the parsed value (PathBuf, usize, etc) directly in the `ArgMatches`.  `get_one::<T>()` accepts what the type is stored inside of `ArgMatches`.

I was a bit concerned with the extra ceremony around this but we decided it was worth it
- For the extra features this enabled
  - Avoids double parsing
  - Bypasses problems we had with the `value_of_t` API we previously maintained
  - We're able to integrate flags directly into clap without a lot of back and forth (previously clap treated them special by not storing any value but separately tracking the presence and the number of occurrences, now its directly supported allowing integration with other clap features)
- It was limited to the Builder API where ease of use is a lower priority compared to the Derive API

> If we need to pass a type argument to .get_one, is it possible to directly .get_one::<&str>?

I considered playing with trait to allow mapping `&str` to `String` but I was concerned with usability issues
- If the user specified the type (`arg.value_parser(value_parser!(usize)))`, then that type wouldn't match with what is used at the access
- When we report an assert on type mismatches, we would be suggesting to the user the owned type and not the borrowed type

> If not, can .map(|s| s.as_str()) be replaced by .as_deref()?

`&String` does not deref to `&str` but to itself

```
error[E0308]: mismatched types
   --> src/bin/bat/main.rs:231:43
    |
231 |         bat::config::get_pager_executable(app.matches.get_one::<String>("pager").as_deref())
    |         --------------------------------- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `str`, found struct `std::string::String`
    |         |
    |         arguments to this function are incorrect
    |
    = note: expected enum `Option<&str>`
               found enum `Option<&std::string::String>`
note: function defined here
   --> /home/epage/src/personal/bat/src/config.rs:92:8
    |
92  | pub fn get_pager_executable(config_pager: Option<&str>) -> Option<String> {
    |        ^^^^^^^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0308`.
```

sharkdp: Thank you for the detailed explanation. Not to digress too much, but now I'm interested :smile:. What about the additional allocations? I imagine it was previously possible to directly access the (Os)Str-slices inside the environment (`std::env::args`) which are already in memory when the program is started. If every argument is now converted to a `Os(Str)`**`ing`**, wouldn't that lead to a performance impact for programs with *many* arguments (e.g. when doing something like `find … | xargs <myprog>`)? Or is the time for those N allocations insignificant compared to everything else (CLI parsing etc.)

epage: > What about the additional allocations? I imagine it was previously possible to directly access the (Os)Str-slices inside the environment (std::env::args) which are already in memory when the program is started. 

`std::env::args_os` returns newly allocated `OsString`s.  clap does take ownership of those but, at the moment, we do not move those directly into `ArgMatches` but will do a clone anyways.  Things get complicate with ownership when you support moving items out of the input, adding items back in, dealing with short flags, and the extra ownership complications caused by needing to use `os_str_bytes`.

I did leave the `ValueParser` trait open to us moving `OsString` into `ArgMatches` if we can do that at a later time.  We can then avoid an extra allocation for `OsString`, `String`, and `PathBuf`.

On a related note, there is an extra source of allocations if you use `ArgGroup`.  An argument inside of an `ArgGroup` will have all of its values cloned into the `ArgGroup`s spot in `ArgMatches`.  I don't think I've even seen this documented anywhere.  I'm changing the behavior to instead put the argument's ID in the `ArgGroup`s match.  This makes it independent of how many values `xargs` passed in, still allows you to look up the value, and also let's you see which group was selected.

epage: iirc gumpdrop does not do any extra allocations than what `std::env::args_os` forces on it and [it still has similar performance to clap in the xargs case](https://github.com/rosetta-rs/argparse-rosetta-rs)

sharkdp: > std::env::args_os returns newly allocated OsStrings

Oh, my bad. For some reason, I thought that Rust provided direct read-only "`OsStr`" access to the underlying "`const char* argv[]`", like in C/C++, but that's not the case. For some more context on this, see https://github.com/rust-lang/rfcs/issues/862 or https://github.com/rust-lang/rust/issues/47164.

>  Things get complicate with ownership when you support moving items out of the input, adding items back in, dealing with short flags, and the extra ownership complications caused by needing to use os_str_bytes.
> I did leave the ValueParser trait open to us moving OsString into ArgMatches if we can do that at a later time. We can then avoid an extra allocation for OsString, String, and PathBuf.

I see, thank you for the detailed explanation!

Issue #2239, opened 2022-07-07
Add single character argument for --wrap never


jguhlin:
I primarily use less -S to view everything due to long lines (bioinformatics, mostly). Bat is incredibly fast, but I'd appreciate a single character argument to substitute for --wrap never (such as -S to match less functionality, or something else).

Cheers

sharkdp:
I'm usually careful when providing new short flags, but this seems reasonable, given that it would be in line with `less`s `-S` option. Maybe should also support `--chop-long-lines`, which is the long-form of `-S`.

johnmatthiggins:
I would like to work on this one. What should the single letter abbreviation be for setting wrapping to off. Should it be implemented with -S or would a different letter work better?

sharkdp:
`-S` seems good.

johnmatthiggins:
Sounds good. I made a pull request (#2309) for adding `-S` and the `--chop-long-lines` option.

PR #2309, opened 2022-09-10
Added -S flag for truncating long lines

johnmatthiggins:
Implementation of feature mentioned in issue #2239. Added both `--chop-long-lines` and `-S` as flags.


Enselic (changes requested):
Thanks! Please also add one or two basic regression tests for this change. And also add an entry to CHANGELOG.md please.

johnmatthiggins:
I added some tests and added the change to the `CHANGELOG.md` file.

johnmatthiggins:
Oops it looks like I had reversed logic and set `-S` and `--chop-long-lines` as aliases for `wrap=character` instead of `wrap=never`. This has now been remedied. Also for the sake of clarity, I added four tests to the project and I'll explain each of them.
* The first test verifies that no line wrapping happens when the `wrap=never` command line flag is present.
* The second test verifies that line wrapping happens when the `wrap=auto` command line flag is present.
* The third and fourth tests verify that no line wrapping happens when  `-S` or `--chop-long-lines` is specified.

I added some tests that cover historical functionality not implemented by this PR because those cases were not yet covered and I wanted to make sure that my changes didn't alter existing functionality.

tests/examples/80-columns.txt
```diff
@@ -0,0 +1 @@
+abcdefghigklmnopqrstuvxyzabcdefghigklmnopqrstuvxyzabcdefghigklmnopqrstuvxyzabcdefghigklmnopqrstuvxyz
```
Enselic, line 1: Shouldn't this file be called `100-columns.txt`? Another proposal: `long-single-line.txt`.

tests/integration_tests.rs
```diff
@@ -1499,6 +1499,80 @@ fn ignored_suffix_arg() {
...
+        .stdout(expected)
+        .stderr("");
+}
+
+#[test]
+fn no_wrapping_with_chop_long_lines() {
+    let expected =
+        "abcdefghigklmnopqrstuvxyzabcdefghigklmnopqrstuvxyzabcdefghigklmnopqrstuvxyzabcdefghigklmnopqrstuvxyz\n";
+
+    bat()
```
Enselic, line 1563: There is a substantial amount of copy-paste here. Would be nice to introduce a helper function that takes as parameter the only thing that varies among the tests (expected output and the wrapping arg).

Enselic (commented):
Looks good! Just a couple of nit-picks.

tests/examples/80-columns.txt
johnmatthiggins: Oh, I like `long-single-line.txt`, I'll go with that.

tests/integration_tests.rs
johnmatthiggins: Added a helper function.

johnmatthiggins: I added the helper function in the `tests/integration_tests.rs` file, but is this the correct procedure? I don't see any other helper functions in this file, so I figured I should ask.

src/bin/bat/clap_app.rs
```diff
@@ -201,6 +201,13 @@ pub fn build_app(interactive_output: bool) -> Command<'static> {
                            The '--terminal-width' option can be used in addition to \
                            control the output width."),
         )
+        .arg(
+            Arg::new("chop-long-lines")
+                .long("chop-long-lines")
+                .short('S')
+                .takes_value(false)
+                .help("Truncate all lines longer than screen width. Alias for '--wrap=character'."),
```
Enselic, line 209:
```suggestion
                .help("Truncate all lines longer than screen width. Alias for '--wrap=never'."),
```

Enselic (commented):
The test code looks much better now, thanks! I found one more comment to make

tests/integration_tests.rs
Enselic: Yep it was correct to put the helper next to the functions that need it

I don't expect any other code to ever need that helper, so it makes sense to put it where you put it

Enselic (approved):
Looks mergeable to me, thanks!

sharkdp:
Thank you very much

PR #2330, opened 2022-09-29
Make `-pp` override `--paging` and vice versa when passed as a later argument.

vishal1132:
The other part of #2128 to make `-pp` override `--paging` and `--paging` override `-pp` whichever is passed as a later argument wins.


sharkdp:
Thank you. It would be great if we could add a regression test for this.

vishal1132:
that makes sense @sharkdp, have a holiday here tomorrow. Will add it then.
Thanks for the reply.

Enselic:
Closing for now to keep PR inbox clean. I considered merging this without a regression test, but by my estimation, the use case is too complicated for there to not be a regression test.

Of course feel free to re-open this if work resumes on this PR.

PR #2365, opened 2022-10-09
Update CONTRIBUTING.md

Sylk:
Verbiage change, and slight markdown adjustments.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,7 +33,7 @@ section in the README.
 
 Please consider opening a
 [feature request ticket](https://github.com/sharkdp/bat/issues/new?assignees=&labels=feature-request&template=feature_request.md)
-first in order to give us a chance to discuss the feature first.
+first in order to give us a chance to discuss the details and specifics of the potential new feature before you go and build it.
 
 
 ## Adding new syntaxes/languages or themes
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -50,12 +50,12 @@ first.
 
 ## Regression tests
 
-You are strongly encouraged to add regression tests. Regression tests are great,
+You are **strongly encouraged** to add regression tests. Regression tests are great,
 not least because they:
 
-* ensure that your contribution will never completely stop working,
+* Ensure that your contribution will never completely stop working.
```
keith-hall, line 56: I personally believe that starting the bullet points with a lowercase letter makes sense, as it's like a continuation of the sentence - if you take the punctuation/markdown formatting away, you get "not least because they ensure[...]" which is easier on the eyes/more natural than "not least because they Ensure[...]"

keith-hall (changes requested):

CONTRIBUTING.md
Sylk: I see your point. I didn't know which was dramatically correct but I went and did a quick lookup and found this. The points you make are valid and grammatically correct. Neat, the more you know I guess!

https://www.imperial.ac.uk/brand-style-guide/writing/grammar/punctuation/bullet-points/

sharkdp (approved):

sharkdp:
Thank you

PR #2366, opened 2022-10-12
Add VimHelp.sublime-syntax and its test

Freed-Wu:
screen-2022-10-30-13-22-55

### DIFF
--- a/assets/syntaxes/02_Extra/VimHelp.sublime-syntax
+++ b/assets/syntaxes/02_Extra/VimHelp.sublime-syntax
@@ -1,0 +1,104 @@
+%YAML 1.2
+---
+# http://www.sublimetext.com/docs/syntax.html
+scope: source.vimhelp
+file_extensions:
+  # shortname
+  - vimhelp
+
+# $VIMRUNTIME/syntax/help.vim
+contexts:
+  main:
+    - match: '(?<=^\s*)(vim?|ex):\s*([a-z]+(=[^\s:]+)?(\s+|:))+'
+      scope: comment.line.modeline.vimhelp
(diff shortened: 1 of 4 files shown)


sharkdp:
Please also follow the usual practices here (see comments in your other PR)

sharkdp:
Also, I'm not sure if this satisfies our policies for syntax inclusion: https://github.com/sharkdp/bat/blob/master/doc/assets.md#Criteria-for-inclusion-of-new-syntaxes

Freed-Wu:
> More than 10,000 downloads at Package Control

I think sublime users have no any need to view a vimhelp file :smile:

sharkdp:
Fair point, but I also fail to see why someone would want to view a vim help text in bat, and not in vim directly

Freed-Wu:
less is a pager and vim is an editor. If users just want to take a look at a file, they can just less it.

sharkdp:
Yeah but... if I feel the need to look at vims help texts, I'm usually inside vim - no?

Freed-Wu:
> if I feel the need to look at vims help texts, I'm usually already inside vim

Maybe any one who need to look at vimhelp install vim in their machine and can
use vim to look at helpfile, however, it only be convenient for those helpfiles
have been installed. For a vim plugin users don't ensure should they install,
before installation, use less to see it is also a reasonable situation.

BTW, I think vimhelp for vim is like orgmode for emacs, and bat have supported
orgmode, so perhaps vimhelp can be supported, too.

sharkdp:
Ok :smile: 

Thank you for your contribution. In the future, please open an issue first before you invest a lot of time in things that might not be accepted into `bat` in the end.

Issue #2262, opened 2022-08-06
Detect binary file takes too long


balroggg:
**What steps will reproduce the bug?**

1. Run bat file.ova

**What happens?**

Take too long to output something

**What did you expect to happen instead?**

Print that it is a binary file

**How did you install `bat`?**

pacman, Archlinux

**bat version and environment**

bat --diagnostic
#### Software version

bat 0.21.0 (405e5f74)

#### Operating system

Linux 5.18.16-arch1-1

#### Command-line

```bash
bat --diagnostic 
```

#### Environment variables

```bash
SHELL=/bin/bash
PAGER=<not set>
LESS=<not set>
LANG=en_US.UTF-8
LC_ALL=<not set>
BAT_PAGER=<not set>
BAT_CACHE_PATH=<not set>
BAT_CONFIG_PATH=<not set>
BAT_OPTS=<not set>
BAT_STYLE=<not set>
BAT_TABS=<not set>
BAT_THEME=<not set>
XDG_CONFIG_HOME=<not set>
XDG_CACHE_HOME=<not set>
COLORTERM=truecolor
NO_COLOR=<not set>
MANPAGER=<not set>
```

#### Config file

```
# This is `bat`s configuration file. Each line either contains a comment or
# a command-line option that you want to pass to `bat` by default. You can
# run `bat --help` to get a list of all possible configuration options.

# Specify desired highlighting theme (e.g. "TwoDark"). Run `bat --list-themes`
# for a list of all available themes
#--theme="OneHalfLight"

# Enable this to use italic text on the terminal. This is not supported on all
# terminal emulators (like tmux, by default):
#--italic-text=always

# Uncomment the following line to disable automatic paging:
#--paging=never

# Uncomment the following line if you are using less version >= 551 and want to
# enable mouse scrolling support in `bat` when running inside tmux. This might
# disable text selection, unless you press shift.
#--pager="--RAW-CONTROL-CHARS --quit-if-one-screen --mouse"

# Syntax mappings: map a certain filename pattern to a language.
#   Example 1: use the C++ syntax for .ino files
#   Example 2: Use ".gitignore"-style highlighting for ".ignore" files
#--map-syntax "*.ino:C++"
#--map-syntax ".ignore:Git Ignore"
```

#### Custom assets metadata

Could not read contents of '/home/balrog/.cache/bat/metadata.yaml': No such file or directory (os error 2).

#### Custom assets

'/home/balrog/.cache/bat' not found

#### Compile time information

[...]

PR #2369, opened 2022-10-19
Speed up binary file read

lshamis:
Test setup:
```
$ truncate -s 10G example_file
```

Before change:
```
$ time bat example_file
───────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: example_file   <BINARY>
───────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

real	0m12.840s
user	0m3.923s
sys	0m8.911s
```
After update
```
$ time bat example_file
───────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: example_file   <BINARY>
───────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

real	0m0.037s
user	0m0.036s
sys	0m0.002s
```

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -272,7 +272,7 @@ impl<'a> InputReader<'a> {
 
         if content_type == Some(ContentType::UTF_16LE) {
             reader.read_until(0x00, &mut first_line).ok();
-        } else if content_type != Some(ContentType::BINARY) {
+        } else if content_type != Some(ContentType::BINARY) && first_line.last() != Some(&b'\n') {
             reader.read_until(b'\n', &mut first_line).ok();
         }
 
--- a/src/input.rs
+++ b/src/input.rs
@@ -258,7 +258,11 @@ pub(crate) struct InputReader<'a> {
 impl<'a> InputReader<'a> {
     fn new<R: BufRead + 'a>(mut reader: R) -> InputReader<'a> {
         let mut first_line = vec![];
-        reader.read_until(b'\n', &mut first_line).ok();
+        reader
+            .by_ref()
+            .take(1024)
+            .read_until(b'\n', &mut first_line)
+            .ok();
 
         let content_type = if first_line.is_empty() {
             None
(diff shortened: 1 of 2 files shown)


lshamis:
This should resolve https://github.com/sharkdp/bat/issues/2262

src/input.rs
```diff
@@ -258,7 +258,11 @@ pub(crate) struct InputReader<'a> {
 impl<'a> InputReader<'a> {
     fn new<R: BufRead + 'a>(mut reader: R) -> InputReader<'a> {
         let mut first_line = vec![];
-        reader.read_until(b'\n', &mut first_line).ok();
+        reader
+            .by_ref()
+            .take(1024)
```
keith-hall, line 263: forgive my ignorance, but does this not mean that if a text file with a first line longer than 1024 chars is read, that first line would get truncated at 1024 chars?

lshamis: Only for the purposes of checking whether the file is binary. The check down below:
```
else if content_type != Some(ContentType::BINARY) && first_line.last() != Some(&b'\n') {
    reader.read_until(b'\n', &mut first_line).ok();
}
```
will complete the first line if the content is not binary. If the content is binary, bat wont show it anyway.

Actually, maybe I need to also check for `self.config.show_nonprintable`.

lshamis:
I'm not sure how I broke `no_args_doesnt_break` on some architectures.

sharkdp:
> I'm not sure how I broke `no_args_doesnt_break` on some architectures.

It might be broken on *all* platforms where the test actually runs? Not that it does not run on Windows. And we don't execute unit tests on all platforms...

lshamis:
Sorry, I don't have time right now to debug. I close this PR for now and will try to complete it in the future. Thanks.

sharkdp:
No worries. And no rush. I suggest we simply keep it open. Maybe someone else wants to look into it

Enselic:
I'm closing this PR for now to keep the PR inbox clean.

I have moved/copied the "help wanted" label to the issue itself: https://github.com/sharkdp/bat/issues/2262

Of course feel free to re-open this PR if work resumes on it!

PR #2375, opened 2022-10-22
Add Todo.txt syntax

BANOnotIT:
Todo.txt has [11k installs](https://packagecontrol.io/packages/TodoTxt%20Syntax) on PackageControl


BANOnotIT:
@sharkdp I need help with debugging syntax since one part of it doesn't work.

This is what I have in sublime:
image

And this is what I get with `bat`:
image


The problem is that `bat` doesn't understand `key:pair` somehow.

`key:pair` syntax have this definition:
```yaml
    - match: (\s+[^\s:]+:[^\s:]+)+\s*$
      comment: Custom attributes
      scope: variable.annotation.todotxt.attribute
```

Do you know what can be a problem? RegEx doesn't seem to be hard

keith-hall:
I think the answer is that `Monokai Extended.tmTheme` doesn't target the `variable.annotation` scope, so that text stays with the color defined for `entity[.task]`.

BANOnotIT:
@keith-hall looks like `entity` [is defined](https://github.com/jonschlinkert/sublime-monokai-extended/blob/0ca4e75291515c4d47e2d455e598e03e0dc53745/Monokai%20Extended.tmTheme#L338-L350).

BANOnotIT:
@keith-hall so what if I create a patch that changes `variable.annotation` to `variable.other`?

BANOnotIT:
I created a patch for theme and this is what it looks like
image

sharkdp:
Thank you!

PR #2399, opened 2022-11-12
Disable completion of `cache` subcommand

cyqsimon:
See https://github.com/sharkdp/bat/issues/2085#issuecomment-1271646802.

TLDR: autocompleting the `cache` subcommand causes usability issues. This PR disables it.

I am not terribly familiar with completion scripts TBH. If you are more competent please help check my edits, thanks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Other
 
 - Various bash completion improvements, see #2310 (@scop)
+- Disable completion of `cache` subcommand, see #2399 (@cyqsimon)
 
 ## Syntaxes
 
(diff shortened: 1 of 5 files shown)


Enselic (approved):
I'm no shell completion expert, but these changes look good to me. And I am strongly in favor of the goal of the change.

I'll set this as Approved. I'll give some more time for people to look at this. If no more input comes in, I'll merge this.

I think it is better to merge it so that problems can be found, if any, rather than letting the PR sit untouched for much longer.

Ideally we would have regression tests for completion scripts, but life is not always ideal.

Enselic:
@cyqsimon Can you also add an entry to CHANGELOG.md please?

sharkdp:
I agree. Thank you @cyqsimon. I would appreciate if we could remove the code in question, instead of commenting it. We would still keep the comments explaining why we're not completing `cache`, of course.

cyqsimon:
@Enselic @sharkdp all done.

PR #2406, opened 2022-11-19
Make `PrettyPrinter::header` an alias for `header_filename`

### LINKED ISSUE #2378: `PrettyPrinter::header` does not display a header
### **What steps will reproduce the bug?**

Compile and run the following program:

```rust
fn main() {
    bat::PrettyPrinter::new()
        .input_file("Cargo.toml")
        .header(true)
        .print()
        .unwrap();
}
```

### **What happens?**

The file is printed without a header.

### **What did you expect to happen instead?**

The file would be printed with a header containing the filename.

### **How did you install `bat`?**

Via cargo:

```toml
[dependencies]
bat = "0.22.1"
```

---

It looks like the issue is due to the fact that there exist both "header filename" and "header filesize" style components. `PrettyPrinter::header` adds `StyleComponent::Header` to the list of components:

https://github.com/sharkdp/bat/blob/f215d07366dac50aa7068e9c3f8d5f46b09b4c1a/src/pretty_printer.rs#L279-L281

But when the components are checked, `bat` only looks for `StyleComponent::HeaderFilename` or `StyleComponent::HeaderFilesize`:

https://github.com/sharkdp/bat/blob/f215d07366dac50aa7068e9c3f8d5f46b09b4c1a/src/style.rs#L109-L119

I'm not sure what the best way to tackle this would be. Should we also check for `StyleComponent::Header` in `StyleComponents::header`? Should the `PrettyPrinter::header` function push in `StyleComponent::HeaderFilename`? Should `PrettyPrinter` expose a `header_filename(bool)` function?

cstyles:
Fixes #2378

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,8 @@
 
 ## `bat` as a library
 
+- `PrettyPrinter::header` correctly displays a header with the filename, see #2378 and #2406 (@cstyles)
+
 
 # v0.22.1
 
--- a/src/pretty_printer.rs
+++ b/src/pretty_printer.rs
@@ -19,7 +19,7 @@ use crate::paging::PagingMode;
 
 #[derive(Default)]
 struct ActiveStyleComponents {
-    header: bool,
+    header_filename: bool,
     #[cfg(feature = "git")]
     vcs_modification_markers: bool,
     grid: bool,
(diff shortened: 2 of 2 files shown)


Enselic:
Thank you for working on this :pray: 

Would be great if you also could add a regression test for the fix. That also makes the PR much easier to review.

h3ndrk:
I would love to have this! What is the state? Thanks!

sharkdp:
I resolved conflicts, reviewed and tested this. Will merge it without a regression test.

PR #2430, opened 2022-12-30
Require changes to `-h` and `--help` to be blessed

Enselic:
From now on (if this PR is merged), any changes to the help texts will be visible in PR diffs, which will make it very easy to review, and very hard to accidentally miss changes to help texts.

If a contributor makes a change to help texts, the `cargo test` failure text they will see contains instructions on how to update the blessed help texts:

    error: expect test failed
       --> ../doc/long-help.txt

    You can update all `expect!` tests by running:

        env UPDATE_EXPECT=1 cargo test

In short, to update blessed help texts, one simply does

    UPDATE_EXPECT=1 cargo test

Do not run the tests if the `git` feature is missing, since then `--diff` will be missing from `--help`. And do not run the tests on Windows, because then the help text will contain the term `.exe`.

Move `man` page step to after `cargo doc` step in CI so that the man page exists when we look for it.

(The [`expect-test`](https://crates.io/crates/expect-test) crate is maintained under the `rust-analyzer` organization, and I've been using it and this help text setup in `cargo-public-api` for a while, and it has worked great so far.)

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -146,18 +146,6 @@ jobs:
         toolchain: stable
         default: true
         profile: minimal
-    - name: Print -h
-      uses: actions-rs/cargo@v1
-      with:
-        command: run
-        args: --locked -- -h
-    - name: Print --help
-      uses: actions-rs/cargo@v1
-      with:
-        command: run
-        args: --locked -- --help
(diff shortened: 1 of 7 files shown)


sharkdp (approved):
Cool! Thank you.

PR #2442, opened 2023-01-05
Directly read macOS theme preference instead of using defaults

BlackHoleFox:
Follows on https://github.com/sharkdp/bat/pull/2197 by improving the implementation of `macos_dark_mode_active`.

For reference there are a lot of ways to read preferences on macOS. Here's the 4 I know of:
1. `defaults` CLI
2. NSUserDefaults via Objective-C
3. CFPreferences via CoreFoundation (C/C++)
4. Directly reading and parsing the file from disk

Out of all of these, `defaults` is the slowest because it involves spawning and executing a whole separate executable to read one value. `defaults` is implemented via 2, which is implemented via 3 at the end. In order to speed this up, I just used option 4 as `bat`'s needs are fairly simple. Using `CFPreferences` would have also worked but the use of `#![deny(unsafe_code)]` indicated that would have been in conflict with the crate's goals since calling C APIs requires `unsafe` by nature. 

If `bat` wanted to do this 100% as-intended by Apple, and possibly take advantage of any OS-level caching, `CFPreferences.framework` would be the way. If you're good with the `unsafe` required, happy to swap out the implementation. Though I don't ever really expect the file reading method to break.

So this new implementation reads the hardcoded settings path and directly parses the plist data with the `plist` crate (already in the dependency tree from another crate, which was nice). As a result, in a release build, reading the theme is now 12x faster and pretty much imperceptible. The lowest time I saw with the old implementation was 4-5ms, while a release build call to `macos_dark_mode_active` now takes <500us on an M1 Max MacBook.

### DIFF
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -401,11 +401,21 @@ fn asset_from_cache<T: serde::de::DeserializeOwned>(
 
 #[cfg(target_os = "macos")]
 fn macos_dark_mode_active() -> bool {
-    let mut defaults_cmd = std::process::Command::new("defaults");
-    defaults_cmd.args(&["read", "-globalDomain", "AppleInterfaceStyle"]);
-    match defaults_cmd.output() {
-        Ok(output) => output.stdout == b"Dark\n",
-        Err(_) => true,
+    const PREFERENCES_FILE: &str = "Library/Preferences/.GlobalPreferences.plist";
+    const STYLE_KEY: &str = "AppleInterfaceStyle";
+
+    let preferences_file = dirs_next::home_dir()
+        .map(|home| home.join(PREFERENCES_FILE))
+        .expect("Could not get home directory");
+
+    match plist::Value::from_file(preferences_file).map(|file| file.into_dictionary()) {
+        Ok(Some(preferences)) => match preferences.get(STYLE_KEY).and_then(|val| val.as_string()) {
+            Some(value) => value == "Dark",
+            // If the key does not exist, then light theme is currently in use.
+            None => false,
+        },
+        // Unreachable, in theory. All macOS users have a home directory and preferences file setup.
+        Ok(None) | Err(_) => true,
     }
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 
 - Various bash completion improvements, see #2310 (@scop)
 - Disable completion of `cache` subcommand, see #2399 (@cyqsimon)
+- Signifigantly improve startup performance on macOS, see #2442 (@BlackHoleFox)
 
 ## Syntaxes
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -96,6 +96,7 @@ dependencies = [
  "nix",
  "once_cell",
  "path_abs",
+ "plist",
  "predicates",
  "regex",
  "semver",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -81,6 +81,10 @@ version = "4.0.2"
 optional = true
 features = ["wrap_help", "cargo"]
 
+[target.'cfg(target_os = "macos")'.dependencies]
+dirs-next = "2.0.0"
+plist = "1.3"
+
 [dev-dependencies]
 assert_cmd = "2.0.5"
 serial_test = "0.6.0"


src/assets.rs
```diff
@@ -401,11 +401,21 @@ fn asset_from_cache<T: serde::de::DeserializeOwned>(
 
 #[cfg(target_os = "macos")]
 fn macos_dark_mode_active() -> bool {
-    let mut defaults_cmd = std::process::Command::new("defaults");
-    defaults_cmd.args(&["read", "-globalDomain", "AppleInterfaceStyle"]);
-    match defaults_cmd.output() {
-        Ok(output) => output.stdout == b"Dark\n",
-        Err(_) => true,
+    const PREFERNECES_FILE: &str = "Library/Preferences/.GlobalPreferences.plist";
```
Enselic, line 404: Is this always the path? Regardless of Mac OS version and regardless of system/user settings?

Enselic (commented):
Very nice 🤩

Unsure when I will get the time to do a detailed review, but on a high level this looks good.

I have two main concerns initially. I wrote one in a comment, and the other is: if we should try to find a crate to use for "is dark mode on". If you upstream your fix, then more projects and users would benefit from it.

That said, if there is no such crate mature enough for us to use, then personally I am fine with rolling our own code. I think this code makes a lot of sense to optimize.

src/assets.rs
BlackHoleFox: > Regardless of mac OS version

Yes, to the best of what I know. Its been at this path since at least 10.7 in 2011 (which is the lowest Rust supports) but probably even before that so I'm confident this will stick around.

> regardless of system/user settings?

I think so? I don't have any documentation to back it up but I have never seen it anywhere else different. If this isn't enough, then we should just use CFPreferences so this is abstracted away.

BlackHoleFox:
>  if we should try to find a crate to use for "is dark mode on". If you upstream your fix, then more projects and users would benefit from it.

The one I know about is `dark-light`, which uses some Objective-C via `objc` to read the app's current appearance from the system. I didn't list this one originally because its not reading any preferences directly. imo this fix wouldn't be upstreamable to a crate like that anyway which is meant to be used in more traditional applications with UI since for them, using AppKit via ObjectiveC _is_ the Apple endorsed way to get the app's appearance for a few reasons (user app overrides being the notable one). 

The upside is that the function here becomes a single match block, but the downsides are:
- Extra third-party dependency for a very small implementation need
    - This also bloats the lockfile because of the other platform's dependencies
- Significantly slower then directly reading the value (~50ms)

Enselic (approved):
This looks good to me, so I'm going to set this as Approved. With one big caveat. Which is that I have not verified or benchmarked this change. The reason being that I have switched jobs, and lost access to my Macs. I currently don't plan on buying a Mac in the near term.

So I'll set this as Approved now, and give other maintainers a chance to try this on a Mac if they have access to one (I'm unsure if they do).

If none of us has access to any Mac any longer, I think we need to trust you and blindly merge this :)

Enselic:
@BlackHoleFox I almost forgot: Can you add any entry about this change to CHANGELOG.md please?

BlackHoleFox:
Sure. I added a line to the `Other` category.

> If none of us has access to any Mac any longer, I think we need to trust you and blindly merge this :)

In the very unlikely case this breaks on someone else's computer, and/or no one else does have a Mac, I can probably do a bugfix or create a different implementation

Enselic:
@sharkdp @keith-hall Does any one of you still run a Mac? Due to a series of coincidences, I have ended up running Windows (and Ubuntu via WSL2 as my day to day Rust dev environment).

If none of you run Mac and are able to verify this change, I think our only option is to blindly merge this.

src/assets.rs
sharkdp, line 404: Typo:
```suggestion
    const PREFERENCES_FILE: &str = "Library/Preferences/.GlobalPreferences.plist";
```

src/assets.rs
```diff
@@ -401,11 +401,21 @@ fn asset_from_cache<T: serde::de::DeserializeOwned>(
...
+    const STYLE_KEY: &str = "AppleInterfaceStyle";
+
+    let preferences_file = dirs_next::home_dir()
+        .map(|home| home.join(PREFERNECES_FILE))
+        .expect("Could not get home directory");
```
sharkdp, line 409: I think this could be

```suggestion
    let preferences_file = dirs_next::home_dir()
        .expect("Could not get home directory")
        .join(PREFERNECES_FILE);
```

We could also think about silently ignoring this error (returning a default value). Not sure what a valid scenario for this is, but maybe we shouldn't crash with a panic in this case.

BlackHoleFox: I can also do that but code elsewhere panics in this case so I figured it wasn't something `bat` wanted to handle.

src/assets.rs
```diff
@@ -401,11 +401,21 @@ fn asset_from_cache<T: serde::de::DeserializeOwned>(
...
+
+    match plist::Value::from_file(preferences_file).map(|file| file.into_dictionary()) {
+        Ok(Some(preferences)) => match preferences.get(STYLE_KEY).and_then(|val| val.as_string()) {
+            Some(value) => value == "Dark",
+            // If the key does not exist, then light theme is currently in use.
+            None => false,
+        },
```
sharkdp, line 416: The inner match could probably be replaced by sth. like `.map(|value| value == "Dark").unwrap_or(false)`, but I'm not sure if that necessarily makes it better.

sharkdp:
Thank you for working on this :+1: 

> Does any one of you still run a Mac?

I don't, but I'm okay with trusting @BlackHoleFox on this.

src/assets.rs
BlackHoleFox: Hm imo its slightly worse since then the comment about what the key missing means has no good spot anymore.

sharkdp: Okay :+1:

src/assets.rs
sharkdp: Ok. Thank you. Let's keep the `.expect(…)`.

BlackHoleFox:
Just realized there isn't much paper trail about the functionality here, my bad. Here's the detection on macOS 13:

Dark theme on:

image

Light theme on:

image

Enselic:
Let's get this merged now, and handle any follow-up problems (if any) in separate PRs.

Issue #2429, opened 2022-12-28
Redefine how non-printable chars are printed


yermulnik:
Is this possible to redefine how non-printable chars are displayed?
Builtin solution is a way unreadable for my terminal configuration and I'd like an option to show non-printables like most of UNIX tools does.

![image](https://user-images.githubusercontent.com/1274789/209837833-2dfcdd78-9526-4c11-8e13-7b6e9aac34b4.png)

ps: Thanks for the great tool =)

keith-hall:
This is currently not configurable, but the relevant code is at https://github.com/sharkdp/bat/blob/b6b9d3a629bd9b08725df2a4e7b92c3023584a89/src/preprocessor.rs#L52

yermulnik:
@keith-hall Thanks for the pointer.
Is it possible to put this on a feature requests list? Sort of `--output-style ascii` or something like this.

PR #2443, opened 2023-01-07
Print non-printable characters using caret notation

einfachIrgendwer0815:
Implements feature request #2429.

When using `--show-all`, non-printable characters are printed using the Unicode Control Pictures. This pr adds the flag `--caret-notation`, which can only be used together with `--show-all`. When `--caret-notation` is set, non-printable characters are printed using caret notation instead.

For example, without `--caret-notation`:
```
This·is·an·example·text.␊␍
```
With `--caret-notation`:
```
This·is·an·example·text.^J^M
```

Instead of `--caret-notation`, the alias `--caret` can be used too.


src/preprocessor.rs
```diff
@@ -100,7 +85,53 @@ pub fn replace_nonprintable(input: &[u8], tab_width: usize) -> String {
...
+                            '\x00' => output.push_str("^@"),
+                            // bell
+                            '\x07' => output.push_str("^G"),
+                            // backspace
+                            '\x08' => output.push_str("^H"),
+                            // escape
+                            '\x1B' => output.push_str("^["),
+
+                            // everything else
+                            c => output.push_str(&c.escape_unicode().collect::<String>())
```
Enselic, line 109: To reduce the diff and levels of indentation I would prefer if you added `if use_caret` to each existing character arm instead is of doing the "refactoring" in the current PR

Enselic (commented):
Thanks, looks sensible on first glance! Before a detailed review, please also add regression tests to integration_tests.rs

src/preprocessor.rs
einfachIrgendwer0815: I'm sorry, I don't quite understand your comment. So should I refactor it or not?

Enselic: Sorry for being vague. I mean that I would prefer e.g.

```
'\x1B' => output.push('␛'),
```

to be changed to

```
'\x1B' => output.push_str(if use_caret { "^[" } else  { "␛" }),
```

leandro86elpiojo:
si

Enviado desde Correo<https://go.microsoft.com/fwlink/?LinkId=550986> para Windows

________________________________
De: Martin Nordholts ***@***.***>
Enviado: Sunday, January 8, 2023 1:37:14 PM
Para: sharkdp/bat ***@***.***>
Cc: Subscribed ***@***.***>
Asunto: Re: [sharkdp/bat] Print non-printable characters using caret notation (PR #2443)


@Enselic commented on this pull request.

________________________________

In src/preprocessor.rs<https://github.com/sharkdp/bat/pull/2443#discussion_r1064170853>:

> +                                output.push_str("^J\x0A");

+                                line_idx = 0;

+                            }

+                            // carriage return

+                            '\x0D' => output.push_str("^M"),

+                            // null

+                            '\x00' => output.push_str("^@"),

+                            // bell

+                            '\x07' => output.push_str("^G"),

+                            // backspace

+                            '\x08' => output.push_str("^H"),

+                            // escape

+                            '\x1B' => output.push_str("^["),

+

+                            // everything else

+                            c => output.push_str(&c.escape_unicode().collect::<String>())



Sorry for being vague. I mean that I would prefer e.g.

'\x1B' => output.push('␛'),



to be changed to

'\x1B' => output.push_str(if use_caret { "^[" } else  { "␛" }),

Enselic:
Thank you for the update. I will review as soon as I get some time over.

src/bin/bat/clap_app.rs
```diff
@@ -59,6 +59,18 @@ pub fn build_app(interactive_output: bool) -> Command {
                      Use '--tabs' to control the width of the tab-placeholders.",
                 ),
         )
+        .arg(
+            Arg::new("caret-notation")
+                .long("caret-notation")
+                .alias("caret")
```
Enselic, line 65: I would prefer if we didn't make `--caret` a short form of `--caret-notation`. "caret notation" indeed seems to be the [official term](https://en.wikipedia.org/wiki/Caret_notation) for this, and I don't think `--caret` is clear enough on its own. I wouldn't rule out us adding an alias in the future, but since it is much easier to add stuff than to remote stuff, I would prefer if we started out without it.

Enselic (commented):
This looks good! To be cautious, I would like us to not have the `--caret` alias, at least not in the initial version. Otherwise this is Approved from me.

Enselic (dismissed):
👍

sharkdp:
This looks good - thank you! Just one question regarding the CLI. To avoid introducing (1) a new option and (2) a new option that can only be used together with another option, we could also think about modifying the `--show-all` option. We could change it such that it takes a value (the "non printable character style"). For now, this could have two options: `unicode` and `caret`/`caret-notation`. The default could be `unicode` to maintain backwards-compatibility.

Downside: this might break `cat` compatibility in some edge cases?!

What do others think of this option?

Enselic:
I think we should leave `--show-all` as is for compatibility with `cat`.

But I like the idea of making the name of the new option more future-proof. Both OpenBSD `cat` and GNU `cat` use the terminology `nonprinting`, so I propose that we rename `--caret-notation` to `--nonprinting-notation=caret`, with `--nonprinting-notation=unicode` being default.

Enselic (changes requested):
Since this will affect the public API of the `bat` library, it is worth putting some extra effort into designing the public API.

My preferred way to see how the public API is impacted is to run `cargo public-api diff latest` after installing with `cargo install --locked cargo-public-api`.

Here is the output I get when diffing the public API of your branch against the public API of the latest `bat` release on crates.io (in the terminal there is syntax highlighting, but that gets lost on GitHub):
```console
$ cargo public-api diff latest
Resolved `diff latest` to `diff 0.22.1`
 Documenting bat v0.22.1
    Finished dev [unoptimized + debuginfo] target(s) in 0.56s
 Documenting bat v0.22.1 (/home/martin/src/bat)
    Finished dev [unoptimized + debuginfo] target(s) in 0.63s
(none)

(none)

Added items to the public API
=============================
+pub bat::config::Config::nonprinting_notation: alloc::string::String
+pub fn bat::style::StyleComponents::clear(&mut self)
+pub fn bat::style::StyleComponents::insert(&mut self, component: bat::style::StyleComponent)
```

As we can see, `bat::config::Config::nonprinting_notation: alloc::string::String` has been added to the public API. 

I really think we should use an enum instead of a free form `String`. See e.g. https://docs.rs/bat/latest/bat/enum.WrappingMode.html.

So can you introduce an enum please? Thanks!

Enselic:
Thank you for the reminder. I am unusually busy both privately and professionally, so I do not have time to look againt at this PR right now.

I have not forgotten about it though. I would be surprised if it takes longer than a month before I get some time over. (Or maybe some other maintainer will take a look before that.)

src/nonprinting_notation.rs
```diff
@@ -0,0 +1,11 @@
...
+pub enum NonprintingNotation {
+    Caret,
+    Unicode,
+}
+
+impl Default for NonprintingNotation {
+    fn default() -> Self {
+        NonprintingNotation::Unicode
+    }
+}
```
Enselic, line 11: Can you add docs and automatically derive Default please? Proposal:

```suggestion
/// How to print non-printable characters with
/// [crate::config::Config::show_nonprintable]
#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
pub enum NonprintingNotation {
    /// Use caret notation (^G, ^J, ^@, ..)
    Caret,

    /// Use unicode notation (␇, ␊, ␀, ..)
    #[default]
    Unicode,
}
```

Enselic (changes requested):
I just realized we already use the term `nonprintable` in `Config::show_nonprintable`. So to be consistent with ourselves, I think we should rename `NonprintingNotation` to `NonprintableNotation`.

Sorry for changing my mind again :(

einfachIrgendwer0815:
> Can you add docs and automatically derive Default please? Proposal:
> 
> ```rust
> /// How to print non-printable characters with
> /// [crate::config::Config::show_nonprintable]
> #[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
> pub enum NonprintingNotation {
>     /// Use unicode notation (␇, ␊, ␀, ..)
>     #[default]
>     Unicode,
> 
>     /// Use caret notation (^G, ^J, ^@, ..)
>     Caret,
> }
> ```

As the failed check shows, this is not compatible with the minimum supported Rust version (1.60), as this feature wasn't stabilized until version [1.62](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1620-2022-06-30)

Enselic:
> As the failed check shows, this is not compatible with the minimum supported Rust version (1.60), as this feature wasn't stabilized until version [1.62](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1620-2022-06-30)

Oh, I see. Well, that's annoying. Let's bump MSRV then! PR: https://github.com/sharkdp/bat/pull/2496

Enselic (approved):
Thanks a lot for the updates, this looks good to me now!

For the record, here is the `bat` public API diff:

```console
$ cargo public-api --omit blanket-impls --omit auto-trait-impls --omit auto-derived-impls diff origin/master..einfachIrgendwer0815/feature_caret_notation
(none)

(none)

Added items to the public API
=============================
+pub bat::config::Config::nonprintable_notation: bat::NonprintableNotation
+pub enum bat::NonprintableNotation
+pub bat::NonprintableNotation::Caret
+pub bat::NonprintableNotation::Unicode
```

src/nonprintable_notation.rs
```diff
@@ -0,0 +1,11 @@
+/// How to print non-printable characters with
+/// [crate::config::Config::show_nonprintable]
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+pub enum NonprintableNotation {
```
Enselic, line 4: I forgot one thing! We need to make the new enum `#[non_exhaustive]`. Otherwise adding more notations will require a major semver version bump.

```suggestion
#[non_exhaustive]
pub enum NonprintableNotation {
```

You don't have to add it though. I can add it before I merge this. I plan on merging it early next week, unless something else comes up.

Enselic (changes requested):

Enselic (approved):

PR #2446, opened 2023-01-13
Fix code fence in doc comment

rhysd:
This PR fixes the code fence in doc comment. This broke syntax highlighting when editing this file in Vim.


Enselic:
Thanks!

PR #2447, opened 2023-01-15
Remove unnecessary heap allocations in `PrettyPrinter::print` for style components

rhysd:
Current `PrettyPrinter::print` implementation causes two redundant heap allocations for building style components.

- At first it collects all style components in `Vec` then it converts `Vec` into `HashSet`
- `HashSet` instance is re-created per `print` call

However both heap allocations are unnecessary because

- the `Vec` instance is not necessary if inserting components to `HashSet` directly
- the heap-allocated memory in `HashSet` instance can be reused


keith-hall (approved):

sharkdp:
Thank you!

PR #2496, opened 2023-03-03
Bump MSRV to 1.62

Enselic:
Mainly to be able to derive default enum variants. See https://github.com/sharkdp/bat/pull/2443#issuecomment-1452334545


sharkdp (approved):
Thanks

PR #2501, opened 2023-03-14
Replace unmaintained GitHub Actions

nickelc:
This PR fixes all current deprecation warnings by the build jobs.

The unmaintained `actions-rs/*` actions are replaced with [`dtolnay/rust-toolchain`],
[`taiki-e/install-action`] and plain cargo commands.

The deprecated `set-output` command is replaced by writing the output variables into a file. 
See [announcement].
```bash
echo "NAME=${value}" >> $GITHUB_OUTPUT
```

[`dtolnay/rust-toolchain`]: https://github.com/dtolnay/rust-toolchain
[`taiki-e/install-action`]: https://github.com/taiki-e/install-action
[announcement]: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -215,8 +215,8 @@ jobs:
         fi
 
         # Let subsequent steps know where to find the (stripped) bin
-        echo ::set-output name=BIN_PATH::${BIN_PATH}
-        echo ::set-output name=BIN_NAME::${BIN_NAME}
+        echo "BIN_PATH=${BIN_PATH}" >> $GITHUB_OUTPUT
+        echo "BIN_NAME=${BIN_NAME}" >> $GITHUB_OUTPUT
 
     - name: Set testing options
       id: test-options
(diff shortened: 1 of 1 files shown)


.github/workflows/CICD.yml
```diff
@@ -18,11 +18,8 @@ jobs:
     name: Ensure 'cargo fmt' has been run
     runs-on: ubuntu-20.04
     steps:
-    - uses: actions-rs/toolchain@v1
+    - uses: dtolnay/rust-toolchain@stable
```
sharkdp, line 21: Just as a hint for other reviewers: this action seems to use a "minimal" profile by default: https://github.com/dtolnay/rust-toolchain/issues/41#issuecomment-1291231129

.github/workflows/CICD.yml
```diff
@@ -195,12 +153,20 @@ jobs:
...
+    - name: Install cross
+      if: matrix.job.use-cross
+      uses: taiki-e/install-action@v2
+      with:
+        tool: cross
+
+    - name: Overwrite build command env variable
+      if: matrix.job.use-cross
+      shell: bash
+      run: echo "BUILD_CMD=cross" >> $GITHUB_ENV
```
sharkdp, line 169: I would have expected "cargo cross", but it seems to work with "cross" alone..

sharkdp (approved):
Thank you very much. That is so much cleaner than before.

PR #2503, opened 2023-03-14
Set `rust-version` in Cargo metadata and use it in the MSRV build job

nickelc:
It limits clippy to only warn about lints that are compatible with the current MSRV.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -1,7 +1,6 @@
 name: CICD
 
 env:
-  MIN_SUPPORTED_RUST_VERSION: "1.62.0"
   CICD_INTERMEDIATES_DIR: "_cicd-intermediates"
 
 on:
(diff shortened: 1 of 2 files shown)


.github/workflows/CICD.yml
```diff
@@ -42,10 +41,13 @@ jobs:
     - name: Checkout source code
       uses: actions/checkout@v3
 
-    - name: Install rust toolchain (v${{ env.MIN_SUPPORTED_RUST_VERSION }})
+    - name: Get the MSRV from the package metadata
+      id: msrv
+      run: cargo metadata --no-deps --format-version 1 | jq -r '"version=" + .packages[0].rust_version' >> $GITHUB_OUTPUT
```
Enselic, line 46:
```suggestion
      run: cargo metadata --no-deps --format-version 1 | jq -r '"version=" + (.packages[] | select(.name = "bat").rust_version)' >> $GITHUB_OUTPUT
```

Enselic (approved):
Nice!

Just had a proposal for a longer but (IMHO) clearer one-liner. But if you don't like it I'm fine with keeping what you have.

nickelc:
> Just had a proposal for a longer but (IMHO) clearer one-liner. But if you don't like it I'm fine with keeping what you have.

That's fine by me.

sharkdp:
Thank you

PR #2504, opened 2023-03-15
Update `clap` to 4.1.8

nickelc:
This requires bumping the MSRV to 1.64.

It also updates the `terminal_size` crate that is pulled in by `clap` to reduce duplicate dependencies.


Enselic (approved):
LGTM!

Cargo.toml
```diff
@@ -78,7 +78,7 @@ default-features = false
 features = ["parsing"]
 
 [dependencies.clap]
-version = "4.0.32"
+version = "4.1.8"
```
sharkdp, line 81: I think we could potentially put 
```suggestion
version = "4"
```
here in order to avoid having to update this manually in the future (unless there's a new major version).

Cargo.toml
```diff
@@ -98,7 +98,7 @@ tempfile = "3.3.0"
 nix = { version = "0.26.2", default-features = false, features = ["term"] }
 
 [build-dependencies.clap]
-version = "4.0.32"
+version = "4.1.8"
```
sharkdp, line 101: same here

sharkdp (approved):
Thank you

Cargo.toml
Enselic: In general it is better to put a lower bound on minors versions, otherwise cargo can resolve versions in the wrong way.

More details and further discussion here: https://users.rust-lang.org/t/psa-please-specify-precise-dependency-versions-in-cargo-toml/71277

Enselic: ...and dependabot does the bumps for us anyway, right? In this case I think we asked dependabot to stop since bumping required manual work. From what I understand, us doing the bump manuwally now will make dependabot start bumping clap for us again.

sharkdp: Ok, thanks for the explanation. I reverted the change

Issue #2166, opened 2022-04-25
Use strip option from Rust 1.59


wcampbell0x2a:
You could replace your strip scripts with the official strip provided in Cargo profiles if you bump MSRV to 1.59.

https://doc.rust-lang.org/cargo/reference/profiles.html#strip

I'd be happy to try this out, unless you have a MSRV policy that I couldn't find.

Enselic:
Our MSRV policy is basically "as small as possible without causing trouble". In practice this usually means setting the MSRV to the highest MSRV of our dependencies. So I think it is too early to set MSRV to 1.59.

But once we do, I agree it would be nice to remove our own strip calls. So we can keep this issue open.

PR #2506, opened 2023-03-16
Use cargo's `strip` option to remove symbols+debuginfo from the binaries

nickelc:
- #2166

closes: #2166


Enselic (approved):
Clear improvement, and artifact sizes looks correct 👍

sharkdp:
Thanks!

PR #2515, opened 2023-03-25
Allow customizing output directory for generated assets

tranzystorek-io:
Makes packaging easier, e.g. by doing `export BAT_ASSETS_GEN_DIR=${PWD}/gen` instead of looking in the target dir.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ## Other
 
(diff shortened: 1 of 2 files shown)


sharkdp:
But... you can already do that using `OUT_DIR`, no?

tranzystorek-io:
Not really, that's an env var that `cargo` sets for the build script, overwriting anything from the current shell

sharkdp:
Ah, right. Thank you for the clarification. In that case, this looks good.

PR #2516, opened 2023-03-25
CI improvements

sharkdp:
- Extract metadata using `cargo metadata` and `jq`, instead of grepping
- Use macos-12 image

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -14,6 +14,26 @@ on:
       - '*'
 
 jobs:
+  crate_metadata:
+    name: Extract crate metadata
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/CICD.yml
```diff
@@ -127,7 +126,7 @@ jobs:
           - { target: i686-pc-windows-msvc        , os: windows-2019                  }
           - { target: i686-unknown-linux-gnu      , os: ubuntu-20.04, use-cross: true }
           - { target: i686-unknown-linux-musl     , os: ubuntu-20.04, use-cross: true }
-          - { target: x86_64-apple-darwin         , os: macos-10.15                   }
```
sharkdp, line 130: 10.15 will be deprecated soon

PR #2528, opened 2023-04-05
Include aarch64-apple-darwin

bjw-s:


bjw-s:
Nevermind... looks like it's still not as easy as just adding the arch :(

PR #2530, opened 2023-04-05
Use the `is-terminal` crate instead of `atty`

nickelc:
The crate is already used by `clap` and a similar trait is about to be stabilized in `std`.

https://doc.rust-lang.org/stable/std/io/trait.IsTerminal.html


Enselic (approved):
Thanks!

sharkdp:
Thank you

PR #2532, opened 2023-04-05
Update `plist` to 1.4

nickelc:
The version replaced `xml-rs` with the `quick-xml` crate. That makes -1 warning for an unmaintained crate from `cargo audit`.


Enselic (approved):
👍

PR #2533, opened 2023-04-05
Use valid SPDX license expression

nickelc:


Enselic (approved):
This is normally what I see too in other crates

PR #2536, opened 2023-04-07
feat: map nix-store style *-nginx.cofig

blaggacao:
```console
./target/debug/bat /nix/store/sm8lcsj94pqsb43ssb7cx3r424zs71p1-nginx.conf
```

is now properly styled.

This is a very typical nix store autogenerated NixOS nginx config naming scheme.

```console
./target/debug/bat -L
[ ... ]
nginx                     conf.erb, nginx.conf, mime.types, fastcgi_params, scgi_params, uwsgi_params, /etc/nginx/**/*.conf,
                          /etc/nginx/sites-*/**/*, **nginx.conf
```

This means the removal of `mime.types` and `nginx.conf` has been merely a deduplication.

This is a small QoL improvement, but let's say that many NixOS users _love_ `bat`.

### DIFF
--- a/src/syntax_mapping.rs
+++ b/src/syntax_mapping.rs
@@ -89,8 +89,7 @@ impl<'a> SyntaxMapping<'a> {
         for glob in &[
             "/etc/nginx/**/*.conf",
             "/etc/nginx/sites-*/**/*",
-            "nginx.conf",
-            "mime.types",
+            "**nginx.conf",
         ] {
             mapping.insert(glob, MappingTarget::MapTo("nginx")).unwrap();
         }


src/syntax_mapping.rs
```diff
@@ -89,8 +89,7 @@ impl<'a> SyntaxMapping<'a> {
         for glob in &[
             "/etc/nginx/**/*.conf",
             "/etc/nginx/sites-*/**/*",
-            "nginx.conf",
```
keith-hall, line 92: Thanks for your contribution! I see that the duplication was from [the nginx sublime syntax file specifying these under `file_extensions`](https://github.com/brandonwamboldt/sublime-nginx/blob/65f5a63c0d1760c5db2264e50e3b14a7a4567cc0/Syntaxes/nginx.sublime-syntax#L9). But `mime.types` is a full file name, so I am wondering if that is why it was also in this mapping, maybe it behaves differently than expected?

Thus I would feel more confident about this change if we had a `mime.types` file in our syntax tests (https://github.com/sharkdp/bat/tree/master/tests/syntax-tests/source/nginx) to prove that they are still highlighted correctly.

blaggacao: I happen to have (a real) one at hand, so I can add this. Good idea!

sharkdp:
Closing this for now, happy to reopen once we have test for this.

PR #2537, opened 2023-04-08
Clean-up bash completion

adrianbiro:
"cword" is unset and even is not used in completion for other shells, so it probably shouldn't be here either. If it for some reason should be there add the directive "# shellcheck disable=SC2034" to the line above it.


Enselic (approved):
LGTM, thanks.

PR #2548, opened 2023-04-22
Replace `dirs` crate with `etcetera` crate

utkarshgupta137:
The cache & config paths are unchanged. The `etcetera` crate allows you to use the XDG spec on macOS too without any hassle. In addition, it is slightly smaller.

Disclaimer: I've recently become a maintainer of the `etcetera` crate. AFAIK, it is the only crate that can replace this logic.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -129,13 +129,14 @@ dependencies = [
  "clircle",
  "console",
  "content_inspector",
- "dirs",
  "encoding_rs",
+ "etcetera",
  "expect-test",
  "flate2",
  "git2",
  "globset",
  "grep-cli",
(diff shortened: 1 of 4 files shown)


eth-p:
Thanks for the contribution!

@(the other maintainers):
I can confirm it works as intended for me on MacOS :)

utkarshgupta137:
@sharkdp This is the same PR as https://github.com/sharkdp/fd/pull/1304 & https://github.com/sharkdp/vivid/pull/103

sharkdp (approved):
Thank you. I'm okay with this change but it needs a rebase. Sorry for the delay in reviewing.

PR #2549, opened 2023-04-22
Update `clap` to 4.2

nickelc:
The new `clap` version is using the new `anstyle` & `anstream` crates for its ANSI text styling and I want to replace the `nu-ansi-term` crate with `anstyle` in a follow-up PR.

Blog: [`anstream`: simplifying terminal styling](https://epage.github.io/blog/2023/03/anstream-simplifying-terminal-styling/)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,10 +27,59 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstream"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


nickelc:
If I remove 6039152 then we can keep the MSRV.

Enselic (approved):
Well, I think we inevitably are going to need to update clap at the some point. Might as well do it now.

Enselic:
Merging now since I want to merge dependabot PRs, but I do not want to cause conflict with this PR

PR #2567, opened 2023-05-12
Replace `nu-ansi-term` with `anstyle` crates

nickelc:
The `nu-ansi-term` crate can now be replaced by the `anstyle` crates which are used by `clap` for its ANSI text styling.


nickelc:
So, I looked into the failing regression tests and the difference is that `anstyle` writes the effects for bold, italic etc. separately.

[playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0cb844213f10725af2b42d4e4f13b05b)

sharkdp:
Thank you and sorry for the review delay. Can you please go into more details why this is necessary/beneficial?

Enselic:
I think the benefit is that we get rid of a dependency (`nu-ansi-term`) without adding any new ones, since the "new" ones are already needed by clap?

Enselic (approved):
This looks good to me. The ANSI color diffs all seem to be only differences in how the colors are encoded, not how they are displayed (at least in my terminal I did not see any actual color change).

Maybe there is a mistake among all those lines. That can be fixed with a follow up PR if needed. When I glance it over slowly it looks reasonable to me.

Can you rebase this on latest master please?

sharkdp:
> So, I looked into the failing regression tests and the difference is that `anstyle` writes the effects for bold, italic etc. separately.

This could have adverse effects though, right? We need to write more output. And terminal emulators might take a longer time to parse the input and/or to set up the proper style.

I would appreciate if we could run a benchmark to see that this doesn't have any negative effects. Since we want to include possible effects of the terminal emulator, we can use `hyperfine`s `--show-output` option.

Enselic:
Good point about performance. I didn't think about that. Let's see some numbers before we merge.

PR #2574, opened 2023-05-22
Fix paging not happening when stdout is interactive but stdin is not

### LINKED ISSUE #2572: regression: bat fails to page - will not read from stdin if no file (or `-`) specified
`bat` (no longer) reads from `stdin` if no file or `-` is specified. This breaks my `git(1)` workflow.

```bash
# Works
$ bat big-file.txt

# Fails to page
$ cat big-file.txt | bat

# Fails to page
$ cat big-file.txt | bat -

# Works (because an actual file is specified)
$ cat big-file.txt | bat /dev/fd/0
```

**What happens?**

`bat` doesn't page - the file is just dumped to `stdout` _a la `cat(1)`_.

`bat` should page the file.

```bash
$ git clone https://github.com/sharkdp/bat
$ cargo install --path .
```

---

```bash
$ bat --version
bat 0.23.0 (b420c42)
```

Nigecat:
This seems to have been introduced in 57cc0d843567eef566540074b391f8212d859871 since the condition `!atty::is(Stream::Stdin)` was inverted to `std::io::stdin().is_terminal()`. This seems to be a mistake since the comment says
```
// If we are reading from stdin, only enable paging if we write to an
// interactive terminal and if we do not *read* from an interactive
// terminal.
```
but the current code enables paging if stdin **is interactive**. 

This PR just flips the condition back to how it was to restore the original behavior. 
No changelog update as there hasn't been a release since the regression was introduced.  

Fixes #2572

### DIFF
--- a/src/bin/bat/app.rs
+++ b/src/bin/bat/app.rs
@@ -104,7 +104,7 @@ impl App {
                     // If we are reading from stdin, only enable paging if we write to an
                     // interactive terminal and if we do not *read* from an interactive
                     // terminal.
-                    if self.interactive_output && std::io::stdin().is_terminal() {
+                    if self.interactive_output && !std::io::stdin().is_terminal() {
                         PagingMode::QuitIfOneScreen
                     } else {
                         PagingMode::Never


Enselic:
Oops! Thank you 🙏

jamesodhunt:
Thanks @Nigecat ! :smile:

PR #2577, opened 2023-05-28
add NSIS support

idleberg:
NSIS is a scripting language to create installers for Windows applications. It has been around for a very long time and still has a wide user base. I'm a frequent contributor to the submodule added in this PR and would love to see it supported by `bat`.


keith-hall:
It looks like https://packagecontrol.io/packages/NSIS meets our criteria of download count for inclusion, so no objections here.

Please add an example NSIS file to the https://github.com/sharkdp/bat/tree/master/tests/syntax-tests/source folder, and run the [`update.sh` script in the folder above](https://github.com/sharkdp/bat/blob/master/tests/syntax-tests/update.sh#L7) to generate the highlighted file and then add them both to this PR. This allows us to more easily track regressions when color schemes change, or the syntax highlighting library we used is updated etc.

idleberg:
@keith-hall Added the files. Any idea why the generated file doesn't highlight? The highlighter works fine in Sublime Text versions 3 and 4.

keith-hall:
If I understand correctly, the highlighting didn't work for you locally - it seems CI highlights it fine and that's why it complains about the difference between the "expected" highlighted file which is included in this PR, which seems to have only one color.
Maybe try running step 3 from https://github.com/sharkdp/bat/blob/master/doc/assets.md#adding-new-builtin-languages-for-syntax-highlighting again and see if it helps.

idleberg:
Unfortunately, it did not help. The generated file still looks all gray.

idleberg:
@keith-hall I don't know why the automatically generated file continues to be all gray, but I did manual adjustments and now the test runs through. I'm not happy with this process, but this should be sufficient to test possible future regressions.

PR #2580, opened 2023-05-31
Enable `bat` to find `more` binary on Windows

### LINKED ISSUE #2570: `bat` refuses to page content

`bat some-long-test-file.txt --pager more --paging always`

**What happens?**

Contents of the file are dumped to the terminal all at once.

Paging with `more`.

Cargo.

---

#### Software version

bat 0.23.0

#### Operating system

Windows 6.2.9200

#### Command-line

```bash
bat test.cmd --pager more --paging always --diagnostic
```

#### Environment variables

```bash
SHELL=<not set>
PAGER=<not set>
LESS=<not set>
LANG=<not set>
LC_ALL=<not set>
BAT_PAGER=more
BAT_CACHE_PATH=<not set>
BAT_CONFIG_PATH=<not set>
BAT_OPTS=<not set>
BAT_STYLE=<not set>
BAT_TABS=<not set>
BAT_THEME=<not set>
XDG_CONFIG_HOME=<not set>
XDG_CACHE_HOME=<not set>
COLORTERM=<not set>
NO_COLOR=<not set>
MANPAGER=<not set>
```

#### System Config file

Could not read contents of 'C:\ProgramData\bat\config': System nie może odnaleźć określonej ścieżki. (os error 3).

#### Config file

Could not read contents of 'C:\Users\Mataha\AppData\Roaming\bat\config': System nie może odnaleźć określonej ścieżki. (os error 3).

#### Custom assets metadata

Could not read contents of 'C:\Users\Mataha\AppData\Local\bat\metadata.yaml': System nie może odnaleźć określonej ścieżki. (os error 3).

#### Custom assets

'C:\Users\Mataha\AppData\Local\bat' not found

#### Compile time information

- Profile: release
- Target triple: x86_64-pc-windows-msvc
- Family: windows
- OS: windows
- Architecture: x86_64
- Pointer width: 64
- Endian: little
- CPU features: fxsr,sse,sse2
- Host: x86_64-pc-windows-msvc

mataha:
On Windows, `more` is a COM executable (i.e. its extension is `com`). As `resolve_binary()` tries to emulate the behaviour of Win32 API's `CreateProcess()`, it will search for `more.exe` and fail. This commit adds an exception to that by specifying the pager's binary explicitly.

Fixes #2570. See also BurntSushi/ripgrep#2523.

I'd like to add that `more` is not as bad as README makes it to be - it *does* support colors:

cmd_COSn24wvTg

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,7 +62,7 @@ path_abs = { version = "0.5", default-features = false }
 clircle = "0.4"
 bugreport = { version = "0.5.0", optional = true }
 dirs = { version = "5.0.0", optional = true }
-grep-cli = { version = "0.1.7", optional = true }
+grep-cli = { version = "0.1.9", optional = true }
 regex = { version = "1.8.3", optional = true }
 walkdir = { version = "2.3", optional = true }
 bytesize = { version = "1.2.0" }
--- a/src/output.rs
+++ b/src/output.rs
@@ -114,6 +114,8 @@ impl OutputType {
                 p.args(args);
             }
             p.env("LESSCHARSET", "UTF-8");
+        } else if pager.kind == PagerKind::More && cfg!(windows) && args.is_empty() {
+            p.arg("/E"); // Enable extended features of Windows' 'more' by default
         } else {
             p.args(args);
         };
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -106,17 +106,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "atty"
-version = "0.2.14"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
-dependencies = [
- "hermit-abi 0.1.19",
- "libc",
- "winapi",
-]
-
-[[package]]
 name = "autocfg"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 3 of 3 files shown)


src/output.rs
```diff
@@ -63,7 +63,15 @@ impl OutputType {
...
-        let resolved_path = match grep_cli::resolve_binary(&pager.bin) {
+        // On Windows 'resolve_binary()' won't be able to find 'more' as it is a COM executable
+        // (i.e. its extension is 'com') and thus we need to specify its binary name explicitly
+        let bin = if pager.kind == PagerKind::More && cfg!(windows) {
+            "more.com"
+        } else {
+            &pager.bin
+        };
+
+        let resolved_path = match grep_cli::resolve_binary(&bin) {
```
Enselic, line 74: Thank you for taking the time to fix this.

I think your change to bat will be much smaller, if needed at all, if we wait for your upstream fix https://github.com/BurntSushi/ripgrep/pull/2523? If so, I think we should wait for the upstream fix.

mataha: I don't think that fix will make it through if the intent is to emulate the behaviour of `CreateProcess()` API. Eh, DOS compatibility...

mataha: Upstream fix has arrived.

Cargo.toml
```diff
@@ -62,7 +62,7 @@ path_abs = { version = "0.5", default-features = false }
 clircle = "0.4"
 bugreport = { version = "0.5.0", optional = true }
 dirs = { version = "5.0.0", optional = true }
-grep-cli = { version = "0.1.7", optional = true }
+grep-cli = { version = "0.1.9", optional = true }
```
mataha, line 65: This requires Rust 1.70 and I believe such a change shouldn't be done in this PR...

mataha:
Closing in favor of #2651.

PR #2587, opened 2023-06-01
Add `os-release` syntax mapping

cyqsimon:
`/etc/os-release` and `/usr/lib/os-release` are files found on basically every Linux distro that describe the OS. As far as I know, there is no other use case for the file name `os-release`.

See https://www.freedesktop.org/software/systemd/man/os-release.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ## Syntaxes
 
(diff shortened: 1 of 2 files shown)


keith-hall (approved):

sharkdp:
Thank you

Issue #2561, opened 2023-05-05
`[bat error]: 'conf.d' is a directory.` is written to STDOUT instead of STDERR


JanPokorny:
**What steps will reproduce the bug?**

1. Invoke bat on a directory and redirect stderr, e.g. `bat /bin 2>/dev/null`

**What happens?**

bat outputs to the terminal: `[bat error]: '/bin' is a directory.`

**What did you expect to happen instead?**

The STDERR should be redirected to `/dev/null`, thus no error output should be visible. Looks like bat writes that error to STDOUT instead. Other errors (like when the file does not exist) get written to STDERR, though.

sharkdp:
Thank you for reporting this.

saguywalker:
Hello @sharkdp, I'd like to work on this. Is it ok for me to pick it up?

deboard:
Please assign me to this issue.

deboard:
This issue appears to be tied to if "attached_to_pager" is set to true in Controller, stdout gets passed to default_error_handler (in error.rs) if so. A good fix may be to just always write "[bat error] {error}" to stderr in in the default section of the match in default_error_handler (error.rs@70).

I have tried this. It fixes the problem and doesn't doesn't mess with the pager functionality. I you run bat like "bat --pager=never /tmp 2> /tmp/bat.txt", it runs as expected.

zzhaolei:
Will this be fixed?

PR #2590, opened 2023-06-03
Set PagerMode to Never when none of input files exists

saguywalker:
This PR fixes https://github.com/sharkdp/bat/issues/2561
 
Behavior when input is directory and existed

Master branch: `call_pager` is `true` because it only checks whether the path exists or not.
This PR: adds extra check to make sure that it's a file

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Features
 
 ## Bugfixes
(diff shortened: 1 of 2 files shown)


Enselic:
Thank you! Can you please also add a regression test for this bugfix that fails without the fix? See integration_tests.rs

Enselic:
Closing for now to keep the PR inbox clean. Of course feel free to re-open if you resume work on this!

PR #2594, opened 2023-06-07
feat: describe WinGet method for Windows install

maroskukan:
As WinGet gets more common and included in newer Windows 10/11 builds, I think it deserves a mention it in the installation methods.


Enselic:
Oops! Me and @sharkdp merged different versions (https://github.com/sharkdp/bat/pull/2519, https://github.com/sharkdp/bat/pull/2677) at the same time, and never noticed the accidental duplication.

Let's keep the other version though. It is shorter, and I like short. Or are there good reasons to keep the other one?

Issue #2552, opened 2023-04-28
Column misalignment past 9999 lines


DianaNites:
**What steps will reproduce the bug?**

1. `seq 1 10010 > file`
2. `bat file`
3. Go to the end of the pager

**What happens?**

Columns become misaligned from this point onwards despite there being plenty of space.

In practice i'm encountering this from `journalctl` / kernel logs / logs in general. logs big.

![2023-04-28-044052](https://user-images.githubusercontent.com/5275194/235138372-f44126c2-0645-4d11-8fa7-b39ed42c5508.png)

**What did you expect to happen instead?**

Normal and aligned line number column at least up to 6 digits, given the space already reserved is enough for that and still leaves the `│` space separated.

![2023-04-28-044129](https://user-images.githubusercontent.com/5275194/235138498-8d21856e-e602-4654-89ab-9f24ac1c1d10.png)

**How did you install `bat`?**

`cargo install bat`

**bat version and environment**

<details>
<summary>Details</summary>

#### Software version

bat 0.23.0 (871abd2)

#### Operating system

Linux 6.2.12-arch1-1

#### Command-line

```bash
bat --diagnostic 
```

#### Environment variables

```bash
SHELL=/usr/bin/fish
PAGER=<not set>
LESS=<not set>
LANG=en_US.UTF-8
LC_ALL=<not set>
BAT_PAGER='less -RFi'
BAT_CACHE_PATH=<not set>
BAT_CONFIG_PATH=<not set>
BAT_OPTS=<not set>
BAT_STYLE=<not set>
BAT_TABS=<not set>
BAT_THEME=<not set>
XDG_CONFIG_HOME=<not set>
XDG_CACHE_HOME=<not set>
COLORTERM=truecolor
NO_COLOR=<not set>
MANPAGER=<not set>
```

#### System Config file

Could not read contents of '/etc/bat/config': No such file or directory (os error 2).

#### Config file

Could not read contents of '/home/diana/.config/bat/config': No such file or directory (os error 2).

#### Custom assets metadata

Could not read contents of '/home/diana/.cache/bat/metadata.yaml': No such file or directory (os error 2).

#### Custom assets

'/home/diana/.cache/bat' not found

#### Compile time information

- Profile: release
- Target triple: x86_64-unknown-linux-gnu
- Family: unix
- OS: linux
- Architecture: x86_64
- Pointer width: 64
- Endian: little
- CPU features: fxsr,sse,sse2
[...]

PR #2597, opened 2023-06-13
Solution for Column misalignment past 9999 lines #2552

michabs:
Check to see that the number of lines in the file is more than 9999 then fix column misalignment to be according to the file number of lines

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -67,6 +67,18 @@ impl InputDescription {
             Some(kind) => format!("{} '{}'", kind.to_lowercase(), self.name),
         })
     }
+
+    pub fn file_num_lines(&self) -> Result<usize> {
+        let file = File::open(&self.name)
+            .map_err(|e| format!("file open error : {}",  e))?;
+        let file_reader = BufReader::new(file);
+        let mut total_lines = 0;
+        for _ in  file_reader.lines() {
+            total_lines += 1;
+        }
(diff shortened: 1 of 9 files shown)


michabs:
Solution for Column misalignment past 9999 lines #2552 fix unit tests

michabs:
fix fmt --check

michabs:
Check to see that the number of lines in the file is more than 9999 then fix column misalignment to be according to the file number of lines

src/input.rs
```diff
@@ -67,6 +67,16 @@ impl InputDescription {
...
+
+    pub fn file_num_lines(&self) -> Result<usize> {
+        let file = File::open(&self.name).map_err(|e| format!("file open error : {}", e))?;
+        let file_reader = BufReader::new(file);
+        let mut total_lines = 0;
+        for _ in file_reader.lines() {
+            total_lines += 1;
+        }
+        Ok(total_lines)
+    }
```
sharkdp, line 79: This will slow down `bat` considerably and it will most probably break some use cases like reading from a (named) pipe. We do not want (and for streaming reasons can not) open the file to count the number of lines before starting to display it.

sharkdp:
Thank you for your contribution, but I will have to close this for reasons given above. Please feel free to appeal if you think this can be fixed.

Issue #2592, opened 2023-06-06
Unit test failed on Windows


vidhyasasi:
Cargo test failed on Windows machine
**What steps will reproduce the bug?**

cargo test 

**What happens?**

test result: FAILED. 101 passed; 9 failed; 2 ignored; 0 measured; 0 filtered out; finished in 3.65s

**How did you install `bat`?**

Build from source

**bat version and environment**
bat 0.23.0 (876f5d4)

**Windows version**

10.0.20348 Build 20348

The failure message looks like:

failures:

---- alias_pager_disable_long_overrides_short stdout ----
thread 'alias_pager_disable_long_overrides_short' panicked at 'Unexpected stdout, failed var == "pager-output\n"
├── var: "hello world\n"
└── var as str: hello world

command=`"C:\\Users\\Administrator\\Documents\\rust_artefacts\\packages\\bat\\target\\debug\\bat.exe" "--no-config" "-P" "--paging=always" "test.txt"`
code=0
stdout="hello world\n"
stderr=""
', /rustc/90c541806f23a127002de5b4038be731ba1458ca\library\core\src\ops\function.rs:250:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- config_read_arguments_from_file stdout ----
thread 'config_read_arguments_from_file' panicked at 'Unexpected stdout, failed var == "dummy-pager-from-config\n"
├── var: "hello world\n"
└── var as str: hello world

command=`"C:\\Users\\Administrator\\Documents\\rust_artefacts\\packages\\bat\\target\\debug\\bat.exe" "test.txt"`
code=0
stdout="hello world\n"
stderr=""
', /rustc/90c541806f23a127002de5b4038be731ba1458ca\library\core\src\ops\function.rs:250:5

---- pager_arg_override_env_noconfig stdout ----
thread 'pager_arg_override_env_noconfig' panicked at 'Unexpected stdout, failed var == "pager-output\n"
├── var: "hello world\n"
└── var as str: hello world

command=`"C:\\Users\\Administrator\\Documents\\rust_artefacts\\packages\\bat\\target\\debug\\bat.exe" "--no-config" "--pager=echo pager-output" "--paging=always" "test.txt"`
code=0
stdout="hello world\n"
stderr=""
', /rustc/90c541806f23a127002de5b4038be731ba1458ca\library\core\src\ops\function.rs:250:5

---- pager_arg_override_env_withconfig stdout ----
[...]

PR #2602, opened 2023-06-22
#2592: Fixes unit tests dependent on echo on windows

boyvanduuren:
Leverage the existing `mock_pages` to put an echo batch script on the path and use that as a pager on Windows.


keith-hall (approved):
Looks good to me, thanks for your contribution

sharkdp:
Why did we not see those test failures in CI?

tests/integration_tests.rs
```diff
@@ -589,37 +597,49 @@ fn do_not_exit_directory() {
...
+#[serial]
 fn pager_basic() {
-    bat()
-        .env("PAGER", "echo pager-output")
-        .arg("--paging=always")
-        .arg("test.txt")
-        .assert()
-        .success()
-        .stdout(predicate::eq("pager-output\n").normalize());
+    mocked_pagers::with_mocked_versions_of_more_and_most_in_path(|| {
```
Enselic, line 602: Thank you for your contribution. I don't quite follow along here though. Why do we need to change for example this test?

boyvanduuren: You're welcome. The reason we're using the mocked versions here, is so that we actually have an `echo` available for powershell. See my comment on the original bug report: https://github.com/sharkdp/bat/issues/2592. `echo` is not resolved, so on powershell tests that rely on it fail.

boyvanduuren:
@sharkdp good question, quick glance at the pipeline and it seems like everything is ran in bash, also for Windows? That would explain it, seeing as this issue pops up on powershell (or anywhere tests rely on a binary that's not on the path, really).

tests/integration_tests.rs
Enselic: That makes sense. I never tried to run `cargo test` in PowerShell on Windows before. But now when I do these tests indeed fail. Any chance you would be willing to look into running our tests in CI in PowerShell on Windows? Some instructions: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-powershell

In our `.github\workflows\CICD.yml` and `matrix:` I think we would want to parameterize on `shell:` so that Windows tests run with `shell: pwsh`. Unless we continuously run the tests in PowerShell, it will for sure break again without anyone not noticing.

If you are not interested in doing this, that's totally fine. Just let me know, and I will do a more detailed review of this PR as it is.

boyvanduuren: Sure, I don't mind taking a look. Do you want it as part of this PR?

Enselic: Great! Yes I think that would be good, that way we know the tests pass in CI PowerShell before we merge.

boyvanduuren: Made a slight change to the test step so that tests are ran in powershell on windows targets. However, it seems a new test introduced with a merge on `master` yesterday seems to be failing intermittently. The test is `lessopen_and_lessclose_stdin_temp`.
It's also failing in other pipelines (see https://github.com/sharkdp/bat/actions/runs/6059140072), and I can reproduce it locally by running `cargo test` a few times on `master`.

Enselic: @Anomalocaridid Hello! Are you by any chance able to look into the reported flakiness of `lessopen_and_lessclose_stdin_temp`? Would be great if you could 🙏

Anomalocaridid: @Enselic After taking a quick look at the test output and code, I am not entirely sure what the cause is.

Also, strangely, if I only run the integration tests with `cargo test --test integration_tests` it does not appear to fail on my system. Even when running it repeatedly using `chronic` from moreutils.

If I had to make an educated guess, it's some sort of data race that's interfering with `$LESSCLOSE`. Likely by either keeping it from executing or by just preventing its output from reaching the test runner. I am not sure if the stdin and temp file parts of the specific test are to blame or if it's somehow related to what order `cargo` runs the tests in.

I'll try to take a closer look later when I get a chance, but I am honestly not even sure how to go from here.

Enselic: I can't reproduce any flakiness unfortunately. Not in WSL 2 nor in PowerShell.

@boyvanduuren Feel free to add an `#[ignore = "failing intermittently on some systems"]` in your PR for now.

tests/utils/mocked_pagers.rs
```diff
@@ -17,10 +17,20 @@ fn get_mocked_pagers_dir() -> PathBuf {
 /// On Unix: 'most' -> 'most'
 /// On Windows: 'most' -> 'most.bat'
 pub fn from(base: &str) -> String {
-    if cfg!(windows) {
-        format!("{}.bat", base)
-    } else {
-        String::from(base)
+    let mut cmd_and_args = base.split(" ");
```
Enselic, line 20: Why do we need to check for `" "` here? Isn't `base` only ever "more", "less", or "echo"?

Enselic: Nevermind, pretend you never saw this comment (I will delete it now)

tests/integration_tests.rs
boyvanduuren: All green.

sharkdp: @boyvanduuren Thanks

@Anomalocaridid: if I run that test "manually", I get an error message from `less`:
```
> export LESSOPEN="-echo empty.txt"
> export LESSCLOSE="echo lessclose: %s %s"
> echo "test.txt" | bat
LESSOPEN ignored: must contain exactly one %s
```

Could this be the issue? Maybe the race condition is somehow related to whether or not that error is printed first.

Anomalocaridid: @sharkdp Possibly? However, it looks like that error also appears when manually running the other tests where `$LESSCLOSE` has a dash as part of its prefix. But those do not have the same issue, right?

A [recent comment](https://github.com/sharkdp/bat/pull/2444#issuecomment-1704811314) on my PR for `$LESSOPEN` support gave me a helpful lead.

Basically, they said that the files were going through `$LESSOPEN` twice, through both `bat` and `less`. Sure enough, by running the test manually with `less` instead of bat, the same `LESSOPEN ignored` message appears. Also it may be worth mentioning that the message is not present in the code for bat's `$LESSOPEN` implementation.

So for some reason, the `$LESSOPEN` environment variable is being passed on to `less`, but only when `$LESSOPEN` has a dash in it that indicates that stdin is to be preprocessed as well as files.

When I get a chance, I'll try to fix both `$LESSOPEN` being passed to `less` as well as make `bat` more closely match `less`'s behavior for `$LESSOPEN` without a `%s`.

tests/utils/mocked_pagers.rs
```diff
@@ -17,10 +17,20 @@ fn get_mocked_pagers_dir() -> PathBuf {
...
+    let suffix = if cfg!(windows) { ".bat" } else { "" };
```
Enselic, line 21: Since we already depend on `shell_words::split` I think this code would end up simpler if we used it here too? If it ends up being worse, we can keep this as it is.

Enselic (changes requested):
Looks close to being mergeable, just had some more comments.

tests/integration_tests.rs
```diff
@@ -2095,6 +2139,7 @@ fn lessopen_and_lessclose_file_piped() {
 #[cfg(feature = "lessopen")]
 #[test]
 #[serial] // Randomly fails otherwise
+#[ignore = "randomly failing on some systems"]
```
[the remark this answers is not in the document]
boyvanduuren: Done

tests/utils/mocked_pagers.rs
boyvanduuren: Done, refactored it a bit further as well.

tests/utils/mocked_pagers.rs
```diff
@@ -17,10 +17,20 @@ fn get_mocked_pagers_dir() -> PathBuf {
...
+            let mut out_cmd = format!("{}{}", s, suffix);
+            let remainder: Vec<&str> = cmd_and_args.collect();
+            if (remainder.len() > 0) {
+                out_cmd.push_str(" ");
+                out_cmd.push_str(remainder.join(" ").as_str());
+            }
+
+            out_cmd
+        }
+        None => String::from(base),
```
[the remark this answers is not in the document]
boyvanduuren: Yes we do :-)

Enselic:
Looks good, thanks!

Issue #956, opened 2020-04-27
`std::fmt` traits compatible version of `PrettyPrinter::print`


yaahc:
As far as I can tell bat as a library only supports writing to stdout, I'm interested in seeing if I can integrate `bat` with `color-backtrace`/`color-spantrace` and `eyre`. To do this I need bat to write to a `std::fmt::Formatter` and return a `fmt::Result` instead of printing to `stdout` (i assume?) and returning a `bat::error::Error`.

So assuming I'm not misreading the docs...

## Proposal

Implement the `Display` trait for `PrettyPrinter` where `<PrettyPrinter as fmt::Display>::fmt` acts as an alternative to `PrettyPrinter::print`.

sharkdp:
Sounds good to me :+1: 

>  eth-p added the "good first issue" label

are you sure? :smile: I'm afraid that might be a bit more involved (especially the error handling part), but I could be wrong.

eth-p:
It might not be the *easiest* first issue out there, but it's a great way to learn Rust "by fire", so to speak 😂. The way everything comes together in `bat` is also pretty simple to understand, and it shouldn't take too long for someone to figure out how it works.

If you think we should remove the label because of how involved the process of integrating it would be, I wouldn't be against it though haha.

yaahc:
> are you sure?

I should note that I had to make similar changes to `color-backtrace` and there are ... problems ... that need to be resolved.

Specifically, I don't know what kind of windows support y'all have, but it could _not_ be fixed in a backwards compatible way for `color-backtrace` because old windows color support uses sys calls between prints which is not easily (at all?) compatible with `std::fmt`.

https://github.com/athre0z/color-backtrace/issues/27 for reference

A quick look at your toml only shows ansi color code deps so this may be a non issue.

eth-p:
> A quick look at your toml only shows ansi color code deps so this may be a non issue.

To do color, we enable ANSI support in the Windows terminal, so there shouldn't be any issues as far as side-effects go. The error handling might be a bit more complex, but capturing the output as a string (to write to the formatter) doesn't seem like it would require too much more than creating a new [`OutputType`](https://github.com/sharkdp/bat/blob/master/src/output.rs) for it.

gsomix:
Hey, I'd like to work on it. Could you please help me with advice: where to start or what might be needed to implement the feature? Thank you!

niklasmohrin:
Hey there,

[...]

PR #2618, opened 2023-07-09
feat: Optionally write to a buffer instead of stdout

Piturnah:
Following on from my comment in #956, this PR adds the ability to write output to a `std::fmt` buffer instead of stdout. It does this by changing the `handle` used by the printer to be a mutable reference to either one of `dyn fmt::Write` or `dyn io::Write`.

Included is also an added example to show that it works. I didn't add it to the `PrettyPrinter` but I can do if it seems like a good idea.

Anything else let me know!

### DIFF
--- a/examples/buffer.rs
+++ b/examples/buffer.rs
@@ -1,0 +1,17 @@
+use bat::{assets::HighlightingAssets, config::Config, controller::Controller, Input};
+
+fn main() {
+    let mut buffer = String::new();
+    let config = Config {
+        colored_output: true,
+        ..Default::default()
+    };
+    let assets = HighlightingAssets::from_binary();
+    let controller = Controller::new(&config, &assets);
+    let input = Input::from_file(file!());
+    controller
+        .run(vec![input.into()], Some(&mut buffer))
+        .unwrap();
+
+    println!("{buffer}");
+}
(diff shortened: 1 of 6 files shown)


keith-hall (approved):

sharkdp:
Can we please run a simple benchmark (e.g. `bat large-plaintext-file.txt`) to make sure this doesn't lead to any regression.

Piturnah:
|:---|---:|---:|---:|---:|
| `bat-main` | 797.0 ± 6.1 | 790.5 | 817.9 | 1.01 ± 0.01 |
| `bat-2618` | 792.9 ± 2.6 | 788.7 | 800.6 | 1.00 |

Both compiled in release mode, `bat-main` comes from 8676bbf97f2832ad2231e102ca9c9b7b72267fda (the commit this PR is based on)

5 warmup runs, 20 runs

Command used to generate the file:
`tr -dc "A-Za-z 0-9" < /dev/urandom | fold -w100|head -n 1000000 > large-plaintext-file.txt`

sharkdp:
Thank you very much!

> `tr -dc "A-Za-z 0-9" < /dev/urandom | fold -w100|head -n 1000000 > large-plaintext-file.txt`

:heart:

sharkdp:
I'm just rebasing to current master to see if the tests still pass.

PR #2619, opened 2023-07-10
docs: Add better snippet for colorizing help pages

chrisgrieser:


README.md
```diff
@@ -228,6 +228,15 @@ help() {
 
 Then you can do `$ help cp` or `$ help git commit`.
 
+Alternatively, you can also use global aliases to override `-h` and `--help` entirely:
+
+```bash
+alias -g -- -h='-h | bat --language=help --style=plain'
```
keith-hall, line 234: `alias -g` doesn't seem to be a valid option for me on Linux...

> `bash: alias: -g: invalid option
alias: usage: alias [-p] [name[=value] ... ]`

veltza:
I would also like to point out that the -h option is not standardized. It can mean anything. For example, -h means --human-readable in ls command.

README.md
sharkdp: `alias` is a shell command. The example here is probably zsh-specific?

I'm a zsh user, but I didn't know I could do that. That is pretty cool!

@chrisgrieser We should document that this is zsh-only. Are there any cases where this can fail?

To make this less "invasive", one could also use sth like `--bathelp`
```
alias -g -- --bathelp='--help | bat --language=help --style=plain'
```
or `--chelp` (colored help)
```
alias -g -- --chelp='--help | bat --language=help --style=plain'
```

sharkdp: yes there are. if a process writes its help text to stderr (`xxd`, for example).

We could use
```zsh
alias -g -- -h='-h 2>&1 | bat --language=help --style=plain'
```

chrisgrieser: all fixed!

chrisgrieser:
Good points everyone. All right, I pushed a commit which:
- adds `2>&1`
- mentions potential conflicts with `-h` meaning something other than `--help`
- mention that global aliases are only a zsh thing.

> I'm a zsh user, but I didn't know I could do that. That is pretty cool!
In fact, one of my favorite use cases for global aliases is `alias -g B="| bat"`, which makes piping to `bat` just so much more convenient considering how often I do it 🙂

PR #2620, opened 2023-07-14
Add install instructions for RHEL8 etc.

tkb-github:


Enselic:
Thank you for the contribution. Not sure about our docs recommending adding a personal repo tbh. It would be much better if you could get bat into some kind of official repo.

sharkdp:
Yes, I agree. Let's wait until it's officially supported. Thank you for your effort though, @tkb-github

tkb-github:
+1 for inclusion in EPEL8. In this case, since Fedora/EPEL9's version doesn't have the jemalloc issue, performance should be on par with the binary release. 

Someone will have to request it on Bugzilla, though, as Rust packages aren't automatically branched for EPEL8 (the issue referenced above has discussions). 

Until then, we'll keep maintaining our in-house COPR repo (which packages the binary release and isn't built from source like EPEL).

PR #2626, opened 2023-07-21
Replace unmaintained `encoding` dependency with `encoding_rs`

rhysd:
bat depends on `encoding` crate but it is [no longer maintained](https://github.com/lifthrasiir/rust-encoding/issues/127). This is reported as [RUSTSEC-2021-0153](https://rustsec.org/advisories/RUSTSEC-2021-0153) in the advisory DB. The advisory suggests [`encoding_rs` crate](https://crates.io/crates/encoding_rs) as alternative.

This PR replaces the unmaintained crate with the alternative.

As a side effect, this patch makes bat's UTF16 support a bit more efficient:

- Decoding with UTF16LE or UTF16BE is now handled with `Cow` instead of `String`. It means that it can avoid extra allocation by reusing the original string buffer when no character in the buffer was replaced while encoding.
- Removing BOM is now handled by `encoding_rs` more efficiently thanks to [`encode_with_bom_removal`](https://docs.rs/encoding_rs/latest/encoding_rs/struct.Encoding.html#method.decode_with_bom_removal) methods. The previous implementation always allocated new buffer only for removing BOM.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -130,7 +130,7 @@ dependencies = [
  "console",
  "content_inspector",
  "dirs",
- "encoding",
+ "encoding_rs",
  "expect-test",
  "flate2",
  "git2",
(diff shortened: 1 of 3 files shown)


sharkdp (approved):

Also great to see that this makes the code easier.

Issue #2633, opened 2023-07-31
Map ksh files to bash syntax


jwbowen:
If there are no objections to the idea, I'd like to add ksh to the syntax
mapping for bash highlighting. I was going to do it myself and just send
a PR, but it wasn't readily apparent what the best way to add it would be.

I'd still be happy to add it with a bit of guidance on where to make changes.

keith-hall:
For those that have never heard of `ksh` before, https://en.wikipedia.org/wiki/KornShell says:

> KornShell is backward-compatible with the Bourne shell

So no objection here to the mapping.  Probably the easiest way is to take inspiration from a previous PR adding a mapping:
https://github.com/sharkdp/bat/pull/2615/files

CosmicHorrorDev:
Looks like this can be closed now that #2638 is merged 👀

PR #2638, opened 2023-08-02
fix for issue #2633

johnmatthiggins:
Adds syntax mapping for KSH based on Bourne Again Shell. Solves issue #2633


keith-hall (approved):
Thanks for your contribution

Issue #2095, opened 2022-02-28
[windows] fzf integration sample from readme does not work


bw1faeh0:
Hi,

On windows the fzf integration sample does not work out of the box.

https://github.com/sharkdp/bat/blob/7be9485c3690799ee204ec5195172f9e90f49924/README.md?plain=1#L121

````
❯ fzf --preview 'bat --color=always --style=numbers --line-range=:500 {}'
invalid color specification: always
````

You have to use `"` instead of `'`:
````
❯ fzf --preview "bat --color=always --style=numbers --line-range=:500 {}"
````
Maybe you can leave a note for windows users.

sharkdp:
Thanks. The version with `"…"` should work just fine on Linux. There are no special shell characters that need escaping here. So we could just change this instead of adding an extra note for Windows users.

PR #2639, opened 2023-08-02
fix for issue #2095

johnmatthiggins:
Changed single quotes to double quotes in README.md fzf integration example for better compatibility across platforms. See issue #2095


keith-hall (approved):
Thanks for your contribution

PR #2641, opened 2023-08-05
Add environment variable `BAT_PAGING`

### LINKED ISSUE #2629: Bug?: disabling paging is impossible without overriding the pager
Currently, it is possible to disable paging in CLI via `bat --paging=never`. To make this configuration global, the help page / man pages suggest overriding the pager environment variable, `BAT_PAGER`. However, this behavior seems to be error-prone:

If `BAT_PAGER` is overridden, it is hard to "fix" it in CLI. Normally, env variables are overridden by CLI keys. E.g. with the standard CLI conventions, if you have `BAT_PAGING=never` and you'd run `bat --paging-never`, you'd get paging. With the current version of `bat` (`0.23.0 (871abd27)`), this is impossible.

This might be low-prio because I'm actually intending to set `bat` to non-paging mode by default and may not even switch back. I'm leaving this to the author's discretion. I'm also not sure if the issue is a "bug" (inconsistency with the standard CLI conventions), or simply a feature request. Triaging is welcomed

Thanks for the nice tool! :heart:

einfachIrgendwer0815:
Fixes #2629

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Features
 
+- Add environment variable `BAT_PAGING`, see #2629 (@einfachIrgendwer0815)
+
 ## Bugfixes
 
 ## Other
--- a/doc/long-help.txt
+++ b/doc/long-help.txt
@@ -89,8 +89,8 @@ Options:
 
       --paging <when>
           Specify when to use the pager. To disable the pager, use --paging=never' or its
-          alias,'-P'. To disable the pager permanently, set BAT_PAGER to an empty string. To control
-          which pager is used, see the '--pager' option. Possible values: *auto*, never, always.
+          alias,'-P'. To disable the pager permanently, set BAT_PAGING to 'never'. To control which
+          pager is used, see the '--pager' option. Possible values: *auto*, never, always.
 
       --pager <command>
           Determine which pager is used. This option will override the PAGER and BAT_PAGER
--- a/src/bin/bat/clap_app.rs
+++ b/src/bin/bat/clap_app.rs
@@ -311,7 +311,7 @@ pub fn build_app(interactive_output: bool) -> Command {
                 .long_help(
                     "Specify when to use the pager. To disable the pager, use \
                     --paging=never' or its alias,'-P'. To disable the pager permanently, \
-                    set BAT_PAGER to an empty string. To control which pager is used, see the \
+                    set BAT_PAGING to 'never'. To control which pager is used, see the \
                     '--pager' option. Possible values: *auto*, never, always."
                 ),
         )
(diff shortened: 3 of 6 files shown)


Enselic:
Thank you for the PR!

I'm tempted to approve this because of how straightforward and symmetrical it is.

But I don't quite understand what is wrong with the existing method to disable that pager with an env var: `BAT_PAGER= bat some-long-file.txt`. Could you explain a bit more perhaps?

einfachIrgendwer0815:
I think setting `BAT_PAGER=''` to disable paging is a somewhat indirect way of doing so. If you want to disable paging, you want to change the paging behaviour. But by using `BAT_PAGER` you are actually changing which pager to use. So, correct me if I'm wrong, you're basically saying to bat, "Do paging, but without a pager" which in turn forces bat not to use paging. So `BAT_PAGER` kind of controls two options here. But saying, "Don't do paging" by setting `BAT_PAGING=never` would make the intent clearer.

In addition, it is currently not possible to set `BAT_PAGER=''` to disable paging, but at the same time set a pager to use in case you exceptionally want paging. If there was `BAT_PAGING`, you could set `BAT_PAGING=never` to disable paging in general. But you could also set `BAT_PAGER` to the pager you want when exceptionally using paging. Then when using `bat --paging=always`, bat would use the pager set by `BAT_PAGER`.

Also, `BAT_PAGING=always` allows to set `--paging always` in the environment, which is currently not possible.

Enselic:
Makes sense! Would be great if you could a simple regression test for this change. A test that runs with BAT_BAGER set but uses BAT_PAGING=never to disable it. The regression test should also fail without the fix. Thanks!

einfachIrgendwer0815:
I have just come across this:
> https://github.com/sharkdp/bat/blob/6dd8b28b0d0569f876cc85390292f69fdc29e607/tests/utils/command.rs#L6-L21

> https://github.com/sharkdp/bat/blob/6dd8b28b0d0569f876cc85390292f69fdc29e607/tests/benchmarks/run-benchmarks.sh#L32-L43

> https://github.com/sharkdp/bat/blob/6dd8b28b0d0569f876cc85390292f69fdc29e607/tests/syntax-tests/create_highlighted_versions.py#L41-L53

Should `BAT_PAGING` be added?

Enselic:
You'll want to add a test to https://github.com/sharkdp/bat/blob/master/tests/integration_tests.rs, search for BAT_PAGER in that file

einfachIrgendwer0815:
> Makes sense! Would be great if you could a simple regression test for this change. A test that runs with BAT_BAGER set but uses BAT_PAGING=never to disable it. The regression test should fail without the fix. Thanks!

Unfortunately, that test doesn't fail without the fix applied because the tests don't run bat on an interactive terminal, which means the default is no paging, so setting BAT_PAGING=never doesn't change anything.

So instead the test I wrote sets BAT_PAGING=always to make sure this actually has an effect.

Enselic:
Thank you for the effort. Let's merge.

PR #2657, opened 2023-09-01
Enable extended features of `more` on Windows

mataha:
This is an attempt to bring user experience on Windows in parity with other systems as by default `more` can't even skip n arbitrary lines.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 - Output directory for generated assets (completion, manual) can be customized, see #2515 (@tranzystorek-io)
 - Bump MSRV to 1.70, see #2651 (@mataha)
+- `more` on Windows now has extended features enabled, see #2657 (@mataha)
 
 ## Syntaxes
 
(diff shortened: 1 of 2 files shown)


sharkdp:
For `less`, we do query the version of `less` in order to find out if a particular command line option is supported or not. Has this `more` option been available for a long time? Is there any risk of people still having old versions of `more` that didn't support `/E`?

mataha:
> Has this `more` option been available for a long time?

Since New Technology was a thing. I don't know about Windows NT, but 2000 and XP had that option for sure. I think it even predates Command Extensions.

> Is there any risk of people still having old versions of `more` that didn't support `/E`?

No. While some DOS implementations offer changing tab width only, those don't qualify as Windows.

sharkdp:
What about Windows users with mingw/cygwin/…? Is there any risk that they would use other `more` versions? Where `/E` doesn't exist or needs to be specified as `-E` or similar?

mataha:
> What about Windows users with mingw/cygwin/…? Is there any risk that they would use other `more` versions? Where `/E` doesn't exist or needs to be specified as `-E` or similar?

Yes. Though that's trivially solvable by checking its extension.

That said - I didn't realize it at first, but (thanks to a bug) extended features are by default always enabled, rendering this PR pointless...

PR #2659, opened 2023-09-05
Do not enable `lessopen` feature by default

Enselic:
See https://github.com/sharkdp/bat/pull/2444#issuecomment-1704802144


Enselic:
Looks like GitHub Actions is globally broken right now...

PR #2660, opened 2023-09-06
Make -pp override --paging and vice versa when passed as a later argument.

J-Kappes:
- should close #2128
- based on #2330
- added regression tests


tests/integration_tests.rs
```diff
@@ -832,6 +832,42 @@ fn disable_pager_if_disable_paging_flag_comes_after_paging() {
         .stdout(predicate::eq("hello world\n").normalize());
 }
 
+#[test]
+fn disable_pager_if_pp_flag_comes_after_paging() {
+    bat()
+        .env("PAGER", "echo pager-output")
```
Enselic, line 838: (Looks like this has a semantic merge conflict with https://github.com/sharkdp/bat/pull/2602, so I'll wait with reviewing this PR until that PR is merged and this PR is rebased on top)

Enselic: It has now been merged, so can you rebase on the main branch please?

tests/integration_tests.rs
```diff
@@ -873,6 +873,42 @@ fn disable_pager_if_disable_paging_flag_comes_after_paging() {
...
+        .arg("test.txt")
+        .assert()
+        .success()
+        .stdout(predicate::eq("hello world\n").normalize());
+}
+
+#[test]
+fn enable_pager_if_disable_paging_flag_comes_before_paging() {
+    bat()
+        .env("PAGER", "echo pager-output")
```
Enselic, line 891: Thank you for the rebase @J-Kappes .

@boyvanduuren I don't understand how this new test can pass on Windows. It is rebased on your work, so shouldn't this need `mocked_pagers::from()`? Is CI not working as we expect?

boyvanduuren: > Is CI not working as we expect?

Seems so indeed. Running the tests locally on Windows/PS cause them to fail. I'll create a bug for this and take another look.

Enselic:
I merged this for now, we can always debug CI problems later. Thank you.

Issue #1739, opened 2021-07-20
Intelligent support for `LESSOPEN` and `LESSCLOSE`


lilyball:
macOS has 3 different plist formats, two of them textual (NeXTStep and XML) and one binary. The textual plists are human-readable already, but the binary one isn't, and trying to run `bat` on it just warns me that it's a binary file.

What I'd really like is a way to configure `bat` to run plists through `plutil -p` to get a human-readable textual output. It could note that the file has been preprocessed this way in the header. And there should be an easy way to disable this (such as a flag that says "disable file preprocessors"). This would be similar to `LESSOPEN` (which has the flag `-L` to disable it), except defined on a per–file extension basis rather than being a global option that requires the script to decide what to do (this way `bat` can know and display whether the file has been preprocessed).

Optionally, this `plutil -p` preprocessing could be done only for binary files, this way running `bat` on an XML plist would show the XML. I'm not sure whether this is a good idea. This approach could then mean the flag to disable preprocessing could look like `--preprocess=always|binary|never`, though I suppose having that as a global option is unfortunate if I want to only preprocess binary plists but always preprocess some other file type. I'm not aware of any precedent for setting per-language options though (just for mapping file globs to syntax names). Perhaps the syntax to define a preprocessor could allow for selecting whether it should only apply to binary files. Or maybe the preprocessor itself should just be a script that makes that call, though in this case `bat` doesn't know whether the file has been modified (which seems like a very useful thing to know for displaying in the header). Or maybe this is just overcomplicating it and a preprocessor should just be defined for a file pattern and I can use the appropriate flag to disable it if I want to see the original file.

Side note: if `bat` ever becomes a pager in its own right instead of relying on `less` it could then have an interactive option to toggle the preprocessor off or on.

Also, while I've been talking about plists, I could imagine this ability being useful for e.g. displaying metadata about images or videos, or displaying some sort of summary of sqlite files, or other similar preprocessors.

### Alternatives

While a generic preprocessor feature seems useful, as an alternative `bat` could simply bake in knowledge of plists on macOS only (along with a macOS-only flag to disable plist transformations). It could either still just run them through `plutil -p`, or it could call into CoreFoundation to parse the plist directly, though I'm not aware offhand of any way to get the `plutil -p` output format that way without reimplementing it by hand.

lilyball:
[...]

PR #2662, opened 2023-09-06
Fix `$LESSOPEN` tests and behavior

Anomalocaridid:
This PR has a handful of fixes for the `$LESSOPEN` functionality that addresses concerns brought up in #2444 and #2602:

- Pass `--no-lessopen` to `less` to prevent it from preprocessing input a second file. ~~This should fix the intermittently failing test, assuming we correctly figured out the cause.~~ **Unfortunately, this did not appear to fix the intermittently failing test.**
- Make `$LESSOPEN` support opt-in rather than on by default. I added a `--lessopen` flag to enable this and kept the `--no-lessopen` flag for people that add `--lessopen` to their config file, but want to override it for whatever reason. Also changed the tests, man page, and completions accordingly.
- Require that `$LESSOPEN` has exactly one `%s`, otherwise ignoring it and with a warning message much like `less` does. Also changed the tests accordingly.
- Along with updating CHANGELOG.md, I took the liberty of moving the entry for the `$LESSOPEN` PR from v0.22.0 to the unreleased features section since it wasn't updated before it was merged recently.
- ~~I re-added the `lessopen` feature to `default` since you need to enable it with an option now anyways. I made sure to have this in its own commit so it can be easily removed if anyone objects to it.~~ Per @Enselic's [request](https://github.com/sharkdp/bat/pull/2662#discussion_r1318883698), I removed it so it has a chance to get more testing before being enabled by default.


Anomalocaridid:
In the absence of a better solution, I ignored `lessopen_and_lessclose_stdin_temp()`. I figure that if we find a better solution, we can just remove it, and if we do not find one before this gets merged, we'll probably opt for that instead anyways to prevent random pipeline failures.

CHANGELOG.md
```diff
@@ -3,10 +3,12 @@
 ## Features
 
 - Add environment variable `BAT_PAGING`, see #2629 (@einfachIrgendwer0815)
+- Add support for `LESSOPEN` and `LESSCLOSE`. See #1597, #1739, and #2444 (@Anomalocaridid)
 
 ## Bugfixes
 
 - Fix `more` not being found on Windows when provided via `BAT_PAGER`, see #2570, #2580, and #2651 (@mataha)
+- Fix `less` preprocessing with `LESSOPEN` again and causing a test to intermittently fail, and incorrect `LESSOPEN` behavior see #2444, #2602, and #2662 (@Anomalocaridid)
```
Enselic, line 11: This was was never present in a release version of bat, right? In that case I don't think we should have this as a bugfix. But feel free to list the PRs in the "Features" item.

src/lessopen.rs
```diff
@@ -32,10 +35,18 @@ enum LessOpenKind {
 
 impl LessOpenPreprocessor {
     /// Create a new instance of LessOpenPreprocessor
-    /// Will return Ok if and only if $LESSOPEN is set
+    /// Will return Ok if and only if $LESSOPEN is set and contains exactly one %s
```
Enselic, line 38: What do you mean with "exactly one %s"?

tests/integration_tests.rs
```diff
@@ -2213,12 +2256,49 @@ fn lessopen_uses_shell() {
...
+        .arg("test.txt")
+        .assert()
+        .success()
+        .stdout("hello world\n")
+        .stderr(
+            "\u{1b}[33m[bat warning]\u{1b}[0m: LESSOPEN ignored: must contain exactly one %s\n",
+        );
+
+    bat()
+        .env("LESSOPEN", "|echo File is %s")
```
Enselic, line 2286: I am completely unfamiliar LESSOPEN and with %s. Can you link to a good online resource where I can learn more please?

Enselic (changes requested):
Thank you for the quick update. I'm not sure about needing `--lessopen` to enable. Long term we want to let the `LESSOPEN` env var decide if the feature is on or off, I think? Or could someone want `LESSOPEN` set but not want `bat` to process it?

Anomalocaridid:
> Thank you for the quick update. I'm not sure about needing `--lessopen` to enable. Long term we want to let the `LESSOPEN` env var decide if the feature is on or off, I think? Or could someone want `LESSOPEN` set but not want `bat` to process it?

Actually, there is a use case where one would want `LESSOPEN` but not want `bat` to use it. Specifically, I added `--lessopen` to address [this concern](https://github.com/sharkdp/bat/pull/2444#issuecomment-1704802144) that using `bat` in `LESSOPEN` breaks `bat` with infinite recursion.

Of course, if you still have reservations, we could try alternative approaches, like [this suggestion](https://github.com/sharkdp/bat/issues/1739#issuecomment-987352129) to use separate variables like `BATOPEN`. (Although in my opinion, the "special case syntax" for having `bat` use `LESSOPEN` is redundant and can be replaced with normal shell variable assignment.)

Or maybe something else entirely if you have any suggestions.

tests/integration_tests.rs
Anomalocaridid: I would recommend the [`INPUT PREPROCESSOR` section of the `less(1)` man page](https://www.man7.org/linux/man-pages/man1/less.1.html#INPUT_PREPROCESSOR). It is what I used to implement `$LESSOPEN` support.

src/lessopen.rs
Anomalocaridid: Let's say you want to use `batpipe` as to preprocess `less` or `bat`'s input.

`batpipe %s` is a valid command for `$LESSOPEN`.

The following are not valid and will result in `less`/`bat` ignoring it and printing a warning:

- `batpipe`
- `batpipe %s %s`
- `batpipe %s %s (...) %s`

CHANGELOG.md
Anomalocaridid: I moved the PRs to the "Features" item. How is it now?

Cargo.toml
```diff
@@ -21,6 +21,7 @@ application = [
     "build-assets",
     "git",
     "minimal-application",
+    "lessopen",
```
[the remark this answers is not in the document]
Anomalocaridid: Now that you mention it, that seems like a much better option. I have removed the commit that re-enabled the `lessopen` feature by default.

Enselic (approved):

Enselic:
Thanks! Let's merge this.

sharkdp:
Thank you @Anomalocaridid the update and @Enselic for the review.

Anomalocaridid:
Thank both of you as well!

PR #2677, opened 2023-09-29
[docs] Remove duplicate winget

connorjs:
Removes duplicate winget instructions for Windows install.

### DIFF
--- a/README.md
+++ b/README.md
@@ -423,13 +423,6 @@ You can install `bat` via [scoop](https://scoop.sh/):
 scoop install bat
 ```
 
-#### With Winget
-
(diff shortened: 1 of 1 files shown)


connorjs:
The difference lies in preference for `-e` (exact). The important part is that it’s already qualified (not just `fd`), so I’m happy to keep either given your preference.

I always use `-e --id`, but personal preference to avoid typos when not copying.

connorjs:
Switched to shorter one, but kept it in the first position (for Windows) + “WinGet” casing. I generally see WinGet, Choco, and Scoop as the order.

Enselic (approved):
Thanks!

Issue #2675, opened 2023-09-27
Search feature searches on line numbers


dmell:
**What steps will reproduce the bug?**

1. open a file with `bat`
2. search for a number, e.g. `4`, using `/`

**What happens?**

`bat` researches the pattern also in the title and line numbers (so 4, 14, 24 etc. will be considered results).

**What did you expect to happen instead?**

I expected to only find results in the content of the file.

**How did you install `bat`?**

`sudo dpkg -i <latest release>`

**bat version and environment**

#### Software version

bat 0.23.0 (871abd2)

#### Operating system

Linux 5.15.0-84-generic

#### Command-line

```bash
bat test --diagnostic 
```

#### Environment variables

```bash
SHELL=/usr/bin/zsh
PAGER=less
LESS=-R
LANG=en_US.UTF-8
LC_ALL=<not set>
BAT_PAGER=<not set>
BAT_CACHE_PATH=<not set>
BAT_CONFIG_PATH=<not set>
BAT_OPTS=<not set>
BAT_STYLE=<not set>
BAT_TABS=<not set>
BAT_THEME=<not set>
XDG_CONFIG_HOME=<not set>
XDG_CACHE_HOME=<not set>
COLORTERM=truecolor
NO_COLOR=<not set>
MANPAGER=<not set>
```

#### System Config file

Could not read contents of '/etc/bat/config': No such file or directory (os error 2).

#### Config file

Could not read contents of '/home/damiano/.config/bat/config': No such file or directory (os error 2).

#### Custom assets metadata

Could not read contents of '/home/damiano/.cache/bat/metadata.yaml': No such file or directory (os error 2).

#### Custom assets

'/home/damiano/.cache/bat' not found

#### Compile time information

- Profile: release
- Target triple: x86_64-unknown-linux-gnu
- Family: unix
- OS: linux
- Architecture: x86_64
- Pointer width: 64
- Endian: little
- CPU features: fxsr,sse,sse2
- Host: x86_64-unknown-linux-gnu

#### Less version

```
> less --version 
less 551 (GNU regular expressions)
Copyright (C) 1984-2019  Mark Nudelman

less comes with NO WARRANTY, to the extent permitted by law.
For information about the terms of redistribution,
see the file named README in the less distribution.
Home page: http://www.greenwoodsoftware.com/less
```

PR #2689, opened 2023-10-01
fix for issue #2675

tdtrung17693:
Adding the arguments for `less` to fix the issue #2675 :
- `--headers` for the headers and line numbers
- `--no-search-headers` to ignore the headers when searching


sharkdp:
Thank you very much — this looks great! Didn't know `less` had those new features. Do you think it would be possible to write (integration) tests for? I think we mock pager calls in some other tests as well. We could do something similar and check that the mocked `less` pager is called with the correct offsets.

tdtrung17693:
Ok. Let me try that.

Enselic:
Hi! Did you have time to look into writing integration tests?

Enselic:
Closing for now to keep PR inbox trimmed, but feel free to re-open once you get around writing a test!

PR #2690, opened 2023-10-01
Updated version of `serde_yaml` to 0.9

### LINKED ISSUE #2627: Update for serde-yaml 0.9
Hi.

In debian we are currently working on updating the serde-yaml crate, while we do have a mechanism for packaging multiple versions of a crate it is something we try to avoid where possible. So I have been testing reverse-dependencies of serde-yaml with the dependency bumped.

After bumping the dependency, bat itself built fine and it's tests passed. However one of the examples needed a small patch to avoid using a removed function.

https://salsa.debian.org/rust-team/debcargo-conf/-/blob/3a24d2e9344d0a204bab9af9bc9cadfb375f81fc/src/bat/debian/patches/serde-yaml-0.9.diff

Raghav-Bell:
It fixes #2627

### DIFF
--- a/examples/yaml.rs
+++ b/examples/yaml.rs
@@ -23,7 +23,8 @@ fn main() {
         }],
     };
 
-    let bytes = serde_yaml::to_vec(&person).unwrap();
+    let mut bytes = Vec::with_capacity(128);
+    serde_yaml::to_writer(&mut bytes, &person).unwrap();
     PrettyPrinter::new()
         .language("yaml")
         .line_numbers(true)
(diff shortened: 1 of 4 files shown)


keith-hall (approved):
Thanks for your contribution!

examples/yaml.rs
```diff
@@ -23,7 +23,8 @@ fn main() {
         }],
     };
 
-    let bytes = serde_yaml::to_vec(&person).unwrap();
```
Enselic, line 26: I don't think we need to change this code? Better keep it as it is to reduce risk IMHO. Or do we have good reason to change it? I seem to be able to build with serde_yaml 0.9 without this change.

sharkdp (approved):

examples/yaml.rs
Enselic: Sorry, my bad, I missed that this was just changing example code. I can reproduce the build problem when I build this example. Thanks for fixing.

PR #2692, opened 2023-10-02
Add support for WebGPU Shader Language

rhysd:
I'm recently learning [WebGPU](https://www.w3.org/TR/webgpu/). It introduces its own shader language [WGSL](https://www.w3.org/TR/WGSL/).

This PR adds the syntax highlight support for WGSL. [The VSCode extension](https://marketplace.visualstudio.com/items?itemName=PolyMeilex.wgsl) was installed 25267 times at the time of writing this PR. Though the language is not so popular yet, it will become popular since it is defined by long live W3C Web Standards. It would be great to have support for the language by bat.


Enselic (approved):
Looks good, thanks!

PR #2693, opened 2023-10-03
Fix syntax tests failure on macOS during the daytime

rhysd:
I'm setting 'Auto' appearance mode in the OS settings. It automatically sets the light mode during the daytime and switches to the dark mode at night.

image

Since syntax tests use `default` theme, bat automatically switches the default color theme looking at the OS settings on macOS. However syntax tests assume the theme is always dark. So the tests fail during the daytime in my environment.

This PR fixes the issue by setting `Monokai Extended` (which is the default color theme in dark mode) explicitly.

### DIFF
--- a/tests/syntax-tests/create_highlighted_versions.py
+++ b/tests/syntax-tests/create_highlighted_versions.py
@@ -8,11 +8,13 @@ import os
 import argparse
 from multiprocessing import Pool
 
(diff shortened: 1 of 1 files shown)


keith-hall (approved):

Enselic:
Thanks!

PR #2695, opened 2023-10-04
Associate `*.sarif` files with `json` syntax

rhysd:
[SARIF](https://docs.oasis-open.org/sarif/sarif/v2.1.0/csprd01/sarif-v2.1.0-csprd01.html) is a format for reporting static analysis results. It is [used by GitHub CodeQL](https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning?learn=code_security_integration&learnProduct=code-security) for example.

Here are some samples from Microsoft's VSCode extension:

https://github.com/microsoft/sarif-vscode-extension/tree/main/samples

The SARIF format is built on top of JSON so the `.sarif` files can be highlighted as JSON. This PR adds support for the file extension.


Enselic:
Thank you

PR #2696, opened 2023-10-04
Replace our own macros to write buffers with standard macros

rhysd:
`printer.rs` defines `write_handle` and `writeln_handle` for `OutputHandle` but actually we can reuse standard `write` and `writeln` macros by defining `write_fmt` method for the struct. Starndard macros are better in terms of readability since they are well-known. In contrast defining our own macros means that readers need to know how they are expanded.


sharkdp (approved):
Thank you very much. This looks much better

PR #2697, opened 2023-10-04
Update the ssh-config submodule

### LINKED ISSUE #2359: Incomplete syntax highlighting for ssh config?
_Sorry if this is a duplicate, I could not find an issue that matched this._

There seem to be a missing keyword match for `User` while parsing ssh configurations:
Screenshot_20221006_100551

While not explicitly referenced in [ssh_config(5)](https://man.archlinux.org/man/ssh_config.5), it is mentioned in [the Arch Linux wiki](https://wiki.archlinux.org/title/OpenSSH#Configuration) and does appear to work:

```~ ∙ ssh -v zarya 2>&1 | grep DOES_THIS                                                                         130 ↵
debug1: Control socket "/home/xenhat/.ssh/sockets/DOES_THIS_WORK/" does not exist
unix_listener: cannot bind to path /home/xenhat/.ssh/sockets/DOES_THIS_WORK/.iZ9SgH1lGBRJVscA: No such file or directory
```

Something to note, however: neovim also fails to apply the ssh-specific parsing:
Screenshot_20221006_100805

So after all, I'm not sure if this actionable at all, but I thought I should mention it.

mrmeszaros:
Pull in the fix for the syntax.
Fixes #2359


sharkdp:
Thank you. Can we maybe update `tests/syntax-tests/source/SSH Config/ssh_config` to include a `User` section? See https://github.com/sharkdp/bat/blob/master/doc/assets.md#syntax-tests for details

mrmeszaros:
Did you think something like this?

mrmeszaros:
Got it to work! Steps:

1. Run `cargo build` and add `./target/debug` to the PATH
2. Run `assets/create.sh` - this will update the compiled `syntaxes.bin`
3. Run `cargo build` to include the changed `syntaxes.bin` in the built `bat` binary
4. Run `tests/syntax-tests/update.sh` to update the regression tests (with the just built `bat` binary)

sharkdp:
@mrmeszaros Thank you!

michaelblyons:
Congrats!

PR #2698, opened 2023-10-04
Fix arithmetic overflow in `LineRange::parse_range`

### LINKED ISSUE #2674: Arithmetic overflow occurs while using API LineRange::from()
I executed fuzz testing with `bat` public APIs and found crash case.

~~~
let fuzz_arg1 = "18446744073709551615:+1";
LineRange::from(fuzz_arg1);
~~~

**What happens?**

~~~
Thread '<unnamed>' panicked at 'attempt to add with overflow', /rustc/871b5952023139738f72eba235063575062bc2e9/library/core/src/ops/arith.rs:109
~~~

The overflow occurs with this statement : 

https://github.com/sharkdp/bat/blob/5a240f36b9251fd6513316333acb969a4d347b9d/src/line_range.rs#L56

Need assertion to prevent arith overflow, or need to explicitly mention panic condition.

skoriop:
Fixes #2674

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 - Fix `more` not being found on Windows when provided via `BAT_PAGER`, see #2570, #2580, and #2651 (@mataha)
 - Switched default behavior of `--map-syntax` to be case insensitive #2520
 - Updated version of `serde_yaml` to `0.9`. See #2627 (@Raghav-Bell)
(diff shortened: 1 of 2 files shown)


sharkdp (approved):
Thank you. I think this is a reasonable fix for this pathological edge case, even if it doesn't lead to "correct" behavior. But avoiding the panic is probably a good idea.

Even better would be to show some error message, but that would probably require larger-scale changes. And it's probably not worth it for this case.

PR #2712, opened 2023-10-15
Add unicode/caret replacements for all control characters

einfachIrgendwer0815:
With PR #2443 printing non-printable characters using caret notation was added besides the already in existence printing using Unicode Control Pictures. However, both applied only for six control characters: LF, CR, NUL, BEL, BS and ESC.

This PR adds replacements for the remaining control characters, including DEL. To achieve this in a code-efficient way, the previous if-else logic is replaced by a calculation of the respective replacement character based on the character that is to be replaced.


keith-hall (approved):
I quite like this solution, and it seems useful, so it gets my approval. We'll see what the other maintainers think.
Feel free to add a changelog entry for it. :)

Enselic:
Looks good! With the new https://github.com/sharkdp/bat/blob/master/doc/release-checklist.md this will get an entry in the CHANGELOG later so I think we can merge. I have not done a detailed review of the diff of the highlighted file, but if both of you think it looks god, then I'm sure it's good.

PR #2750, opened 2023-10-29
fix(completions): add missing fish completions

MariaSolOs:
As advertised by the title :)


Enselic (approved):
I'm too lazy to verify the change, but it does look reasonable. Thanks!

MariaSolOs:
> I'm too lazy to verify the change, but it does look reasonable. Thanks!

Fair enough haha

PR #2755, opened 2023-10-31
More extensible syntax mapping mechanism

### LINKED ISSUE #2747: Need a more sustainable way to manage known syntax mappings
Currently everything is done in `src/syntax_mapping.rs` using bespoke code for each syntax mapping we want to add, which is starting to become a problem.

The `SyntaxMapping::builtin` function is already 160 lines long, and is nowhere near complete. As a daily user of `bat`, on average I run into this situation several times a week when I'm thinking, "man, this well-known file path really should have syntax mapping built in". But then I remember how it's implemented and think, "I really don't want to make the maintainability any worse than it already is".

What I'm saying is, there has to be a better way to code this to make it more maintainable and extensible (specifically, easier to add more mappings and harder to accidentally add duplicates). So I would like to ask the community for your opinion (how to organise the mappings, what patterns/libraries to use, etc.). I'm happy to help implement the changes if we come to some technical consensus.

cyqsimon:
Will close #2747.

## Checklist

- [x] Ratify specification
- [x] Implementation
- [x] Add tests


cyqsimon:
@keith-hall would you please take a look at `README.md` and see if there are any obvious problems with this architecture? Thanks.

src/syntax_mapping/builtins/README.md
```diff
@@ -0,0 +1,108 @@
...
+# 99-unset-ambiguous-extensions.toml
+[mappings]
+"MappingTarget::MapExtensionToUnknown" = [
+    "*.conf",
+]
+```
+
+## Ordering
+
+At compile time, all TOML files are processed in filesystem order. So
```
keith-hall, line 98: "filesystem order" could potentially be construed ambiguously, for example I think I've seen before that listing files in a folder can give different results on a Mac compared to Linux. So perhaps clearer to say "lexicographical filename order"

keith-hall:
I don't see any problems with this approach 👍

cyqsimon:
I'll leave it here for a week or two for people to take a look. If there are no major objections I will go ahead with the implementation.

Enselic:
I like the idea in general. My main concern is startup time, but in the linked issues you are talking about doing clever stuff at compile time, which has the potential to even improve startup time.

I don't think we can at this point "we promise to include your solution in bat", but I am very interested in looking at a prototype and evaluating that, then we can decide on the next steps.

I hope am not sounding discouraging! Would be great if we could solve syntax mapping more elegantly.

cyqsimon:
I was investigating the possibility of doing this with proc macros, because I suspected that would make the code somewhat more elegant.  In conclusion I decided that using a build script is probably still the way to go (at least for the moment). This comment is just to document my research; I will add more info if they are worthy of documenting.

1. I initially thought I would be using [phf_codegen](https://docs.rs/phf_codegen/), but I soon realised that a hashmap would be poorly suited for this job because glob patterns are not appropriate keys. 
2. Seems like someone has tried to do similar (unsurprisingly) and [has written an article about it](https://dev.to/rustyoctopus/generating-static-arrays-during-compile-time-in-rust-10d8). It is a bit old (2020-09) but most things in it are still valid.
3. I tried to look for a crate that does similar to `phf_codegen` but for arrays. Unfortunately I was not able to find any, but honestly I suspect it's because it's just too simple.

cyqsimon:
Okay, I have an initial prototype of the codegen part working. Right now it makes something like this:
```rust
// I formatted it manually to make it easier for the eye
static STATIC_RULES: [(&str, MappingTarget); 6] = [
    (
        r#"${XDG_CONFIG_HOME}/git/config"#,
        MappingTarget::MapTo(r#"Git Config"#),
    ),
    (
        r#"${HOME}/.config/git/config"#,
        MappingTarget::MapTo(r#"Git Config"#),
    ),
    (
        r#"${XDG_CONFIG_HOME}/git/ignore"#,
        MappingTarget::MapTo(r#"Git Ignore"#),
    ),
    (
        r#"${HOME}/.config/git/ignore"#,
        MappingTarget::MapTo(r#"Git Ignore"#),
    ),
    (
        r#"${XDG_CONFIG_HOME}/git/attributes"#,
        MappingTarget::MapTo(r#"Git Attributes"#),
    ),
    (
        r#"${HOME}/.config/git/attributes"#,
        MappingTarget::MapTo(r#"Git Attributes"#),
    ),
];
```

I'm thinking, I can actually take this one step further by moving the string searches to compile time.

cyqsimon:
Okay, I think I have a decent implementation of the aforementioned idea. This is what the generated code currently looks like:
```rust
// formatted
static STATIC_RULES: [(BuiltinMatcher, MappingTarget); 7] = [
    (
        BuiltinMatcher::Dynamic(Lazy::new(|| {
            join_segments(&[
                MatcherSegment::Env(r#"XDG_CONFIG_HOME"#),
                MatcherSegment::Text(r#"/git/config"#),
            ])
        })),
        MappingTarget::MapTo(r#"Git Config"#),
    ),
    (
        BuiltinMatcher::Dynamic(Lazy::new(|| {
            join_segments(&[
                MatcherSegment::Env(r#"HOME"#),
                MatcherSegment::Text(r#"/.config/git/config"#),
            ])
        })),
        MappingTarget::MapTo(r#"Git Config"#),
    ),
    (
        BuiltinMatcher::Dynamic(Lazy::new(|| {
            join_segments(&[
                MatcherSegment::Env(r#"XDG_CONFIG_HOME"#),
                MatcherSegment::Text(r#"/git/ignore"#),
            ])
        })),
        MappingTarget::MapTo(r#"Git Ignore"#),
    ),
    (
        BuiltinMatcher::Dynamic(Lazy::new(|| {
            join_segments(&[
                MatcherSegment::Env(r#"HOME"#),
                MatcherSegment::Text(r#"/.config/git/ignore"#),
            ])
        })),
        MappingTarget::MapTo(r#"Git Ignore"#),
    ),
    (
        BuiltinMatcher::Dynamic(Lazy::new(|| {
            join_segments(&[
                MatcherSegment::Env(r#"XDG_CONFIG_HOME"#),
                MatcherSegment::Text(r#"/git/attributes"#),
            ])
        })),
        MappingTarget::MapTo(r#"Git Attributes"#),
    ),
    (
        BuiltinMatcher::Dynamic(Lazy::new(|| {
            join_segments(&[
                MatcherSegment::Env(r#"HOME"#),
                MatcherSegment::Text(r#"/.config/git/attributes"#),
            ])
        })),
        MappingTarget::MapTo(r#"Git Attributes"#),
    ),
    (
        BuiltinMatcher::Fixed(r#"*.conf"#),
        MappingTarget::MapExtensionToUnknown,
    ),
];
```

The downside is that this uses a couple of new types and functions, which doesn't feel quite right wherever I put them:
- They are only used by the generated code, so I put them in `src/syntax_mapping.rs`, future maintainers will probably find it odd/confusing.
- They are static code blocks, so they don'

cyqsimon:
Rebased onto `master`.

Enselic:
I skimmed over the code and it looks like you know what you are doing. Nothing jumped out at me as a red flag. I didn't look very close at the code though.

I don't want to waste your time, but how much work would it be to port all of the existing syntax mappings to the new system? So that we can see how it impacts performance. Or maybe you can partly port it? So that half of syntax mappings use new and half use old. Maybe that is enough to git a sense for how this impacts performance.

To check performance, do this (approximate instructions):

```
cargo install --locked hyperfine
cargo build --release
tests/benchmarks/run-benchmarks.sh --release
```

cyqsimon:
Thanks for the compliment.

I'm not ready to start porting just yet - currently the codegen is working, but the `include!`-ed mappings are not yet being used. I need to add that bit of code first; shouldn't be too difficult though.

As of the porting, there's not actually that much to port. Maybe 20 files (which can be done in like 30 minutes). I'll run the benchmark after it's all done.

cyqsimon:
Impl is all done. Tests are failing because mappings haven't been ported. Will do said porting tomorrow.

src/syntax_mapping.rs
```diff
@@ -29,204 +40,63 @@ pub enum MappingTarget<'a> {
...
+    ///
+    /// The `GlabMatcher`s are lazily compiled.
+    ///
+    /// If a mapping rule requires an environment variable that is unset, it
+    /// will be ignored.
+    pub fn builtin_mappings(&self) -> impl Iterator<Item = (&GlobMatcher, &MappingTarget<'a>)> {
+        BUILTIN_MAPPINGS
+            .iter()
+            .filter_map(|(matcher, target)| matcher.as_ref().map(|glob| (glob, target)))
     }
```
cyqsimon, line 88: Anyone familiar with lifetime downcasts?

Is this possible / easy to do?

cyqsimon: Okay, found a hack. `Iterator::map` seems to downcast lifetimes. Why and how I have no clue.

cyqsimon: I asked in the community Discord and got an explanation:

> It's not that Iterator::map is "downcasting lifetimes", it's just that impl Iterator<Item = T> is invariant in T, so you need a closure to "do" the variance

https://discord.com/channels/273534239310479360/1120124565591425034/1170543402870382653

cyqsimon: I asked in the community Discord and got an explanation:

> It's not that Iterator::map is "downcasting lifetimes", it's just that impl Iterator<Item = T> is invariant in T, so you need a closure to "do" the variance

https://discord.com/channels/273534239310479360/1120124565591425034/1170543402870382653

cyqsimon:
I ran the benchmark like so (using patch in #2768):

```bash
cd "$(git rev-parse --show-toplevel)/tests/benchmarks"
rm -rf results
mkdir -p results/{pre,post}

git checkout master
cargo build -r
RUN_COUNT=50 ./run-benchmarks.sh --release
cp benchmark-results/report.md results/pre/

git checkout syntax-mapping-refactor
cargo build -r
RUN_COUNT=50 ./run-benchmarks.sh --release
cp benchmark-results/report.md results/post/

bat results/*/*.md
```

### Current `master`

|:---|---:|---:|---:|---:|
| `bat` | 3.9 ± 0.9 | 2.4 | 7.8 | 1.00 |
| `bat … small-CpuInfo-file.cpuinfo` | 20.4 ± 1.7 | 18.6 | 28.4 | 1.00 |
| `bat … small-Markdown-file.md` | 25.4 ± 1.8 | 23.5 | 34.1 | 1.00 |
| `bat … --language=txt numpy_test_multiarray.py` | 5.4 ± 2.4 | 3.7 | 43.3 | 1.00 |
| `bat … grep-output-ansi-sequences.txt` | 38.6 ± 1.8 | 36.3 | 44.0 | 1.00 |
| `bat … jquery.js` | 520.4 ± 5.8 | 514.5 | 533.6 | 1.00 |
| `bat … miniz.c` | 50.1 ± 1.7 | 48.1 | 56.0 | 1.00 |
| `bat … numpy_test_multiarray.py` | 688.9 ± 5.0 | 683.3 | 697.1 | 1.00 |
| `bat … grep-output-ansi-sequences.txt` | 36.1 ± 2.3 | 33.3 | 45.2 | 1.00 |
| `bat … jquery.js` | 515.6 ± 5.0 | 511.1 | 527.1 | 1.00 |
| `bat … miniz.c` | 49.6 ± 2.3 | 47.6 | 59.6 | 1.00 |
| `bat … numpy_test_multiarray.py` | 688.0 ± 6.1 | 680.7 | 697.3 | 1.00 |
| `bat … --language=txt *.txt` | 4.6 ± 1.0 | 3.3 | 8.8 | 1.00 |

### `syntax-mapping-refactor`

|:---|---:|---:|---:|---:|
| `bat` | 2.2 ± 0.7 | 1.1 | 5.6 | 1.00 |
| `bat … small-CpuInfo-file.cpuinfo` | 20.3 ± 1.6 | 18.6 | 28.2 | 1.00 |
| `bat … small-Markdown-file.md` | 25.9 ± 1.6 | 24.3 | 31.7 | 1.00 |
| `bat … --language=txt numpy_test_multiarray.py` | 3.6 ± 1.1 | 2.2 | 8.3 | 1.00 |
| `bat … grep-output-ansi-sequences.txt` | 38.3 ± 1.4 | 36.6 | 45.6 | 1.00 |
| `bat … jquery.js` | 522.3 ± 4.7 | 517.0 | 532.5 | 1.00 |
| `bat … miniz.c` | 48.9 ± 2.5 | 46.4 | 59.8 | 1.00 |
| `bat … numpy_test_multiarray.py` | 690.9 ± 4.2 | 685.4 | 700.7 | 1.00 |
| `bat … grep-output-ansi-sequences.txt` | 36.4 ± 1.8 | 34.1 | 42.3 | 1.00 |
| `bat … jquery.js` | 517.1 ± 5.6 | 513.0 | 532.4 | 1.00 |
| `bat … miniz.c` | 48.8 ± 1.8 | 46.1 | 53.6 | 1.00 |
| `bat … numpy_test_multiarray.py` | 689.0 ± 3.7 | 683.8 | 694.9 | 1.00 |
| `bat … --language=txt *.txt` | 3.1 ± 1.1 | 1.7 | 7.8 | 1.00 |


---

Unsurprisingly, the simpler the task, the bigger the speedup. For heavier tasks there is barely a difference.

cyqsimon:
One thing I was somewhat concerned about was a possible increase of the binary size due to the pervasive use of `once_cell::sync::Lazy`. Fortunately it seems like the increase is not too alarming.

```bash
cd "$(git rev-parse --show-toplevel)"
rm -rf res
mkdir -p res/{pre,post}

git checkout master
cargo build -r
cp target/release/bat res/pre/

git checkout syntax-mapping-refactor
cargo build -r
cp target/release/bat res/post/

du --summarize --bytes res/*/bat
```

```
5379928	res/post/bat
5359128	res/pre/bat
```

Eh. 20KB difference seems fine.

cyqsimon:
This is what the codegen looks like right now (Linux). I think this is pretty much what we will end up with:

```rust
/// Generated by build script from /src/syntax_mapping/builtins/.
pub(crate) static BUILTIN_MAPPINGS: [(Lazy<Option<GlobMatcher>>, MappingTarget); 56] = [
(Lazy::new(|| Some(build_matcher_fixed(r#"httpd.conf"#))), MappingTarget::MapTo(r#"Apache Conf"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"/etc/apache2/**/*.conf"#))), MappingTarget::MapTo(r#"Apache Conf"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"/etc/apache2/sites-*/**/*"#))), MappingTarget::MapTo(r#"Apache Conf"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"**/bat/config"#))), MappingTarget::MapTo(r#"Bourne Again Shell (bash)"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"Containerfile"#))), MappingTarget::MapTo(r#"Dockerfile"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"*.h"#))), MappingTarget::MapTo(r#"C++"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#".clang-format"#))), MappingTarget::MapTo(r#"YAML"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"*.fs"#))), MappingTarget::MapTo(r#"F#"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"fish_history"#))), MappingTarget::MapTo(r#"YAML"#)),
(Lazy::new(|| build_matcher_dynamic(&[MatcherSegment::Env(r#"XDG_CONFIG_HOME"#), MatcherSegment::Text(r#"/git/config"#)])), MappingTarget::MapTo(r#"Git Config"#)),
(Lazy::new(|| build_matcher_dynamic(&[MatcherSegment::Env(r#"HOME"#), MatcherSegment::Text(r#"/.config/git/config"#)])), MappingTarget::MapTo(r#"Git Config"#)),
(Lazy::new(|| build_matcher_dynamic(&[MatcherSegment::Env(r#"XDG_CONFIG_HOME"#), MatcherSegment::Text(r#"/git/ignore"#)])), MappingTarget::MapTo(r#"Git Ignore"#)),
(Lazy::new(|| build_matcher_dynamic(&[MatcherSegment::Env(r#"HOME"#), MatcherSegment::Text(r#"/.config/git/ignore"#)])), MappingTarget::MapTo(r#"Git Ignore"#)),
(Lazy::new(|| build_matcher_dynamic(&[MatcherSegment::Env(r#"XDG_CONFIG_HOME"#), MatcherSegment::Text(r#"/git/attributes"#)])), MappingTarget::MapTo(r#"Git Attributes"#)),
(Lazy::new(|| build_matcher_dynamic(&[MatcherSegment::Env(r#"HOME"#), MatcherSegment::Text(r#"/.config/git/attributes"#)])), MappingTarget::MapTo(r#"Git Attributes"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"*.jsonl"#))), MappingTarget::MapTo(r#"JSON"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"*.ksh"#))), MappingTarget::MapTo(r#"Bourne Again Shell (bash)"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"/var/spool/mail/*"#))), MappingTarget::MapTo(r#"Email"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"/var/mail/*"#))), MappingTarget::MapTo(r#"Email"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"nginx.conf"#))), MappingTarget::MapTo(r#"nginx"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"mime.types"#))), MappingTarget::MapTo(r#"nginx"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"/etc/nginx/**/*.conf"#))), MappingTarget::MapTo(r#"nginx"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"/etc/nginx/sites-*/**/*"#))), MappingTarget::MapTo(r#"nginx"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"*.nse"#))), MappingTarget::MapTo(r#"Lua"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"/etc/os-release"#))), MappingTarget::MapTo(r#"Bourne Again Shell (bash)"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"/usr/lib/os-release"#))), MappingTarget::MapTo(r#"Bourne Again Shell (bash)"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"/etc/initrd-release"#))), MappingTarget::MapTo(r#"Bourne Again Shell (bash)"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"/usr/lib/extension-release.d/extension-release.*"#))), MappingTarget::MapTo(r#"Bourne Again Shell (bash)"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"/usr/share/libalpm/hooks/*.hook"#))), MappingTarget::MapTo(r#"INI"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"/etc/pacman.d/hooks/*.hook"#))), MappingTarget::MapTo(r#"INI"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"*.pac"#))), MappingTarget::MapTo(r#"JavaScript (Babel)"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"*.ron"#))), MappingTarget::MapTo(r#"Rust"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"*.sarif"#))), MappingTarget::MapTo(r#"JSON"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"/etc/profile"#))), MappingTarget::MapTo(r#"Bourne Again Shell (bash)"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"**/.ssh/config"#))), MappingTarget::MapTo(r#"SSH Config"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"**/systemd/**/*.conf"#))), MappingTarget::MapTo(r#"INI"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"**/systemd/**/*.example"#))), MappingTarget::MapTo(r#"INI"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"*.automount"#))), MappingTarget::MapTo(r#"INI"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"*.device"#))), MappingTarget::MapTo(r#"INI"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"*.dnssd"#))), MappingTarget::MapTo(r#"INI"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"*.link"#))), MappingTarget::MapTo(r#"INI"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"*.mount"#))), MappingTarget::MapTo(r#"INI"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"*.netdev"#))), MappingTarget::MapTo(r#"INI"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"*.network"#))), MappingTarget::MapTo(r#"INI"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"*.nspawn"#))), MappingTarget::MapTo(r#"INI"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"*.path"#))), MappingTarget::MapTo(r#"INI"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"*.service"#))), MappingTarget::MapTo(r#"INI"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"*.scope"#))), MappingTarget::MapTo(r#"INI"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"*.slice"#))), MappingTarget::MapTo(r#"INI"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"*.socket"#))), MappingTarget::MapTo(r#"INI"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"*.swap"#))), MappingTarget::MapTo(r#"INI"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"*.target"#))), MappingTarget::MapTo(r#"INI"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"*.timer"#))), MappingTarget::MapTo(r#"INI"#)),
(Lazy::new(|| Some(build_matcher_fixed(r#"*.conf"#))), MappingTarget::MapExtensionToUnknown),
(Lazy::new(|| Some(build_matcher_fixed(r#"build"#))), MappingTarget::MapToUnknown),
(Lazy::new(|| Some(build_matcher_fixed(r#"rails"#))), MappingTarget::MapToUnknown)
];
```

cyqsimon:
Maintainers, can you please take a look at my porting? I've classified the rules according to their applicable platforms; please help check if all mappings are in their correct subdirs. Thanks.

cyqsimon:
Hmm. Seems like `rusty-fork` has issues with arm targets. And looking at the repo, it looks all but abandoned. Are there any good alternatives? If not I will fork.

Enselic:
Thank you for the updates. The status here is that I intend to do a somewhat detailed review of this, and play around a bit with the code. Since that is a substantive effort I can't say when I will get the time and energy to do it. I also hesitate a bit to put too much effort into this myself before sharkdp has given this his blessing. But he is currently busy on other projects (last time I spoke to him) and I don't want to ping him in when it is not urgent.

In short: don't worry if someone does not look immediately at this, it _will_  be done eventually, but it might take longer than a few days.

Enselic:
I began playing around with the code a bit. I tried this:

```sh
hyperfine -L bat target/release/bat-master,target/release/bat-new -L color always,never "{bat} --no-config --color={color} tests/examples/single-line.txt"
```

<details>
<summary>I get these numbers</summary>

```
Benchmark 1: target/release/bat-master --no-config --color=always tests/examples/single-line.txt
  Time (mean ± σ):       4.4 ms ±   0.2 ms    [User: 4.0 ms, System: 0.3 ms]
  Range (min … max):     4.0 ms …   5.4 ms    628 runs

  Warning: Command took less than 5 ms to complete. Note that the results might be inaccurate because hyperfine can not calibrate the shell startup time much more precise than this limit. You can try to use the `-N`/`--shell=none` option to disable the shell completely.

Benchmark 2: target/release/bat-new --no-config --color=always tests/examples/single-line.txt
  Time (mean ± σ):       4.4 ms ±   0.1 ms    [User: 4.1 ms, System: 0.4 ms]
  Range (min … max):     4.1 ms …   5.3 ms    598 runs

  Warning: Command took less than 5 ms to complete. Note that the results might be inaccurate because hyperfine can not calibrate the shell startup time much more precise than this limit. You can try to use the `-N`/`--shell=none` option to disable the shell completely.

Benchmark 3: target/release/bat-master --no-config --color=never tests/examples/single-line.txt
  Time (mean ± σ):       1.3 ms ±   0.1 ms    [User: 1.3 ms, System: 0.1 ms]
  Range (min … max):     1.2 ms …   2.2 ms    1640 runs

  Warning: Command took less than 5 ms to complete. Note that the results might be inaccurate because hyperfine can not calibrate the shell startup time much more precise than this limit. You can try to use the `-N`/`--shell=none` option to disable the shell completely.
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Benchmark 4: target/release/bat-new --no-config --color=never tests/examples/single-line.txt
  Time (mean ± σ):       0.6 ms ±   0.1 ms    [User: 0.6 ms, System: 0.0 ms]
  Range (min … max):     0.5 ms …   1.4 ms    2981 runs

  Warning: Command took less than 5 ms to complete. Note that the results might be inaccurate because hyperfine can not calibrate the shell startup time much more precise than this limit. You can try to use the `-N`/`--shell=none` option to disable the shell completely.
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Summary
  target/release/bat-new --no-config --color=never tests/examples/single-line.txt ran
    2.35 ± 0.32 times faster than target/release/bat-master --no-config --color=never tests/examples/single-line.txt
    7.71 ± 0.95 times faster than target/release/bat-master --no-config --color=always tests/examples/single-line.txt
    7.83 ± 0.95 times faster than target/release/bat-new --no-config --color=always tests/examples/single-line.txt
```

</details>

What puzzles me is that startup time seems greatly improved with your code with --color=never, but there seems to be no difference with --color=always. I would expect the speed improvement to be noticeable also with --color=always. Do you know why this is not the case?

cyqsimon:
I know why this is the case, and in fact I believe this is the expected result for most input files.

From a high-level perspective, the reason this code saves any time at all with syntax highlighting disabled is that it skipped building the `GlobMatcher`s, which is the expensive step. However, this is not achieved via building them at compile time, rather by putting them in lazy cells which are not evaluated unless iterated.

With syntax highlighting enabled though, we necessarily have to iterate through all the rules (or until a rule hits) to see if any matches. Since most files do not hit, this means evaluating all the lazy cells, building all `GlobMatcher`s in the process. This case is functionally no different than the old implementation, which explains the similar performance.

As of why building `GlobMatcher`s cannot happen at compile time, this is due to how the `globset` crate works. It actually uses Regex underneath, which is a state machine so has to be built at run time as far as I know. If there's a better way I'd love to know about it.

However, there are a few cases where you should see a performance improvement with syntax highlighting enabled:
1. if the input file does hit a built-in rule (particularly if said rule is ordered near the front of the list):
```sh
bat Containerfile --color=always
```
2. when using `--language` to set a syntax override:
```sh
bat foo --color=always --language=ini
```
3. when using `--map-syntax` to set a custom mapping, and it hits:
```sh
bat bar --color=always --map-syntax 'bar:ini'
```
In all three cases the performance improvement comes from skipping building some or all `GlobMatcher`s.

Cargo.toml
```diff
@@ -67,6 +67,7 @@ bytesize = { version = "1.3.0" }
 encoding_rs = "0.8.33"
 os_str_bytes = { version = "~6.6", optional = true }
 run_script = { version = "^0.10.1", optional = true}
+rusty-fork = "0.3.0"
```
Enselic, line 70: This can be  a dev-dependency I think?

Enselic (commented):
Makes sense, thanks! I really like that your solution improves performance in some situations, even if not in all situations.

I had a drive-by comment, but I still haven't done a detailed review of the code. But when I glance over it it looks very reasonable.

src/syntax_mapping.rs
```diff
@@ -251,10 +128,76 @@ impl<'a> SyntaxMapping<'a> {
...
+        // fixed builtin mappings will panic if they fail to compile
+        let _mappings = map.builtin_mappings().collect::<Vec<_>>();
+    }
+
+    // disabled for ARM builds because of an issue with `rusty_fork`
+    #[cfg(not(any(target_arch = "arm", target_arch = "aarch64")))]
+    // lazy initialisation test needs to be run on a separate instance because
+    // it will race with other tests
+    // see: https://github.com/rust-lang/rust/issues/47506
+    rusty_fork_test! {
```
Enselic, line 159: So, I looked in more detail at this test and  have some concerns. The first concern relates to the rusty-fork crate itself, which has not been updated in 3 years and thus seems unmaintained. That it does not work on arm also indicates that it is unmaintained.

Luckily I think we don't need rusty-fork, because I don't think we need this test. Don't get me wrong, I am a big fan of regression tests. But I am also a fan of long-term maintainability. And in my experience, one big factor in how to minimize long-term maintenance is to write tests that do not depend on implementations details.

This particular test is closely tied to implementation details, because it tests *how* fast startup is ensured in certain cases, rather than testing *that* fast startup is possible.

But, performance tests are tricky to run in a free CI where runners come in different capacities. The best mechanism we currently have to compare performance is tests/benchmarks/run-benchmarks.sh. It is not great, but IMHO it is better than tests that tests implementation details. So my preference would be to remove this test (and the rusty-fork depenency) and instead set up a test in tests/benchmarks/run-benchmarks.sh that looks something like this:

* measure the startup performance of the bat binary when run in a way where lazy mappings are not needed
* measure the startup performance of the bat binary when run in a way where all mappings are needed
* test that the bat binary starts faster in the first case than in the second case

This is a good test because it does not care *how* fast startup is achieved, only *that* it is achieved. We can completely change the implementation without changing the test at all.

I hope you don't find this comment discouraging! Keep up the good work :)

(I have not yet had time to look at any other code in the PR in detail)

cyqsimon:
I've got a rough benchmark prototype implemented, but it's giving me some surprising results. 

So I would like to bring in the changes in #2768 onto this branch to help me figure out what's wrong. @Enselic how would you prefer I do this? Rebase or merge?

Enselic:
Feel free to rebase since I have not started a detailed review yet 👍

cyqsimon:
Here's the benchmark prototype, but I am unable to figure out how to make the results predictable/reliable.

The problem is that the results swing about very erratically when run consecutively, to the point where any of the three tests could be the fastest or the slowest. I suspect there's something to do with CPU caching. Maybe `hyperfine` has some way of mitigating the influence of this, but I doubt it's going to be reliable.

Can you please try running these benchmarks several times to see whether this problem is reproducible? Thanks all.

Cargo.lock
```diff
@@ -17,6 +17,21 @@ dependencies = [
  "memchr",
 ]
 
+[[package]]
+name = "android-tzdata"
```
Enselic, line 21: How come we need android-tzdata in the lock file? In general there are many new dependencies that seem not obviously needed. Maybe we can disable some features of the new dependencies?

(I figured it would be better to ask than figuring out for myself, because I never manage to find the time to figure it out by myself.)

cyqsimon: It's an indirect dependency of `serde_with`'s `std` feature, although curiously it refuses to show up in `cargo tree` regardless of what flags I pass it.

Anyways, it's indeed unnecessary. I've updated the dependency declaration of `serde_with` to exclude it along with a few others.

cyqsimon:
Force pushed to trigger CI rerun.

cyqsimon:
Okay something weird is going on with the CI. I don't think this is my fault?

Enselic:
I think the CI failure will be fixed by https://github.com/sharkdp/bat/pull/2811.

Cargo.toml
```diff
@@ -100,6 +100,14 @@ nix = { version = "0.26.4", default-features = false, features = ["term"] }
 
 [build-dependencies]
 anyhow = "1.0.75"
+indexmap = { version = "2.1.0", features = ["serde"] }
```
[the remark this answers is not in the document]
cyqsimon: > Unfortunately I have quite a lot on my plate right now, so it's unclear when I will get the time to continue review.

No pressure! I'm in no rush.

> In general, the fewer dependencies the better, so I'm wondering if we need to depend on indexmap? Is it needed to make syntax mappings deterministic maybe?

Yes it is exactly as you stated. Without `IndexMap`, there would be no guarantee that a rule defined in front of another rule would be matched first, if their targets are different. For example:

```toml
[mappings]
"INI" = [ "foo0", "foo1" ]
"XML" = [ "bar0", "bar1" ]
```

Without `IndexMap`, it would still be guaranteed that `foo0` comes before `foo1` (because they are serialised as a `Vec`), but `bar0` and `bar1` may end up in front of `foo0` and `foo1`.

CHANGELOG.md
```diff
@@ -15,13 +15,20 @@
 - Minor benchmark script improvements #2768 (@cyqsimon)
 - Update Arch Linux package URL in README files #2779 (@brunobell)
 - Update and improve `zsh` completion, see #2772 (@okapia)
+- More extensible syntax mapping mechanism #2755 (@cyqsimon)
+  - [BREAKING] Precedence order of user-defined syntax mappings (using `-m/--map-syntax`) has been reversed
+    - `-m '*.foo:alpha' -m '*.foo:bravo'` now maps `bar.foo` to `alpha`
```
sharkdp, line 20: I think I've read through the majority of the communication on this changeset, but I haven't seen any reason for this. Can you remind me why this is necessary? I would personally prefer it to have the old behavior, I think. Typically, later CLI arguments overwrite previous ones.

tests/benchmarks/run-benchmarks.sh
```diff
@@ -117,6 +124,49 @@ hyperfine \
...
+for KIND in indeterminant manually-set mapped; do
+	JSON_SRC="$RESULT_DIR/startup-time-with-$KIND-syntax.json"
+	AVG=$(jq -r '.results[0].mean' "$JSON_SRC")
+	AVG_TIMES+=("$AVG")
+done
+# indeterminant should be slower, because it necessarily has to evaluate all rules
+# to ensure that nothing matches; manually-set and mapped should both be faster
+# because most or all GlobMatcher builds are skipped
+python3 -c "if ${AVG_TIMES[0]} < ${AVG_TIMES[1]}: print('WARN: indeterminant syntax has faster startup than manually set syntax!')"
+python3 -c "if ${AVG_TIMES[0]} < ${AVG_TIMES[2]}: print('WARN: indeterminant syntax has faster startup than mapped syntax!')"
```
sharkdp, line 167: Is this something we want to keep in here permanently? If so, maybe we could move it to a more specialized benchmarking script? What do you think?

src/syntax_mapping/builtins/README.md
```diff
@@ -0,0 +1,114 @@
+# `/src/syntax_mapping/builtins`
```
sharkdp, line 1: Thank you for this excellent piece of documentation :heart_eyes:

build/syntax_mapping.rs
```diff
@@ -0,0 +1,294 @@
...
+/// Parse a matcher.
+///
+/// Note that this implementation is rather strict: it will greedily interpret
+/// every valid environment variable replacement as such, then immediately
+/// hard-error if it finds a '$', '{', or '}' anywhere in the remaining text
+/// segments.
+///
+/// The reason for this strictness is I currently cannot think of a valid reason
+/// why you would ever need '$', '{', or '}' as plaintext in a glob pattern.
+/// Therefore any such occurrences are likely human errors.
```
sharkdp, line 61: :+1:

CHANGELOG.md
cyqsimon: > but I haven't seen any reason for this.

Yeah. This was a conscious decision by me to do so. My reasoning was that it maintains consistency with the ordering of the builtin mappings. Although when I come to think about it, the user doesn't really know or care about the implementation, so this justification is a bit far-fetched 🤔.

Okay yeah you've convinced me. I'll restore the old behaviour.

tests/benchmarks/run-benchmarks.sh
sharkdp: Ok, so I did some benchmarks myself. Measuring time differences on the order of ~1 millisecond is challenging. There are two main things that I usually do: you need to work on a quiet machine. Close applications if possible. Definitely close your browser. It makes a huge difference.

Second, you can run a `--shell=none` benchmark if the command is not too fancy (doesn't use any shell syntax, variables, etc). This allows hyperfine to run the command directly, without going through an intermediate shell (and having to correct for that by measuring the shell spawning time).

Finally, for benchmarks you always need to look at the full distribution if the noise level is this high. Sometimes it's more useful to look at the *minimum*, or the *median*, not at the mean — because it's influenced by outliers.

Here are my results using

```
hyperfine \
	--shell=none \
	--warmup 200 \
	--runs 2000 \
	-L version master,2755 \
	--export-json=indeterminant.json \
	'./bat-{version} --no-config mystery-file'

hyperfine \
	--shell=none \
	--warmup 200 \
	--runs 2000 \
	-L version master,2755 \
	--export-json=manually-set.json \
	'./bat-{version} --no-config mystery-file --language=Dockerfile'

hyperfine \
	--shell=none \
	--warmup 200 \
	--runs 2000 \
	-L version master,2755 \
	--export-json=mapped.json \
	'./bat-{version} --no-config Containerfile --language=Dockerfile'
```

and plotted using hyperfine's `plot_histogram.py` script:

indeterminate:

image

manually set:

image

mapped:

image


I think it's clear that there is a (statistically) significant advantage!

sharkdp: Also: note that bat (and a lot of other programs) change their behavior when the output is piped to /dev/null, which is hyperfine's default setting. For this benchmark, you'll get different results when running with `bat`s `--color=always` or hyperfine's `--output=inherit` option.

They still show an advantage for this branch (see below), except for the "indeterminate" benchmark — where they are on par.

I think those are pretty much the results you expected?

indeterminant:

image

manually set:

image

image

cyqsimon: This is great to know! Thanks for the help.

> For this benchmark, you'll get different results when running with `bat`s `--color=always` or hyperfine's `--output=inherit` option.

Ah yes good catch. Without `--color=always`, I imagine syntax mappings wouldn't need to be checked at all, which means all 3 benchmarks would behave identically. Coincidentally this is still relevant in verifying that lazy evaluation works, just not what I intended to test.

I've added `--color=always` to these 3 benchmarks, and created another test named (uncreatively) "Startup time without syntax highlighting" as a comparison against "Startup time with syntax highlighting".

> I think those are pretty much the results you expected?

Yes indeed. Just to correct a detail - for the third test (mapped), you shouldn't specify `--language=Dockerfile`. With this option included, you are essentially repeating the second test (manually-set). That being said, I do expect the results to be very similar.

> Is this something we want to keep in here permanently? If so, maybe we could move it to a more specialized benchmarking script? What do you think?

After some thought, I realised this bit of comparison somewhat pointless. After all, the point of this script is to compare across commits, not to compare across different usage scenarios within the same commit. I've decided to remove this section altogether.

sharkdp (approved):

sharkdp:
Thank you for the update

cyqsimon:
> A more advanced approach might even make use of that (the short circuiting) and evaluate batches of patterns at once (in parallel), but return early once a match is found. Worst case, this would evaluate num_threads-1 patterns that would not have been needed.

I was thinking about this proposal, and I actually don't think it's as intimidating as it sounds. All this requires is a custom iterator adapter to manage the thread pool and the message passing; the caller would require very few changes.

I'll try to write a prototype today or in the next few days.

sharkdp:
Maybe the thread spawning overhead is actually so high that it's not worth it. Starting a thread pool typically also takes a few milliseconds.

tests/benchmarks/run-benchmarks.sh
cyqsimon: Out of curiosity, how did you @sharkdp plot this particular style of graph (sorry I don't know its proper name)? I'm trying to benchmark my multithreaded offloading prototype but couldn't find the appropriate option in `plot_histogram.py`.

PR #2756, opened 2023-10-31
Build script refactor

cyqsimon:
The original intent behind this PR is to accommodate for #2747 (PR: #2755), which requires significant code additional to the build script. Hence it would make sense to clean up the build script beforehand and make it organizationally amenable to said addition (i.e. splitting the code into multiple files and modules). There should be no functional changes.

While doing so, I also did some maintenance work:
- cleaned out the old compgen code that's unlikely to ever be used again
- use `anyhow` to make life a bit easier
- fix inaccurate use of the `option_env!` and `env!` macros (the environment variables should be read at build script run time, not build script compile time)
- use newer code style, specifically
  - implicit variable capture in format string ([since 1.58.0](https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html#captured-identifiers-in-format-strings))
  - use `[].into_iter().collect()` to immutably construct a `HashMap` ([since 1.56.0 & 2021 edition](https://doc.rust-lang.org/edition-guide/rust-2021/IntoIterator-for-arrays.html)) 
- only rerun build script when relevant source files/env change

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -75,6 +75,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anyhow"
+version = "1.0.75"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a4668cab20f66d8d020e1fbc0ebe47217433c1b6c8f2040faf858554e394ace6"
+
+[[package]]
 name = "assert_cmd"
 version = "2.0.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


Enselic (approved):
LGTM! I will merge this after the weekend unless other maintainers raise concerns.

keith-hall (approved):

Enselic:
With two maintainer approvals, let's merge already now. I don't want to hold you back unnecessarily.

PR #2766, opened 2023-11-03
CI: add changelog check

cyqsimon:
This workflow ensures that every PR has a corresponding entry in CHANGELOG.md that includes the PR number and the submitter ID.

### DIFF
--- a/.github/workflows/require-changelog-for-PRs.yml
+++ b/.github/workflows/require-changelog-for-PRs.yml
@@ -1,0 +1,24 @@
+name: Changelog
+
+on:
+  pull_request:
+
+jobs:
+  check-changelog:
+    name: Check for changelog entry
+    runs-on: ubuntu-latest
+    env:
+      PR_SUBMITTER: ${{ github.actor }}
+      PR_NUMBER: ${{ github.event.number }}
+      PR_BASE: ${{ github.base_ref }}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Fetch PR base
+        run: git fetch --no-tags --prune --depth=1 origin
+      - name: Search for added line in changelog
+        run: |
+          ADDED=$(git diff -U0 "origin/${PR_BASE}" HEAD -- CHANGELOG.md | grep -P '^\+[^\+].+$')
+          echo "Added lines in CHANGELOG.md:"
+          echo "$ADDED"
+          echo "Grepping for PR info:"
+          grep "#${PR_NUMBER}\\b.*@${PR_SUBMITTER}\\b" <<< "$ADDED"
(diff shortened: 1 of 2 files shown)


cyqsimon:
I apologise for the messy commit log. You may want to use squash merge 😅.

Enselic:
No worries :) Feel free to force-push to fix it! After a first code review it could be could to avoid force push though to make incremental review easier.

This seems like a good idea, but we need to make an exception for auto-merged dependabot PRs. Example: https://github.com/sharkdp/bat/pull/2761. Either based on branch (starts with `dependabot/...` or based on that the author is dependabot. We handle this by drafting release notes right before a release: https://github.com/sharkdp/bat/blob/master/doc/release-checklist.md#changelogmd-updates. But it's nice if the CHANGELOG is prepared as much as possible anyway.

cyqsimon:
Rebased on `master`.

cyqsimon:
`git rebase -i master` and squashed some messy commits.

cyqsimon:
Ah shit. I think I messed up. `branches-ignore:` refers to the target branch, not the source branch.

.github/workflows/require-changelog-for-PRs.yml
```diff
@@ -0,0 +1,26 @@
...
+
+on:
+  pull_request:
+
+jobs:
+  check-changelog:
+    name: Check for changelog entry
+    runs-on: ubuntu-latest
+    # dependabot PRs are automerged if CI passes; we shouldn't block these
+    if: github.actor == 'dependabot[bot]'
```
Enselic, line 11: You want "not equal" here, don't you?

Enselic:
This is hard to verify before it's merged I suppose? So I'm tempted to merge it after you fix the comment so we can verify and iterate on it.

Enselic:
Oh wait, right, it runs already (after you fix the bug)

.github/workflows/require-changelog-for-PRs.yml
cyqsimon: 🤦. Fixed.

cyqsimon:
> This is hard to verify before it's merged I suppose? So I'm tempted to merge it after you fix the comment so we can verify and iterate on it.

I made sure this workflow errors properly before I added the changelog for this PR. So if it passes now (as it should), I think it should be mostly fine.

Enselic:
@cyqsimon Hi! Any chance you could fix a bug in the script? Right now it uses the last commit in a PR to determine author. That will give the wrong author if a maintainer adds a commit to the PR. We likely need to use `gh pr view 2779 --json author | jq -r .author.login` or something similar. Thanks in advance!

Example of where the check fail when it shouldn't: https://github.com/sharkdp/bat/pull/2779

cyqsimon:
Arrrrgh. There's always an edge case that slips by. I'll take a look.

PR #2767, opened 2023-11-04
fix: display color when `NO_COLOR` is an empty string

acuteenvy:
https://no-color.org

> Command-line software which adds ANSI color to its output by default should check for a NO_COLOR environment variable that, when present ***and not an empty string*** (regardless of its value), prevents the addition of ANSI color.

The current behavior of `bat` is to check whether or not `NO_COLOR` is set, and even if it's an empty string, it still disables color.


keith-hall (approved):
Looks good to me. Probably worth adding an integration test for (see `tests/integration_tests.rs`) :)

acuteenvy:
> Looks good to me. Probably worth adding an integration test for (see `tests/integration_tests.rs`) :)

I tried, but it's quite hard to do - checking if standard output is a tty disables color in every test regardless of `NO_COLOR`.

src/bin/bat/app.rs
```diff
@@ -207,7 +207,10 @@ impl App {
                 || match self.matches.get_one::<String>("color").map(|s| s.as_str()) {
                     Some("always") => true,
                     Some("never") => false,
-                    Some("auto") => env::var_os("NO_COLOR").is_none() && self.interactive_output,
```
Enselic, line 210: Thanks! Would be very nice if you could add a helper function to src/bin/bat so we don't have to duplicate this logic 🙏

Enselic (approved):
Looks good, thanks!

PR #2772, opened 2023-11-12
Improvements and fixes to zsh completion

okapia:
* Use correct return status so that approximate completion isn't broken. Zsh completions should not return 0 if they fail to add matches. Supporting this typically involves a local variable named `ret` as in the patch. Someone was complaining on a forum about this not working which is what motivated me to take a look 
* Follow zsh conventions on description forms. Not using capitalisation is the convention going back a long time and most people agree that consistency is more important than any preference one way or another. There was also some things which look like copy-paste from help output. We don't put types in angle brackets, e.g. <mode> - users can configure that if they like it and references to something like N only make sense in help output where N also appears in a syntax summary.
* Some options can be used multiple times, e.g. -H, -r, --file-name. That's the initial `*` in the spec.
* Set completion context correctly for the cache subcommand. This matters for configuration with `zstyle`.
* Better completion for --map-syntax argument
* Add --nonprintable-notation option. There's also -S/--chop-long-lines but I'm not sure whether these are deprecated forms of `--wrap`.
* Correct some of the mutual exclusion lists for options. Options don't need to exclude themselves - that's the default. And some options were excluding `:` instead of `*` which doesn't make sense with `_files` being called from a `*` spec - perhaps that was reflecting an old form the function took.

I also put in a fallback to complete "cache" if `_files` fails. I think this addresses the concerns from the old linked github issue while allowing `bat cac` to complete to 'cache' if there aren't already files. The normal zsh approach would actually be to use `_alternative` and allow the user to configure it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Refactor and cleanup build script #2756 (@cyqsimon)
 - Checks changelog has been written to for PRs in CI #2766 (@cyqsimon)
 - Minor benchmark script improvements #2768 (@cyqsimon)
+- Update and improve `zsh` completion, see #2772 (@okapia)
 
 ## Syntaxes
 
--- a/assets/completions/bat.zsh.in
+++ b/assets/completions/bat.zsh.in
@@ -1,19 +1,20 @@
 #compdef {{PROJECT_EXECUTABLE}}
 
-local context state state_descr line
+local curcontext="$curcontext" ret=1
+local -a state state_descr line
 typeset -A opt_args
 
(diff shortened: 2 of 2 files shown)


keith-hall (approved):
The usual caveat of me not being a ZSH user and not knowing much about completions applies, but the PR description indicates you know what you are doing, and the diff seems reasonable, so you get my approval. Thanks for your contribution. Note that in order to pass CI checks, you'll need to [add a changelog entry](https://github.com/sharkdp/bat/blob/master/CONTRIBUTING.md#add-an-entry-to-the-changelog).

okapia:
Thanks, as requested, I've added a changelog entry.

PR #2780, opened 2023-11-22
Add context-around-highlight option

dipeshkaphle:
The motivation behind this PR is mostly for my own convenience. 

I use ripgrep + fzf a lot in my workflow(where I use bat for previewing result). 

I have one function in my zshrc which looks like this:

image

So when I search for something using `ff nvim`, I get something like this:

image

I was quite frustrated by this because I have context of what's happening after the matching line but not before. 
This could probably be done by using some very very hacky script but I just thought adding this feature to bat might be more easier.

I'm making this PR to solve that issue because it might be helpful to other folks too.

After this change and making some change to my zsh function, I get something like this for preview
image

Feel free to close this if you think this won't be useful or is not needed.

If you think it'll be useful, I'd be very happy to make necessary changes for this to be merged

### DIFF
--- a/src/bin/bat/app.rs
+++ b/src/bin/bat/app.rs
@@ -7,6 +7,7 @@ use crate::{
     clap_app,
     config::{get_args_from_config_file, get_args_from_env_opts_var, get_args_from_env_vars},
 };
+use bat::line_range;
 use clap::ArgMatches;
 
 use console::Term;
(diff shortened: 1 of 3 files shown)


Enselic:
Just a quick drive-by comment:

It would help a lot to understand what this is doing if you added a regression test. It also ensure we don't regress on this feature.

IMPORTANT NOTE: A regression test might not be sufficient to make this mergable at the PR might still be rejected (I want to be frank). Maybe clear step-by-step instructions on how to reproduce the problem could make it clearer when this change is useful? Is it possible for you to solve this by adjusting your script maybe? Or using a custom pager?

Again, this is a just a quick drive-by comment rather than keep saying nothing. I haven't looked very close at the code for example.

keith-hall:
Note that we have recently merged https://github.com/sharkdp/bat/pull/3345 which I believe offers similar functionality, just without a new argument - hopefully it suits your use case too 🙂

Enselic:
Assuming it did. Closing.

PR #2782, opened 2023-11-23
Visualization of hexadecimal RGB color values (#2705)

### LINKED ISSUE #2705: Render visualization of hexadecimal colors (or other common formats) using true color ANSI escape sequences
Hi there. This feature request might be a bit unorthodox, but I'm sharing it in case you like it.

**Example:** Given a file like this:

image

Render an output like this:

image

See also this [video demo](https://www.youtube.com/watch?v=y542zPAPeG4).

Here's an implementation example of the above, on top of `cat`:

```bash
#!/usr/bin/env bash

# All my gist code is licensed under the MIT license.

# Video demo: https://www.youtube.com/watch?v=y542zPAPeG4

# colorcat
# - cats a file, but if any line contains N hex colors, it appends the colors
#   (rendered as ansi escape sequences) to the end of the line.
# - input can be stdin, a file, or a hex color in plain text
if [[ "$#" -eq 1 && ! -f "$1" ]]; then
  echo "$1"
else
  cat "$@"
fi | while read -r line; do
  local colors=""
  for word in $line; do
    if [[ "$word" =~ ^[^A-Fa-f0-9]*#?([A-Fa-f0-9]{6})[^A-Fa-f0-9]*$ ]]; then
      hex=${BASH_REMATCH[1]}
      local r=$((16#${hex:0:2}))
      local g=$((16#${hex:2:2}))
      local b=$((16#${hex:4:2}))
      local truecolor="\033[48;2;${r};${g};${b}m"
      local reset="\033[0m"
      colors="${colors}${truecolor}  ${reset} "
    fi
  done
    echo -e "$line $colors" 
done
```

For `bat`, rendering in the gutter next to the line number might be preferred.

Thanks.

einfachIrgendwer0815:
Adds a new option (`--render-hex-colors`) that renders hexadecimal RGB color values.

An example of this can be seen by running:
```bash
cargo run -- --render-hex-colors tests/examples/hex_colors.txt
```

### Limitations

 - color values must be in the form of #rrggbb or #rgb (e.g. #f2a023 or #fa2)
 - only works on true-color terminals

Resolves #2705.


sharkdp:

I'm very sorry, but I'd rather not merge custom code for a particular (edge) use case into `bat`. If we can implement this as a custom syntax, great. But if we need a modification to `bat` itself for this feature, I think it would be better left to other, more specialized tools.

PR #2791, opened 2023-12-01
Use GitHub API to get correct PR submitter

cyqsimon:
Amends #2766.


cyqsimon:
On this branch:

```bash
git commit --amend
git push -f
```

This should be good enough to verify the patch.

Enselic:
Seems to work, thanks again!

Issue #2793, opened 2023-12-05
Integrating with google/oss-fuzz for continuous fuzz testing?


silvergasp:
Hey David,

I hope this message finds you well. I've been following along with bat for some time now, and I'm a daily user for bat. I'd like to suggest and champion an effort to set up some basic fuzz-testing and combine it with google/oss-fuzz for continuous fuzzing. I'm fully aware that you are a very busy person and I don't want to overload your review/maintenance capacity by introducing too many new ideas. Is this a bad time to discuss potential security/reliability improvements? 

If you're not familiar with fuzzing or oss-fuzz I've included a few brief notes below.

#### **Benefits of Fuzz-Testing**

- **Dynamic Code Testing**: Fuzz-testing challenges systems with unexpected data, aiming to identify vulnerabilities. It’s akin to an exhaustive stress-test for the code.
- **Detecting Hidden Vulnerabilities**: It can uncover potential weaknesses that may not be evident in routine tests.
- **Continuous and Automated Testing**: With tools like Google’s OSS-Fuzz, fuzz-testing can be automated, running continuously on distributed systems, ensuring daily resilience checks.

#### **Google/oss-fuzz for Continuous Fuzzing**

- **Automated Fuzzing**: OSS-Fuzz undertakes comprehensive fuzz-testing daily on a distributed cluster.
- **Security Boost**: It provides enhanced security measures free of cost, thanks to Google’s backing.
- **Detailed Reporting**: OSS-Fuzz offers exhaustive reports in case of detected anomalies, enabling effective action.

I’d be more than happy to lead the effort in integrating fuzz testing with bat and assist in any way required.

silvergasp:
I've added a small proof of concept to demonstrate what this would all look like in #2794.

PR #2794, opened 2023-12-05
Add a basic fuzz-harness

silvergasp:
Adds a basic fuzz testing harness.

Ref: #2793

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 
 - Add environment variable `BAT_PAGING`, see #2629 (@einfachIrgendwer0815)
 - Add opt-in (`--features lessopen`) support for `LESSOPEN` and `LESSCLOSE`. See #1597, #1739, #2444, #2602, and #2662 (@Anomalocaridid)
+- Add basic fuzz-testing support. See #2794 (@silvergasp)
 
 ## Bugfixes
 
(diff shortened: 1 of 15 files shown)


Enselic:
Interesting. I skimmed it over a bit. In (not unsafe) Rust, buffer overflows and null pointer dereferences can not happen, but it can still be useful to fuzz, I think.

Just curios: Did the fuzzer find any bugs yet?

silvergasp:
> Just curios: Did the fuzzer find any bugs yet?

Sort of... I found a lot of different ways to crash bat. e.g. if the terminal/tab width in the config is really large (larger than any normal terminal width) it'll crash with an out of memory condition. I didn't think that these where "real" bugs, as they would end up being pretty unlikely, and usually down to user error. I ended up constraining the fuzzer to [keep these below specific values](https://github.com/sharkdp/bat/pull/2794/files#diff-cba64c21ab992eaad29fce147a08f4560a4769bc14682b8a96081a5fd02dbecdR48). But it depends on your definition of a bug.

sharkdp:
I love fuzz testing and used it extensively in [another project of mine](https://github.com/sharkdp/numbat/tree/master/numbat/fuzz). I found quite a few interesting bugs using it.

For bat, the fact that you didn't find anything of substance (according to what you wrote above, please feel free to correct me if you think otherwise)  is *probably* a hint that fuzz-testing is not as useful for this project. This does not mean that bat is bug free of course. But maybe it's a project that doesn't lend itself well to fuzz-testing.

In this sense: thank you very much for this experiment. But I think I'll close this without merging, in order for this not to turn into future maintenance work.

Issue #2706, opened 2023-10-11
`lessopen` feature causes tests to fail


tranzystorek-io:
**What steps will reproduce the bug?**

N/A

**What happens?**

When attempting to build and test the bat pacakge for Void Linux with the `lessopen` feature enabled, the `long_help` and `short_help` tests fail, see CI logs: https://github.com/void-linux/void-packages/actions/runs/6487011019/job/17616392685

**What did you expect to happen instead?**

Tests run successfully.

**How did you install `bat`?**

N/A

---

**bat version and environment**

bat 0.24.0
Void Linux

keith-hall:
thanks for reporting. I'm assuming that the `LESSOPEN` env var is non-empty in the Void Linux CI runner? Probably it makes sense for the integration tests to clear the `LESSOPEN`  env var (except for the tests which already set it), as they likely assume it is blank except for those tests.

tranzystorek-io:
Hm, I assumed it was a simple issue with those help output tests being hardcoded for the default-features case, where clap generates the --lessopen flag help conditionally.

keith-hall:
Oh that would make more sense, somehow I didn't really understand the diff so skipped over that possibility. Thanks for reporting, it makes sense to get that fixed.

PR #2805, opened 2023-12-16
Rework `lessopen` implementation to use `execute` crate instead of `run_script`

Anomalocaridid:
I replaced the `lessopen` implementation to use a crate called `execute` instead of the currently-used `run_script`. This has the benefit of simplifying the code a little, but more importantly fixes a few weird issues I noticed.

The issues I noticed that these changes fix are:

- [`batpipe`](https://github.com/eth-p/bat-extras/blob/master/doc/batpipe.md) not working right when using `fish` as a shell. Currently, `bat` shows the commands that `batpipe` prints to configure `$LESSOPEN` when you evaluate it in your startup config rather than the preprocessed file output.
- `bat` not showing any output at all when using [`glow`](https://github.com/charmbracelet/glow) to preprocess markdown files.

Edit: [Also appears to fix some known issues with `lessopen`-related integration tests.](https://github.com/sharkdp/bat/pull/2805#issuecomment-2509550516)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 - Fix panel width when line 10000 wraps, see #2854 (@eth-p)
 - Fix compile issue of `time` dependency caused by standard library regression #3045 (@cyqsimon)
 - Fix override behavior of --plain and --paging, see issue #2731 and PR #3108 (@einfachIrgendwer0815)
+- Fix bugs in `$LESSOPEN` support, see #2805 (@Anomalocaridid)
 
 ## Other
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -124,6 +124,7 @@ dependencies = [
  "content_inspector",
  "encoding_rs",
  "etcetera",
+ "execute",
  "expect-test",
  "flate2",
  "git2",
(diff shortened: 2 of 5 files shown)


Anomalocaridid:
Marking as draft because I realized my changes cause some of the integration tests to fail.

Anomalocaridid:
I fixed the issues with the integration tests.

However, while troubleshooting them, I came across another potential concern that I am unsure how to resolve. I noticed that unlike what I assumed in my initial implementation, `less` does not appear to send `$LESSCLOSE`'s output to stdout. If I change this to match `less`'s behavior, then that would break the integration tests that test `$LESSCLOSE` because they rely on `bat` outputting `$LESSCLOSE`'s output. I am not sure how to change them to work around this, though.

Also, it may be worth seeing at some point if using `execute` instead of `run_script` also fixed the issue with the randomly failing integration test.

Anomalocaridid:
After some testing, as far as I can tell, the `#[serial]` attribute on some of the `lessopen`-related tests is not needed anymore. In addition, the one `lessopen`-related tests that was ignored because it randomly fails appears to work reliably.

Also while I was at it, I disabled the `long-help` and `short-help` tests for `lessopen` in order to address #2706.

keith-hall (approved):
Thanks!

Anomalocaridid:
My bad, I saw it said auto-merge was enabled and it would be merged when it meets the requirements and I saw the thing saying my PR branch was behind on changes from master and I assumed that being up to date was one of the requirements.

Issue #2783, opened 2023-11-24
Set file name as terminal title


Talinx:
Currently bat does not change the terminal title. Showing the file name in the terminal can be useful for navigating between terminal tabs.

For example, the following script sets the terminal title to the file name before displaying it with bat:

```bash
#!/bin/bash
echo -ne "\033]0;$1\007"
bat $1
```

The actual implementation inside bat should be more sophisticated by setting the title appropriately when viewing multiple files, resetting the title on exit etc.

Oliver-Looney:
I'd like this feature as well, I'm going to look at adding it over the next few days!

PR #2807, opened 2023-12-18
2783 setting terminal title

Oliver-Looney:
This PR is to resolve issue https://github.com/sharkdp/bat/issues/2783

This change sets the terminal title to the input file names when the pager is being used & --set-terminal-title flag is used, to make this feature opt in

https://github.com/sharkdp/bat/assets/88001922/4e20a7e7-62b5-4407-8647-57287c471d08


Oliver-Looney:
I am not sure how to fix this failing action `CICD / arm-unknown-linux-gnueabihf (ubuntu-20.04) (pull_request)`  

Error:

```shell
With the provided path, there will be 1 file uploaded
Artifact name is valid!
Root directory input is valid!
Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run
```

How to fix this?

Enselic:
I think the CI failure will be fixed by https://github.com/sharkdp/bat/pull/2811.

src/bin/bat/main.rs
```diff
@@ -227,9 +227,24 @@ pub fn list_themes(cfg: &Config, config_dir: &Path, cache_dir: &Path) -> Result<
...
 
+fn set_terminal_title_to_inputs_names(inputs: &Vec<Input>) {
+    let mut input_names = "bat: ".to_string();
+    for (index, input) in inputs.iter().enumerate() {
+        input_names += &input.description.name.to_string();
+        if index < inputs.len() - 1 {
+            input_names += ", ";
+        }
+    }
+    print!("\x1b]2;{}\x07", input_names);
```
Enselic, line 238: Will this behave well in all major terminal emulators on all major OSes?

Enselic (commented):
Thank you for taking the time to create a PR. I skimmed it over and have one initial question.

src/bin/bat/main.rs
Oliver-Looney: I believe ANSI escape characters are standard across terminal emulators on different OSes, https://en.wikipedia.org/wiki/ANSI_escape_code#OSC_(Operating_System_Command)_sequences

I have updated the `2` in this line to `0` to match those docs

Oliver-Looney:
Thanks for the review!

sharkdp:
Thank you for your contribution. I would also be worried about portability of this specific ANSI escape code. If we can not test this across a wide range of terminal emulators at the moment, let's maybe make this opt-in at first?

Oliver-Looney:
Yes, that makes sense. I'll update this pr to make this opt-in

src/bin/bat/app.rs
```diff
@@ -287,6 +287,7 @@ impl App {
             use_custom_assets: !self.matches.get_flag("no-custom-assets"),
             #[cfg(feature = "lessopen")]
             use_lessopen: self.matches.get_flag("lessopen"),
+            set_terminal_title: self.matches.get_flag("set_terminal_title"),
```
eth-p, line 290: Using hypens instead of underscores (i.e. `--set-terminal-title`) would make the option more consistent with the rest of `bat`'s options.

src/input.rs
```diff
@@ -94,7 +94,7 @@ pub(crate) struct InputMetadata {
 pub struct Input<'a> {
     pub(crate) kind: InputKind<'a>,
     pub(crate) metadata: InputMetadata,
-    pub(crate) description: InputDescription,
+    pub description: InputDescription,
```
eth-p, line 97: I'm not a big fan of changing the visibility of this. It wasn't meant to be directly accessible through the public API, and increasing the surface area of the API will make it harder for us to change implementation details in the future without them becoming breaking changes.

You can already use `my_input.description()` to get a `&InputDescription`, anyway :)

src/input.rs
```diff
@@ -13,7 +13,7 @@ use crate::error::*;
 /// This tells bat how to refer to the input.
 #[derive(Clone)]
 pub struct InputDescription {
-    pub(crate) name: String,
+    pub name: String,
```
eth-p, line 16: Same for this. It shouldn't be directly accessible as part of the API, and we already have `my_input_description.title()` as a way to get the appropriate title for the file (or name if no title is specified).

eth-p (changes requested):
Thanks for adding this! There are a couple of tiny changes that I would like you to make, but this looks great otherwise!

eth-p:
> Thank you for your contribution. [...] let's maybe make this opt-in at first?

I'm not sure if it would be a good idea to transition this to becoming opt-out in the future without more feedback from users. I don't believe it's common for terminal programs other than the shell to set the title by default; `vim` requires you to give permission by adding `set title` to its config to enable it, for instance. 

There's also the chance that setting the terminal title using a fixed pattern could break someone's workflow. For example, I have `tmux` configured to have different key bindings depending on the foreground application, which is determined by the pane's title. It would be an easy fix for me since I'm aware of the change, but it might be frustrating to others who aren't aware of the change.

But then again, it would likely be a small minority that has that issue. At some point, improving the UX for the many outweighs breaking the obscure workflows of the few.

> I would also be worried about portability of this specific ANSI escape code. If we can not test this across a wide range of terminal emulators at the moment.

It should be safe. OSC sequences follows the form of `INTRODUCER` `ID` `;` `CONTENT` `TERMINATOR`, with the introducer being `ESC ]` and terminator being either `BEL` or `ESC \`. This allows the terminal emulator to know where the sequence starts and ends, which means it can ignore any commands that it doesn't recognize. A good example of this would be hyperlinks, which are implemented using an OSC sequence:

image

VS Code (on the left) supports and recognizes hyperlinks, while Terminal.app (on the right) does not. This also applies to any random command that doesn't exist:

image

sharkdp:
> > Thank you for your contribution. [...] let's maybe make this opt-in at first?
> 
> I'm not sure if it would be a good idea to transition this to becoming opt-out in the future without more feedback from users. I don't believe it's common for terminal programs other than the shell to set the title by default; `vim` requires you to give permission by adding `set title` to its config to enable it, for instance.
> 
> There's also the chance that setting the terminal title using a fixed pattern could break someone's workflow. For example, I have `tmux` configured to have different key bindings depending on the foreground application, which is determined by the pane's title. It would be an easy fix for me since I'm aware of the change, but it might be frustrating to others who aren't aware of the change.

Good points. Let's design it as a opt-in feature.

Oliver-Looney:
> Thanks for adding this! There are a couple of tiny changes that I would like you to make, but this looks great otherwise!

Thanks for the review! I think I have implemented the changes. I'll get the tests fixed in the next day or two

eth-p (approved):
Looks good to me! Thanks for make the changes :)

PR #2816, opened 2023-12-29
Eliminate dependency on syn v1 by updating git-version

dtolnay:
Generated by running `cargo update -p git-version`, which picks up https://github.com/fusion-engineering/rust-git-version/pull/20. This was the last crate in bat's dependency graph still pulling in syn v1.


Enselic:
Thank you 🙇

PR #2836, opened 2024-01-22
Offload building built-in glob matchers to worker threads

cyqsimon:
This is an exploratory implementation of [this idea](https://github.com/sharkdp/bat/pull/2755#pullrequestreview-1828155377).

As mentioned in that thread, it is possible that this change brings about no performance improvements at all, hence "exploratory". However, it is likely that as the mapping rule list grows, the benefits of leveraging multiple threads will (eventually, if we ever get there) start to outweigh the cost of spawning them.

## Checklist

- [x] Implementation
- [ ] Tests

### DIFF
--- a/src/syntax_mapping.rs
+++ b/src/syntax_mapping.rs
@@ -3,7 +3,7 @@ use std::path::Path;
 use globset::{Candidate, GlobBuilder, GlobMatcher};
 
 use crate::error::Result;
-use builtin::BUILTIN_MAPPINGS;
+use builtin::OffloadIter;
 use ignored_suffixes::IgnoredSuffixes;
 
 mod builtin;
@@ -91,8 +91,7 @@ impl<'a> SyntaxMapping<'a> {
     pub fn builtin_mappings(
         &self,
     ) -> impl Iterator<Item = (&'static GlobMatcher, &'static MappingTarget<'static>)> {
-        BUILTIN_MAPPINGS
-            .iter()
+        OffloadIter::new()
             .filter_map(|(matcher, target)| matcher.as_ref().map(|glob| (glob, target)))
     }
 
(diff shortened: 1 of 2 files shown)


cyqsimon:
I benchmarked this against `master` (2086cd2) with the following command after minimising the noise on my system:

```bash
hyperfine \
  --shell=none \
  --warmup=500 \
  --runs=5000 \
  -L version master,offload \
  --export-json=indeterminant.json \
  'target/release/bat-{version} --no-config --color=always tests/benchmarks/startup-time-src/mystery-file'
```

whisker plot

[indeterminant.json.gz](https://github.com/sharkdp/bat/files/14011604/indeterminant.json.gz)

It does indeed seem like it's slightly slower than `master` in its current state, although I do believe the implementation can be improved quite a bit.

cyqsimon:
Closing in favour of v2.

hamirmahal:
> I benchmarked this against `master` ([2086cd2](https://github.com/sharkdp/bat/commit/2086cd2668df6064cc2d1f3e5026532ac96cb398)) with the following command after minimising the noise on my system:
> 
> ```shell
> hyperfine \
>   --shell=none \
>   --warmup=500 \
>   --runs=5000 \
>   -L version master,offload \
>   --export-json=indeterminant.json \
>   'target/release/bat-{version} --no-config --color=always tests/benchmarks/startup-time-src/mystery-file'
> ```
> 
> whisker plot
> 
> [indeterminant.json.gz](https://github.com/sharkdp/bat/files/14011604/indeterminant.json.gz)
> 
> It does indeed seem like it's slightly slower than `master` in its current state, although I do believe the implementation can be improved quite a bit.

Out of curiosity, how did you get that really nice graphic?

hamirmahal:
```
$  hyperfine 'node e'
Benchmark 1: node e
  Time (mean ± σ):      23.4 ms ±   1.1 ms    [User: 17.6 ms, System: 7.2 ms]
  Range (min … max):    22.1 ms …  31.5 ms    120 runs
```

is the most I was able to get working.

PR #2838, opened 2024-01-24
display which theme is the default one in colored output

sblondon:
The user knows which theme is used by default with `--list-themes` parameter.

This is a partial screenshot of the output:
screenshot_theme_list

Only the `config.colored_output` displays the ` (default)` string. I didn't add it in the other case because it's really minimal but I will add it if you want.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@
 - Relax syntax mapping rule restrictions to allow brace expansion #2865 (@cyqsimon)
 - Apply clippy fixes #2864 (@cyqsimon)
 - Faster startup by offloading glob matcher building to a worker thread #2868 (@cyqsimon)
-- Display which theme is the default one, see #2838 (@sblondon)
+- Display which theme is the default one in colored output, see #2838 (@sblondon)
 
 ## Syntaxes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@
 - Relax syntax mapping rule restrictions to allow brace expansion #2865 (@cyqsimon)
 - Apply clippy fixes #2864 (@cyqsimon)
 - Faster startup by offloading glob matcher building to a worker thread #2868 (@cyqsimon)
+- Display which theme is the default one, see #2838 (@sblondon)
 
 ## Syntaxes
 
--- a/src/bin/bat/main.rs
+++ b/src/bin/bat/main.rs
@@ -30,6 +30,7 @@ use directories::PROJECT_DIRS;
 use globset::GlobMatcher;
 
 use bat::{
+    assets::HighlightingAssets,
     config::Config,
     controller::Controller,
     error::*,
@@ -200,11 +201,18 @@ pub fn list_themes(cfg: &Config, config_dir: &Path, cache_dir: &Path) -> Result<
     let mut stdout = stdout.lock();
 
     if config.colored_output {
+        let default_theme = HighlightingAssets::default_theme();
         for theme in assets.themes() {
+            let default_theme_info = if default_theme == theme {
+                " (default)"
+            } else {
+                ""
+            };
             writeln!(
                 stdout,
-                "Theme: {}\n",
-                Style::new().bold().paint(theme.to_string())
+                "Theme: {}{}\n",
+                Style::new().bold().paint(theme.to_string()),
+                default_theme_info
             )?;
             config.theme = theme.to_string();
             Controller::new(&config, &assets)
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -273,6 +273,24 @@ fn squeeze_limit_line_numbers() {
 }
 
 #[test]
+fn list_themes() {
+    #[cfg(target_os = "macos")]
+    let default_theme_chunk = "Monokai Extended Light\x1B[0m (default)";
+
+    #[cfg(not(target_os = "macos"))]
+    let default_theme_chunk = "Monokai Extended\x1B[0m (default)";
+
+    bat()
+        .arg("--color=always")
+        .arg("--list-themes")
+        .assert()
+        .success()
+        .stdout(predicate::str::contains("DarkNeon").normalize())
+        .stdout(predicate::str::contains(default_theme_chunk).normalize())
+        .stdout(predicate::str::contains("Output the square of a number.").normalize());
+}
+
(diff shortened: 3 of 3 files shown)


Enselic (commented):
Thanks, seems like a nice feature. Would be great if you could add a regression in tests/integration_tests.rs too.

sblondon:
@Enselic I added an integration test.

keith-hall (approved):

sblondon:
@Enselic @keith-hall I rebased and fixed the conflicts occurred since the first commits. Do you think about blocking elements preventing the merge of this PR?

tests/integration_tests.rs
```diff
@@ -272,6 +272,19 @@ fn squeeze_limit_line_numbers() {
...
+#[test]
+fn list_themes() {
+    bat()
+        .arg("--color=always")
+        .arg("--list-themes")
+        .assert()
+        .success()
+        .stdout(predicate::str::contains("DarkNeon").normalize())
+        .stdout(predicate::str::contains("Monokai Extended").normalize())
+        .stdout(predicate::str::contains("Output the square of a number.").normalize())
```
Enselic, line 284: Hmm what is this?

tests/integration_tests.rs
```diff
@@ -272,6 +272,19 @@ fn squeeze_limit_line_numbers() {
...
+        .stdout(predicate::str::contains(" (default)").normalize());
```
Enselic, line 285: I would expect a single `contains` that contains "Monokai Extended (default)".

Right now your test will pass even if the wrong theme is marked as default, right?

Enselic (changes requested):

tests/integration_tests.rs
sblondon: @Enselic You're right. However, the default theme could be `Monokai Extended Light` instead of `Monokai Extended` for MacOS with preferences setup to [light mode](https://github.com/sharkdp/bat/blob/e8d777b73a270330e240a64635851e2342c56323/src/assets.rs#L103) (according to [macos_dark_mode_active()](https://github.com/sharkdp/bat/blob/e8d777b73a270330e240a64635851e2342c56323/src/assets.rs#L403)).

Several ways to fix it are possible:
1. no change: the check is kept weak.
2. assert "Monokai Extended (default)". The test will break on some configuration. It could pass currently the CI, and, one day, will break because the MacOS image used has changed preferences setup.
3. assert "Monokai Extended (default)" but ignore the test on MacOS with `cfg_attr()`. It's future proof but not tested on MacOS.
4. assert a regexp like "Monokai Extended.* (default)". 
5. assert "Monokai Extended (default)" and mock `macos_dark_mode_active()` to be sure the function returns `true`.
6. other way I didn't think about?

The regexp solution (4) is my preferred one.

Which solution do you want I implement?

Enselic: If the default ends up different on macOS, I think it makes sense to `[cfg(...)]` the expected default based on platform.

sblondon: On MacOS, the default can be `Monokai Extended` or `Monokai Light Extended` (if some preferences are defined).

Do you mean you prefer solution 3?
Or doing 3 and adding another test only for macOS?

Enselic: In CI I'm pretty sure the default on macOS is always the same? If that unexpectedly changes, I think it is good if we are notified in the form of a failing test.

So I don't think you need to ignore the test on macOS, I think it should work to cfg the expected default theme.

sblondon: The macOS output in CI is different from other operating systems so I used `cfg` attribute. Is it good for you or did you expect [`cfg!` macro](https://doc.rust-lang.org/rust-by-example/attribute/cfg.html)?

A bold style is applied to the theme name. The `\x1B[0m` characters are the removal of this bold style.

tests/integration_tests.rs
```diff
@@ -272,6 +272,28 @@ fn squeeze_limit_line_numbers() {
...
+fn list_themes() {
+    let default_theme_chunk: &str;
+    #[cfg(target_os = "macos")]
+    {
+        default_theme_chunk = from_utf8(b"Monokai Extended Light\x1B[0m (default)").unwrap();
+    }
+    #[cfg(not(target_os = "macos"))]
+    {
+        default_theme_chunk = from_utf8(b"Monokai Extended\x1B[0m (default)").unwrap();
+    }
```
Enselic, line 285:
```suggestion
    #[cfg(target_os = "macos")]
    let default_theme_chunk = "Monokai Extended Light\x1B[0m (default)";

    #[cfg(not(target_os = "macos"))]
    let default_theme_chunk = "Monokai Extended\x1B[0m (default)";
```

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@
 - Relax syntax mapping rule restrictions to allow brace expansion #2865 (@cyqsimon)
 - Apply clippy fixes #2864 (@cyqsimon)
 - Faster startup by offloading glob matcher building to a worker thread #2868 (@cyqsimon)
+- Display which theme is the default one, see #2838 (@sblondon)
```
Enselic, line 36:
```suggestion
- Display which theme is the default one in colored output, see #2838 (@sblondon)
```

src/bin/bat/main.rs
```diff
@@ -200,11 +201,18 @@ pub fn list_themes(cfg: &Config, config_dir: &Path, cache_dir: &Path) -> Result<
     let mut stdout = stdout.lock();
 
     if config.colored_output {
```
Enselic, line 203: Would be nice to cover the "not colored output" case, but I don't think you need to do that in this PR. This is clearly an improvement on its own in my view.

Enselic (changes requested):
Thanks, this looks reasonable! I only had a suggestion on a simplification that I think is worthwhile.

tests/integration_tests.rs
sblondon: Nice idea. I didn't know block (`{}`) was optional with `#[cfg()]`. :-)
I squashed it with the commit which adds the test.

src/bin/bat/main.rs
sblondon: I agree to implement the same feature for the "not colored output" case but I prefer to do it in another PR after this one is merged into `master`.

Enselic (approved):

PR #2818, opened 2023-12-29
upgrade `plist` to 1.6

CGMossa:

PR #2839, opened 2024-01-27
Change CONTRIBUTING.md according to new changelog requirements

einfachIrgendwer0815:
In PR #2766 ([`bcc2de86`](https://github.com/sharkdp/bat/commit/bcc2de86b47c26ad5393588aaead540e348dfc3c)) a new workflow was added which ensures that the `CHANGELOG` stays up to date and that entries follow a certain format. However, this has led to quite a bit of confusion among contributors (https://github.com/sharkdp/bat/pull/2827#issuecomment-1874709439). For example, the authors of #2827, #2807, #2835 and #2818 had to modify the changelog multiple times to match the requirements.

To solve that, this PR changes CONTRIBUTING.md according to the new format requirements. Also, a note is added to the workflow to point contributors in the right direction.

But I do have a question that I'm myself confused about. The CONTRIBUTING.md states:
> If your contribution changes the behavior of `bat` (as opposed to a typo-fix
in the documentation), please update the [`CHANGELOG.md`](CHANGELOG.md) file
and describe your changes.

So, do PRs like this one need a changelog entry? I think CONTRIBUTING.md says no, but the workflow will fail nonetheless.

### DIFF
--- a/.github/workflows/require-changelog-for-PRs.yml
+++ b/.github/workflows/require-changelog-for-PRs.yml
@@ -29,5 +29,5 @@ jobs:
           ADDED=$(git diff -U0 "origin/${PR_BASE}" HEAD -- CHANGELOG.md | grep -P '^\+[^\+].+$')
           echo "Added lines in CHANGELOG.md:"
           echo "$ADDED"
-          echo "Grepping for PR info:"
+          echo "Grepping for PR info (see CONTRIBUTING.md):"
           grep "#${PR_NUMBER}\\b.*@${PR_SUBMITTER}\\b" <<< "$ADDED"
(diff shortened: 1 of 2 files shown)


keith-hall (approved):
Nice, this should bring some clarification, thanks. :+1:
I agree that the Github Actions workflow doesn't match the contributing guidelines, perhaps it will make sense to edit the workflow to make updating the changelog optional for purely documentation changes etc.

Enselic:
> So, do PRs like this one need a changelog entry? I think CONTRIBUTING.md says no, but the workflow will fail nonetheless.

The way I see it, it's fine if the Changelog job fails if there is no need to add something to the changelog. It can simply be ignored. 

Maybe we can clarify that in the docs too somehow?

Enselic (approved):
Thank you

PR #2848, opened 2024-02-05
Add Prolog syntax highlighting

### LINKED ISSUE #1409: Prolog
**Syntax:**
Prolog (GNU Prolog, SWI-Prolog)

**Guideline Criteria:**
https://packagecontrol.io/packages/Prolog

laalsaas:
I found [this Prolog Syntax definition](https://packagecontrol.io/packages/Prolog) and added it to bat. It is under MPL, so I added support for generating Acknowledgement for that license. Feel-free to double-check the legal part.

I had a bit of a hard time creating this PR because much of the tooling in this repo relies on the `bat`-binary in `$PATH` instead of the locally built one. This was troubling when using my own acknowledgement code, as well as when generating the syntax test. I ended up hard-coding the path of the binary into the scripts (which I didn't submit for obvious reasons). I'm mentioning this in case anyone runs into the same problem.

Fixes #1409


sharkdp:
Thank you. Can you look into getting the CI tests to pass? (add changelog entry, look into license check)

laalsaas:
There are 3 Failing tests:
- Changelog entry, I made one but I think it's missing the PR Number, no problem, can add that.
- The conflicting file endings: Prolog and Perl both claim the .pl extension, I guess it's more sane to leave this to Perl. The interesting question is where to change that, should I just update the .sublime-syntax file or also the .tmLanguage file from which it was generated.
- The License Check: I think Ican't resolve that. The Prolog Syntax is under MPL. The MPL explicitly mentions the GPL in its text. Since the Test just greps for GPL, It fails, even though I didn't include any GPL code.

sharkdp:
> The conflicting file endings: Prolog and Perl both claim the .pl extension, I guess it's more sane to leave this to Perl.

Agreed.

> The interesting question is where to change that, should I just update the .sublime-syntax file or also the .tmLanguage file from which it was generated.

Editing .sublime-syntax only is fine, thank you!



> The License Check: I think Ican't resolve that. The Prolog Syntax is under MPL. The MPL explicitly mentions the GPL in its text. Since the Test just greps for GPL, It fails, even though I didn't include any GPL code.

@Enselic Any idea how we could resolve this?

Enselic:
Maybe we could try to use cargo deny for license checks. I use it in other projects and it is more sophisticated than our custom solution. Unsure if it supports submodules though.

If that doesn't work, I think we have to make our check more sophisticated.

laalsaas:
Hey, is there any chance moving this forward? i.e. would it be an option to add this file to the [excluded files](https://github.com/sharkdp/bat/blob/master/tests/scripts/license-checks.sh#L5)

Enselic:
I think adding a general exception to MPL is a better approach, but since I think this syntax would be the first MPL syntax, I don't want to approve that myself. New licenses in bat needs to be approved by sharkdp in my view.

cafkafk:
> I think adding a general exception to MPL is a better approach, but since I think this syntax would be the first MPL syntax, I don't want to approve that myself. New licenses in bat needs to be approved by sharkdp in my view.

@sharkdp sorry to ping but if this got stuck just on you needing to see it I thought it made sense to connect you to this thread.

Enselic:
Since we can include MPL licensed syntaxes without needing to relicense the rest of bat, I think it's fine to include MPL licensed syntaxes like Prolog.

I made a quick sketch in https://github.com/sharkdp/bat/pull/3283. I haven't verified it at all. Feel free to take that code without attribuation and incorpoprate it in this PR to make CI pass.

Enselic:
closing due to inactivity

cafkafk:
Will anyone do a follow up on this? If not, it seems sad to see this disappear when it's functional code that can just be merged

PR #2849, opened 2024-02-08
Clarify when/how default args are added to less.

eth-p:
I rewrote part of the README to explain when arguments will not be passed to `less` by default, and to describe the purpose of the `-S` option and when it is used.


keith-hall (approved):

sharkdp (approved):
Excellent — thank you!

Issue #2810, opened 2023-12-21
Add flag for git blame


Oliver-Looney:
I'd like to implement git blame like JetBrains ide, e.g:
<img width="406" alt="image" src="https://github.com/sharkdp/bat/assets/88001922/efcf8e94-f675-4b93-adac-f03d962102fe">

Show git blame information to the left of line numbers.  Ideally be able to take in an optional format like how git --log -pretty="<format>" works
[git log docs](https://git-scm.com/docs/pretty-formats)

Would this be feature be welcomed/ any feedback?
I am planning to start working on this over the holidays

Issue #1536, opened 2021-02-04
Add a flag to show `git blame`


danielhoherd:
It would be awesome to see the syntax highlighted output of `bat` with the info from `git blame` included.

sharkdp:
Thank you for your request.

I think I like this idea! We could implement this as a new `--style` component (`blame`). Layout-wise, we could probably include a second side-bar to the right of the line-number+git-changes side bar. Another idea could be to include it on the right hand side of the screen. Having the source code way out on the right is actually something I don't really like about the default `git blame` output.

The `git2` crate (that we already use) should provide us with all the necessary information.
* [`blame.rs`](https://github.com/rust-lang/git2-rs/blob/master/examples/blame.rs) example application
* [`git2::Blame` struct](https://docs.rs/git2/0.13.17/git2/struct.Blame.html)

sharkdp:
What I really like about this idea is that I typically use `git blame` when I'm *reading* source code. If I use `bat` for this already, I can simply append `--blame` when I need it.

Enselic:
Since it already is within scope of bat to integrate with git, I'd say this is a perfectly reasonable feature request.

However, in my opinion, a good git blame tool needs enable easy navigation back and forth in the git log, i.e. have e.g. a "blame parent commit of the commit that last changed this line" feature, like `git gui blame` and many IDE integrations have. I don't see how such a feature could be implemlemented in a good way with `bat`.

So for me the question boils down to "Is it ok for bat to only be a mediocre git blame tool?", and the answer I'm currently leaning towards is "If you can't do it really well, don't do it at all". There could of course be a clever way to implement "blame parent commit" that I'm not thinking about however, that would change my mind.

danielhoherd:
FWIW, my original reason for requesting this feature is because I was looking through a bunch of code with `git blame`, and its lack of syntax highlighting was frustrating, but I just needed to see when the last commit of certain lines were. To me, that alone is good enough. I never considered this as being a blame history explorer, and I wouldn't expect such a feature from `bat`. I see it as being similar to this one feature, <https://marketplace.visualstudio.com/items?itemName=eamodio.gitlens#current-line-blame>, which just gives hints to the reader about the current code.

Enselic:
Thanks for elaborating on your use case! That particular VS Code extension is what I personally use. For the record, I would like to point out that it has a "Blame Previous Revision" button if you hover over its annotation, which works the same as "blame parent commit" I described above.

[...]

PR #2851, opened 2024-02-09
Add Git Blame style option.

userwiths:
# Related Issues

#2810 #1536 

# Quick Summary

* Added a new decorator.
* Added required changes in `config` in order to recognize 2 new CLI options/arguments.
* Added a simple format ability with the placeholders from [git's pretty-formats](https://git-scm.com/docs/pretty-formats)
* Updated doc tests to take into consideration new options.

# Visual Example
```bat ./README.md --style=full,blame --blame-format="%h %cn"```
image

### DIFF
--- a/doc/long-help.txt
+++ b/doc/long-help.txt
@@ -138,6 +138,7 @@ Options:
             * rule: horizontal lines to delimit files.
             * numbers: show line numbers in the side bar.
             * snip: draw separation lines between distinct line ranges.
+            * blame: show Git blame information.
 
   -r, --line-range <N:M>
           Only print the specified range of lines for each file. For example:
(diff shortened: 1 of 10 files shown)


Enselic:
The code looks completely reasonable. It would be nice with a regression test, but I think I'd rather merge this and let others (or preferably: you : ) fix follow-up problems, than to let this PR sit as Draft forever.

So unless other maintainers object, I suggest we merge this after you rebase it and resolve conflicts.

I don't use the git functionality myself so I don't want to put so much time into the feature, but it seems sad to let this PR sit here forever, and it seems quite useful, even if I probably won't use it myself.

Enselic:
Closing for now to keep PR inbox clean but am happy to re-open if you get around to rebasing!

PR #2853, opened 2024-02-11
Document default style components

einfachIrgendwer0815:
Adds a short note to the help text of `--style` explaining which style components are enabled by default.


sharkdp:
Thank you

PR #2857, opened 2024-02-12
skip highlighting when no color

eth-p:
- Split syntax highlighting from print_line into fn
- Skip syntax parsing when color output disabled


eth-p:
@sharkdp and @Enselic, a small question for context:

Over here, there's [this code](https://github.com/sharkdp/bat/blame/5a2a20af42e7a426943122524287844acc5ee975/src/printer.rs#L534-L536) introduced in https://github.com/sharkdp/bat/commit/ce96df00b6e77841f7c77fdc50e5b6e5b944e375 which returns from `print_line` when no `HighlighterFromSet` is available on the printer. Based on the commit message, this is meant to exit early in order to prevent the printer from displaying raw binary.

We also have [this code just above it](https://github.com/sharkdp/bat/blame/5a2a20af42e7a426943122524287844acc5ee975/src/printer.rs#L512-L514) which returns early if the detected content type is binary.

Lastly, [this code](https://github.com/sharkdp/bat/blame/5a2a20af42e7a426943122524287844acc5ee975/src/printer.rs#L225-L243) is what determines whether the `highlighter_from_set` is `None`.

In this pull request, I modified that last bit of code to also provide `None` when color output is disabled. Then, when `highlighter_from_set` is `None`, the newly-refactored `highlight_regions_for_line` function will return a fake `(Style, &str)` region with a dummy value for the `Style`. This shouldn't cause any problems, correct? I looked through the code, and nothing else seemed to rely on `highlighter_from_set`.

sharkdp (approved):
I agree with your analysis, it looks like we handled the binary case twice before, so this should be fine. Maybe do a quick test to make sure the new code (around `EMPTY_SYNTECT_STYLE`) will not be called in interactive-use binary-file scenarios?

eth-p:
Seems to work as intended :)

image

No color + binary = just the header.
No color + binary + `-A` = the contents without color.
Color + binary + `-A` = the contents with color.
Color + binary = just the header with color.

PR #2865, opened 2024-02-20
Relax syntax mapping rule restrictions to allow brace expansion

cyqsimon:
It turns out there is a valid reason for glob rules to contain `{` and `}`. So I relaxed the compile time restriction.

See https://globster.xyz/.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - Pull in fix for unsafe-libyaml security advisory, see #2812 (@dtolnay)
 - Update git-version dependency to use Syn v2, see #2816 (@dtolnay)
 - Update git2 dependency to v0.18.2, see #2852 (@eth-p)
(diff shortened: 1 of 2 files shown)


keith-hall (approved):

Enselic:
Thank you for the fix! Would be great if you could add a regression test. Prefer a "real" test over a unit test.

cyqsimon:
> Thank you for the fix! Would be great if you could add a regression test. Prefer a "real" test over a unit test.

Hmm do we need it? It's part of the build script so it just won't build if there's a regression somehow.

Enselic:
I would not call it "needed", but I do think it would be good to have. For one, it would make it easy to understand what the fix does and why it is needed (I currently don't understand this). And if someone undo your change either accidentally or deliberately, CI will fail, it and will (again) be clear why the code needs to look the way it does.

sharkdp:
Merging this, since we're discussing the tests in #2866

PR #2866, opened 2024-02-20
Add syntax mapping for podman quadlets

cyqsimon:
See https://www.redhat.com/sysadmin/quadlet-podman

Queued up after #2865 as this syntax mapping uses brace expansion.

### DIFF
--- a/src/syntax_mapping/builtins/linux/50-podman-quadlet.toml
+++ b/src/syntax_mapping/builtins/linux/50-podman-quadlet.toml
@@ -1,0 +1,7 @@
+# see `man quadlet`
+[mappings]
+"INI" = [
+    "**/containers/systemd/*.{container,volume,network,kube,image}",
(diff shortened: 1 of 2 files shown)


src/syntax_mapping/builtins/linux/50-podman-quadlet.toml
```diff
@@ -0,0 +1,7 @@
+# see `man quadlet`
+[mappings]
+"INI" = [
+    "**/containers/systemd/*.{container,volume,network,kube,image}",
```
Enselic, line 4: This seems easy to create a regular syntax regression test for? That would be an excellent test for the fix I think.

sharkdp: I think the argument is that the existence of this mapping alone is kind of a regression test for the `{…}`-based glob patterns. Of course, it would be nice to have syntax tests on top, but I'm personally fine with merging this as is.

PR #2867, opened 2024-02-20
Map containers `.conf` files to TOML syntax

cyqsimon:
See https://github.com/containers/image/tree/main/docs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 
 - `cmd-help`: scope subcommands followed by other terms, and other misc improvements, see #2819 (@victor-gp)
 - Upgrade JQ syntax, see #2820 (@dependabot[bot])
+- Map containers .conf files to TOML syntax #2867 (@cyqsimon)
 
 ## Themes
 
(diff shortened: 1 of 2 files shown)


Enselic (approved):
Probably non-trivial to add tests since absolute paths are involved, so I'll just assume this works. No objections from me.

cyqsimon:
> Probably non-trivial to add tests since absolute paths are involved, so I'll just assume this works. No objections from me.

Ideas for a testing infrastructure for this sort of thing:
- `chroot`? This will still require privileges.
- Test containers? This will add a lot of complexity, but it might be worth it if it's also useful in other integration testing.

eth-p:
> `chroot`? This will still require privileges.

There's also [PRoot](https://proot-me.github.io/), which uses ptrace to emulate chroot. I don't believe there are any alternatives for MacOS or Windows, but it could work when running tests on Linux systems.

Enselic:
Probably easier to set `XDG_CONFIG_HOME` in the test to some path that the test suite controls? (But again, not needed for merge as far as I'm concerned.)

sharkdp:
I will merge it without the tests, if that's okay with everyone.

PR #2868, opened 2024-02-21
Faster startup by offloading glob matcher building to a worker thread

cyqsimon:
This is a much simpler and unsurprisingly better try than #2836.

Instead of an expensive thread pool, we simply create one simple worker thread to blindly build every builtin glob matcher until completion (or made to stop). Doing this early results in effective parallelization and therefore saves time. This may end up doing more work than strictly necessary, but that's not what we care about.

### DIFF
--- a/src/syntax_mapping.rs
+++ b/src/syntax_mapping.rs
@@ -1,6 +1,14 @@
-use std::path::Path;
+use std::{
+    path::Path,
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
+    thread,
+};
 
 use globset::{Candidate, GlobBuilder, GlobMatcher};
+use once_cell::sync::Lazy;
 
 use crate::error::Result;
 use builtin::BUILTIN_MAPPINGS;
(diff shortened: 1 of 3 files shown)


cyqsimon:
Using the same benchmark procedure as https://github.com/sharkdp/bat/pull/2836#issuecomment-1904170496, I got the following results:

benchmark results

I would consider this a success.

src/syntax_mapping.rs
```diff
@@ -44,14 +52,45 @@ pub struct SyntaxMapping<'a> {
...
 
+    /// Start a thread to build the glob matchers for all builtin mappings.
+    ///
+    /// The use of this function while not necessary, is useful to speed up startup
+    /// times by starting this work early in parallel.
+    ///
+    /// The thread halts if/when `halt_glob_build` is set to true.
+    pub fn start_offload_build_all(&self) {
+        let halt = Arc::clone(&self.halt_glob_build);
+        thread::spawn(move || {
```
sharkdp, line 84: I've had problems in the past with threads that are never joined. Are we sure this can not cause any problems?

cyqsimon: I gave this some thought, but in the end couldn't see any possible failure mode. The thread is so simple that it's trivial to reason about its exact behaviour.

I believe it's okay to forgo joining the thread because there's no resource the thread holds that needs to be released upon halting - `BUILTIN_MAPPINGS` is `'static`, and the halt flag is safely behind an `Arc`. The only synchronization requirement is to make sure that it indeed does halt, which it obviously does.

If you prefer defensive coding just to be safe, I can do that too. But in this case I really don't think it's necessary.

sharkdp: We recently had a problem in another project of mine where a non-joined thread caused problems on Windows (https://github.com/sharkdp/numbat/pull/298). But that might have been only due to the fact that the background thread was accessing the network(?).

sharkdp: Anyway, I guess I'm okay with giving this a shot :smile:. But those type of things can be hard to track down (spurious crashes that depend on precise timings).

cyqsimon: > We recently had a problem in another project of mine where a non-joined thread caused problems on Windows ([sharkdp/numbat#298](https://github.com/sharkdp/numbat/pull/298)). But that might have been only due to the fact that the background thread was accessing the network(?).

Yeah in that case releasing the socket (a shared resource) would require synchronization. We don't have that problem here.

> Anyway, I guess I'm okay with giving this a shot 😄. But those type of things can be hard to track down (spurious crashes that depend on precise timings).

I guess I'll add a comment in the threading code to remind future maintainers to watch out for synchronization issues if it ever starts to hold interesting resources in the future.

cyqsimon: Also, that's a really interesting project you've got there! When I've got time I would like to take a look at the possibility of integrating it with Krunner.

sharkdp:
Thank you!

src/syntax_mapping.rs
sharkdp: Great idea. Let me know how if you have problems with the current API. It's not really stable and/or particularly nice at the moment, but I have built several frontends for Numbat so far. And there is a 3rd party integration into a krunner-like launcher here: https://github.com/MatthiasGrandl/Loungy/blob/main/src/commands/root/numbat.rs

PR #2888, opened 2024-03-09
Print $TERM in --diagnostics

roguh:
- **Sort env vars printed by --diagnostic**
- **Print $TERM with --diagnostic**

I would also like `--diagnostics` to print the current terminal emulator, but I can't find a portable way to do it.

Maybe `readlink "/proc/$(cat /proc/$(echo $$)/stat|cut -d ' ' -f 4)/exe"` for Linux?


Enselic (approved):
Seems reasonable

PR #2902, opened 2024-03-18
Prevent out-of-memory crashes using line length limits

### LINKED ISSUE #636: files of infinite length cause bat to crash

einfachIrgendwer0815:
Fixes #636 with the solution proposed [in this comment](https://github.com/sharkdp/bat/issues/636#issuecomment-1887097455).

The limit at which reading continues but content gets discarded is now called "soft limit". Respectively, the limit at which bat will abort is now called "hard limit".

The default values for the soft and hard limits can be overwritten using the `--soft-line-limit` and `--hard-line-limit` cli options. `--disable-line-limits` will disable both limits.


doc/long-help.txt
```diff
@@ -155,6 +155,17 @@ Options:
...
           auto, always, *never*.
 
+      --disable-line-limits
+          Disables all line limits. Short for `--soft-line-limit 0 --hard-line-limit 0`.
+
+      --soft-line-limit <BYTES>
+          Line length (in bytes) at which the line will be ignored. Zero disables this limit.
+          Default: 64 kB
+
+      --hard-line-limit <BYTES>
```
Enselic, line 165: In what scenario do we prefer aborting bat over ignoring the line?

doc/long-help.txt
```diff
@@ -155,6 +155,17 @@ Options:
           escape sequences unless the syntax highlighting language is plain text. Possible values:
           auto, always, *never*.
 
+      --disable-line-limits
```
Enselic, line 158: The fewer options the better IMO. So I don't think we need this one, since I don't think we need two settings (see comment below).

doc/long-help.txt
```diff
@@ -155,6 +155,17 @@ Options:
...
+          Disables all line limits. Short for `--soft-line-limit 0 --hard-line-limit 0`.
+
+      --soft-line-limit <BYTES>
```
Enselic, line 161: I'm thinking about the name for this. What name for this setting do other editors use?

tests/snapshots/output/grid_header_numbers.snapshot.txt
```diff
@@ -1,27 +1,27 @@
...
-  20 │ 
-  21 │ fn perimeter(rectangle: &Rectangle) -> u32 {
-  22 │     (rectangle.width + rectangle.height) * 2
-  23 │ }
-─────┴──────────────────────────────────────────────────────────────────────────
+───────┬────────────────────────────────────────────────────────────────────────
+       │ File: sample.rs
+───────┼────────────────────────────────────────────────────────────────────────
+   1   │ /// A rectangle. First line is changed to prevent a regression of #1869
+   2   │ struct Rectangle {
```
Enselic, line 5: Can we prevent formatting to change in this file (and others?)

Enselic (changes requested):
Hi! A took a quick look and had some drive-by comments.

doc/long-help.txt
einfachIrgendwer0815: This is helpful for files (especially pseudo-files) that have very long lines or do not contain any LF characters at all. One example given in #636 is `/dev/zero`. Let us assume there is the soft limit but not the hard limit. When reading such a file, bat would hang forever while wasting CPU time.

So the soft limit prevents out-of-memory events, the hard limit additionally prevents hanging infinitely.

Admittedly, the use case is tiny. Removing the hard limit is probably fine, since oom crashes are already prevented by the soft limit alone.

tests/snapshots/output/grid_header_numbers.snapshot.txt
einfachIrgendwer0815: The inserted spaces are there to leave room for the `!` that is placed in front of replaced lines, so that the formatting does not look off in those cases. Without these placeholder spaces the formatting would stay as it was, however, the grid would look weird for replaced lines.

It would look something like this:
```
 51 │ normal content
 52 ! │ <line too long>
 53 │ normal content
```

The other alternative is to not add the `!` altogether:
```
 51 │ normal content
 52 │ <line too long>
 53 │ normal content
```
But, especially without color, it would be somewhat easy to overlook that a line was replaced.

Issue #2894, opened 2024-03-11
Allow turning on and off styles, relative to current settings


docwhat:
It would be nice to be able to turn on and off styles relative to the current settings.

For example, using `-` could disable a style, and `+` could enable a style.

Some examples:

- `--style=full --style=-numbers` -- Everything except line numbers.
- `--style=full,-numbers` -- Everything except line numbers.
- `--style=+headers,+rule,-changes` -- Default, with headers and rules without changes.

Related issues:
- #5
- #72
- #1032
- #1161
- #1741

einfachIrgendwer0815:
Also related to #1032

docwhat:
> Also related to #1032

I missed that. Sorry!

Issue #1741, opened 2021-07-20
Disabling individual style components without otherwise affecting existing style


lilyball:
I'd really like a way to turn off individual style components (such as `numbers`) without otherwise affecting the style. This way I can e.g. disable `numbers` and `changes` (thus removing the gutter) while still respecting user choice for header/grid.

This could be implemented like `--style=-numbers`, to remove `numbers` from the existing style, or `--style=+numbers` to add it. These could of course be grouped, like `--style=-numbers,+grid`. Mixing this syntax with an absolute component should be prohibited (e.g. `--style=changes,-numbers`), and this can be done in separate flags if desired (e.g. `--style=auto --style=-numbers` to override style completely but disable numbers). Alternatively the syntax could look like `--style+=numbers,grid --style-=changes` but I can't think offhand of any precedent for this option syntax, so I think just supporting -/+ prefixes on the components makes more sense.

The motivation here is I want numbers off by default, but I don't want to override any other style choices set in the `BAT_STYLE` env var (experimentally, the `BAT_STYLE` var is applied before reading my config file). Setting `--style=-numbers` also shouldn't disable the `--style=auto` behavior, whereas `--style=changes,header,grid,snip` would.

### Open question

What should `--style=auto --style=+numbers` do when piped? Should it retain the auto behavior overall, or should it always display numbers? If it's the former, how about `--style=auto --style=-numbers --style=+numbers`? I can't decide which way is better. Though really the problem here is that `auto` is a specific style instead of being a separate flag, or even a modifier to individual components (that would actually be really nice, if I could e.g. turn off numbers/changes when piped but keep header/grid/snip).

I suppose right now I'm inclined to say `--style=auto --style=-numbers --style=+numbers` should definitely display numbers when piped, and `--style=auto --style=+numbers` probably should as well just for consistency (i.e. `--style=+numbers` means display numbers even when piped, no matter what the inherited style is).

And then I recommend considering the idea of having a per-component modifier for auto behavior (such as `--style=header,grid,snip,*numbers,*changes` and `--style=+*numbers`; or maybe some other sigil that doesn't need shell escaping). I'll file another issue on this. (Edit: #1743)

sharkdp:
Thank you for your request.

I haven't fully read it but it looks like it's related to #1032 and #1161 (and tickets linked therein). It would be great if you could comment on #1032 instead, if your request adds new information (it looks like it does).

Issue #1161, opened 2020-08-31
Subtractive --style arguments


ssokolow:
Currently, it seems that I either have to always specify `changes` and `header`, write a wrapper which builds a `--style` string based on the arguments passed, or live with `grid`, which draws too many lines.

What I'd really like is the ability to specify something like `--style auto,-numbers,-grid`.

ssokolow:
...and I just realized that this is a duplicate of #1032 which I missed because, somehow, your issue tracker is bugged out and producing open non-pull-requests which fail to show up under the default `is:issue is:open` filter.

PR #2929, opened 2024-04-07
Merge `--style`s from different places using `+`/`-`

eth-p:
This commit aims to solve #1032, #1161, #1741, and #2894 by introducing a mechanism for merging the styles specified in the config file, `BAT_STYLE` environment variable, and the `--style` command-line argument.

This does so by treating each `--style` argument as a list of style components, and introducing a new `+` and `-` component prefix (e.g. `+numbers`, `-grid`) that allows `--style` lists to be merged together. When a style list consists only of prefixed components, it will be merged into the list that comes before it. If it contains any non-prefixed components, it will *replace* the list before it instead.

Ultimately, the style lists are merged into a final set of style components roughly following this pseudocode algorithm:

```rust
lists.fold(
    init=set(),
    |accumulator, list| {
        if list.components.has_any(|component| component.prefix is "") {
            accumulator.clear()
        }
        
        for component in list.components {
            if component.prefix == "-" {
                accumulator.remove_all(component.expand())
            } else {
                accumulator.add_all(component.expand())
            }
        }
        
    }
)
```

This preserves the existing behavior of `--style` taking precedence over `BAT_STYLE` or the config file, while also allowing users to add or remove styles from individual runs of `bat`.

Examples:

* `BAT_STYLE=full bat --style=plain` => (none)  
  Overriding the style from the config file or environment variable. (current behavior)
  image

* `bat --style=default,-header,-numbers` => `changes,grid,snip`  
  Removing components from the default/full style.
  image

* `BAT_STYLE=full bat --style=-grid` => `changes,header-filename,header-filesize,numbers,snip`  
  Removing components specified from the config file or environment variable.
  image

* `BAT_STYLE=grid,numbers bat --style=+header` => `grid,numbers,header-filename`  
  Adding components from the config file or environment variable. (current behavior)
  image

## Questions & Answers

**Q. Why introduce `+component` instead of using `component` without the prefix?**  
**A.** In part due to technical limitations, and in part to avoid breaking users' existing configurations by preserving the original behavior of `--style`.

**Q. Why did you remove `overrides_with`?**  
**A.** It forced clap's argument matcher to discard all but the most recent occurrence of `--style`.

## Unresolved Questions

* ~~What should we do when there is only one `--style`, and it contains something like `+numbers`? It's currently treated like `--style=numbers`; would it be better to treat it like `default,+numbers`?~~

* ~~Why is [this test](https://github.com/sharkdp/bat/actions/runs/8585106737/job/23526154120?pr=2929) failing under the MSRV toolchain?~~

### DIFF
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -2810,3 +2810,71 @@ fn strip_ansi_auto_does_not_strip_ansi_when_plain_text_by_option() {
 
     assert!(output.contains("\x1B[33mYellow"))
 }
+
+// Tests that style components can be removed with `-component`.
+#[test]
+fn style_components_can_be_removed() {
+    bat()
+        .arg({
+            #[cfg(not(feature = "git"))]
+            {
+                "--style=full,-grid"
+            }
+            #[cfg(feature = "git")]
+            {
+                "--style=full,-grid,-changes"
+            }
+        })
+        .arg("--decorations=always")
+        .arg("--color=never")
+        .write_stdin("test")
+        .assert()
+        .success()
+        .stdout("     STDIN\n     Size: -\n   1 test\n")
+        .stderr("");
+}
+
+// Tests that style components are chosen based on the rightmost `--style` argument.
+#[test]
+fn style_components_can_be_overidden() {
+    bat()
+        .arg("--style=full")
+        .arg("--style=header,numbers")
+        .arg("--decorations=always")
+        .arg("--color=never")
+        .write_stdin("test")
+        .assert()
+        .success()
+        .stdout("     STDIN\n   1 test\n")
+        .stderr("");
+}
+
+// Tests that style components can be merged across multiple `--style` arguments.
+#[test]
+fn style_components_will_merge() {
+    bat()
+        .arg("--style=header,grid")
+        .arg("--style=-grid,+numbers")
+        .arg("--decorations=always")
+        .arg("--color=never")
+        .write_stdin("test")
+        .assert()
+        .success()
+        .stdout("     STDIN\n   1 test\n")
+        .stderr("");
+}
+
+// Tests that style components can be merged with the `BAT_STYLE` environment variable.
+#[test]
+fn style_components_will_merge_with_env_var() {
+    bat()
+        .env("BAT_STYLE", "header,grid")
+        .arg("--style=-grid,+numbers")
+        .arg("--decorations=always")
+        .arg("--color=never")
+        .write_stdin("test")
+        .assert()
+        .success()
+        .stdout("     STDIN\n   1 test\n")
+        .stderr("");
+}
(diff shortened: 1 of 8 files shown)


eth-p:
Converting to a draft until I can figure out why one of the tests is failing when building with the MSRV toolchain.

tests/integration_tests.rs
```diff
@@ -2631,3 +2631,62 @@ fn highlighting_independant_from_map_syntax_case() {
...
+    bat()
+        .arg("--style=full,-grid")
+        .arg("--decorations=always")
+        .arg("--color=never")
+        .write_stdin("test")
+        .assert()
+        .success()
+        .stdout("       STDIN\n       Size: -\n   1   test\n")
+        .stderr("");
+}
```
einfachIrgendwer0815, line 2647: On the MSRV toolchain, `cargo test` is run without the `git` feature. Therefore, the style component `full` does not contain `changes`, so less space is allocated for the sidebar.
```suggestion
#[test]
fn style_components_can_be_removed() {
    bat()
        .arg({
            #[cfg(not(feature = "git"))]
            {
                "--style=full,-grid"
            }
            #[cfg(feature = "git")]
            {
                "--style=full,-changes,-grid"
            }
        })
        .arg("--decorations=always")
        .arg("--color=never")
        .write_stdin("test")
        .assert()
        .success()
        .stdout("     STDIN\n     Size: -\n   1 test\n")
        .stderr("");
}
```

tests/integration_tests.rs
```diff
@@ -2631,3 +2631,62 @@ fn highlighting_independant_from_map_syntax_case() {
         .stdout(expected)
         .stderr("");
 }
+
+// Tests that style components can be removed with `-component`.
+#[test]
+fn style_components_can_be_removed() {
+    bat()
+        .arg("--style=full,-grid")
```
einfachIrgendwer0815, line 2639:
```suggestion
        .arg({
            #[cfg(not(feature = "git"))]
            {
                "--style=full,-grid"
            }
            #[cfg(feature = "git")]
            {
                "--style=full,-changes,-grid"
            }
        })
```

tests/integration_tests.rs
```diff
@@ -2631,3 +2631,62 @@ fn highlighting_independant_from_map_syntax_case() {
...
+        .arg("--decorations=always")
+        .arg("--color=never")
+        .write_stdin("test")
+        .assert()
+        .success()
+        .stdout("       STDIN\n       Size: -\n   1   test\n")
```
einfachIrgendwer0815, line 2645:
```suggestion
        .stdout("     STDIN\n     Size: -\n   1 test\n")
```

einfachIrgendwer0815:
The test run on the MSRV toolchain runs `cargo test` without the `git` feature. Therefore, the style component `full` does not contain `changes`, so less space gets allocated to the sidebar. I commented two suggestions that should fix the error.

eth-p:
@einfachIrgendwer0815 Ah, that explains a lot! Thank you so much for pointing that out; it saved me a lot of debugging time. Rebased with your suggestion to fix the failing test :)

keith-hall (approved):

eth-p:
Rebased on top of https://github.com/sharkdp/bat/commit/b7e44c76dc7bfe1d4bf6693ac20abdbbc48754b9.

Also made a new commit to answer the question of how `--style=+numbers`/`--style=-numbers` would behave if `--style` was only specified once. It was more consistent and intuitive to apply the change to the "default" style, rather than applying it to an empty set of component.

PR #2930, opened 2024-04-07
docs: add install instructions to Exherbo Linux in README.md

davlgd:


Enselic:
Hmm there are hundreds if not thousands of linux distros and I don't think we should have an entry in our README.md for every single one. Not sure what other maintainers think though.

It's hard to define a rule for what distros should be in the readme but as an aproximation I'll use "I personally have heard of it before".

So I suggest we close this PR I'm afraid.

PR #2932, opened 2024-04-08
glob is called twice

rtczza:
In this context, matcher.glob() already returns a Cow<'a, str> type, so there is no need to call the glob method again. The glob method has been mistakenly called twice in a row here. The correct code should directly convert the Cow<'a, str> to a String type and then push it into Vec<String>. The to_string() method can be used here (which can accept Cow<'a, str> and clone or borrow as needed) to complete the conversion.

The above is my own understanding, not necessarily correct, welcome to discuss.


rtczza:
has been rebased and this submission has been removed.

PR #2937, opened 2024-04-14
display which theme is the default one in basic output

sblondon:
The previous [PR #2838](https://github.com/sharkdp/bat/pull/2838) shows which theme is the default one in colored output when `--list-themes` argument is used. This PR does the same with not colored output (as suggested by @Enselic ).

The output is:

bat_default_theme

I think about another commit to refactor the internal of `list_themes()` function (so it should not the squashed with the previous one).

Current code: 

```rust
    if config.colored_output {
        for theme in assets.themes() {
            // display colored theme
            // display "Further themes can be installed..."
    } else {
        for theme in assets.themes() {
           // display basic theme
        }
    }
```

To:
```rust
    for theme in assets.themes() {
        if config.colored_output {
            // display colored theme
        } else {
           // display basic theme
       }
    }
    if config.colored_output {
            // display "Further themes can be installed..."
    } 

```

What do you think about it?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 - Relax syntax mapping rule restrictions to allow brace expansion #2865 (@cyqsimon)
 - Apply clippy fixes #2864 (@cyqsimon)
 - Faster startup by offloading glob matcher building to a worker thread #2868 (@cyqsimon)
+- Display which theme is the default one in basic output (no colors), see #2937 (@sblondon)
 - Display which theme is the default one in colored output, see #2838 (@sblondon)
 
 ## Syntaxes
--- a/src/bin/bat/main.rs
+++ b/src/bin/bat/main.rs
@@ -200,8 +200,8 @@ pub fn list_themes(cfg: &Config, config_dir: &Path, cache_dir: &Path) -> Result<
     let stdout = io::stdout();
     let mut stdout = stdout.lock();
 
+    let default_theme = HighlightingAssets::default_theme();
     if config.colored_output {
-        let default_theme = HighlightingAssets::default_theme();
         for theme in assets.themes() {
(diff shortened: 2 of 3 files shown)


sblondon:
I added a [new commit](https://github.com/sharkdp/bat/pull/2937/commits/778e8731624b85d2f5f2ea6327ef788e34843c06) with the refactoring explained in the previous comment. So you can review it directly.

IMO, it's better with this refactoring.
However if you don't like it, I will drop the commit.

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@
 - Relax syntax mapping rule restrictions to allow brace expansion #2865 (@cyqsimon)
 - Apply clippy fixes #2864 (@cyqsimon)
 - Faster startup by offloading glob matcher building to a worker thread #2868 (@cyqsimon)
+- Display which theme is the default one in basic output (no colors), see #2937 (@sblondon)
```
Enselic, line 37: We could combine these two lines into one and reference both PRs on the same line, but let's do that right before release when we need to clean up the changelog anyway.

Enselic (approved):
Looks reasonable! I had one nitpick but we can fix that later, to avoid another round of back and forth.

Thanks!

CHANGELOG.md
sblondon: I'm not sure who is 'we'. Do you mean the `bat` maintainers? I can do a PR for that but I doubt it worth the time.

Enselic: Yep I mean the bat maintainers, but feel free to do a PR if you like

sblondon: I done it but the CI requires to add a line in Changelog for the PR so it's not useful. I closed the PR and you or another bat maintainers will be more efficient on this topic.

PR #2938, opened 2024-04-15
enable the 'conversions' feature

bin-ly:
use of deprecated associated function `os_str_bytes::RawOsString::assert_from_raw_vec`


Enselic (commented):

I'm not quite following what you are doing here. In what way is bat better after this change than it was before? For example from the perspective of a user, or from the perspective of a developer.

Enselic:
Closing as inactive

PR #2947, opened 2024-04-26
Make .debdiff imply .diff using syntax_mapping toml

### LINKED ISSUE #2940: Please support .debdiff files
Debian source diff files are often named `.debdiff`. They are normal diff files. Please support coloring them by just treating them as diff files.

jacg:
Closes #2940

I would still like some guidance on what tests need to be added.


cyqsimon:
If it's a Debian-specific extension, it should go into `/src/syntax_mapping/builtins/linux/` rather than `unix-family/` no?

cyqsimon:
And I don't think we need tests for trivial additions like these. The build will fail if the TOML file is bad.

keith-hall (approved):

PR #2948, opened 2024-04-30
Add list of possible styles to README

ghostdevv:
I noticed it wasn't present in the README but is in the man pages (which I missed first time reading)

### DIFF
--- a/README.md
+++ b/README.md
@@ -505,7 +505,7 @@ You can use the `--style` option to control the appearance of `bat`s output.
 You can use `--style=numbers,changes`, for example, to show only Git changes
 and line numbers but no grid and no file header. Set the `BAT_STYLE` environment
 variable to make these changes permanent or use `bat`s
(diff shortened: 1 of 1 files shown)


ghostdevv:
Looks like CI is failing on checking for changelog entry, but [`CONTRIBUTING.md`](https://github.com/sharkdp/bat/blob/b4e3a84e1a6e68b2211bdebbbb0873f1d4aa2f8d/CONTRIBUTING.md#add-an-entry-to-the-changelog) says that a docs change doesn't need one. Let me know if I should add anyway!

Enselic:
if we want to add more info I think we should add them to the linked section. but I don't think we need to add them. the instructions says to check `--help`. and keeping this list up to date will just be extra work. sorry and thank you for understanding.

ghostdevv:
no worries! iirc I added the change since at the time I found it difficult to find the information - so perhaps it wasn't in the --help. glad it is now!

PR #2956, opened 2024-05-08
Only start offload worker thread when there's more than 1 core

cyqsimon:
An oversight on my part that only came to mind when I'm working on a single core transit VPS today.


cyqsimon:
Rebased.

cyqsimon:
On further thought, maybe `available_parallelism().map(|n| n.get() > 1).unwrap_or(false)` is better.

Per [documentation](https://doc.rust-lang.org/beta/std/thread/fn.available_parallelism.html#errors):

> This function will, but is not limited to, return errors in the following cases:
> - If the amount of parallelism is not known for the target platform.
> - If the program lacks permission to query the amount of parallelism made available to it.

In neither case should we really fail fast IMO. What do you think?

Issue #2969, opened 2024-05-27
Maybe patch the newest lisp syntaxes?


ccqpein:
**What steps will reproduce the bug?**

Create new lisp file by:

`echo '(defun test () :a-from-b)' > ./test.lisp`

**What happens?**

Then, when I use `bat ./test.lisp`, the `-from-` inside `a-from-b` is highlighted like keyword. But it shouldn't be.

**What did you expect to happen instead?**

Render it like the normal value.

**How did you install `bat`?**

`cargo`

**bat version and environment**

`v0.24.0` and `macOS 14.5`

**how I fix it**

I reinstall the newest syntaxes from sublime as the README provided [adding-new-syntaxes--language-definitions](https://github.com/sharkdp/bat?tab=readme-ov-file#adding-new-syntaxes--language-definitions)

```
➜ ls "$(bat --config-dir)/syntaxes"
Lisp.sublime-syntax #<= the master version of sublime Package repo
```

Then run the `bat cache --build`. And the highlight is good enough.

So, does Bat need one more patch to update the `Lisp.sublime-syntax`? I noticed that the version of `assets/syntaxes/01_Packages` was years ago.

Thank you

ccqpein:
Gave a potential fix in PR #2970

PR #2970, opened 2024-05-27
Patch the lisp syntax

ccqpein:
The potential solution of the issue #2969.

I am not sure if it is the right way to update syntax of bat, and it is my first git patch. Feel free to change anything. 

Thanks


keith-hall:
this looks okay to me, this `sublime-syntax` doesn't seem to use any more recent features which the syntax highlighting engine `syntect` doesn't support yet.
In order to get the CI step to pass, you'll need to update the "highlighted file" - see https://github.com/sharkdp/bat/blob/master/doc/assets.md#syntax-tests. I would also recommend including some code in the "source file" which this upgrade solves highlighting for, like you mentioned `(defun test () :a-from-b)`

ccqpein:
Sure, let me check the test file.

PS: I think I don't need to add this code snippet as the new test. Because the current test sets already show the issue (except it is `-test-` rather than the `-from-`). 

Screenshot 2024-05-28 at 21 05 17

keith-hall (approved):

keith-hall:
Thanks for your contribution 🙂

ccqpein:
@keith-hall no problem. Thanks for your project maintaining too!

PR #2973, opened 2024-05-29
Add 'c' parameter to 'less' command

KaKi87:
As per https://github.com/sharkdp/bat/discussions/2912#discussioncomment-9596674

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -89,7 +89,7 @@ impl OutputType {
                 }
 
                 if wrapping_mode == WrappingMode::NoWrapping(true) {
-                    p.arg("-S"); // Short version of --chop-long-lines for compatibility
+                    p.arg("-Sc"); // Short version of --chop-long-lines --clear-screen for compatibility
                 }
 
                 // Passing '--no-init' fixes a bug with '--quit-if-one-screen' in older


src/output.rs
```diff
@@ -89,7 +89,7 @@ impl OutputType {
                 }
 
                 if wrapping_mode == WrappingMode::NoWrapping(true) {
-                    p.arg("-S"); // Short version of --chop-long-lines for compatibility
+                    p.arg("-Sc"); // Short version of --chop-long-lines --clear-screen for compatibility
```
keith-hall, line 92: I wonder if it's worth trying to add an integration test for this... I guess probably more hassle than it's worth :) 
can we be confident that `-c` flag is standard across most `less` versions/implementations in the wild?

KaKi87: > I wonder if it's worth trying to add an integration test for this... I guess probably more hassle than it's worth :)

Well, in any case, that's not something I could help with, I only made this edit because you showed me where to make it :sweat_smile: 

> can we be confident that `-c` flag is standard across most `less` versions/implementations in the wild?

Definitely 🤣

[`v160   7/29/90  Clear screen before initial output w/ -c.`](https://github.com/gwsw/less/blob/25d08b356e98b34ee2c1cc63d7170e0dea87290d/version.c#L266)

keith-hall (approved):

keith-hall:
Probably it could be worth mentioning this at https://github.com/sharkdp/bat?tab=readme-ov-file#using-less-as-a-pager

KaKi87:
Hello,
Any news on this ?
Thanks

Enselic:
this is a minefield, we fix a bug for someone we introduce a bug for someone else

can we write a regression test for your use case_ that we will at least not break it in the future if this turns out to be problematic and we found other solutions

I don't fully understand the problem, it would be helpful if you could elaborate

KaKi87:
> we fix a bug for someone we introduce a bug for someone else

What bugs are you referring to ?

> can we write a regression test for your use case_ that we will at least not break it in the future if this turns out to be problematic and we found other solutions

Sure.

> I don't fully understand the problem, it would be helpful if you could elaborate

I provided an explanation with an attached GIF there : https://github.com/sharkdp/bat/discussions/2912#discussion-6405610

Enselic:
I'm not referring to specific bugs but to the risk of introducing bugs by changing behavior.

But if we (with "we" I really mean "you" ;) add a regression test for this case, that's fine. Then it's up to those who depend on the current behavior to fix any follow-up bugs and write regression tests for their uses cases.

I can reproduce the problem with this btw:

1. `seq 1 200 | tr '\n' ' ' | bat --wrap=never`
2. Scroll around

**Expected result:**
The line is always at the top of the terminal.

**Actual result:**
The line starts at the bottom and jumps to the top, as in your GIF.

KaKi87:
> (with "we" I really mean "you" ;)

Oh 😅 I don't how to do that lol

I don't do Rust, but even if I did or if this was a JS project (that's the one language I'm fluent at), I actually don't see how this could be checked, as it's not just manipulating strings, but checking scroll (and text-based scroll sounds even harder than web scroll) ?

Enselic:
For the record, if you are comfortable to merge this without a regression test @keith-hall, that's fine with me! (I noticed your PR Approval still stands.)

Enselic:
Closing for now to keep PR inbox clean but am of course happy to re-open if you get around writing a test.

KaKi87:
Well, you pinged someone about merging this without a test, and they never responded :sob: 

Plus, I also said I wouldn't know how to do that :sob:

PR #2976, opened 2024-05-31
Allow printing binary content by treating it the same as text

### LINKED ISSUE #2709: Allow `bat` to render some binary content
```
$ bat /proc/3151/cmdline
───────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: /proc/3151/cmdline   <BINARY>
───────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
$ /bin/cat /proc/3151/cmdline;
sleep100%
```

I was trying to read `/proc/[pid]/cmdline` using `bat` and expected it to behave like `cat`, but it doesn't.

Should any exceptions be made for the `/proc` filesystem?

See `man 5 proc`:

```
PROC(5)                                                   Linux Programmer's Manual                                                   PROC(5)

NAME
       proc - process information pseudo-filesystem

DESCRIPTION
       The  proc  filesystem  is a pseudo-filesystem which provides an interface to kernel data structures.  It is commonly mounted at /proc.
       Typically, it is mounted automatically by the system, but it can also be mounted manually using a command such as:

           mount -t proc proc /proc

       Most of the files in the proc filesystem are read-only, but some files are writable, allowing kernel variables to be changed.
```

### LINKED ISSUE #2974: Add Flag to Avoid Treating NUL Separated Input as Binary
### Discussed in https://github.com/sharkdp/bat/discussions/2971

---

### Issue

Currently, running a command like the following will print a warning:
- Related Issue #823

```sh
printf "First\0" | bat -p
```

SCR-20240529-trdo

The warning is defined in `src/printer.rs`:

https://github.com/sharkdp/bat/blob/8f8c953ab69403a7acf516e65265b7f70bc3ec73/src/printer.rs#L435-L444

The decision to label the input as `BINARY` seems to be made in `src/input.rs`:

https://github.com/sharkdp/bat/blob/8f8c953ab69403a7acf516e65265b7f70bc3ec73/src/input.rs#L260-L271

A _hacky_ workaround is to make the first line empty, use `bat`, and then remove the first line:

```sh
printf "\nFirst\0" | bat -p | sed '1d'
```

---

### Proposed solution

A new flag that doesn't label `content_type` as `BINARY` when the first line ends with a `NUL` byte:

```sh
# naming the flag '--text' to align with 'grep/git diff'
printf "First\0" | bat -p --text
```

The `crate` [^1] used to determine if content is `binary` states:

```rs
//! encoding). Note that **this analysis can fail**. For example, even if unlikely, UTF-8-encoded
//! text can legally contain NULL bytes. Conversely, some particular binary formats (like binary
```

Based on this, a `--text` flag would be very appropriate, similar to how `grep` and `git diff` have one as well.

```sh
printf "First\0" | grep 'First'
# grep: (standard input): binary file matches

printf "First\0" | grep --text 'First'
# First
```

[^1]: [sharkdp/content_inspector: Fast inspection of binary buffers to guess/determine the type of content](https://github.com/sharkdp/content_inspector)

einfachIrgendwer0815:
Adds a CLI option called `--binary` which controls how binary content is handled. Possible values:
  * `no-printing`: do not print any content from a binary file (default; basically what bat currently does)
  * `as-text`: treat the binary as if it was text -> it will get printed as is

Fixes #2974.
Fixes #2709.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Syntax highlighting for JavaScript files that start with `#!/usr/bin/env bun` #2913 (@sharunkumar)
 - `bat --strip-ansi={never,always,auto}` to remove ANSI escape sequences from bat's input, see #2999 (@eth-p)
 - Add or remove individual style components without replacing all styles #2929 (@eth-p)
+- Add option `--binary=as-text` for printing binary content, see issue #2974 and PR #2976 (@einfachIrgendwer0815)
 
 ## Bugfixes
 
--- a/doc/long-help.txt
+++ b/doc/long-help.txt
@@ -20,6 +20,13 @@ Options:
             * unicode (␇, ␊, ␀, ..)
             * caret   (^G, ^J, ^@, ..)
 
+      --binary <behavior>
+          How to treat binary content. (default: no-printing)
+          
+          Possible values:
+            * no-printing: do not print any binary content
+            * as-text: treat binary content as normal text
+
   -p, --plain...
           Only show plain style, no decorations. This is an alias for '--style=plain'. When '-p' is
           used twice ('-pp'), it also disables automatic paging (alias for '--style=plain
--- a/doc/short-help.txt
+++ b/doc/short-help.txt
@@ -11,6 +11,8 @@ Options:
           Show non-printable characters (space, tab, newline, ..).
       --nonprintable-notation <notation>
           Set notation for non-printable characters.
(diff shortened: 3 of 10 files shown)


einfachIrgendwer0815:
Rebased onto `master` to resolve conflicts

keith-hall (approved):

Issue #2935, opened 2024-04-11
compatibility issue with cat while piping in scripts


arnvgh:
**What steps will reproduce the bug?**

1. get into any folder with some files and execute `ls | cat -n | less` it should give output with numbers (as the `cat` man page says `-n` for number all output lines) 
```
     1  bar
     2  foo
(END)
```
**NOTE:** even if we don't pipe it with `less` it will give the same output. (This is where the incompatibility comes in)

2. Doing the same thing with bat `ls | bat -n | less` omits the numbers and prints
```
bar
foo
(END)
```
while when I don't pipe the same command in `less` it outputs the following 
```
   1 bar
   2 foo
``` 

**What did you expect to happen instead?**

As `bat` meant to be a `cat(1) clone` I assume the behavior should be consistent (same in both the cases as `cat`). 
It should not remove the number lines when it is called with the `--number|-n` flag even while piping into other command such as `less`.

**How did you install `bat`?**
```bash
sudo dnf install bat
```
**bat version and environment**
#### Software version

bat 0.24.0

#### Operating system

Linux 6.8.4-200.fc39.x86_64

#### Command-line

```bash
bat --diagnostic
```

#### Environment variables

```bash
SHELL=/usr/bin/zsh
PAGER=less
LESS=-R
LANG=en_US.UTF-8
LC_ALL=<not set>
BAT_PAGER=<not set>
BAT_PAGING=<not set>
BAT_CACHE_PATH=<not set>
BAT_CONFIG_PATH=<not set>
BAT_OPTS=<not set>
BAT_STYLE=<not set>
BAT_TABS=<not set>
BAT_THEME=<not set>
XDG_CONFIG_HOME=/home/ag/.config
XDG_CACHE_HOME=/home/ag/.cache
COLORTERM=truecolor
NO_COLOR=<not set>
MANPAGER=<not set>
```

#### System Config file

Could not read contents of '/etc/bat/config': No such file or directory (os error 2).

#### Config file

Could not read contents of '/home/ag/.config/bat/config': No such file or directory (os error 2).

#### Custom assets metadata

Could not read contents of '/home/ag/.cache/bat/metadata.yaml': No such file or directory (os error 2).

#### Custom assets

'/home/ag/.cache/bat' not found

#### Compile time information

- Profile: release
- Target triple: x86_64-unknown-linux-gnu
- Family: unix
- OS: linux
- Architecture: x86_64
- Pointer width: 64
- Endian: little
[...]

PR #2983, opened 2024-06-01
Adding line number printing when output is piped out

domenicomastrangelo:
Updating print_line function in SimplePrinter to use line numbers when printing a line if line numbers are desired (self.config.style_components.numbers() / StyleComponent::LineNumbers).

This was a compatibility issue with cat until now as describe in issue [#2935](https://github.com/sharkdp/bat/issues/2935).

Updating of some integration tests was necessary as now we expect the output to contain line numbers.

### DIFF
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -48,7 +48,7 @@ fn basic() {
         .arg("test.txt")
         .assert()
         .success()
-        .stdout("hello world\n")
+        .stdout("   1 hello world\n")
         .stderr("");
 }
 
@@ -58,7 +58,7 @@ fn stdin() {
         .write_stdin("foo\nbar\n")
         .assert()
         .success()
-        .stdout("foo\nbar\n");
+        .stdout("   1 foo\n   2 bar\n");
 }
 
 #[test]
@@ -68,7 +68,7 @@ fn concatenate() {
         .arg("test.txt")
         .assert()
         .success()
-        .stdout("hello world\nhello world\n");
+        .stdout("   1 hello world\n   1 hello world\n");
 }
 
 #[test]
@@ -80,7 +80,7 @@ fn concatenate_stdin() {
         .write_stdin("stdin\n")
         .assert()
         .success()
-        .stdout("hello world\nstdin\nhello world\n");
+        .stdout("   1 hello world\n   1 stdin\n   1 hello world\n");
 }
 
 #[test]
@@ -90,7 +90,7 @@ fn concatenate_empty_first() {
         .arg("test.txt")
         .assert()
         .success()
-        .stdout("hello world\n");
+        .stdout("   1 hello world\n");
 }
 
 #[test]
@@ -100,7 +100,7 @@ fn concatenate_empty_last() {
         .arg("empty.txt")
         .assert()
         .success()
-        .stdout("hello world\n");
+        .stdout("   1 hello world\n");
 }
 
 #[test]
@@ -121,7 +121,7 @@ fn concatenate_empty_between() {
         .arg("test.txt")
         .assert()
         .success()
-        .stdout("hello world\nhello world\n");
+        .stdout("   1 hello world\n   1 hello world\n");
 }
 
 #[test]
@@ -132,7 +132,7 @@ fn concatenate_empty_first_and_last() {
         .arg("empty.txt")
         .assert()
         .success()
-        .stdout("hello world\n");
+        .stdout("   1 hello world\n");
 }
 
 #[test]
@@ -142,7 +142,7 @@ fn concatenate_single_line() {
         .arg("single-line.txt")
         .assert()
         .success()
-        .stdout("Single LineSingle Line");
+        .stdout("   1 Single Line   1 Single Line");
 }
 
 #[test]
@@ -153,7 +153,7 @@ fn concatenate_single_line_empty() {
         .arg("single-line.txt")
         .assert()
         .success()
-        .stdout("Single LineSingle Line");
+        .stdout("   1 Single Line   1 Single Line");
 }
 
 #[test]
@@ -174,7 +174,7 @@ fn line_range_2_3() {
         .arg("--line-range=2:3")
         .assert()
         .success()
-        .stdout("line 2\nline 3\n");
+        .stdout("   2 line 2\n   3 line 3\n");
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


tests/integration_tests.rs
```diff
@@ -1779,7 +1779,7 @@ fn file_with_invalid_utf8_filename() {
         .arg(file_path.as_os_str())
         .assert()
         .success()
-        .stdout("dummy content\n");
+        .stdout("   1 dummy content\n");
```
keith-hall, line 1782: I'm curious, is this really the expected behavior? The test invokes `bat` without any line number style component arguments. I think it should stay as it was before - according to the issue being solved, line numbers should only be displayed when piping when the command line argument was explicitly passed, no?

domenicomastrangelo: good catch!

I added a check for handle [here](https://github.com/sharkdp/bat/pull/2983/files#diff-64932fd51f70fd457567303b06d7e4e92efb03febb303d5a1e0ba7ac82d4014dR168)

domenicomastrangelo: this doesn't pass the test now, it seems like there may be an issue with the filename and the is_terminal() function or something like that

> fn is_terminal(&self) -> bool
> 
> ───────────────────────────────────────────
> Returns `true` if the descriptor/handle refers to a terminal/tty.
> 
> On platforms where Rust does not know how to detect a terminal yet, this will return
`false`. ***This will also return `false` if an unexpected error occurred, such as from
passing an invalid file descriptor.***

domenicomastrangelo: debugging it shows self.config.style_components.numbers() returns true, is there a different way to know if line numbers are requested that i'm missing?

einfachIrgendwer0815: My guess would be the following:

The `SimplePrinter` is used whenever stdout is not interactive (so it's paging), `--color` is not set to `always`, `--decorations` is not set to `always` and `--force-colorization` is not set either (see [app.rs](https://github.com/sharkdp/bat/blob/b7e44c76dc7bfe1d4bf6693ac20abdbbc48754b9/src/bin/bat/app.rs#L225-L232), [config.rs](https://github.com/sharkdp/bat/blob/b7e44c76dc7bfe1d4bf6693ac20abdbbc48754b9/src/config.rs#L53-L54) and [controller.rs](https://github.com/sharkdp/bat/blob/b7e44c76dc7bfe1d4bf6693ac20abdbbc48754b9/src/controller.rs#L177-L187)). `--style` controls which style components (such as header or line numbers) are used.  The default for `--style` is `changes,grid,header-filename,numbers,snip`. This default currently stays the same even if `loop_through == true` (which means that `SimplePrinter` will be used). Currently, `SimplePrinter` just does not care about `--style`.

So, when `loop_through == true`, the default for `--style` should be `plain` instead. Then your changes should work since they do when you specify `--style=plain` explicitly.

domenicomastrangelo: Thank you for the tip @einfachIrgendwer0815, I updated the code now and it's working :)

einfachIrgendwer0815:
I think I've found a simpler approach. The following should do the trick:

```patch
diff --git a/src/bin/bat/app.rs b/src/bin/bat/app.rs
index 8f69870f..cf385c88 100644
--- a/src/bin/bat/app.rs
+++ b/src/bin/bat/app.rs
@@ -251,9 +251,10 @@ impl App {
             loop_through: !(self.interactive_output
                 || self.matches.get_one::<String>("color").map(|s| s.as_str()) == Some("always")
                 || self
                     .matches
                     .get_one::<String>("decorations")
                     .map(|s| s.as_str())
                     == Some("always")
-                || self.matches.get_flag("force-colorization")),
+                || self.matches.get_flag("force-colorization")
+                || self.matches.get_flag("number")),
             tab_width: self
```

That way, whenever `-n`/`--number` is specified the default `InteractivePrinter` is used instead of the `SimplePrinter`.

Enselic:
I certainly prefer things to be simple, thanks for the insight. So this PR is not needed then? I'll close it but let me know if I misunderstood something.

lmmx:
I just came across this via #3316 and repeated the same idea in #3438 :upside_down_face: 

The issue reported in #2935 which remains in #3316 is that `-n` line numbers do not pipe.

It looks like a proposal was given above to continue this work (by editing or shipping as a fresh PR):

> I think I've found a simpler approach. The following should do the trick:
> 
> ```diff
> diff --git a/src/bin/bat/app.rs b/src/bin/bat/app.rs
> index 8f69870f..cf385c88 100644
> --- a/src/bin/bat/app.rs
> +++ b/src/bin/bat/app.rs
> @@ -251,9 +251,10 @@ impl App {
>              loop_through: !(self.interactive_output
>                  || self.matches.get_one::<String>("color").map(|s| s.as_str()) == Some("always")
>                  || self
>                      .matches
>                      .get_one::<String>("decorations")
>                      .map(|s| s.as_str())
>                      == Some("always")
> -                || self.matches.get_flag("force-colorization")),
> +                || self.matches.get_flag("force-colorization")
> +                || self.matches.get_flag("number")),
>              tab_width: self
> ```
> 
> That way, whenever `-n`/`--number` is specified the default `InteractivePrinter` is used instead of the `SimplePrinter`.

The logic for this now lives in the `controller.rs` module not `app.rs` but the same general idea as given above could be achieved by doing this at the branch point here where the SimplePrinter is gated:

https://github.com/sharkdp/bat/blob/a730eaae0a17de2c2fd8099471a49e7419db1b54/src/controller.rs#L179-L189

I can put that in a fresh PR (#3438), it sounds like this is a viable approach

That is the same idea I took away from reading through, emphasis added:

> My initial impression is that if this were to change it may surprise people if regular piping suddenly started decorating style.
> 
> But in my case I want it to, at least when I actively pass in a style like `-n`. **Not sure if it'd be possible to distinguish someone passing in the default style vs. it being defaulted to.**
> 
> Maybe we need:
> - a 3rd printer, neither Simple nor Interactive, but de-coloured?
> - an extra config flag, to record if the user explicitly set a flag?
>   - As an initial approximation we could simply set this based on the presence of non-default/non-empty style components
>     - ...in which case it doesn't need to be a new struct field at all
> - _❇️ something else... ❇️_ 

 _Originally posted by @lmmx in [#3316](https://github.com/sharkdp/bat/issues/3316#issuecomment-3410348881)_

`matches.get_flag` indeed gives such a way to distinguish active style setting!

PR #2995, opened 2024-06-09
Update the man page of bat (`assets/manual/bat.1.in`)

akinomyoga:
The man page of `bat` (`assets/manual/bat.1.in`) doesn't seem to be up-to-date compared to the output obtained by `bat --help` and the options completed by `assets/completion/bat.bash.in`.

----

Originally reported by https://github.com/akinomyoga/ble.sh/issues/458#issuecomment-2156378867. A Bash interactive setting `ble.sh` extracts the descriptions of the options generated by programmable completions from the man page, but the discrepancy between the programmable completion `assets/completion/bat.bash.in` and the man page `assets/manual/bat.1.in` causes missing descriptions for some options.

However, I still see discrepancies between different files:

- `doc/long-help.txt` and `bat --help`
- `doc/short-help.txt` and `bat -h`
- `assets/completion/bat.bash.in`
- `assets/completion/bat.fish.in`
- `assets/completion/bat.zsh.in`
- `assets/manual/bat.1.in`

Not a single one out of the above list matches another, which means that there are six different sets of options. Also, some options appearing in several files have different descriptions in the above files. Which one would be considered the up-to-date information that reflects the public interface? Or may all of the options defined in `src/bin/bat/clap_app.rs` be used by the users (i.e., there are no hidden options)?

----

Fixes: #2996 ... I initially didn't include the fix for #2996, but I decided to include also the fix for #2996 in this PR cf https://github.com/sharkdp/bat/pull/2995#issuecomment-2156815629.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@
 - Update the Lisp syntax, see #2970 (@ccqpein)
 - Use bat's ANSI iterator during tab expansion, see #2998 (@eth-p)
 - Support 'statically linked binary' for aarch64 in 'Release' page, see #2992 (@tzq0301)
+- Update options in shell completions and the man page of `bat`, see #2995 (@akinomyoga)
 
 ## Syntaxes
 
--- a/assets/completions/bat.bash.in
+++ b/assets/completions/bat.bash.in
@@ -76,6 +76,7 @@ _bat() {
 	-m | --map-syntax | \
 	--ignored-suffix | \
 	--list-themes | \
+	--squeeze-limit | \
 	--line-range | \
 	-L | --list-languages | \
 	--lessopen | \
@@ -157,6 +158,7 @@ _bat() {
 			--diff-context
 			--tabs
 			--wrap
+			--chop-long-lines
 			--terminal-width
 			--number
 			--color
@@ -169,12 +171,15 @@ _bat() {
 			--ignored-suffix
 			--theme
 			--list-themes
+			--squeeze-blank
+			--squeeze-limit
 			--style
 			--line-range
 			--list-languages
 			--lessopen
 			--diagnostic
 			--acknowledgements
+			--set-terminal-title
 			--help
 			--version
 			--cache-dir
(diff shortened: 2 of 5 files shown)


keith-hall (approved):

keith-hall:
> Or may all of the options defined in `src/bin/bat/clap_app.rs` be used by the users (i.e., there are no hidden options)?

I believe this to be the case, nothing is hidden

akinomyoga:
> I believe this to be the case, nothing is hidden

Thanks! I now checked `clap_app.rs` in detail and noticed that some options are explicitly marked as `.hide(true)`:

- `--no-config` (added in #338)
- `--no-custom-assets` (added in #1758)
- `--no-lessopen` (added in #2444)
- `-P, --no-paging` (added in #1082)
- `--generate-config-file` (added in #885)
- `--config-file` (added in 8dc7e2efa349e372ee359ce38fe6c5555c2d294d)
- `--config-dir` and `--cache-dir` (added in #498)

They seem to be intentionally hidden from the help text. However, the completion settings seem to generate such "hidden" options.

- The first three `--no-*` options were added in the `fish` completion in #2750.
- `-P` is generated by the `fish` completion
- The last three items in the above list are generated by all the `bash`/`zsh`/`fish` completions except for `--cache-dir` missing in the `fish` completion.

I also note that the last four items (`-P`, ... `--config-dir` and `--cache-dir`) are mentioned in the text in the man page (as discussed in #2996).

akinomyoga:
I added the other missing options (except for `--no-paging`) to the man page and the completions in commits 8e00b9ffccb6647fe8e5c69aa6f679594183ad28, 090b60557f65f680108dc99cb27a33c4c4c28755, 02e1188006525e3885e944f5df694ec3d3762cca. Each commit can be dropped if it is unwanted.

akinomyoga:
May I ask what is blocking this PR? Is there anything I can do for this PR?

keith-hall (approved):

akinomyoga:
@keith-hall Wow, thank you for your quick response!

PR #2999, opened 2024-06-11
Add option to remove ANSI escape sequences from bat's input.

eth-p:
This pull request introduces a new `--strip-ansi` option to `bat`.

This option tells `bat` whether to strip ANSI/VT escape sequences from the input file. The following values are supported:
 - `always`: Always strip escape sequences.
 - `never`: Never strip escape sequences. (default)
 - `auto`: Strip escape sequences when printing any language that is *not* "Plain Text".  
    This preserves bat's ability to print piped commands (or pre-highlighted files) and keep their existing colors.

### Disabled, Regular File

image

No significant difference.

### Enabled, Regular File

image

A minor reduction in performance.

### Disabled, Pre-highlighted File

image

No significant difference.

### Enabled, Pre-highlighted File

image

It's certainly an improvement.

### Auto Mode

image

The detection part of `--strip-ansi=auto` causes no significant difference when syntax highlighting is used.

*However*, it does add a few milliseconds of startup time when used with `--color=never`. This is because `--strip-ansi=auto` relies on detecting the syntax in order to determine if a file is plain text or not (and thus whether to strip the ANSI sequences).

image

## TODO

 - Support stripping sequences in `SimplePrinter` (maybe. thoughts?)
 - ~~Wait for #2998.
    It contains the changes made in b4fe182960e45d4fd9ebd5ac2c4d2e4b4522b144, which this PR depends on.~~ 
 - ~~Update `long_help` integration test~~
 - ~~Write tests~~
 - ~~Benchmarks~~
 - ~~Run `cargo fmt`~~
 - ~~Allow `--strip-ansi` to be passed multiple times (so it can be set to auto in config and overridden on the cli)~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -759,8 +759,13 @@ bat() {
 
 If an input file contains color codes or other ANSI escape sequences or control characters, `bat` will have problems
 performing syntax highlighting and text wrapping, and thus the output can become garbled.
-When displaying such files it is recommended to disable both syntax highlighting and wrapping by
+
+If your version of `bat` supports the `--strip-ansi=auto` option, it can be used to remove such sequences
+before syntax highlighting. Alternatively, you may disable both syntax highlighting and wrapping by
 passing the `--color=never --wrap=never` options to `bat`.
+
+> [!NOTE]
+> The `auto` option of `--strip-ansi` avoids removing escape sequences when the syntax is plain text.
 
 ### Terminals & colors
 
--- a/src/bin/bat/clap_app.rs
+++ b/src/bin/bat/clap_app.rs
@@ -403,6 +403,18 @@ pub fn build_app(interactive_output: bool) -> Command {
                 .hide_short_help(true)
         )
         .arg(
+            Arg::new("strip-ansi")
+                .long("strip-ansi")
+                .overrides_with("strip-ansi")
+                .value_name("when")
+                .value_parser(["always", "never"])
+                .default_value("never")
+                .hide_default_value(true)
+                .help("Strip colors from the input (always, *never*)")
+                .long_help("Specify when to strip ANSI escape sequences from the input. Possible values: always, *never*.")
+                .hide_short_help(true)
+        )
+        .arg(
             Arg::new("style")
                 .long("style")
                 .value_name("components")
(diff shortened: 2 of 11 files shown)


eth-p:
Rebased to c264ecd2 and fixed some clippy warnings.

src/bin/bat/clap_app.rs
```diff
@@ -402,6 +402,20 @@ pub fn build_app(interactive_output: bool) -> Command {
...
+            Arg::new("strip-ansi")
+                .long("strip-ansi")
+                .overrides_with("strip-ansi")
+                .value_name("when")
+                .value_parser(["auto", "always", "never"])
+                .default_value("never")
+                .hide_default_value(true)
+                .help("Strip colors from the input (auto, always, *never*)")
+                .long_help("Specify when to strip ANSI escape sequences from the input. \
+                The automatic mode will remove escape sequences unless the detect language \
```
keith-hall, line 415:
```suggestion
                The automatic mode will remove escape sequences unless the syntax highlighting language \
```
"unless the detect language" doesn't quite read right to me. "detected" could work but as it can be explicitly set, I feel calling it "the language used for syntax highlighting" or similar flows better despite being more verbose

README.md
```diff
@@ -759,9 +759,14 @@ bat() {
 
 If an input file contains color codes or other ANSI escape sequences or control characters, `bat` will have problems
 performing syntax highlighting and text wrapping, and thus the output can become garbled.
-When displaying such files it is recommended to disable both syntax highlighting and wrapping by
+
+If your version of `bat` supports the `--strip-ansi=auto` option, it can be used to remove such sequences
```
keith-hall, line 763: Interesting idea to add a hint about version in the readme, especially for features not part of a release yet 🙂

keith-hall (approved):
Nice work, I'm sure lots of users will find this extra functionality useful

src/bin/bat/clap_app.rs
eth-p: Oops, good catch. That was supposed to be "detected," but appears to have lost the "-ed" somewhere along my revisions to it.

README.md
eth-p: It was relevant to the section, and I didn't want to forget to add it later. Had to word it so it didn't seem like it was already available, and that's what I came up with :)

PR #3011, opened 2024-06-28
Fix cmder link on readme file

DrissBoumlik:


Enselic (approved):
I checked with curl and the old page clearly is spam. The new page looks reasonable. Without a doubt it seems like an improvement at least.

PR #3012, opened 2024-06-29
Update winget-releaser to latest

vedantmgoyal9:


vedantmgoyal9:
@sharkdp

Enselic:
why does the user change? also please pin to a specific commit to reduce risk of supply chain attacks (I will audit the action once the commit is pinned)

Enselic:
Thank you for the update. I can see that the pinned action brings in other dependencies that are not locked. Since I don't know about these projects, I can't blindly trust them. Sorry to be boring.

Just an idea, not a decision: Maybe put this CI outside of the bat repository? So that it becomes clear that it is not an official packaging. I'm afraid I personally don't have the time to fully vet this release pipeline. If other maintainers are comfortable merging this, they may of course do so.

Just wanted to write down where I stand rather than say nothing. Sorry if it's disappointing.

Enselic:
Thinking some more about this, I'd say that this PR makes the situation better, since it at least locks to a certain commit. And when I do a casual diff from what I assume was the initial v2 release, it looks OK.

So let's merge this after all. Sorry for the back and forth.

For the record: I haven't audited all involved components of this release process. Merging this should be seen as being part of https://github.com/sharkdp/bat/pull/2519 (which probably at least should have locked the commit in the first place.)

PR #3015, opened 2024-06-30
Bump themes/Nord-submlime from `0d655b2` `bf92a9e`

### LINKED ISSUE #2987: no license text for bundled Nord-sublime theme
The "Nord-sublime" theme bundled [here](https://github.com/sharkdp/bat/tree/v0.24.0/assets/themes) keeps the license header from the original project before the fork (`License:    MIT`), but it doesn't contain a copy of the MIT license text. However, the MIT license *requires* that redistributed sources contain a copy of the license text.

I would suggest to file an issue with the [upstream project](https://github.com/crabique/Nord-plist) about explicitly adding the license text as a file. If that doesn't happen, I don't think bat can redistribute this theme.

CosmicHorrorDev:
(Disclaimer: I am not dependabot)

Resolves #2987 

This updates the Nord theme to the latest commit which adds the MIT license from before the repo was forked. The repo owner did suggest pointing to a specific commit in the original repo from before the theme format changed instead of pointing to their repo, but I don't really see a difference at this point


CosmicHorrorDev:
It looks like other submodule bumps don't get a changelog entry, so I'm assuming it's fine if I don't add one

keith-hall (approved):

keith-hall (approved):

PR #3062, opened 2024-08-05
Minor README fixes

injust:
- Use relative links to fix unnecessary page navigation
- Add missing apostrophe in "`bat`s"

### DIFF
--- a/README.md
+++ b/README.md
@@ -506,7 +506,7 @@ highlighting. To select the `TwoDark` theme, call `bat` with the
 `--theme=TwoDark` option or set the `BAT_THEME` environment variable to
 `TwoDark`. Use `export BAT_THEME="TwoDark"` in your shell's startup file to
 make the change permanent. Alternatively, use `bat`s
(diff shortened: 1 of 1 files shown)


Enselic (changes requested):
we can keep the 's but let's also keep links absolute because the readme will show up in many places outside of GitHub context

injust:
> but let's also keep links absolute because the readme will show up in many places outside of GitHub context

The links should work, as long as those places render Markdown properly. For example, the relative links on https://crates.io/crates/bat work as expected.

The problem right now is that clicking the absolute links (e.g. on https://github.com/sharkdp/bat/blob/master/README.md or https://crates.io/crates/bat) will force a page navigation. So if the README appears on another website, the link will unnecessarily take you to GitHub.

(Actually, because of GitHub's `?tab=readme-ov-file` shenanigans, even links on https://github.com/sharkdp/bat will do a janky page navigation...)

Enselic (approved):
ok you seem sure this is better, so let's try

I noticed it is already split up in two commits so that reverthing the links is easy if needed, thanks

PR #3064, opened 2024-08-08
Update Chinese version translation of README.md

flytam:


Enselic (approved):
I have no idea what the new version says but I'm going to assume it's better.

PR #3067, opened 2024-08-09
Base2Tone-bat duotone colorscheme repo as submodule 

atelierbram:
I would like to see duotone color themes included in bat. These are only dark versions of [Base2Tone colorschemes](https://base2t.one). Because these colorschemes have such a limited color palette, but come in lots of variations, it will make the choice for people to find something that they like and/or matches the look of their terminal emulator much easier imo.


Enselic:
hi, did you know you can use this theme with bat without adding it upstream first? why does all users need this theme?

atelierbram:
> hi, did you know you can use this theme with bat without adding it upstream first? why does all users need this theme?

Hi @Enselic Yes I know, this is mentioned in [the repo](https://github.com/atelierbram/Base2Tone-bat) as well how to add those themes yourself, it's not difficult, but I just want to make it more convenient for users by making these themes easier discoverable. These are popular themes and one may not know these are available and may not think to specifically search for it on the interwebs.

Enselic:
How do you measure popularity of this theme? Both in absolute numbers and relative to other themes.

atelierbram:
> How do you measure popularity of this theme? Both in absolute numbers and relative to other themes.

Hard to answer; impossible to quantify. Maybe it’s not coincidental that a coding platform like CodePen — which may appeal more to designers than to developers — has Duotone themes (among a handful of others) featured for years now.

> Base2Tone colorschemes are created from two base hues, into (32) values with different saturation and lightness levels. Base2Tone is based on [DuoTone themes](http://simurai.com/projects/2016/01/01/duotone-themes) by [Simurai](http://simurai.com/) for Atom:

> … It tones down less important parts (like punctuation and brackets) and highlights only the important ones. This leads to a more calm color scheme, but still lets you find the stuff you’re looking for.

So these themes are most appreciated by those who prefer a less distracting environment while coding. If Bat, and this repository representing Bat, would like to accommodate those users, I would applaud that.

Enselic:
IMHO it would be better to put all themes and syntaxes in a central repository that bat can use. I think there is an issue opened for that somewhere.

That way, adding a new theme would not be up to bat maintainers, but up to the maintainers of such a central repository (that bat would use).

Exactly how to solve that technically is an open question.

Sorry to be boring but my recommendation is to close this as "won't fix", but if other bat maintainers want to maintain the inclusion of these themes, that's fine with me of course.

Perhaps I just don't understand how great the themes are? Maybe it would be easier to convince me (and others) by providing screenshots of how the themes look when used by bat.

keith-hall:
I believe https://github.com/trishume/syntect/issues/168#issuecomment-2561642416 is the issue tracking this

Enselic:
Closing to keep PR inbox clean but we can keep discussing if needed.

PR #3068, opened 2024-08-10
Support relative negative line ranges

ajesipow:
Adds support for relative negative line ranges, e.g. `-10:` or `:-10`.

This is implemented by buffering as many lines as the largest negative offset as a "look ahead" to determine if a line should be printed.

For example:
* for a range `-10:` 10 lines are buffered and only once EOF is reached will the (last) lines in the buffer be emitted
* for a range `:-10` 10 lines are buffered and lines are emitted as the buffer moves through the file, once EOF is reached the last lines in the buffer are discarded

Some benchmarks on a 21MB, 89648 lines log file:

```
❯ bat --version
bat 0.24.0
❯ wc -l test.log
   89648 test.log
❯ hyperfine --warmup 100 'bat ../test/test.log -r 79648:'
Benchmark 1: bat ../test/test.log -r 79648:
  Time (mean ± σ):      15.0 ms ±   0.5 ms    [User: 8.9 ms, System: 5.6 ms]
  Range (min … max):    14.3 ms …  17.0 ms    182 runs
❯ hyperfine --warmup 100 './target/release/bat ../test/test.log -r 79648:'
Benchmark 1: ./target/release/bat ../test/test.log -r 79648:
  Time (mean ± σ):      13.7 ms ±   0.5 ms    [User: 8.5 ms, System: 5.2 ms]
  Range (min … max):    13.0 ms …  16.0 ms    193 runs
❯ hyperfine --warmup 100 './target/release/bat ../test/test.log -r=-10000:'
Benchmark 1: ./target/release/bat ../test/test.log -r=-10000:
  Time (mean ± σ):      16.4 ms ±   0.8 ms    [User: 10.6 ms, System: 5.9 ms]
  Range (min … max):    15.2 ms …  19.8 ms    176 runs
```

### DIFF
--- a/src/line_range.rs
+++ b/src/line_range.rs
@@ -1,16 +1,26 @@
 use crate::error::*;
+use itertools::{Itertools, MinMaxResult};
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Copy, Clone)]
 pub struct LineRange {
-    lower: usize,
-    upper: usize,
+    lower: RangeBound,
+    upper: RangeBound,
+}
+
+/// Defines a boundary for a range
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub(crate) enum RangeBound {
+    // An absolute line number marking the boundary of a range
+    Absolute(usize),
+    // A relative (implicitly negative) offset from the end of the file as a boundary
+    OffsetFromEnd(usize),
 }
 
 impl Default for LineRange {
     fn default() -> LineRange {
         LineRange {
-            lower: usize::min_value(),
-            upper: usize::max_value(),
+            lower: RangeBound::Absolute(usize::MIN),
+            upper: RangeBound::Absolute(usize::MAX),
         }
     }
 }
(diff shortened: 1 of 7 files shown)


src/line_range.rs
```diff
@@ -1,25 +1,35 @@
 use crate::error::*;
+use itertools::{Itertools, MinMaxResult};
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Copy, Clone)]
```
ajesipow, line 4: We may not actually want to add this, since `LineRange` is `pub` and a future impl could make it not `Copy` and therefore require a breaking change.

keith-hall (approved):
Thanks for your hard work on this, and your patience. I think it looks good :) I will ask another maintainer to review before merging though :+1:

Enselic (commented):
I haven't looked closely enough to Approve, but I have no objections to merging this.

The most important part as far as I'm concerned is the presence of integration/regression tests, which is present here.

PR #3070, opened 2024-08-14
Add print_with_writer to PrettyPrint

kojix2:
Hi

Thank you for making such a great tool. `bat` is one of the commands I use most often in my terminal.

This pull request is not about using bat as a command-line tool, but about using it as a library with `PrettyPrinter`.

Could you please add a way to output to a custom writer, like a String or a file, using PrettyPrinter?
I suggest adding a `print_with_writer` function that lets users choose where the output goes.

This change would make `PrettyPrinter` more flexible and useful.
Thank you for considering my request.

---

I am working on libraries like [bat-c](https://github.com/kojix2/bat-c) and [wombat](https://github.com/kojix2/wombat) that use this function.


kojix2:
Updated CHANGELOG.md.

keith-hall (approved):

kojix2:
Fixed the conflict.

kojix2:
Thank you @keith-hall !!

PR #3071, opened 2024-08-14
Update documentation for git diff

HE7086:
Updated documentation for the `git diff` section.

Currently if (for whatever reason) a file in the repo has special characters in its name, including space, new line, etc., batdiff would malfunction. With filenames being handled in form of NULL-terminated strings, the issue is resolved. (delta does not have this problem)

I noticed that this section is not present in Japanese/Russian version of the doc, not sure how to deal with them :P

see also: eth-p/bat-extras#124

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ You can combine `bat` with `git diff` to view lines around code changes with pro
 highlighting:
 ```bash
 batdiff() {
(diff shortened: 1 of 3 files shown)


Enselic (approved):
Sounds reasonable

PR #3072, opened 2024-08-18
Add helper scripts for testing the shell completions

bash:
While working on #2896 I needed a way to test the shell completions.

So I put together a script for each shell (PowerShell is notably missing) that spawns the shell
with the right PATH / completions setup. It might be useful to other people too :)

### DIFF
--- a/scripts/test-completions.bash
+++ b/scripts/test-completions.bash
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+# Spawns a new bash shell that's set up with a PATH that points
+# the `bat` executable and completions from the `target/` dir.
+
+# Requires https://github.com/scop/bash-completion
+
+target_dir="$(dirname "$(realpath "$0")")/../target/debug"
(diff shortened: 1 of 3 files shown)


keith-hall (commented):
Hi, thanks for this. I'm keen to improve the developer experience for testing changes made to completions files, to help encourage us all to do a better job of keeping them up to date. (re: https://github.com/sharkdp/bat/issues/3170)
Now that we have https://github.com/sharkdp/bat/pull/3126, do you think any changes need to be made to these test helper scripts?
I had the idea that it might be nice to create a subfolder and put them all together there to reduce clutter, but maybe it's not necessary as they all have the same name just different extension :thinking:

epage:
btw https://docs.rs/completest-pty/latest/completest_pty/ can be used to write tests runnable in `cargo test` for completions.  Its what `clap_complete` is using (to a limited degree; its slow)

Enselic:
Closing as inactive since this did not receive a response:

> Now that we have https://github.com/sharkdp/bat/pull/3126, do you think any changes need to be made to these test helper scripts?

PR #3094, opened 2024-10-01
Update README.md -- Added info on escaping help aliasing in zsh

terr-steak:
Added note about escaping alias when using help aliasing in zsh


Enselic (approved):
nice tip

PR #3099, opened 2024-10-07
Added catppuccin themes for bat

aadirave:
Added catppuccin themes for bat as a default option in /assets/themes/bat. Done by adding the bat repo from the Catppuccin devs as a submodule. Resolves an issue opened on Sep 4.


keith-hall:
Hi, we are considering adding 	Catppuccin themes to `bat`. However, ideally not to `assets/themes/bat` but to `assets/themes/Catppuccin`, If you can change this and fix the conflicts (including moving the changelog entry to the "unreleased" section) then we would be more likely to merge this PR. Tip: when making changes to a fork, it is always best to create a new branch for it rather than doing it on the default branch.

keith-hall:
> I'd really like this theme in bat, can I either help move this PR forward or open a new PR with the updated changes?

I'd suggest to open a new PR 🙂 also, it might be worth checking if it is really the correct repo to reference as a submodule

keith-hall:
superceded by #3317

PR #3101, opened 2024-10-09
Significantly improve performance by using a buffered writer

MoSal:
Before

```
Benchmark 1: seq 10000000 | bat
  Time (mean ± σ):      6.235 s ±  0.052 s    [User: 3.664 s, System: 2.714 s]
  Range (min … max):    6.172 s …  6.355 s    10 runs
```

After

```
Benchmark 1: seq 10000000 | ./target/release/bat
  Time (mean ± σ):     215.9 ms ±   5.1 ms    [User: 275.4 ms, System: 38.8 ms]
  Range (min … max):   210.3 ms … 224.9 ms    10 runs
```

Using `less` for comparison

```
Benchmark 1: seq 10000000 | less
  Time (mean ± σ):     637.3 ms ±  43.3 ms    [User: 642.1 ms, System: 95.6 ms]
  Range (min … max):   584.5 ms … 700.1 ms    10 runs
```

And raw

```
Benchmark 1: seq 10000000
  Time (mean ± σ):      63.1 ms ±   1.3 ms    [User: 57.1 ms, System: 5.9 ms]
  Range (min … max):    62.1 ms …  66.0 ms    10 runs
```

### DIFF
--- a/src/controller.rs
+++ b/src/controller.rs
@@ -230,7 +230,13 @@ impl<'b> Controller<'b> {
                 }
             };
 
-            self.print_file_ranges(printer, writer, &mut input.reader, &line_ranges)?;
+            self.print_file_ranges(
+                printer,
+                writer,
+                &mut input.reader,
+                &line_ranges,
+                input.is_terminal,
+            )?;
         }
         printer.print_footer(writer, input)?;
 
@@ -243,6 +249,7 @@ impl<'b> Controller<'b> {
         writer: &mut OutputHandle<W>,
         reader: &mut InputReader,
         line_ranges: &LineRanges,
+        input_is_terminal: bool,
     ) -> Result<()> {
         let mut line_buffer = Vec::new();
         let mut line_number: usize = 1;
@@ -281,6 +288,11 @@ impl<'b> Controller<'b> {
 
             line_number += 1;
             line_buffer.clear();
+
+            // flush per line if input is terminal to allow interactive use
+            if input_is_terminal {
+                writer.flush()?;
+            }
         }
         writer.flush()?;
         Ok(())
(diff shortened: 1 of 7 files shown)


Enselic:
Those numbers look very nice 👍 

I skimmed over the diff which is quite big. Is it possible to split this up into smaller commits/PRs to make it easier to review/handle?

sharkdp:
Didn't have time to look into the code, but didn't we experiment with this a while ago and decided that we can't use a buffered writer, because we need to support interactive use cases where you run `bat`, enter a single line of input, see the output, enter the next line of input, see the output, …

src/controller.rs
```diff
@@ -88,9 +88,12 @@ impl<'b> Controller<'b> {
             clircle::Identifier::stdout()
         };
 
+        const BUF_W_SZ: usize = 1 << 14;
         let mut writer = match output_buffer {
             Some(buf) => OutputHandle::FmtWrite(buf),
-            None => OutputHandle::IoWrite(output_type.handle()?),
+            None => {
+                OutputHandle::IoWrite(BufWriter::with_capacity(BUF_W_SZ, output_type.handle()?))
```
Enselic, line 95: That might very well be the case. If so, it would be good to add a comment here explaining that. So that anyone that gets the same idea will find the comment when they are about to change the code.

MoSal:
@sharkdp 
Is it an acceptable solution to flush on each read line if stdin is detected as terminal?
I just pushed a commit that does just that.

eth-p:
> Is it an acceptable solution to flush on each read line if stdin is detected as terminal?

That wouldn't account for other programs piping to bat: `sed 's/A/a/' </dev/stdin | bat`

If it's possible to flush at the end of every input line regardless of the source, I would be happy with that compromise. We can always optimize it further in the future to add special exceptions for cases such as "the input source is a regular file."

MoSal:
Closing in favor of a new approach.

PR #3102, opened 2024-10-09
Add --no-syntax option

MoSal:
When using plain output, calls to `highlight_regions_for_line()`
 may cause a lot of overhead for little gain.

 So add a new option `--no-syntax` that skips syntax highlighting.
 It only has effect for plain output.

 This change is made in a way that preserves backwards compatibility and
 doesn't break any tests.

 Here is a test-case.

 Without `--no-syntax`

 ```
 Benchmark 1: seq 10000000 | ./target/release/bat -pf
  Time (mean ± σ):      7.899 s ±  0.195 s    [User: 7.962 s, System: 0.076 s]
  Range (min … max):    7.587 s …  8.095 s    10 runs
 ```

 With `--no-syntax`

 ```
 Benchmark 1: seq 10000000 | ./target/release/bat -pf --no-syntax
  Time (mean ± σ):     636.7 ms ±  42.9 ms    [User: 696.8 ms, System: 41.2 ms]
  Range (min … max):   603.5 ms … 715.5 ms    10 runs
 ```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@
 - Use bat's ANSI iterator during tab expansion, see #2998 (@eth-p)
 - Support 'statically linked binary' for aarch64 in 'Release' page, see #2992 (@tzq0301)
 - Update options in shell completions and the man page of `bat`, see #2995 (@akinomyoga)
+- Significantly improve performance by using a buffered writer, see #3101 (@MoSal)
 
 ## Syntaxes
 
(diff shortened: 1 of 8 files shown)


Enselic:
can we somehow be "smart" so users don't need a new option?

Enselic:
for my understanding: why not use `cat` for that use case?

MoSal:
This should be closed actually, together with the other PR. This one is just wrong. And I've come up with a new approach that works for #3101. I just want to explore alternative ways to do syntax highlighting first, because syntect is the only slow part remaining.

PR #3113, opened 2024-10-24
Bump clircle to 0.6.1

niklasmohrin:
This includes mostly dependency updates. The most significant one is that I removed clircle's dependency on winapi. After this PR, running `cargo update` on bat will remove winapi from the lockfile.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@
 - Update options in shell completions and the man page of `bat`, see #2995 (@akinomyoga)
 - Update nix dev-dependency to v0.29.0, see #3112 (@decathorpe)
 - Bump MSRV to [1.74](https://blog.rust-lang.org/2023/11/16/Rust-1.74.0.html), see #3154 (@keith-hall)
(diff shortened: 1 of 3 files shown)


keith-hall (approved):

keith-hall:
Audit failures have been fixed in https://github.com/sharkdp/bat/pull/3154

niklasmohrin:
Cool, I updated the PR :)

Should we `cargo update` here, or do you want to do it at a later point?

keith-hall:
> Cool, I updated the PR :)
> 
> Should we `cargo update` here, or do you want to do it at a later point?

We can do it separately :) ideally we will make a new release soon, so I guess it makes sense to update dependencies first

niklasmohrin:
@keith-hall It looks like #3161 bumped the version in the meantime, I suppose we can still remove the default features and note the update reason - or not, if you want :)

keith-hall (approved):

keith-hall:
Thanks 👍

PR #3115, opened 2024-10-30
Adding pipe delimeter support for csv files

pratik-m:
bat features delimiter auto detection for csv wherein it supports three delimiters - comma, semicolon and tab. Pipe delimiter is not supported by bat. This PR adds supports for pipe delimiter for csv 

`find . -maxdepth 1 -type f -printf '%p|%s|%m|%i|%Tc\n' | batcat -l csv`

image


keith-hall (approved):
not sure why the single quotes were changed to double quotes in the YAML, but otherwise looks good to me

PR #3128, opened 2024-11-20
Throw an error when try to read a block device

Integral-Tech:
- When reading a block device using `bat`, an error should be thrown.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@
 - Support 'statically linked binary' for aarch64 in 'Release' page, see #2992 (@tzq0301)
 - Update options in shell completions and the man page of `bat`, see #2995 (@akinomyoga)
 - Update nix dev-dependency to v0.29.0, see #3112 (@decathorpe)
+- Throw an error when try to read a block device, see #3128 (@Integral-Tech)
 
 ## Syntaxes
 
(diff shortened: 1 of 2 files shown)


Integral-Tech:
@sharkdp

keith-hall:
Does POSIX `cat` also error out when trying to read a block device? The aim is to maintain compatibility to be a drop in replacement. Also, an integration would be nice if we are to consider merging this :)

Integral-Tech:
> Does POSIX `cat` also error out when trying to read a block device? The aim is to maintain compatibility to be a drop in replacement. Also, an integration would be nice if we are to consider merging this :)

No, but use `bat` to read a block device will make it frozen

keith-hall:
I wonder if there is a better solution that doesn't involve disabling functionality like https://github.com/sharkdp/bat/pull/3101 or https://github.com/sharkdp/bat/pull/2902

Enselic:
> > Does POSIX `cat` also error out when trying to read a block device? The aim is to maintain compatibility to be a drop in replacement. Also, an integration would be nice if we are to consider merging this :)
> 
> No, but use `bat` to read a block device will make it stuck (have to force terminate with Ctrl+C)

If I understand you correctly  `bat` behaves like `cat` in this case?

Compatibility with `cat` is an important property that has served `bat` well by my estimation, so I would like to propose we close this PR as "won't fix". Sorry if that is disappointing.

PR #3142, opened 2024-12-05
Create sparql.sublime-syntax

RoiEXLab:
Hi,
I just had to create my own Variant of a `.sublime-syntax` for the SPARQL language and thought: Maybe I can submit it here so it helps other people in the future.
Disclaimer: I don't know shit about the sublime-syntax Format, I just used https://github.com/stardog-union/stardog-vsc/blob/0ce931c26529fdaf5c6dd58d91e589bbd6944a1d/stardog-rdf-grammars/syntaxes/sparql.tmLanguage.json (thus the comment because I'm not sure if the licensing applies in this case) merged a bunch of files together after doing some research how includes work, and then used a converter to get the sublime-syntax file out of it.
As I see it, it seems to work rather well even though I'm pretty sure it is not completely compliant with the official https://www.w3.org/TR/sparql11-query/#rQueryUnit grammar specification.

If there are any issues with this file, feel free to nudge me in the right direction.

### DIFF
--- a/assets/syntaxes/02_Extra/sparql.sublime-syntax
+++ b/assets/syntaxes/02_Extra/sparql.sublime-syntax
@@ -1,0 +1,140 @@
+%YAML 1.2
+---
+# Based on https://github.com/stardog-union/stardog-vsc/tree/0ce931c26529fdaf5c6dd58d91e589bbd6944a1d/stardog-rdf-grammars/syntaxes
+name: SPARQL
+file_extensions:
+  - rq
+  - sparql
+scope: source.sparql
+contexts:
+  main:
+    - include: rule-constraint
+    - include: iriref
+    - include: prefix
+    - include: prefixed-name
+    - include: comment
+    - include: special-predicate
+    - include: literals
+    - include: language-tag
+    - include: query-keyword-operators
+    - include: functions
+    - include: variables
+    - include: expression-operators
+  comment:
+    - match: "#.*$"
+      scope: comment.line.number-sign.turtle
+  boolean:
+    - match: \b(?i:true|false)\b
+      scope: constant.language.sparql
+  expression-operators:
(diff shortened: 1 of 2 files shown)


assets/syntaxes/02_Extra/sparql.sublime-syntax
```diff
@@ -0,0 +1,140 @@
...
+    - match: '[+-]?(?:\d+|[0-9]+\.[0-9]*|\.[0-9]+(?:[eE][+-]?\d+)?)'
+      scope: constant.numeric.turtle
+  iriref:
+    - match: '<[^\x20-\x20<>"{}|^`\\]*>'
+      scope: entity.name.type.iriref.turtle
+  language-tag:
+    - match: '@(\w+)'
+      scope: meta.string-literal-language-tag.turtle
+      captures:
+        "1": entity.name.class.turtle
```
keith-hall, line 45: Note: Sublime Text would fail to load this syntax definition, due to the quoting on the capture group numbers in various places.

keith-hall (commented):
Thanks for your PR. Generally, before shipping new syntax definitions by default, we [check if they meet certain criteria](https://github.com/sharkdp/bat/blob/master/doc/assets.md#Criteria-for-inclusion-of-new-syntaxes).
Also, we prefer where possible to have the syntax definitions be a submodule so that the burden of maintaining them doesn't fall on us `bat` maintainers. This also has the advantage that other users can more easily gain access to them and try them our with other text editors/highlighting tools.

RoiEXLab:
@keith-hall Thanks for your reply. I must have missed the passage on the inclusion criteria. If you search for "sparql" on Package Control, you get a package from 11 years ago with 3k installs, falling short of the 10k requirement. I don't have any ambitions here, so I will just close the PR. Maybe someone in the future will find this useful.

PR #3150, opened 2024-12-24
[ new ] Add support for Idris 2 programming language

buzden:
[Idris 2](https://www.idris-lang.org/) is a pretty big and well-known programming language among purely functional languages supporting dependent types. But it still lacks support by `bat`, and thus by other tools like `delta` and `typst`. I suggest to add support for it with this PR.


keith-hall:
Hi, thanks for this. It would be nice to get some indication of how popular Idris 2 is, to help us decide whether it meets our [criteria for inclusion by default](https://github.com/sharkdp/bat/blob/master/doc/assets.md#Criteria-for-inclusion-of-new-syntaxes).
Otherwise the next step could be to try get a curated list of packages going and add it there: https://github.com/trishume/syntect/issues/168

buzden:
> It would be nice to get some indication of how popular Idris 2 is

Well, maybe some [statistics from GitHub](https://github.com/search?q=lang%3AIdris%20&type=repositories) may help.

There are ~1600 repositories using this language solely on GitHub, around 10000 issues and PRs in these projects, GitHub has its syntax highlighting for its comments (it also has strong requirements on language popularity).

Also, say, Idris is more recognised than another language with dependent types, Lean, which is present in `bat` but even is not recognised by GitHub as a popular language.

> [criteria for inclusion by default](https://github.com/sharkdp/bat/blob/master/doc/assets.md#Criteria-for-inclusion-of-new-syntaxes).

Well, this criterion is valid for popular languages in Sublime text editor, as far as I understand. Sublime is not supported by Idris (Idris language has interactive compiler and closely works with the editor, supporting only vim, emacs and vscode). Thus, sublime highlighting was created specially to be supported by `bat`/`syntect` from scratch based on ideas from existing highlighting, e.g. for Haskell, with which Idris shares a lot of common syntax, but still differs significantly.

Additionally, I dared to suggest to include Idris while formally not meeting this criterion because I checked some already added languages and they also did not meet it formally, thus I thought that it is not valid anymore.

> Otherwise the next step could be to try get a curated list of packages going and add it there: [trishume/syntect#168](https://github.com/trishume/syntect/issues/168)

Am I right that you suggest adding my syntax definitions [here](https://github.com/getzola/zola/tree/master/sublime/syntaxes/extra) and [here](https://github.com/github-linguist/linguist/tree/main/vendor/grammars) (interestingly, the second one seems to use `tmLanguage` formal instead of `sublime-syntax`)? It looks feasible, at least the first onem since it uses the same format of config files.

CHANGELOG.md
```diff
@@ -94,6 +94,7 @@
 - Adds support for pipe delimiter for CSV #3115 (@pratik-m)
 - Add syntax mapping for `/etc/pacman.conf` #2961 (@cyqsimon)
 - Associate `uv.lock` with `TOML` syntax, see #3132 (@fepegar)
+- Add support for [Idris 2 programming language](https://www.idris-lang.org/) #3150 (@buzden)
```
keith-hall, line 97: Can I ask you to move this up to the relevant part of the "unreleased" section of the changelog please? :slightly_smiling_face:

keith-hall (approved):
> Idris (Idris language has interactive compiler and closely works with the editor, supporting only vim, emacs and vscode). Thus, sublime highlighting was created specially to be supported by `bat`/`syntect` from scratch based on ideas from existing highlighting

Ah, good to know, thanks for the context. 

> Additionally, I dared to suggest to include Idris while formally not meeting this criterion because I checked some already added languages and they also did not meet it formally, thus I thought that it is not valid anymore.

We have indeed unofficially relaxed the requirements a bit since lazy loading of syntaxes was implemented, to reduce the startup cost of having many languages. (https://github.com/sharkdp/bat/pull/2181)

Considering this, and that it was written especially for `bat` and friends, then let's merge it :tada:

CHANGELOG.md
buzden: Sure!

buzden:
Oh, seems I have unintentnially broken the merge process. I saw that auto-merge was not working because `master` branch has already moved forward, I rebased that this has disabled auto-merge.

PR #3154, opened 2024-12-26
[CI] fix cargo audit failing to run successfully

keith-hall:
- `cargo audit` failed to run in CI, presumably the `ubuntu-latest` image version changed and no longer has the ``cargo-audit`` crate installed by default
- audit failed due to unsoundness in `anstream`
- upgrading clap to fix that required bumping MSRV from 1.70 to [1.74, released over a year ago](https://blog.rust-lang.org/2023/11/16/Rust-1.74.0.html)

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,0 +1,2 @@
+[advisories]
+ignore = ["RUSTSEC-2024-0320"]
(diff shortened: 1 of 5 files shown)


keith-hall:
okay, looks like we will need `syntect` to move away from `yaml-rust` https://github.com/trishume/syntect/pull/555

Enselic (approved):
🎉

PR #3168, opened 2025-01-07
Fix name of BAT_THEME_{DARK,LIGHT} env vars

bash:
*this is a bit embarrassing since I'm the person who wrote the code in the first place 😆*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Features
 
 ## Bugfixes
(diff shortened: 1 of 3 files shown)


keith-hall:
Probably we should add a changelog entry, in case we want to push a bugfix release for it? 🙂

keith-hall:
Would it be worth trying to add an integration test to prove these env vars work as expected or not really possible?

bash:
Added a changelog entry and integration tests. TIL about `assert_cmd` :)

keith-hall (approved):

keith-hall:
Nice, thanks! 🙂

tony-sol:
Is it time for 0.25.1 though?

PR #3169, opened 2025-01-08
Bump git2 to 0.20

branchvincent:
Thanks for bat :)

Just bumping `git2` to allow/support latest `libgit2`: https://github.com/rust-lang/git2-rs/commit/68aa1ad420ab0f82277c39d21228eca54c833a2a

Also ran `cargo update` since tests still pass on my machine


branchvincent:
Needed to downgrade `home` to support MSRV, lmk if you'd like a changelog for this

keith-hall (approved):

PR #3174, opened 2025-01-10
Fix tiny accidental repetition in README.md

philipp-tailor:
Fixes copy and paste error, where same option was listed twice, instead of the 2 distinct options of `--theme` for different system appearances.


keith-hall (approved):
Thanks 👍

PR #3177, opened 2025-01-14
add zbat wrapper to decompress files automatically

anarcat:
This relies on bsdcat from archive-tools, which supports more formats than just gzip or, failing that, falls back to gunzip -c with a warning.

Previously[1], @sharkdp stated that bat would not support compressed files, but would be open in implementing a wrapper. There was another implementation proposed elsewhere[2], but it seems overengineered to me.

[1] https://github.com/sharkdp/bat/issues/642#issuecomment-526816644
[2] https://github.com/sharkdp/bat/issues/237#issuecomment-617079288

I don't know how to hook this up in the Rust build system, I'm sorry. It would also be nice to run shellcheck on this in CI, which I'm also unfamiliar with. It is, at least, currently shellcheck-clean. [SC3040](https://www.shellcheck.net/wiki/SC3040) is the pipefail stuff, and I don't go through the silly workaround they propose of doing a tempfile.


Enselic:
I'm sorry but maintenance is hard enough as it is, and I'm afraid having more wrappers scripts to maintain would not be ideal. Thank you for understanding.

PR #3179, opened 2025-01-18
Work around `cargo vendor` losing syntax_mapping builtins directory

dtolnay:
See https://github.com/rust-lang/cargo/issues/15080. This fixes:

```console
error: failed to run custom build command for `bat v0.25.0`

Caused by:
  process didn't exit successfully: `target/debug/build/bat-126fec914d2804b6/build-script-main` (exit status: 1)
  --- stdout
  cargo:rerun-if-changed=build/
  cargo:rerun-if-changed=src/syntax_mapping/builtins/

  --- stderr
  Error: IO error for operation on src/syntax_mapping/builtins/macos: No such file or directory (os error 2)

  Caused by:
      No such file or directory (os error 2)
```

### DIFF
--- a/build/syntax_mapping.rs
+++ b/build/syntax_mapping.rs
@@ -236,8 +236,14 @@ fn get_def_paths() -> anyhow::Result<Vec<PathBuf>> {
     ];
 
     let mut toml_paths = vec![];
-    for subdir in source_subdirs {
-        let wd = WalkDir::new(Path::new("src/syntax_mapping/builtins").join(subdir));
+    for subdir_name in source_subdirs {
+        let subdir = Path::new("src/syntax_mapping/builtins").join(subdir_name);
+        if !subdir.try_exists()? {
+            // Directory might not exist due to this `cargo vendor` bug:
+            // https://github.com/rust-lang/cargo/issues/15080
+            continue;
+        }
+        let wd = WalkDir::new(subdir);
         let paths = wd
             .into_iter()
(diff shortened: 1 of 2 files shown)


keith-hall (approved):

build/syntax_mapping.rs
```diff
@@ -236,8 +236,14 @@ fn get_def_paths() -> anyhow::Result<Vec<PathBuf>> {
...
 
     let mut toml_paths = vec![];
-    for subdir in source_subdirs {
-        let wd = WalkDir::new(Path::new("src/syntax_mapping/builtins").join(subdir));
+    for subdir_name in source_subdirs {
+        let subdir = Path::new("src/syntax_mapping/builtins").join(subdir_name);
+        if !subdir.try_exists()? {
+            // Directory might not exist due to this `cargo vendor` bug:
+            // https://github.com/rust-lang/cargo/issues/15080
+            continue;
```
keith-hall, line 244: Is there a way we can log a build warning here so that it is visible when the mappings are lost?

dtolnay: If I understand the Cargo bug correctly from the Cargo maintainers' responses, the platform-specific mapping directory is only lost when there are no platform-specific mappings (like macos). So the person building shouldn't need to care. I don't think a warning is warranted.

keith-hall: Makes sense, thanks 👍

PR #3201, opened 2025-02-03
Rename submodule paths to fix Dependabot submodule updates

### LINKED ISSUE #3198: Dependabot submodule updates are broken
(It's probably an upstream error. It should be easily mitigable on `bat`'s side but I lack context to say for sure.)

The Dependabot action to update submodules has been failing since [at least the last couple of months](https://github.com/sharkdp/bat/actions/workflows/dependabot/dependabot-updates?query=is%3Afailure).

The two runs we can access both fail on the same step:

```log
2025-02-01T03:34:09.3613620Z   proxy | 2025/02/01 03:34:09 [066] GET https://api.github.com:443/repos/sharkdp/bat/contents/assets/syntaxes/02_Extra/JavaScript%20(Babel)?ref=4fc55cfcec44dba3f67bd79cac9ccf64dc18eb08
2025-02-01T03:34:09.3614825Z   proxy | 2025/02/01 03:34:09 [066] * authenticating github api request with token for api.github.com
2025-02-01T03:34:09.4715821Z   proxy | 2025/02/01 03:34:09 [066] 404 https://api.github.com:443/repos/sharkdp/bat/contents/assets/syntaxes/02_Extra/JavaScript%20(Babel)?ref=4fc55cfcec44dba3f67bd79cac9ccf64dc18eb08
2025-02-01T03:34:09.4737486Z updater | 2025/02/01 03:34:09 ERROR <job_956925857> Error during file fetching; aborting: assets/syntaxes/02_Extra/JavaScript (Babel) not found
```

Previous runs are all deleted because of the GitHub Actions' eviction policy. But [there's a Dependabot message in a previously-created PR](https://github.com/sharkdp/bat/pull/3023#issuecomment-2489983822) that reports an error on fetching this same submodule.

I only noticed because I was expecting `cmd-help` to be auto-bumped and it was not.

I think I have tracked the bug to the GitHub API, [I have reported it here.](https://github.com/orgs/community/discussions/150556)

## Mitigation

The bug seems to affect submodule files/mappings that contain spaces in their name. We could change those paths to remove spaces and it should fix this workflow.

e.g.:

```diff
# .gitmodules
[submodule "assets/syntaxes/JavaScript (Babel)"]
- 	path = assets/syntaxes/02_Extra/JavaScript (Babel)
+ 	path = assets/syntaxes/02_Extra/JavaScript(Babel)
```

If the error persists, we may try removing parentheses too. But the only bug I've been able to prove is with spaces.

victor-gp:
Fixes #3198.

Before:

- [Failing Dependabot submodules job](https://github.com/sharkdp/bat/actions/runs/13084738523)
- Fails because it cannot locate the `Javascript (Babel)` submodule (responds 302 then 404): 
  <https://api.github.com/repos/sharkdp/bat/contents/assets/syntaxes/02_Extra/JavaScript%20(Babel)?ref=4fc55cfcec44dba3f67bd79cac9ccf64dc18eb08>
 - Still failing when the ref is master:
   <https://api.github.com/repos/sharkdp/bat/contents/assets/syntaxes/02_Extra/JavaScript%20(Babel)?ref=master>

After renaming to `Javascript_(Babel)`:

- Request for the submodule succeeds (responds 302 then 200):
  <https://api.github.com/repos/victor-gp/bat/contents/assets/syntaxes/02_Extra/JavaScript_(Babel)/?ref\=fix-submodules-for-dependabot>

After merging to master:

- The request for the submodule should succeed (302 => 200):
  <https://api.github.com/repos/sharkdp/bat/contents/assets/syntaxes/02_Extra/JavaScript_(Babel)/?ref\=master>
- Succesful Dependabot submodules job? :crossed_fingers:

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -65,7 +65,7 @@
 	path = assets/themes/onehalf
 	url = https://github.com/sonph/onehalf
 [submodule "assets/syntaxes/JavaScript (Babel)"]
-	path = assets/syntaxes/02_Extra/JavaScript (Babel)
+	path = assets/syntaxes/02_Extra/JavaScript_(Babel)
 	url = https://github.com/babel/babel-sublime
 [submodule "assets/syntaxes/FSharp"]
 	path = assets/syntaxes/02_Extra/FSharp
@@ -89,7 +89,7 @@
 	path = assets/themes/sublime-snazzy
 	url = https://github.com/greggb/sublime-snazzy
 [submodule "assets/syntaxes/Assembly (ARM)"]
-	path = assets/syntaxes/02_Extra/Assembly (ARM)
+	path = assets/syntaxes/02_Extra/Assembly_(ARM)
 	url = https://github.com/tvi/Sublime-ARM-Assembly
 [submodule "assets/syntaxes/protobuf-syntax-highlighting"]
 	path = assets/syntaxes/02_Extra/Protobuf
@@ -108,7 +108,7 @@
 	path = assets/syntaxes/02_Extra/Fish
 	url = https://github.com/Phidica/sublime-fish.git
 [submodule "assets/syntaxes/Org mode"]
-	path = assets/syntaxes/02_Extra/Org mode
+	path = assets/syntaxes/02_Extra/Org_mode
 	url = https://github.com/jezcope/Org.tmbundle.git
 [submodule "assets/syntaxes/DotENV"]
 	path = assets/syntaxes/02_Extra/DotENV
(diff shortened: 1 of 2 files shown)


victor-gp:
I've noticed that these submodules have a `.sublime-syntax` counterpart, a file owned by `bat` in the same `syntaxes/02_Extras/` directory as the submodule. [^1]

I've also noticed that the repositories the submodules point to don't have any `sublime-syntax` file, they have `.tmLanguage` files and similar. So I gather that the owned file is generated somehow from the submodule's own?

**In such case, should we also rename those files?**

[1] All 4 changed submodules, e.g.:
- https://github.com/sharkdp/bat/tree/master/assets/syntaxes/02_Extra#:~:text=JavaScript%20(Babel)%20%40%20f4579f9
- https://github.com/sharkdp/bat/blob/master/assets/syntaxes/02_Extra/JavaScript%20(Babel).sublime-syntax

keith-hall:
> I've also noticed that the repositories the submodules point to don't have any `sublime-syntax` file, they have `.tmLanguage` files and similar. So I gather that the owned file is generated somehow from the submodule's own?

Yes, we have this documented here: https://github.com/sharkdp/bat/blob/master/doc/assets.md#adding-new-builtin-languages-for-syntax-highlighting

> In such case, should we also rename those files?

I would say no, because unless the file contains a name key inside it's yaml contents, the base file name is the name given to the syntax definition and used when mapping syntaxes, setting language from the command line arg etc. So we don't want to change that 🙂

PR #3213, opened 2025-02-23
Map various .NET file extensions to XML syntax

### LINKED ISSUE #3211: Use XML syntax for MSBuild files
MSBuild project files are a standard in the .NET ecosystem.

File extensions for these include `.props`, `.targets`, `.csproj`, `.vbproj`, and I think `.fsproj` too but would need to check.

Can the existing XML formatting be automatically used for these, rather than requiring users to pass `--language xml`?

keith-hall:
Fixes #3211


ltrzesniewski:
Hi @keith-hall, I'd very much like this to be released if possible. 😇

I've seen the discussion in #2200, so I thought it wouldn't hurt to ask. (I've just spent more time than I'd like to admit trying to figure out why `cargo build` works but `cargo install` doesn't include this feature 😅)

keith-hall:
The plan is to merge https://github.com/sharkdp/bat/pull/3402 and then make a release. I can try to make time for it over the weekend.

ltrzesniewski:
Thanks! Please don't hurry and have a nice week-end. This is just for personal convenience.

keith-hall:
Just a quick update, I didn't get round to this last weekend. I have created a milestone for things I want to get included in the next release and plan to work on them when I get time: https://github.com/sharkdp/bat/milestone/11
I'm hoping they won't take long, if something does turn out to, we can always decide to leave it out of the next release.

ltrzesniewski:
Thanks for the update! 🙂  
Like I said: don't hurry, I just wanted to know when the next release was *roughly* planned to be.

PR #3216, opened 2025-02-26
upgrade to rust 1.85

awol2005ex:
upgrade to rust 1.85


blinxen:
Changing the MSRV to 1.85, without a good reason, will make `bat` fail to build on distros that don't have the latest rust version. Example: `epel9`, `epel10` and probably many more.

Enselic:
Why not use rustup to get easy access to the latest Rust version?

blinxen:
For private or development use, that is indeed a good solution. However for stable Linux distributions (e.g. debian, RHEL, Ubuntu) the Rust toolchain is updated in longer cycles. In those cases, if the MSRV is set to a higher version then `bat` can't be built for them. As the package maintainer of `bat` for Fedora and EPEL, I would like to keep bat up-to-date there.

Enselic:
Why is not bat also updated in longer cycles in that case? Users of such distros by definition value stability over bleeding edge, don't they?

The reason I am asking these questions is because there is a cost (both in time and in "fun factor") at keeping MSRV low. So I hope we can find a pragmatic approach here.

blinxen:
> Why is not bat also updated in longer cycles in that case? Users of such distros by definition value stability over bleeding edge, don't they?

I actually do that but there are some corner cases that could break the build. Sometimes a new distribution release gets branched before a Rust update and you get stuck with an old Rust version until the next big release. This makes it not possible to update `bat` for quite some time. While there are enough workarounds, I would like to always stick to upstream and introduce as little changes when packaging as possible.

> The reason I am asking these questions is because there is a cost (both in time and in "fun factor") of keeping MSRV low. So I hope we can find a pragmatic approach.

I am 100% with you on this. You should not forcibly keep the MSRV low, just don't update it if you don't have to. Sometimes a dependency forces you to raise the MSRV because that dependency requires a new language feature. Maybe you guys are the ones that want to use new language features. These are all valid reasons. I just wanted to shed some light on why raising the MSRV without a valid reason could be bad.

Enselic:
well for https://github.com/sharkdp/bat/pull/3194 for example

can you make CI pass please?

src/assets/serialized_syntax_set.rs
```diff
@@ -19,7 +19,7 @@ impl SerializedSyntaxSet {
     pub fn deserialize(&self) -> Result<SyntaxSet> {
         match self {
             SerializedSyntaxSet::FromBinary(data) => Ok(from_binary(data, COMPRESS_SYNTAXES)),
-            SerializedSyntaxSet::FromFile(ref path) => {
+            SerializedSyntaxSet::FromFile( path) => {
```
Enselic, line 22: space before ` path`

Enselic (changes requested):

src/assets/serialized_syntax_set.rs
Enselic: which CI also complains about

Enselic:
Closing due to inactivity to keep the PR inbox clean, but of course feel free to re-open if you resume working on this.

PR #3235, opened 2025-03-21
Replace libgit2 with gitoxide

blinxen:
As discussed in https://github.com/sharkdp/bat/discussions/3151, I have prepared a PR to replace the `libgit2` usage in `bat` with `gitoxide`. The PR is not finished *yet* because some tests are failing since the produced `git diff` does not match the [exact behavior of `libgit2`](https://github.com/pascalkuthe/imara-diff/issues/21).

I see here two solutions:

* Extend the `imara-diff` crate to generate `git` like diffs
* Adapt the tests to accept the *new* diff format (it is not a format but more like a different diffing algorithm)

What do the maintainers think?

CC: @byron

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,10 +6,6 @@ updates:
     interval: monthly
     time: "04:00"
     timezone: Europe/Berlin
-  ignore:
-  - dependency-name: git2
-    versions:
-    - 0.13.17
 - package-ecosystem: gitsubmodule
   directory: "/"
   schedule:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 
 - Work around build failures when building `bat` from vendored sources #3179 (@dtolnay)
 - CICD: Stop building for x86_64-pc-windows-gnu which fails #3261 (Enselic)
+- Replace `libgit2` with a pure Rust implementation of git called `gitoxide`, see PR #3235 (@blinxen)
 
 ## Syntaxes
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
 
 [[package]]
+name = "ahash"
+version = "0.8.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
+dependencies = [
+ "cfg-if",
+ "once_cell",
+ "version_check",
+ "zerocopy",
+]
+
+[[package]]
 name = "aho-corasick"
 version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -16,6 +28,12 @@ checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
 dependencies = [
  "memchr",
 ]
+
+[[package]]
+name = "allocator-api2"
+version = "0.2.21"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "ansi_colours"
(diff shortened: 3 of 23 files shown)


Byron:
You may also find this document about [diff-sliders](https://github.com/mhagger/diff-slider-tools/blob/master/README.md) interesting. It also details the tuning that Git has received to produce more 'likable' diffs, some of which `libgit2` seems to inherit. It seems like an art as well.

Right now, `imara-diff` has received no time in the direction of optimizing diffs for humans, and I think it would have to be a community effort if that should happen.

As this comes up all the time, I also feel that pressure is mounting, and more so the more people use `imara-diff` :).

blinxen:
I think that trying to improve `imara-diff` would be more interessting. I checked the `git` mailing list and found the initial patch that introduced what the [diff-sliders document](https://github.com/mhagger/diff-slider-tools/blob/master/README.md) proposes. I will check it out and try to implement something similiar for `imara-diff`. At the mean time we can keep this PR open since it does not have any urgency.

blinxen:
Since I have not been able to work much on the `imara-diff` improvements lately, I updated this PR so that it at least passes all tests.

blinxen:
https://github.com/GitoxideLabs/gitoxide/pull/2011 just got merged and I tested the optimized diff locally and it works :D. I will updated this PR once a new version of `gix` is released. Then this PR will be open for review.

Byron:
I am sorry for my negligence and wishful thinking when merging the implementation despite it having only one test, but it does indeed not appear to work correctly.

Having put it into GitButler to get some real-world experience with it, here we have an example where both Git and GitButler (with the standard hunk visualization) agree:

expected
expected-ui

However, when using the new hunk processing to get a Git-like diff, this happens:

wrong-with-git-diff-sink

Of course it could be that the `ChangeKind` needs to be interpreted in some way, but even if so, I think it needs tests to make clear how this is working, the more the merrier.

For now I have removed the implementation from the codebase, but of course would be happy to see a follow-up PR that brings it back, along with tests, ideally even baseline tests where various expectations are produced by Git itself.

blinxen:
That is not so good :grimacing:

I took a quick peek and something with the scoring is weird with this diff. For some reason the implementation prefers moving up to line 132 over staying at line 141. I will look into it and also try to come up with a good test suite.

blinxen:
Already found the issue --> https://github.com/blinxen/gitoxide/commit/d7ab68e659df3cd99bc69fd89932575a28f4691c

>  Maybe one can bring in some tests from [the diff-slider repo](https://github.com/mhagger/diff-slider-tools/blob/master/README.md) to get more coverage.

That will be the next step

keith-hall:
Any new updates on this one? :)

blinxen:
I did not have enough time lately to continue to work on the diff improvements. However, the `imara-diff` maintainer implemented the same algorithm and [merged it into Helix](https://github.com/helix-editor/helix/pull/13722). That might be a better direction to take. Until is solved https://github.com/pascalkuthe/imara-diff/issues/27, we can't use it though.

> Thanks for working on this! Honestly, it feels like the diffs are still valid, so adapting the tests is acceptable in my opinion.

If this statement is still valid then I can solve the merge conflicts and we can just merge it. Improvements can be added in a later PR.

keith-hall:
I do worry that many users will complain that the diffs don't match what `git` shows... Perhaps having a feature flag to decide which implementation to use at compile time and keeping libgit2 in the codebase for now could make sense?
What do you think? Too much work?

blinxen:
>  What do you think? Too much work?

I don't see a benefit in this. I initially started this PR to reduce maintenance work on `bat` in Fedora. So if we are going to have the `gitoxide` feature then I would like to use it in Fedora. That however will make people create issues here that on Fedora they see a different diff output. Therefore I suggest to close this PR and maybe come back again later when the whole situation is better.

Byron:
This makes me sad, but I also have no solution for the diff-slider issue and the Git related modifications.
It's notable though that the version of this that `helix` is using was basically conjured up over night, with no tests that I know of.
And I don't think `gitoxide` will ever test that in production, so it's indeed better to hold back and see how the situation resolves itself.

Byron:
@blinxen Gitoxide now provides Git slider postprocessing for diffs, which seems to be ~95% accurate against baseline Git. You could probably experiment with it for a chance to bring this back.

Byron:
`gix` 0.82 comes with this slider post-processing. It might be worth giving this another try with particular attention to those slider-situations where post-processing should yield better outcomes.

PR #3238, opened 2025-03-24
Always disable paging when listing themes

Sky9x:
I have `--paging=always` in my `.config/bat/config` which makes `bat --list-themes` very not helpful (without doing `bat --list-themes --no-pager`). This force disables paging when listing themes.


einfachIrgendwer0815:
Although this solves the problem of `bat --list-themes` producing unhelpful output when paging, I think the better approach would be to have bat produce the normal `--list-themes` output and pass it to the pager as a whole.

Sky9x:
ooh youre right i didnt think of that

PR #3239, opened 2025-03-24
Add paging to `--list-themes`

einfachIrgendwer0815:
This allows to use `--list-themes` with paging by passing all theme sample outputs to the pager as a whole.

Alternative to #3238
(Closes #3238)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Features
 
+- Add paging to `--list-themes`, see PR #3239 (@einfachIrgendwer0815)
+
 ## Bugfixes
 
 - Fix `BAT_THEME_DARK` and `BAT_THEME_LIGHT` being ignored, see issue #3171 and PR #3168 (@bash)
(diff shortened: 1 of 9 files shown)


keith-hall (approved):
Thanks, looks good. Is it possible to write an integration test for this, I wonder? :thinking:

einfachIrgendwer0815:
I do not think integration testing is possible here. A real pager cannot be used when testing because that would cause the test to hang and there would be no way to access the piped data. When using a "dummy" pager (e. g `cat`)  that simply passes through what is piped into its stdin, both bat and that pager would share the same stdout, so stdout would receive the same content as before. There would be no difference observable for testing.

einfachIrgendwer0815:
(Sorry for closing and reopening, I intended to make this a draft but that does not seem possible on the mobile app)

Looking at this once more, I think it is kinda hacky to write to an intermediate buffer and to just tell the `Controller` not to call the pager itself. I would like to look for a better solution first if you do not mind.

keith-hall (approved):

Enselic:
I wouldn't call myself an expert on this area of the code, but when I skim the code over I can't see anything that looks scary.

I haven't looked close enough to mark as Approved, but if both of you think this code is the right approach, I don't mind you merging it. If something breaks it can always be fixed afterwards.

Issue #3230, opened 2025-03-09
multibytes characters in file path are not taken into account when wrapping the header, which leads to bat crashing


vejkse:
**What steps will reproduce the bug?**

1. Run `touch /tmp/A—B가`, where ‘—’ is a multibyte UTF-8 character and ’가’ is a multibyte double-width CJK character.
2. Run `bat --terminal-width=30 /tmp/A—B가`.

**What happens?**

`bat` crashes:
```
thread 'main' panicked at /usr/src/debug/rust/rustc-1.84.0-src/library/core/src/str/mod.rs:648:21:
byte index 21 is not a char boundary; it is inside '가' (bytes 20..23) of `File: /tmp/A—B가   <EMPTY>`
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
───────┬──────────────────────
```

Different widths give different results:
- with 26 or 27, it crashes with the boundary inside ‘—’;
- with 21 or 22, it doesn’t crash but ANSI colour codes are broken: `146;131;116m│ m   <EMPTY>`.

**What did you expect to happen instead?**

Depending on the width, either not crashing or not breaking ANSI colour codes.

**How did you install `bat`?**

I’m using the latest ArchLinux package.

---

**bat version and environment**

#### Software version

bat 0.25.0 (v0.25.0-modified)

#### Operating system

Linux 6.13.5-arch1-1

#### Command-line

```bash
bat --diagnostic
```

#### Environment variables

```bash
BAT_CACHE_PATH=<not set>
BAT_CONFIG_PATH=<not set>
BAT_OPTS=<not set>
BAT_PAGER=<not set>
BAT_PAGING=<not set>
BAT_STYLE=<not set>
BAT_TABS=<not set>
BAT_THEME=<not set>
COLORTERM=truecolor
LANG=en_US.UTF-8
LC_ALL=<not set>
LESS=' --RAW-CONTROL-CHARS --tilde --incsearch --ignore-case --clear-screen '
MANPAGER='nvim +Man'\!''
NO_COLOR=<not set>
PAGER=less
SHELL=/bin/bash
TERM=alacritty
XDG_CACHE_HOME=/home/vejkse/.cache
XDG_CONFIG_HOME=/home/vejkse/.config
```

#### System Config file

Could not read contents of '/etc/bat/config': No such file or directory (os error 2).

#### Config file

```
--theme=gruvbox-dark
```

#### Custom assets metadata

[...]

PR #3245, opened 2025-04-01
Fix for multibyte characters in file path

HSM95:
Partial fix for #3230.

Bat will no longer crash when multibyte characters are in the file path. However, there may still be misformatting of the header in the output.

### DIFF
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -17,6 +17,7 @@ use content_inspector::ContentType;
 
 use encoding_rs::{UTF_16BE, UTF_16LE};
 
+use unicode_segmentation::UnicodeSegmentation;
 use unicode_width::UnicodeWidthChar;
 
 use crate::assets::{HighlightingAssets, SyntaxReferenceInSet};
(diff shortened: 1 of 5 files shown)


src/printer.rs
```diff
@@ -403,14 +404,14 @@ impl<'a> InteractivePrinter<'a> {
         handle: &mut OutputHandle,
         content: &str,
     ) -> Result<()> {
-        let mut content = content;
         let content_width = self.config.term_width - self.get_header_component_indent_length();
-        while content.len() > content_width {
-            let (content_line, remaining) = content.split_at(content_width);
-            self.print_header_component_with_indent(handle, content_line)?;
-            content = remaining;
+        let mut content_graphemes: Vec<&str> = content.graphemes(true).collect();
```
keith-hall, line 408: thanks for taking a look at fixing this. I wonder if it makes any sense to have an optimization that it only splits into graphemes if the width would overflow? Because for short filenames, I think it can just continue to print it wholly without splitting and then rejoining into a string

HSM95: @keith-hall Yes, I added a quick check using `chars`. It's not perfect because `chars().count()` may exceed the number of graphemes, but it's the best we can do without splitting into graphemes :)

HSM95: @keith-hall I wanted to check in on this to see if there's anything else you think it needs?

keith-hall: No, let's merge it, thanks 🙂

keith-hall (approved):

src/printer.rs
HSM95: 👍 Sounds good, thank you

PR #3236, opened 2025-03-23
Add GDScript Syntax Support (#3233)

chetanjangir0:
This PR adds a new submodule for GDScript syntax definitions (located at assets/syntaxes/02_Extra/GDScript-sublime) and includes a comprehensive test file in tests/syntax-tests/source/GDScript/. The changelog has been updated accordingly. (#3233, @chetanjangir0)

PR #3255, opened 2025-04-01
Make highlight tests fail when new syntaxes don't have fixtures

dan-hipschman:
I stumbled across https://github.com/sharkdp/bat/pull/3236#issuecomment-2746349454 and wanted to fix it.

For more context, a new language was added but the author forgot to run `update.sh` to add a test fixture for it. This should have caused CI to fail, but it passed. The reason is that the highlighting regression test scripts iterate over existing fixtures, and hence don't catch newly added languages that haven't added a fixture.


keith-hall (approved):
Looks good, thanks for taking the initiative on this 🙂

PR #3260, opened 2025-04-07
Cargo.toml: Document that MSRV can be bumped if there is a reason for it

Enselic:
To the best of my knowledge our current MSRV rule is that we can bump MSRV whenever there is a reason for it. Let's document it.

Downstream packagers seem to agree: https://github.com/sharkdp/bat/pull/3216#issuecomment-2781035085


Enselic:
(Changelog entry not needed.)

PR #3261, opened 2025-04-07
CICD: Stop building for x86_64-pc-windows-gnu which fails

Enselic:
Let's not build for platforms that are broken. If someone fixes the build we can of course add it back later.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 ## Other
 
 - Work around build failures when building `bat` from vendored sources #3179 (@dtolnay)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 ## Other
 
 - Work around build failures when building `bat` from vendored sources #3179 (@dtolnay)
+- CICD: Stop building for x86_64-pc-windows-gnu which fails #3261 (Enselic)
```
Enselic, line 17: (Not having @ in front of my name is intentional, because I want to avoid mentions by tools.)

keith-hall (approved):

PR #3263, opened 2025-04-10
Fix copy-paste mistake in bat.1.in

gthb:
Fix apparent copy-paste mistake: the `--theme-light` documentation says “dark” in two places where it should say “light” (presumably copy-pasted from the `--theme-dark` documentation)


gthb:
Ignoring the “Changelog” CI failure, like `CONTRIBUTING.md` says I should, because only changing documentation.

keith-hall (approved):

keith-hall:
Thanks!

Issue #3234, opened 2025-03-21
bat failing to build on Fedora, due to onig_sys build failure.


akulbe:
**What steps will reproduce the bug?**

1. cargo install bat

**What happens?**

see https://gist.github.com/akulbe/eebd60129501fc00778eb44af8404b52

**What did you expect to happen instead?**

for `bat` to build successfully

**How did you install `bat`?**

cargo install bat

---

**bat version and environment**

bat 0.25.0 isn't building

akulbe:
I just tried that, and the results were the same. Would creating another
gist (with the `--locked` option)  be helpful?

keith-hall:
Probably not necessary, thanks. It looks like the same issue as https://github.com/rust-onig/rust-onig/issues/191 to me

perobertson:
> Probably not necessary, thanks. It looks like the same issue as [rust-onig/rust-onig#191](https://github.com/rust-onig/rust-onig/issues/191) to me

It looks like that issue has been fixed but a new release needs to be created for it and that is being asked for in https://github.com/rust-onig/rust-onig/issues/193

Aymen69-LYN:
Le problème vient de la dépendance onig_sys utilisée par bat
Plus précisément, cela casse avec Fedora 42 + GCC 15
La nouvelle version onig_sys 
Suivis sur le ticket rust-onig/rust-onig#193
Options : 
sudo dnf install compat-gcc-12
export CC=gcc-12
cargo install bat
Options 2:
CC=clang cargo install bat

PR #3274, opened 2025-04-29
fix: switch to fancy regex

perobertson:
Related: https://github.com/sharkdp/bat/issues/3234

This switches to using fancy-regex from oniguruma so that the app can compile on more systems.

It also appears that https://github.com/kkos/oniguruma has been archived so it may be some time before the compile errors are fixed with that project or a fork of it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Make highlight tests fail when new syntaxes don't have fixtures PR #3255 (@dan-hipschman)
 - Fix crash for multibyte characters in file path, see issue #3230 and PR #3245 (@HSM95)
 - Add missing mappings for various bash/zsh files, see PR #3262 (@AdamGaskins)
+- Fix compile errors with `oniguruma` by switching to `fancy-regex`, see #3234 (@perobertson)
 
 ## Other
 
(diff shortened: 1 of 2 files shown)


keith-hall:
I would say that fancy-regex isn't quite ready yet, but I'm working on it... I guess the archival of Oniguruma means I need to get it done faster...

perobertson:
I haven't done any comparisons yet, but I wonder if something like [regex](https://crates.io/crates/regex) could be used instead of [oniguruma](https://github.com/kkos/oniguruma) while fancy-regex gets fixed up.

keith-hall:
fancy-regex delegates to the regex(-automata) crate, adding features which require backtracking. Without those, only some syntax definitions which are crafted specifically to avoid backtracking would work, and even then in some cases backreferences are unavoidable...
fancy-regex is currently unoptimized and lacks support for subroutine calls etc, which is why I say it is not ready to replace Oniguruma yet

Enselic:
Thank you for the PR! But it Probably makes sense to close this pr until fancy regex supports needed features?

to keep or inbox trimmed

perobertson:
Good call, and thanks for the reminder about this one @Enselic.

PR #3276, opened 2025-04-30
remove LiveScript submodule

keith-hall:
the repo has gone away, causing all our CI etc. to break

we can keep the highlighting though because we have a .sublime-syntax file


sharkdp (approved):
Thank you

PR #3283, opened 2025-05-01
tests/scripts/license-checks.sh: Add MPL exception

Enselic:
Since the MPL license allows us to include syntaxes and themes without relicensing the rest of bat, it seems fine to allow MPL-licensed assets. Do that. (And if someone disagrees we can reverse this decision later.)

This unblocks https://github.com/sharkdp/bat/pull/2848.

Note: No CHANGELOG entry is needed for this particular PR.

Note: This PR might make sense to fold into the above PR. We'll see.

Note: This is untested


Enselic:
I'll close this for now but it is still fine to begin using MPL syntaxes and themes.

PR #3290, opened 2025-05-11
docs(security): add initial security policy

janderssonse:
This PR adds a SECURITY.md file, battle tested in other projects and orgs, (the construct is CCO ie public domain, for example from here https://raw.githubusercontent.com/itiquette/git-provider-sync/refs/heads/main/SECURITY.md so just reuse)

A SECURITY.md would help anyone assessing the project for use, give a hint of how it handles critical no public security issues, and give anyone a clear instruction on how to report them non public.

**IE, for someone thinking about using bat in an organisation or privately it would give an extra trust factor.**

This policy basically  says "send your findings, and we will see if we handle them, we will notify you".

Besides, being a good FOSS practice, makes the project look more professional and it is heavily supported by GitHub https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/creating-a-default-community-health-file etc as one of the community health files, so it will pop up automatically in the ui for the end user. 

Examples:
Security Tab in project front will be added automatically
Skärmbild från 2025-05-11 05-57-27

Security Policy in the top right corner of UI will be added automatically

Skärmbild från 2025-05-11 05-58-05

Security Policy under Security Overview for the project will have the Security Policy green and enabled.
Skärmbild från 2025-05-11 05-58-23

NOTE: there is a <...> in the text, where the preferred channel for reporting should be added I left that for you, (or tell me what to add there, and I'll rebase with that).

NOTE: I had this in multiple orgs and projects over the years. Only once I had a report, so I dont think one should be worry about getting to much reports from this, this is at least my experience.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,33 @@
+# Security Reporting
+
+If you wish to report a security vulnerability privately, we appreciate your diligence. Please follow the guidelines below to submit your report.
+
+## Reporting
+
+To report a security vulnerability, please provide the following information:
+
+1. **PROJECT**
+   - Include the URL of the project repository - Example: <https://github.com/sharkdp/bat>
+
+2. **PUBLIC**
+   - Indicate whether this vulnerability has already been publicly discussed or disclosed.
+   - If so, provide relevant links.
+
+3. **DESCRIPTION**
+   - Provide a detailed description of the security vulnerability.
+   - Include as much information as possible to help us understand and address the issue.
+
+Send this information, along with any additional relevant details, to <email AT somewhere or other channel>.
+
+## Confidentiality
+
+We kindly ask you to keep the report confidential until a public announcement is made.
+
+## Notes
+
+- Vulnerabilities will be handled on a best-effort basis.
+- You may request an advance copy of the patched release, but we cannot guarantee early access before the public release.
+- You will be notified via email simultaneously with the public announcement.
+- We will respond within a few weeks to confirm whether your report has been accepted or rejected.
+
+Thank you for helping to improve the security of our project!


janderssonse:
It fails without a changelog entry, but I think this applies (I see this as docs). >[!NOTE]
> For PRs, a CI workflow verifies that a suitable changelog entry is
> added. If such an entry is missing, the workflow will fail. If your
> changes do not need an entry to the changelog (see above), that
> workflow failure can be disregarded.

SECURITY.md
```diff
@@ -0,0 +1,33 @@
...
+## Confidentiality
+
+We kindly ask you to keep the report confidential until a public announcement is made.
+
+## Notes
+
+- Vulnerabilities will be handled on a best-effort basis.
+- You may request an advance copy of the patched release, but we cannot guarantee early access before the public release.
+- You will be notified via email simultaneously with the public announcement.
+- We will respond within a few weeks to confirm whether your report has been accepted or rejected.
```
Enselic: I'd like to point out that bat comes with no warranty, and I don't want to make any promises about timeliness

I don't really think there is a problem with our current place and instruction found at https://github.com/sharkdp/bat?tab=readme-ov-file#security-vulnerabilities

Enselic (changes requested):

Enselic:
Closing for now to keep PR inbox clean but we can keep discussing here if needed.

SECURITY.md
janderssonse: @Enselic Then we should just remove the last note /respond paragraph/ maybe. I don't see there are any obligations in this, on purpose. Best-effort means that doing nothing is an option too (no warranty).  There is nothing wrong the current instruction you link too, but this community file is an idiomatic way (like I show with examples in the PR-text - ie both acknowledged by platforms like GitHub, and an end user would expect to look for this information, in a file with this name. )(SECURITY.md).

Enselic: As a first step we could put our current instruction in the separate file and get that merged

janderssonse: I can update this PR suggestion with a PR which is a mix of the both and be careful to remove anything which even remotely implies any promises. (The We will respond... paragraph, or soften it even more... In most cases, we will .. friendly, but no promises?). Sound good?

Enselic: Lets first do it minimally

and to be honest I don't see the point of adding more info

if someone does not know how to write a security bug report, they probably should not

(If it's too easy we end up in https://daniel.haxx.se/blog/2025/07/14/death-by-a-thousand-slops/)

janderssonse: I simplified it and rebased, now it is almost like the original information from the README. Please have a look.

SECURITY.md
```diff
@@ -0,0 +1,7 @@
+# Security Reporting
+
+To report a security vulnerability, please contact [David Peter](https://david-peter.de/) via email.
+
+Vulnerabilities will be handled on a best-effort basis.
```
Enselic: remove this line please

SECURITY.md
```diff
@@ -0,0 +1,7 @@
+# Security Reporting
```
Enselic:
```
# Security vulnerabilities
```

SECURITY.md
```diff
@@ -0,0 +1,7 @@
...
+
+To report a security vulnerability, please contact [David Peter](https://david-peter.de/) via email.
```
Enselic, line 3: This rewording is OK.

SECURITY.md
janderssonse: Yes, fixed

SECURITY.md
janderssonse: Yes, fixed

SECURITY.md
janderssonse: Ack

janderssonse:
> Like I said, I prefer a minimal change.
> 
> I'd also like to keep the header in the README.md to that old links don't break, but it should reference the SECURITY.md file.

Please have a look again:)

Enselic (approved):
thanks for the update, let's merge

PR #3315, opened 2025-05-30
Fix UTF-8 BOM file type detection for first-line syntax patterns

### LINKED ISSUE #3314: Detection fails when file is saved as UTF-8 with BOM
Windows with bat 0.25.0 (25f4f96)

create this file and save as "test.csproj" (yes, csproj) with UTF-8 encoding:

```xml
<?xml version="1.0" encoding="utf-8"?>
<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
</Project>
```

`bat test.csproj` will display as XML.

now resave it with "UTF-8 with BOM" encoding.

`bat test.csproj` will display as plain text.

krikera:
This fix strips the UTF-8 BOM character from the first line before performing syntax detection using `strip_prefix('\u{feff}')`, ensuring that files with BOM are handled correctly.

## Changes

- Modified `get_first_line_syntax` in `src/assets.rs` to strip BOM before pattern matching
- Added comprehensive test coverage for XML, shell scripts, and PHP files with BOM
- Updated CHANGELOG.md

Fixes #3314

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ## Bugfixes
 
+- Fix UTF-8 BOM not being stripped for syntax detection, see #3314 (@krikera)
 - Fix `BAT_THEME_DARK` and `BAT_THEME_LIGHT` being ignored, see issue #3171 and PR #3168 (@bash)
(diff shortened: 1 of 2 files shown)


keith-hall (approved):
Nice, thanks

PR #3336, opened 2025-07-07
Send errors to stderr

### LINKED ISSUE #2561: `[bat error]: 'conf.d' is a directory.` is written to STDOUT instead of STDERR

JerryImMouse:
Resolves #2561

I thought how it can be done better. But the most simple solution for now is this.
The best solution in my opinion is dividing default handle for errors and normal output. But this behaviour will require changing a lot of stuff. 
If this is required - I can do it.


keith-hall (approved):
Thanks

PR #3340, opened 2025-07-08
Build script: replace string-based codegen with `quote`-based codegen

cyqsimon:
This avoids potential issues with patterns that include `#`s (as unlikely as it is), because it would mess with raw string syntax. It's also just the more "proper" way to do things.


build/syntax_mapping.rs
```diff
@@ -116,22 +119,17 @@ impl FromStr for Matcher {
...
-            _ => {
-                let segs = self.0.iter().map(MatcherSegment::codegen).join(", ");
-                format!(r###"Lazy::new(|| build_matcher_dynamic(&[{segs}]))"###)
-            }
-        }
+            segs @ [_, ..] => quote! { Lazy::new(|| build_matcher_dynamic(&[ #(#segs),* ]))},
+        };
+        tokens.append_all(t);
     }
 }
```
cyqsimon: I very much enjoyed finding this nicer way of writing this 😁

cyqsimon:
Rebased.

build/syntax_mapping.rs
```diff
@@ -290,11 +288,13 @@ pub fn build_static_mappings() -> anyhow::Result<()> {
     println!("cargo:rerun-if-changed=src/syntax_mapping/builtins/");
 
     let mappings = read_all_mappings()?;
+    let rs_src = syn::parse_file(&mappings.to_token_stream().to_string())?;
+    let rs_src_pretty = prettyplease::unparse(&rs_src);
```
keith-hall, line 295: parse and then unparse :sweat_smile: I see we have also pulled in a few extra explicit dependencies, but as long as you're happy, it's fine with me :grinning:

keith-hall (approved):

build/syntax_mapping.rs
cyqsimon: Yeah I too thought this was a bit cringe. But it's [what `quote` recommended](https://docs.rs/quote/1.0.40/quote/#non-macro-code-generators), and [`prettyplease` has a very limited set of APIs](https://docs.rs/prettyplease/0.2.35/prettyplease/index.html#functions) to say the least. What to do. Unless there's a better way that I missed (which could be possible), the best can I do is add an `// IMPRV:` note.

As of the dependency situation, all except `prettyplease` were already transiently needed. So I think it's fine.

PR #3342, opened 2025-07-09
docs(man): update --style documentation with priority rules

### LINKED ISSUE #3341: Man page is out-of-date re --style option
[Man page source](https://github.com/sharkdp/bat/blob/master/assets/manual/bat.1.in) is out-of-date with respect to the recent changes to `--style` which are documented in `long-help.txt`.

At least should include

>           When styles are specified in multiple places, the "nearest" set of styles take precedence.
>           The command-line arguments are the highest priority, followed by the BAT_STYLE environment
>           variable, and then the configuration file. If any set of styles consists entirely of
>           components prefixed with "+" or "-", it will modify the previous set of styles instead of
>           replacing them.

krikera:
Update man page to include missing information about style specification priority and the behavior of "+" and "-" prefixes when modifying styles. This aligns the man page with the documentation in long-help.txt.

Fixes #3341


keith-hall (approved):
Thanks

Issue #3344, opened 2025-07-15
Id love to have something like a " bat -r 1050:+-10 filename"


cavanaug:
Id like to be able to say show the context  around a specific line number without having to do the math in my head.

I for example on wsl had a problem that returned from bash as 

$ git browse
/usr/bin/xdg-open: 1050: rundll32.exe: not found

Id love to be able to just show 10 lines of context around 1050

$ bat -r 1050:+-10 /usr/bin/xdg-open

or maybe

$ bat -r 1050:=10 /usr/bin/xdg-open

cavanaug:
I submitted a PR # 3345 that implements this capability as an extra : model to the current line ending.

So you can do

\# Will show lines 1040-1060.   In essence 10 lines of context from the start and end line number ranges.
$ bat -r 1050::10 /usr/bin/xdg-open 

\# Will show lines 990-1110.   In essence 10 lines of context from the start and end line number ranges.
$ bat -r 1000:1100:10 /usr/bin/xdg-open

cavanaug:
#3345 completes this and has been merged

PR #3345, opened 2025-07-15
feat: ✨ add context support to line-range syntax (#3344)

cavanaug:
docs(long-help.txt): 📚 add examples for N::C and N:M:C context syntax
docs(clap_app.rs): 📚 update CLI help text with context syntax examples
feat(line_range.rs): ✨ implement N::C and N:M:C parsing and add tests

I decided to just take a stab at fixing the issue #3344 that I submitted to allow for providing a line number and how many lines of context to show.

### DIFF
--- a/doc/long-help.txt
+++ b/doc/long-help.txt
@@ -194,6 +194,8 @@ Options:
             '--line-range 40:' prints lines 40 to the end of the file
             '--line-range 40' only prints line 40
             '--line-range 30:+10' prints lines 30 to 40
+            '--line-range 35::5' prints lines 30 to 40 (line 35 with 5 lines of context)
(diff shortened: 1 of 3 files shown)


keith-hall (commented):
Thanks, this seems reasonable to me, I like that you explained the use case in the issue 🙂 I wonder, would it be possible to add some integration tests do you think, please? 🙂

cavanaug:
> Thanks, this seems reasonable to me, I like that you explained the use case in the issue 🙂 I wonder, would it be possible to add some integration tests do you think, please? 🙂

Sure, let me look into how to add those tests.    Ill try to get to it over the weekend...

cavanaug:
I added lines to the multiline.txt file so I had more flexibility in testing various ranges and context.   I updated 2 of the tests that used this file previously so things are still appropriate.

PR #3350, opened 2025-07-22
Add Cooklang syntax

dubadub:
Adds syntax highlighting for Cooklang *.cook files https://cooklang.org.


Enselic:
does this language meet the criteria for inclusion? see https://github.com/sharkdp/bat/blob/master/doc/assets.md#adding-new-builtin-languages-for-syntax-highlighting

dubadub:
I think it is, but not in a way as stated in the doc. Sublime isn't widely popular these days (sighs...). VSCode extension more popular with 2.5k installs. Obsidian plugin with 27k installs. There're a few thousand files on GitHub https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.cook+%40+NOT+function&p=2. Of course I'll leave it to your judgement on what you consider acceptable.

Enselic:
I think syntaxes like these are best handled by external bundling of syntaxes and themes

We have an issue for that but I can't find it right now

Thank you for understanding.

PR #3352, opened 2025-07-31
Add --wrap=word option for word wrapping

### LINKED ISSUE #3324: Line wrapping
Currently, when a markdown file has lines that are longer than the screen width, they are split right at screen with, even when that falls in the middle of a word. It would be nice to have the option `--wrap word`, so that the wrapping happens on a word. This would make markdown files more readable.

krikera:
Implements word wrapping functionality that splits text at word boundaries instead of character boundaries, improving readability especially for markdown files.

Changes:
- Add WrappingMode::Word enum variant
- Update clap argument parser to accept 'word' as wrap mode
- Implement word wrapping logic in InteractivePrinter
- Fix printer selection to use InteractivePrinter for explicit wrap modes
- Add comprehensive tests for word vs character wrapping
- Update documentation, help text, and shell completions
- Add changelog entry

Fixes #3324

### DIFF
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -753,42 +753,161 @@ impl Printer for InteractivePrinter<'_> {
 
                             let mut max_width = cursor_max - cursor;
 
-                            // line buffer (avoid calling write! for every character)
-                            let mut line_buf = String::with_capacity(max_width * 4);
+                            match self.config.wrapping_mode {
+                                WrappingMode::Word => {
+                                    // Word wrapping: split text into words and wrap at word boundaries
+                                    let words: Vec<&str> = text.split_whitespace().collect();
+                                    let mut current_line = String::new();
+                                    let mut current_line_width = 0;
 
-                            // Displayed width of line_buf
-                            let mut current_width = 0;
+                                    for word in words.iter() {
+                                        let word_width = word
+                                            .chars()
+                                            .map(|c| c.width().unwrap_or(0))
+                                            .sum::<usize>();
 
-                            for c in text.chars() {
-                                // calculate the displayed width for next character
-                                let cw = c.width().unwrap_or(0);
-                                current_width += cw;
+                                        // Add space before word if it's not the first word in the line
+                                        let space_width =
+                                            if current_line.is_empty() { 0 } else { 1 };
+                                        let total_width = word_width + space_width;
 
-                                // if next character cannot be printed on this line,
-                                // flush the buffer.
-                                if current_width > max_width {
-                                    // Generate wrap padding if not already generated.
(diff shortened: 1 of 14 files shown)


tests/integration_tests.rs
```diff
@@ -2297,6 +2297,38 @@ fn no_wrapping_with_chop_long_lines() {
...
 }
 
+#[test]
+fn word_wrapping() {
+    // Test word wrapping vs character wrapping with a line that contains words
+    bat()
+        .arg("--wrap=word")
+        .arg("--style=plain")
+        .arg("--color=never")
+        .arg("--decorations=never")
```
keith-hall, line 2307: can we add another test with decorations as well please

src/printer.rs
```diff
@@ -753,78 +753,172 @@ impl Printer for InteractivePrinter<'_> {
...
+                                                panel_wrap = if self.panel_width > 0 {
+                                                    Some(format!(
+                                                        "{} ",
+                                                        self.decorations
+                                                            .iter()
+                                                            .map(|d| d
+                                                                .generate(line_number, true, self)
+                                                                .text)
+                                                            .collect::<Vec<String>>()
+                                                            .join(" ")
```
keith-hall: I wonder, as the code above is doing a `split_whitespace()`, and here it is doing a join on space explicitly, will it lose formatting like tabs, non-breaking spaces, double spaces etc.? It would be nice to see more tests here.
Would it be quicker and easier to take a different approach, instead of splitting each word, to find the first whitespace before the terminal width and split there instead?
Also, I would personally be happier if this can be split out into smaller functions, because there is a lot of code here.

keith-hall (commented):
Thanks for having a go at this! I especially like how the completions, help and man pages were all updated! :confetti_ball: 
I have a few concerns and questions on the implementation, please don't let it put you off :slightly_smiling_face:

krikera:
Thank you for the review! I appreciate the feedback.

tests/integration_tests.rs
krikera: yeah sure will

src/printer.rs
krikera: You're absolutely right about the formatting issue. Using split_whitespace() followed by join(" ") will indeed strip tabs, multiple spaces, and other whitespace formatting. 
I'll implement the approach you suggested, and I’ll break this into small, reusable functions.

Enselic:
Labeling as "waiting-on-author" since CI fails (it might become green after rebase).

Enselic:
Closing due to inactivity to keep the PR inbox clean, but of course feel free to reopen if you resume working on this.

PR #3353, opened 2025-07-31
Add syntax mapping for flatpakref

Ferenc-:
* `flatpakref` uses the `.ini` format, see the [man page](https://www.man7.org/linux//man-pages/man5/flatpakref.5.html#FILE_FORMAT).
* `flatpakrepo` also uses the `.ini` format, see the [man page(https://www.man7.org/linux//man-pages/man5/flatpakrepo.5.html#FILE_FORMAT).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@
 - Add syntax highlighting for VHDL, see #3337 (@JerryImMouse)
 - Add syntax mapping for certbot certificate configuration #3338 (@cyqsimon)
 - Update Lean syntax from Lean 3 to Lean 4 #3322 (@YDX-2147483647)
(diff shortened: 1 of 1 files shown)


Ferenc-:
@sharkdp the changelog checker is broken, the regex in `"#${PR_NUMBER}\\b.*${PR_SUBMITTER}\\b"`  is wrong, because using `\b` consumes `-` as a word boundary because `-` is not a word character.
The regex could be `grep -e "#${PR_NUMBER}\\b.*${PR_SUBMITTER}[)]"` and then it would work.

keith-hall (approved):

keith-hall:
> the changelog checker is broken

to be fair it's quite unusual to see a username ending in a dash. But we can look at that CI step separately from this PR. Thanks for adding the mapping :slightly_smiling_face:

PR #3362, opened 2025-08-02
Make batdiff more portable

qianbinbin:
- Use `-0` instead of `--null` for xargs to work on most systems (e.g., `--null` doesn't work on macOS and OpenBSD);
- Add missing `--relative` and `-z` options for git diff.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ You can combine `bat` with `git diff` to view lines around code changes with pro
 highlighting:
 ```bash
 batdiff() {
-    git diff --name-only --relative --diff-filter=d -z | xargs --null bat --diff
+    git diff --name-only --relative --diff-filter=d -z | xargs -0 bat --diff
 }
 ```
 If you prefer to use this as a separate tool, check out `batdiff` in [`bat-extras`](https://github.com/eth-p/bat-extras).
--- a/doc/README-ko.md
+++ b/doc/README-ko.md
@@ -163,7 +163,7 @@ git show v0.6.0:src/main.rs | bat -l rs
 볼 수 있습니다:
 ```bash
 batdiff() {
-    git diff --name-only --diff-filter=d -z | xargs --null bat --diff
+    git diff --name-only --relative --diff-filter=d -z | xargs -0 bat --diff
 }
 ```
 이것을 별도의 도구로 쓰고 싶다면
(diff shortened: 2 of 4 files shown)


keith-hall (approved):
Thanks. I see the CI failures are wholly unrelated to this change, so I logged an issue for it at https://github.com/sharkdp/bat/issues/3363

qianbinbin:
Hi @keith-hall , the command `sed -u -e \"s/\\x1B\[[0-9;]*m//g; s/.\\x08//g\"` for [manpager](https://github.com/sharkdp/bat#man) doesn't work with old BSD sed, for example on macOS Catalina it doens't support:

- `-u` option
- hexadecimal values like `\x1B`

```sh
$ which sed # BSD sed on Catalina
/usr/bin/sed
$ echo -e "abc\033[31mdef\bX" | sed -e "s/\x1B\[[0-9;]*m//g; s/.\x08//g" # 'deX' will be red, backspace kept
abcdeX
$ echo -e "abc\033[31mdef\bX" | sed -e "s/\x1B\[[0-9;]*m//g; s/.\x08//g" | xxd
00000000: 6162 631b 5b33 316d 6465 6608 580a       abc.[31mdef.X.
$ which gsed # GNU sed
/usr/local/bin/gsed
$ echo -e "abc\033[31mdef\bX" | gsed -e "s/\x1B\[[0-9;]*m//g; s/.\x08//g" # colors and backspaces & preceding characters will be removed
abcdeX
$ echo -e "abc\033[31mdef\bX" | gsed -e "s/\x1B\[[0-9;]*m//g; s/.\x08//g" | xxd
00000000: 6162 6364 6558 0a                        abcdeX.
```

Should I add some notes for users to switch to GNU sed, or just replace `sed` with `awk`? `awk` works on Catalina and Linux, though the command will be a bit longer:

```diff
-export MANPAGER="sh -c 'sed -u -e \"s/\\x1B\[[0-9;]*m//g; s/.\\x08//g\" | bat -p -lman'"
+export MANPAGER="sh -c 'awk '\''{ gsub(/\\033\\[[0-9;]*m/, \"\", \$0); gsub(/.\\b/, \"\", \$0); print }'\'' | bat -p -lman'"
```

keith-hall:
Probably in general Mac users should switch to GNU tools lol, but it could be less friction to just have one command to suit all OSes - I personally don't think it is a problem that the awk command is longer, it should be copy/pasted once and then forgotten about 😉

PR #3374, opened 2025-08-13
docs: Add installation with gah

marverix:
Hello,
this PR adds to the documentation the installation method using [gah](https://github.com/marverix/gah/):
> gah is an GitHub Releases app installer, that DOES NOT REQUIRE SUDO! It is a simple bash script that downloads the latest release of an app from GitHub and installs it in ~/.local/bin. It is designed to be used with apps that are distributed as a single binary file.


marverix:
@Enselic I released [gah v1.3.0](https://github.com/marverix/gah/releases/tag/v1.3.0) which adds verification based on sha256 checksum, which is now added to each asset uploaded to GitHub Releases: https://github.blog/changelog/2025-06-03-releases-now-expose-digests-for-release-assets/


For example in [jq](https://github.com/jqlang/jq/releases/tag/jq-1.8.1) is already visible:

Enselic:
cc https://github.com/sharkdp/bat/issues/1867

PR #3375, opened 2025-08-14
README.md: Put some requirements on installation instructions

Enselic:
Adding new installation methods increases risk of supply chain exploits and increases burden of maintainers. Let's put some requirements on what installation methods to have in the README.md

If there is consensus to do this we can close these as "won't fix":
* https://github.com/sharkdp/bat/pull/3374
* https://github.com/sharkdp/bat/pull/2930 (I just now saw I already proposed to close it and there were no objections so I closed it now)

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,6 +265,15 @@ Please report any issues with the help syntax in [this repository](https://githu
 
 ## Installation
 
+<!--
+
+Installation instructions need to:
+* be for widely used systems
+* be non-obvious
+* be from somewhat official sources
+
+-->
+
 [![Packaging status](https://repology.org/badge/vertical-allrepos/bat-cat.svg?columns=3&exclude_unsupported=1)](https://repology.org/project/bat-cat/versions)
 
 ### On Ubuntu (using `apt`)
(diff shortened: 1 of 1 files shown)


marverix:
Hello, as an author of #3374, I would like to know if those changes will also work backwards? Will you doble check all previous installation methods? If no, I think this is very unfair treatment on your side. Of course you are free to do what you want, but I think that in open source world we should be treated equals.

Enselic:
Hi, thank you for the feedback. I agree everyone should be treated equally, but unfortunately my spare time is limited and I don't have time for deep retroactive analyses. I did my best based on quick glances. Perhaps you could help out? What further changes to the installation instructions would you like to suggest and why?

marverix:
@Enselic If your main concern is the security aspect, then your requirement should be that given installator is doing the checksum check. This way you have guarantee (in a certain degree at least) that there is no man in the middle. The problem is that `bat` does not provide such checksums in the assets for the release. And even if you will put such requirement right now, it would be only logical to go thru the list of installators and verify that every one is compliant with new standards.

Enselic:
security is one concern, ease of maintenance and keeping PR inbox trimmed is another, and keeping README.md small and easily digestable is a third.

If someone is using gah and wants to install bat, the first thing they will try is `gah install bat`, and it will work. I don't think that needs to be in the README.

marverix:
Ok, so I will close the PR. `gah install bat` will work, because I added it to the aliases list. Otherwise user will need to run `gah install sharkdp/bat`. So having an alias is just a good will of mine. On one hand I fully understand that you want to keep list small, but on another, I don't understand why now? Well, it's not my concern anymore :) 
Have a nice day!

Enselic:
Have a nice day you too!

Enselic:
Well, since there has been no further objections, let's merge.

PR #3381, opened 2025-08-19
docs: added minor theming documentation notes

### LINKED ISSUE #3288: Documentation: Add a few details
I tried to change bat's default theme to my liking (make header-rules easier to spot) and finally succeeded. Could you add the following items to the documentation? That will make things easier for people that want similar customizations.

* The file name for theme files in "$(bat --config-dir)/themes" must end with ``.tmTheme``
* The default theme is either ``Monokai Extended`` or ``Monokai Extended Light``, depending on the terminal being recognized as having a light or dark theme
* The name of a custom theme is taken from the file name, not from its ``name`` property.
* Most of the theme is for syntax coloring for various specific languages, but the ``settings.settings`` dict of the theme contains general settings for bat's own use. Among these is the key ``gutterForeground``,  which is not set in the default theme because it relies on the hardcoded default value.

Thanks for the great software.

CalebLarsen:
Closes #3288. I found all of their points reasonable except maybe the last one, which I didn't include documentation for. All changes have only been made to the English `README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -503,8 +503,9 @@ bat --completion <shell>
 ### Highlighting theme
 
 Use `bat --list-themes` to get a list of all available themes for syntax
-highlighting. To select the `TwoDark` theme, call `bat` with the
-`--theme=TwoDark` option or set the `BAT_THEME` environment variable to
+highlighting. By default, `bat` uses `Monokai Extended` or `Monokai Extended Light`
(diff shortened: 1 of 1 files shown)


CalebLarsen:
I'm of the opinion that these changes are on the level of correcting typos, but I will make an entry in the `CHANGELOG` if desired.

Enselic (approved):

PR #3399, opened 2025-09-02
dependabot: Stop updating submodules

Enselic:
I personally find dependabot updating submodules to be more annoying than helpful. It is better to let humans do when someone comes around that has an actual problem to solve.

This will make it easier to keep the PR inbox trimmed.


Enselic:
Let's keep status quo for a while longer. Closing for now.

PR #3402, opened 2025-09-09
Add --pager=builtin using the Minus library

### LINKED ISSUE #1540: Use `minus` as a paging library
I don't know what bat currently uses for paging (looks like `less` maybe) but it proposes two main problems. First is cross-compatiblity, bat cannot be packaged for Windows and requires `less` (which requires Mingw) to be installed. Second is less is binary and it's use as a library is not very perfect solution. These are the main problem `minus` solved. It's a pure rust library for paging output in a terminal. It provides great customisation of output which also being very compact

I wanted to know whether the `bat` team could implement this pager inside `bat`

**Disclosure: I am the author of minus**

academician:
This adds a new value for the `--pager` option called "builtin", which uses the the [minus](/AMythicDev/minus) pager library. The Minus crate implements a pure Rust terminal pager with fairly minimal dependencies.

There was a suggestion on the original ticket to change "builtin" to "builtin-minus", but for now I'm leaving it as-is; we can always add "builtin-minus" and "builtin-other" if we add other pagers.

Another future work item might be to use `minus` by default if `less` is not found in the path. We could also choose to use the built-in pager by default if PAGER and BAT_PAGER aren't set.

Changes:
* When the "pager" feature is enabled, includes minus as a dependency
* If --pager or BAT_PAGER are set to "builtin", uses minus for paging
* Update documentation, help text, and shell completions
* Add changelog entry

Closes #1540


Enselic:
I did a quick skim-through and this looks great! (Not sure when I will get time for a detailed look. I also want to give other maintainers some time to chime in if they want.)

Enselic:
@keith-hall feel free to merge this if you're happy with it (I don't need to take a detailed look if this looks good to you)

src/output.rs
```diff
@@ -11,6 +13,36 @@ use crate::paging::PagingMode;
...
+    pager: minus::Pager,
+    handle: Option<JoinHandle<Result<()>>>,
+}
+
+#[cfg(feature = "paging")]
+impl BuiltinPager {
+    fn new() -> Self {
+        let pager = minus::Pager::new();
+        let handle = {
+            let pager = pager.clone();
```
keith-hall, line 27: This bit seems a little odd to me at first glace. We just created a new instance of a minus Pager in the statement above, and here we clone it ... Perhaps there is a way to do this without cloning?

src/output.rs
```diff
@@ -138,31 +176,43 @@ impl OutputType {
...
 impl Drop for OutputType {
     fn drop(&mut self) {
-        if let OutputType::Pager(ref mut command) = *self {
-            let _ = command.wait();
+        match *self {
+            OutputType::Pager(ref mut command) => {
+                let _ = command.wait();
+            }
+            OutputType::BuiltinPager(ref mut pager) => {
+                if pager.handle.is_some() {
```
keith-hall, line 211: Are there any circumstances where the handle would be None?

src/output.rs
```diff
@@ -138,31 +176,43 @@ impl OutputType {
...
+                    let _ = pager.handle.take().unwrap().join().unwrap();
```
keith-hall, line 212: Is there any way to make this cleaner and clearer about fallibility? Would it ever cause a panic? Do we need to be more defensive?

src/output.rs
academician: This is the normal way of doing this since I'm spawning the pager in a thread. The [Pager struct](https://github.com/AMythicDev/minus/blob/b8f8f58d20d720e7a09694d13da9948c51096407/src/pager.rs#L37-L41) just contains a pair of crossbeam channels (examples in [minus's docs](https://docs.rs/minus/latest/minus/#threads) and [crossbeam's docs](https://docs.rs/crossbeam-channel/latest/crossbeam_channel/#sharing-channels)).

We have to pass ownership of one copy of `pager` to the lambda just below this so it can receive messages later on from the other end, which we treat as a `Writer`.

src/output.rs
academician: No, it will always be Some, so I suppose this check strictly necessary and we could unwrap() or expect() instead. I used an Option here because we have to take ownership of the interior value on the line below, but I'm realizing now that OnceCell would be more appropriate, so I could change it to that. It wouldn't make this code much different, though.

src/output.rs
academician: The `join().unwrap()` would panic if the `minus` thread panics, so I suppose that's a concern. Do you have an idea of what you think would be correct in that case?

academician: I think ideally we'd pass any errors to the error handler in controller, but that's a bit difficult to do with how we currently construct OutputType and how these errors happen in `Drop::drop()`. We could give OutputType an error handler function that it could call in case one of failure, but it'll get a bit complicated.

More simply, we could use [`std::panic::catch_unwind`](https://doc.rust-lang.org/stable/std/panic/fn.catch_unwind.html) around `Controller::run_with_handler` to deal with any panics in a generic way, but using that to create great error messages is also challenging.

Either way, we already have a similar issue with `OutputType::Pager`'s drop just above, since `command.wait()` can also return an error that we're not handling. So IMO this might be out of scope here, but I might be able to spend some time on it later.

src/output.rs
keith-hall: Thanks for the explanation 👍

src/output.rs
keith-hall: Let's consider it out of scope for now, thanks for talking it over 🙂 I originally had the idea that if the pager has an error we could switch to no pager, but that's almost certainly infeasible with how the input is read and output gets written line by line. So let's keep it simple. It would be nice to show a friendly error instead of panicking, but if the likelihood of panicking is very low then it isn't worth over-engineering. We can just keep this in mind and watch for feedback/bug reports 👍

src/output.rs
keith-hall: I will leave it up to your discretion, which is the better approach. If you are happy with the current code, we can leave it as is 🙂

Issue #3404, opened 2025-09-12
Option to change the git signs


sand4rt:
Is it possible to add an option to change the git signs?

flashios09:
@sand4rt I can create a PR for this feature request if @sharkdp approve it !

flashios09:
Sneak peek:

<img width="932" height="977" alt="Image" src="https://github.com/user-attachments/assets/7a5105df-793e-43fe-804a-f3f8e08cc868" />

1. Modern **gitsigns**: matching nvim gitsigns plugin(https://github.com/lewis6991/gitsigns.nvim)
2. Updated `printer.rs` too to read the `markup.changed`, `markup.inserted`, `markup.deleted` scopes, otherwise just use the default styles `Yellow.normal()`, `Green.normal()`, `Red.normal()`:
```rust
impl Colors {
    fn plain() -> Self {
        Colors::default()
    }

    fn colored(theme: &Theme, true_color: bool) -> Self {
        let gutter_style = Style {
            foreground: match theme.settings.gutter_foreground {
                // If the theme provides a gutter foreground color, use it.
                // Note: It might be the special value #00000001, in which case
                // to_ansi_color returns None and we use an empty Style
                // (resulting in the terminal's default foreground color).
                Some(c) => to_ansi_color(c, true_color),
                // Otherwise, use a specific fallback color.
                None => Some(Fixed(DEFAULT_GUTTER_COLOR)),
            },
            ..Style::default()
        };

        let scope_style = |scope_name: &str, default_style: Style| {
            theme
                .scopes
                .iter()
                .find(|item| {
                    item.scope
                        .eq(&ScopeSelectors::from_str(scope_name).unwrap())
                })
                .and_then(|item| item.style.foreground)
                .map(|c| Style {
                    foreground: to_ansi_color(c, true_color),
                    ..Style::default()
                })
                .unwrap_or(default_style)
        };

        Colors {
            grid: gutter_style,
            rule: gutter_style,
            header_value: Style::new().bold(),
            git_added: scope_style("markup.inserted", Green.normal()),
            git_removed: scope_style("markup.deleted", Red.normal()),
            git_modified: scope_style("markup.changed", Yellow.normal()),
            line_number: gutter_style,
        }
    }
}
```
I need to add tests before the **PR**, maybe today !

sand4rt:
Nice! Would it be possible to render the icons after the linenumber line instead of before? like gitsigns.nvim?

flashios09:
[...]

PR #3407, opened 2025-09-23
Add `--gitsigns` option (#3404)

flashios09:
See #3404

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,7 @@
 - Update Lean syntax from Lean 3 to Lean 4 #3322 (@YDX-2147483647)
 - Map `.flatpakref` and `.flatpakrepo` files to INI syntax #3353 (@Ferenc-)
 - Update hosts syntax #3368 (@keith-hall)
+- Map `.kshrc` files to Bash syntax #3364 (@ritoban23)
 
 ## Themes
 
--- a/src/syntax_mapping/builtins/unix-family/50-korn-shell.toml
+++ b/src/syntax_mapping/builtins/unix-family/50-korn-shell.toml
@@ -1,3 +1,6 @@
 # KornShell is backward-compatible with the Bourne shell #2633
 [mappings]
-"Bourne Again Shell (bash)" = ["*.ksh"]
+"Bourne Again Shell (bash)" = [
+    "*.ksh",
+    "*.kshrc" 
+]


flashios09:
@sand4rt take a look and tell me what you think :)

Enselic:
Hi! What is the problem with the default signs? If they are problematic somehow it's probably better to fix the problems they have than to add complexity to the code and for users.

flashios09:
> than to add complexity to the code and for users

Hi @Enselic ,
Which complexity are you talking about ?

This PR will just add `--gitsigns` option, just like the `--style` option, used to **customize** user preferences.

This is how gitsigns looks with the **modern** preset(`bat --gitsigns=modern path/to/file`):
image


The whole feature is about **100 lines**, **90%** of those +1000 lines are just **tests** :)

Enselic:
Ok, let's start with the first question.

What is the problem with the default signs?

sand4rt:
> Ok, let's start with the first question.
> 
> 
> 
> What is the problem with the default signs?

They are not the same as other terminal apps like gitsigns.nvim

Enselic:
Why not make gitsigns configurable instead of bat?

sand4rt:
> Why not make gitsigns configurable instead of bat?

It is configurable in gitsigns, but i find the line-based indicators much clearer and easier to parse visually. In my opinion they reduce visual noise compared to the current symbols, and make diffs easier to scan. Popular editors like VSCode and JetBrains IDEs also use line markers as the default.

flashios09:
> Ok, let's start with the first question.
> 
> What is the problem with the default signs?

Hi @Enselic , what is the problem with the default **cat** binary that comes with almost all UNIX based systems ?
The answer is **OUTDATED**, that's why we have **bat**, **rg** instead of **grep**, **fd** instead of **find**, **eza** instead of **ls** ...

What it the problem with the default signs ?

**OUTDATED**, we aren't in **2015** anymore when **SublimeText2** was the most popular text editor(all the respect for SublimeText, still one of the best text editors), at that time, the default gitsigns was:
- **~** in **orange** for modified lines
- **+** in **green** for new(added) lines
- **-** in **red** for deleted lines

Those aren't the **"default"** anymore, with the new era of text editors/IDE, VSCode and all its fork(Cursor ...), Intellij IDE, Neovim with gitsigns.nvim plugin, all are using new gitsigns:
- **|** in **blue** for modified lines
- **|** in **green** for new(added) lines
- **-** in **red** for deleted lines

With this PR, we will be able to:
1. Use the TextMate theme file to apply the **colors** to gitsigns(if the scope exists, otherwise use the current default colors), like this we can change/customize the modified/added/deleted sign colors
2. Use the `--gitsigns` option to customize the signs, via the **classic/modern** preset or a custom 4 comma-separated characters(already mentioned/explained in the help command).

BEFORE(classic gitsigns):
image

AFTER(modern gitsigns):
image

Now tell me, which one is easier to get the git changes from ? **classic** preset or the **modern** one used by all modern text editors/IDE ?

PS:
This PR has **0 breaking changes**, no effect at all on the default **bat** command with all the **already existent** options/config, **classic**(the current one) is still the **default** preset for gitsigns, which means only the user who wants to customize gitsigns, will need to use the new `--gitsigns` option or the `BAT_GITSIGNS` env var, just like the `--style` or `--theme` options.

Enselic:
Sorry for being annoying. I just needed to be convinced this was a worthwhile thing to change, and now you have convinced me. Thank you for taking the time to do so.

I (or some other maintainer) will look at the details later when I get the time.

src/printer.rs
```diff
@@ -892,14 +897,130 @@ impl Colors {
...
         };
 
+        // Get scope style from theme using given `scope_name` or use `default_style`.
+        let get_scope_style = |scope_name: &str, default_style: Style| {
+            let scope_selectors = ScopeSelectors::from_str(scope_name).unwrap();
+
+            theme
+                .scopes
+                .iter()
+                .find(|item| item.scope.eq(&scope_selectors))
```
keith-hall, line 907: Scope selectors from the theme generally shouldn't be expected to *exactly* match a scope, but instead use scope selector matching logic to determine which theme rules match and which has the highest precedence... Probably it doesn't matter too much here, but I would prefer consistency with how highlighting works

flashios09: Sorry I didn't get the point here ?
TextMate theme files usually use `markup.inserted`, `markup.changed` and `markup.deleted` scopes to define colors for gitsigns:
image

keith-hall: It seems like I missed this reply earlier, sorry about that. I meant that the tmTheme may choose to have rules with less specific scope selectors like `markup` which would match `markup.deleted`, `markup.changed` etc. or even multiple rules could theoretically match, and whose styles would get applied/merged together etc. based on precedence order. So my idea was to [use proper scope selector logic](https://github.com/trishume/syntect/blob/master/src/highlighting/selector.rs#L34) instead of checking for string equality, but probably it isn't going to be necessary in practice, because, like you pointed out, most themes have these specific rules.

src/gitsigns.rs
```diff
@@ -0,0 +1,75 @@
...
+            .collect();
+
+        if parts.len() != 4 {
+            return Err("Expected 4 comma-separated signs: `modified,added,removed-above,removed-below`, e.g. `~,+,‾,_`".into());
+        }
+
+        for (i, part) in parts.iter().enumerate() {
+            if part.chars().count() != 1 {
+                return Err(format!(
+                    "Invalid sign at position {}: `{}`. Each sign must be a single character.",
```
keith-hall, line 61: Maybe some unit tests for the error cases could be useful

doc/short-help.txt
```diff
@@ -23,6 +23,8 @@ Options:
           Specify the name to display for a file.
   -d, --diff
           Only show lines that have been added/removed/modified.
+      --gitsigns <gitsigns>
+          Set git `added`, `modified`, `removed-above`, `removed-below` signs. (default: classic)
```
keith-hall, line 27: it would be great if we could also update the auto completion scripts please, so they would suggest the new argument and it's possible values

src/printer.rs
flashios09: > but probably it isn't going to be necessary in practice, because, like you pointed out, most themes have these specific rules

Sorry for late response, do you want/suggest using `ScopeSelector::does_match` instead because i got confused here ?

src/gitsigns.rs
flashios09: Already exist https://github.com/flashios09/bat/blob/b27d65801852828c6026b2c27c09e98405d5217e/tests/integration_tests.rs#L3486

doc/short-help.txt
flashios09: Sorry i didn't get the point here, can you explain more please ? What do you mean by "*update the auto completion scripts*" ?

keith-hall: There are some scripts which are used for completions in various shells like bash, fish, zsh. When typing `bat --t` and pressing the tab key, it will automatically finish writing `--theme` for example. So when adding new command line options like `--gitsigns`, it would be nice to ensure the completion scripts know to offer that as a suggestion. If it's too much hassle, don't worry, we can always do it later. I think I got Copilot or another LLM AI to help me last time I needed to do it because I don't know all the various shell syntaxes.

src/printer.rs
keith-hall: I would prefer it, yes please 👍

keith-hall:
I see there are a couple of conflicting files now... if you are still interested in this change, maybe you can rebase or merge main branch in @flashios09

Issue #3269, opened 2025-04-22
A bad config file makes it impossible to use


spott:
**What steps will reproduce the bug?**

1. put some weird option (or option for another system) in a config file
2. run bat with any command or option.

**What happens?**

the bad option in the config file means bat is unusable.  No -V, --help, or anything else to fix the problem.

**What did you expect to happen instead?**

bat to complain, but ignore the option in the config file... at least for -V or --help.

**How did you install `bat`?**

nix home manager (which is why the config file being bad is kinda a pita.)

---

**bat version and environment**

```
system
------

**$ uname -srm**
Linux 6.6.87 x86_64

bat
---

**$ bat --version**
error: unexpected argument '--theme-dark' found

  tip: a similar argument exists: '--theme'

Usage: bat --theme <theme> [FILE]...

For more information, try '--help'.

**$ env**
PAGER=less
error: unexpected argument '--theme-dark' found

  tip: a similar argument exists: '--theme'

Usage: bat --theme <theme> [FILE]...

For more information, try '--help'.

bat_config
----------
error: unexpected argument '--theme-dark' found

  tip: a similar argument exists: '--theme'

Usage: bat --theme <theme> [FILE]...

For more information, try '--help'.

bat_wrapper
-----------
diagnostic.sh: line 83: file: command not found

No wrapper script for 'bat'.

bat_wrapper_function
--------------------

No wrapper function for 'bat'.

**$ /run/current-system/sw/bin/zsh --login -i -c type cat**
```
cat is an alias for bat
```

tool
----

**$ less --version**
less 668 (PCRE2 regular expressions)
```

(from the diagnostic script.)

PR #3414, opened 2025-09-27
Fix help/version/diagnostic commands with invalid config

keith-hall:
Fixes https://github.com/sharkdp/bat/issues/3269


Enselic (approved):
You can consider yourself to have a perpetual Approve from me on all PRs that adds new regression tests and don't break existing tests 👍

Issue #3293, opened 2025-05-14
update documentation to add bash autocompletion for "help" syntax highlighting


dandrake:
I added the bash alias and help() function so that I get syntax highlighting for "`--help`" stuff. It's great!

In my shell, though, there's no autocompletion on the commands. But it's very easy to make that happen if you already have the bash-completion package installed, which I suspect most users do: just put

```
complete -F _command help
```

in your shell configuration -- `.bashrc` or elsewhere.

Could you update the documentation to mention this? It makes "`help <command>` super useful!

PR #3423, opened 2025-10-01
Update documentation to add bash autocompletion for help

Euge908:
Related to Issue[ #293](https://github.com/sharkdp/bat/issues/3293)

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,6 +231,7 @@ You can use `bat` to colorize help text: `$ cp --help | bat -plhelp`
 
 You can also use a wrapper around this:
 
+
 ```bash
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -241,6 +242,16 @@ help() {
 
 Then you can do `$ help cp` or `$ help git commit`.
 
+
+Bashcompletion might not work properly with the above command, because `help` is aliased as a function. To resolve this, you can also put the following in your `.bashrc`:
```
Enselic, line 246: can the problematic snippet be fixed so no workaround is needed? I don't fully understand why the original snippet has a problem though. What are the steps to reproduce?

PR #2931, opened 2024-04-08
provide syntax highlighting for .spec files

rtczza:
https://github.com/sharkdp/bat/issues/2605

PR #3431, opened 2025-10-09
Add RPMSpec support

### LINKED ISSUE #2605: Please provide syntax highlighting for .spec files
Files like https://build.opensuse.org/package/view_file/utilities/bat/bat.spec

Essentially they work like bash files but allow `Key: value` pairs and macros that work like single line `%macro` or inline `%{macro}`

DarkMatter-999:
closes #2605 
supersedes #2931 

This PR adds syntax definitions for RPM `.spec` files (located at assets/syntaxes/02_Extra/RPMSpec.sublime-syntax).

It also includes tests for syntax highlighting.

ref: https://github.com/dreadatour/RPMSpec

### DIFF
--- a/assets/syntaxes/02_Extra/RPMSpec.sublime-syntax
+++ b/assets/syntaxes/02_Extra/RPMSpec.sublime-syntax
@@ -1,0 +1,66 @@
+%YAML 1.2
+---
+# http://www.sublimetext.com/docs/syntax.html
+name: RPM Spec
+file_extensions:
+  - spec
+scope: source.spec
+contexts:
+  main:
+    - match: '(Icon|ExclusiveOs|ExcludeOs):'
+      scope: keyword.spec
+    - match: '(BuildArch|BuildArchitectures|ExclusiveArch|ExcludeArch):'
+      captures:
+        0: keyword.spec
+      push:
+        - match: $
+          pop: true
+        - include: archs
+        - match: '[^\s,]+'
+          scope: invalid.spec
+    - match: '(Conflicts|Obsoletes|Provides|Requires|Requires\(.*\)|Enhances|Suggests|BuildConflicts|BuildRequires|Recommends|PreReq|Supplements):'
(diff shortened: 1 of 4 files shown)


assets/syntaxes/02_Extra/RPMSpec.sublime-syntax
```diff
@@ -0,0 +1,66 @@
+%YAML 1.2
```
Enselic, line 1: Under what license is this syntax file? I suggest adding a license to the top.

Enselic (changes requested):
Does this syntax fulfil the popularity requirement? See https://github.com/sharkdp/bat/blob/master/doc/assets.md#criteria-for-inclusion-of-new-syntaxes.

assets/syntaxes/02_Extra/RPMSpec.sublime-syntax
DarkMatter-999: Thanks for the review. The code is licensed under the MIT License.

Since this package has only 5K downloads on Package Control, I don’t think it meets the popularity requirement. Please let me know if you’d like me to work on this further.

Enselic: Thank you for looking it up. I'm afraid there are not enough users for this syntax, so I will close this PR to keep the PR inbox clean.

DarkMatter-999: No issues. Thanks ✌️

mcepl:
> Does this syntax fulfil the popularity requirement? See https://github.com/sharkdp/bat/blob/master/doc/assets.md#criteria-for-inclusion-of-new-syntaxes.

Are you certain that making decision for the format specific for the platform [where SublimeText is not used much](https://chat.mistral.ai/work/a560b8d4-e282-4926-ac60-cd8f5050e84e) is wise?

PR #3432, opened 2025-10-11
make --help and -h use pager

### LINKED ISSUE #1587: Automatically use pager on bat (--list-themes|-h|--help)
In bat 0.18 an automatic switch between paged and unpaged output was implemented for `bat -L` (#1394).

For consistency's sake this should also implemented for `bat (--list-themes|-h|--help)`, shouldn't it?

MuntasirSZN:
This pr implements use of pager in help output. Fixes #1587 

Help coloring is the same. 
image


MuntasirSZN:
> A couple of questions - does it honor `--paging=never` on the command line and in the config file(s)? And does/can the help display keep the original colors from clap, or should it apply our help syntax highlighting?

No, it doesn't honor. Its the same behavior as list-themes and list-languages. Colors cannot be shown, as its converted to string. Yes, our help highlighting can be used. Lemme do it asap.

MuntasirSZN:
@keith-hall i think just setting `language: Some("help")` will do, right?

keith-hall:
Yep, that should do the trick 👍 once done, it will be useful to compare the output highlighted from clap and this one, so please post a before and after screenshot 🙂 

Also, I think we should support when the user requests to skip the pager (and colors), especially if someone wants to pipe the output to a file or into another program

Edit: plus, as the pager is configurable, especially now that minus is an option, it would make sense to ensure it uses the configured pager. And let's add an integration test for the help output, to ensure that we catch regressions in the highlighting.

MuntasirSZN:
Yeah now theming is done..

MuntasirSZN:
I couldn't use minus pager. Used --pager builtin but doesnt work. Maybe its just me? @keith-hall i need help..

keith-hall:
> I couldn't use minus pager. Used --pager builtin but doesnt work. Maybe its just me? @keith-hall i need help..

works for me on your branch
image

Do I understand correctly that it only takes config from the command line? Likely the recent changes in https://github.com/sharkdp/bat/pull/3414 have complicated this, but it would be nice to share the same code/logic as the "normal" path, and get config from `BAT_PAGER`,`PAGER` and other environment variables etc. as a fallback when the arguments are not provided in the config file or command line invocation...

MuntasirSZN:
Ok on it @keith-hall

MuntasirSZN:
Done

MuntasirSZN:
@keith-hall is it good now?

keith-hall:
I haven't had much time, but I had a quick look and I don't understand how it handles the env vars, maybe I'm missing something? I had the idea that it might make sense to revert https://github.com/sharkdp/bat/pull/3414 to make this change (for outputting the help via the pager and honoring all existing command line switches and env vars) trivial, and try other approaches to allow showing help when some arguments are invalid - like stripping all invalid arguments when `--help` is in the argument list etc.

MuntasirSZN:
@keith-hall its via bat::config::get_pager_executable which does all the work

MuntasirSZN:
Also all of that stripping is already done. Because all of bats logic comes after help logic. And if help flag is passed, all is cancelled and help is shown.

MuntasirSZN:
@keith-hall is there anything else thats needed in this pr to be merged?

keith-hall:
In my testing, it seems to ignore the `--theme` option, and always uses the default theme. i.e. `bat --theme=TwoDark --help` gives the same output as `bat --theme=ansi --help`, which is unexpected

MuntasirSZN:
Ok done @keith-hall can I get a final review? basically all at once.

keith-hall:
I realize it must be annoying to have reviews come in piecemeal, but sometimes its hard to do a complete review in one pass, especially when things need tweaking. I do appreciate your diligence in this one, however :+1: 
I think it is nearly ready. I notice it currently doesn't respect the theme from the config file or `BAT_THEME` env var, which would be nice to have working... This is kind of what I meant, that it would be easier to maintain if we didn't have to special-case everything for help, just have it working through the same code-path like for other inputs...

MuntasirSZN:
> I realize it must be annoying to have reviews come in piecemeal, but sometimes its hard to do a complete review in one pass, especially when things need tweaking. I do appreciate your diligence in this one, however 👍 I think it is nearly ready. I notice it currently doesn't respect the theme from the config file or `BAT_THEME` env var, which would be nice to have working... This is kind of what I meant, that it would be easier to maintain if we didn't have to special-case everything for help, just have it working through the same code-path like for other inputs...

Its not possible easy way as described in the issue themselves. I also did what you wanted. SO done. @keith-hall

Enselic:
The `pull[bot]` seems buggy? It forced push this branch to a commit already on master, namely https://github.com/sharkdp/bat/commit/53714265880c1485accdc49d708fc65b9665ec6a:
image

This PR now contains 0 commits:
image

I think it was attributed to auto-merge-dependabot-prs since that happened to be the latest commit on master at the time.

MuntasirSZN:
@Enselic i will open another pr. The changes are gone here. Why it was merged? It will take days, I have health issues.

Enselic:
It wasn't merged. The pull[bot] force-pushed away the commit. My suggestion is to stop using pull[bot]. I hope you get better.

Issue #3316, opened 2025-05-30
`--style` documentation is misleading


TheBicPen:
The description on --style is misleading. It says "auto: same as 'full', unless the output is piped." and "full: enables all available components (default)".

This is misleading - I read it as saying that when --style=full is passed, the full decorations are enabled in a command like this: `echo '' | bat --style=full | cat`. This is not the case - this pipeline prints a newline rather than something like "STDIN\n 1 ". Compare the output with `echo '' | bat --style=full`. If the output is piped, loop_through is enabled, the SimplePrinter is selected, and the output is printed without any decorations regardless of the `--style=full` flag. This effectively throws away the style configuration whenever cat-compatibility mode is enabled, which is not what the documentation says.

As a secondary question, what is the point of the `--style=auto` option if the style is ignored when the output is piped? It provides the same functionality as `--style=full --decorations=auto`. It seems like having both is redundant.

lmmx:
Seems like this is the same issue I came to report, please let me know if I should make a separate one, but when piping with the `-n` flag (equivalent to `—style numbers`) I note that no numbers are passed.

(I think this is less a documentation issue and should rather affect the behaviour to be clear!)

keith-hall:
I believe https://github.com/sharkdp/bat/issues/2935 is related

pharmacologic:
i've also recently encountered this issue of style options (`--style=header-filename` in my case) not being passed when piping to another program. while asking an LLM for workarounds it informed me that setting `--color=always` will preserve style options when piped - which works regardless of whether coloring is part of the style, apparently

<img width="640" height="428" alt="Image" src="https://github.com/user-attachments/assets/820c75b5-ef3a-4f3d-aa9a-8f49587dd854" />

lmmx:
I don't know if this is at all related, but I came across [anstream](https://crates.io/crates/anstream) this week, which provides utilities for piping support built into a `println!`/`eprintln!` macro (that for me solved a broken pipe error when piping into `rg -q` which prematurely closes the pipe after a match)

> _anstream_
>
> IO stream adapters for writing colored text that will gracefully degrade according to your terminal's capabilities.

It's used in clap, and I think cargo itself, and I'm wondering now if it'd be useful here. I did a quick search on the repo and can see plain `println!` is used, perhaps this is a source of issues. Or perhaps it is abstracted beyond that and I'm barking up the wrong tree :-)

lmmx:
## Bat vivisection

[...]

Issue #3125, opened 2024-11-15
bat -n doesn't behave like cat -n


depesz:
I'm not entirely sure if it warrants being called a bug, but it's a weird feature that i wasn't expecting.

**What steps will reproduce the bug?**

1. Run `printf '%s\n' a b c | cat -n | head -n 5`
2. Run `printf '%s\n' a b c | bat -n | head -n 5`
3. observe line numbers missing from 2nd command.

**What happens?**

Line numbers present from running cat:

```
=$ printf '%s\n' a b c | cat -n | cat -
     1  a
     2  b
     3  c
```

disappear when running bat:

```
=$ printf '%s\n' a b c | bat -n | head -n 5
a
b
c
```

**What did you expect to happen instead?**

Given that bat is supposedly be replacement of cat, I would expect it to handle cat's options in the same way.

**How did you install `bat`?**

`cargo install bat`

---

**bat version and environment**

#### Software version

bat 0.24.0

#### Operating system

Linux 6.11.0-9-generic

#### Command-line

```bash
bat --diagnostic
```

#### Environment variables

```bash
SHELL=/bin/bash
PAGER=less
LESS=-iMFXSx4R
LANG=en_US.UTF-8
LC_ALL=<not set>
BAT_PAGER=<not set>
BAT_PAGING=<not set>
BAT_CACHE_PATH=<not set>
BAT_CONFIG_PATH=<not set>
BAT_OPTS=<not set>
BAT_STYLE=<not set>
BAT_TABS=<not set>
BAT_THEME=<not set>
XDG_CONFIG_HOME=<not set>
XDG_CACHE_HOME=<not set>
COLORTERM=<not set>
NO_COLOR=<not set>
MANPAGER=<not set>
```

#### System Config file

Could not read contents of '/etc/bat/config': No such file or directory (os error 2).

#### Config file

Could not read contents of '/home/depesz/.config/bat/config': No such file or directory (os error 2).

#### Custom assets metadata

Could not read contents of '/home/depesz/.cache/bat/metadata.yaml': No such file or directory (os error 2).

#### Custom assets

'/home/depesz/.cache/bat' not found

#### Compile time information

- Profile: release
- Target triple: x86_64-unknown-linux-gnu
- Family: unix
- OS: linux
- Architecture: x86_64
- Pointer width: 64
- Endian: little
- CPU features: fxsr,sse,sse2
- Host: x86_64-unknown-linux-gnu

#### Less version

```
> less --version
[...]

PR #3438, opened 2025-10-16
feat: make output pipeable with `-n`, non-auto styles

lmmx:
## Background

I recently noticed I cannot pipe `bat -n` output anywhere (e.g. cat, other Unix tools, the clipboard via xclip) and get line numbers like `nl`, it just looks like `cat` output.

I have come across these duplicates, there may be other reports of the same thing:

- #2935
- #3125
- #3316 
- #2983

I found there was an abandoned PR with the same discussion (#2983) and someone proposed a new approach, but it never got extended to a full solution.

Since that PR, lots of the machinery it proposed was added (passing style components through in the `loop_through` case which occurs when piped), which made this easier this time round.

I've made sure to add tests and to preserve the standard behaviour (the `--style=auto` case, which appears to be the CLI default, not to be confused with `--style=default` which is set explicitly and behaves like `auto` but the docs highlight that auto should be expected to pipe as plain)

## Description

- <strike>Gate the use of the SimplePrinter on `plain` too</strike>
- Gate the use of the SimplePrinter (via the `loop_through` variable) on:
  - the `number` flag being passed
  - the `style` flag being passed with `numbers` in the value, or anything except `auto`
    - Note: the docs say that auto is "same as 'default', unless the output is piped." i.e. should pipe as plain

## Bugs fixed

- [x] `-n`/`--number` should pipe
- [x] `--style=numbers` should pipe
- [x] `--style=header,grid,numbers` should pipe
- [x] `--style=default` should pipe
- `--style=auto` should continue to not pipe (retained existing behaviour, not a bug fix)

## Test cases

- You should be able to set the `-n`/`--number` flag
- You should be able to pass it in `--style` equivalently (as `--style=numbers`)
- You should be able to combine it with other flags (e.g. `--style=header,grid,numbers`)

Integration tests added:

- fn piped_output_with_default_style
- fn piped_output_with_line_number_flag
- fn piped_output_with_line_numbers_style_flag
- fn piped_output_with_line_numbers_with_header_grid_style_flag
- fn piped_output_with_auto_style
- fn piped_output_with_default_style_flag

<details><summary>Click to show demos of existing bat behaviour</summary>

```sh
louis 🌟 ~/dev/bat $ bat Cargo.toml -r :3
───────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: Cargo.toml
───────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ [package]
   2   │ authors = ["David Peter <mail@david-peter.de>"]
   3   │ categories = ["command-line-utilities"]
───────┴────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
louis 🌟 ~/dev/bat $ bat Cargo.toml -r :3 --style=grid
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
[package]
authors = ["David Peter <mail@david-peter.de>"]
categories = ["command-line-utilities"]
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
louis 🌟 ~/dev/bat $ bat Cargo.toml -r :3 --style=header
File: Cargo.toml
[package]
authors = ["David Peter <mail@david-peter.de>"]
categories = ["command-line-utilities"]
louis 🌟 ~/dev/bat $ bat Cargo.toml -r :3 --style=header,grid
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
File: Cargo.toml
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
[package]
authors = ["David Peter <mail@david-peter.de>"]
categories = ["command-line-utilities"]
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
louis 🌟 ~/dev/bat $ bat Cargo.toml -r :3 --style=header,grid,numbers
─────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
     │ File: Cargo.toml
─────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1 │ [package]
   2 │ authors = ["David Peter <mail@david-peter.de>"]
   3 │ categories = ["command-line-utilities"]
─────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
louis 🌟 ~/dev/bat $ bat Cargo.toml -r :3 --style=header,grid,numbers | cat
[package]
authors = ["David Peter <mail@david-peter.de>"]
categories = ["command-line-utilities"]
louis 🌟 ~/dev/bat $ bat Cargo.toml -r :3 --style=header,grid,numbers | cat
[package]
authors = ["David Peter <mail@david-peter.de>"]
categories = ["command-line-utilities"]
louis 🌟 ~/dev/bat $ 
```

</details>

### No flags (implicit style=auto)

- `bat Cargo.toml -r :2 | cat`

Before:

```
[package]
authors = ["David Peter <mail@david-peter.de>"]
categories = ["command-line-utilities"]
description = "A cat(1) clone with wings."
```

After: **(Unchanged)**

### Plain

- `bat Cargo.toml -p -r :4 | cat`

Before:

```
[package]
authors = ["David Peter <mail@david-peter.de>"]
categories = ["command-line-utilities"]
description = "A cat(1) clone with wings."
```

After: **(Unchanged)**

### Line numbers only

- `bat Cargo.toml -n -r :4 | cat` or `bat Cargo.toml --style=numbers -r :4 | cat`

Before: **(As for plain)**

After:

```
   1 [package]
   2 authors = ["David Peter <mail@david-peter.de>"]
   3 categories = ["command-line-utilities"]
   4 description = "A cat(1) clone with wings."
```

### Full style (explicit default)

- `bat Cargo.toml --style=default -r :4 | cat`

Before: **(As for plain)**

After:

```
─────┬──────────────────────────────────────────────────────────────────────────
     │ File: Cargo.toml
─────┼──────────────────────────────────────────────────────────────────────────
   1 │ [package]
   2 │ authors = ["David Peter <mail@david-peter.de>"]
   3 │ categories = ["command-line-utilities"]
   4 │ description = "A cat(1) clone with wings."
─────┴──────────────────────────────────────────────────────────────────────────
```

### DIFF
--- a/assets/manual/bat.1.in
+++ b/assets/manual/bat.1.in
@@ -14,7 +14,8 @@ It also communicates with git(1) to show modifications with respect to the git i
 
 Whenever the output of {{PROJECT_EXECUTABLE}} goes to a non-interactive terminal, i.e. when the
 output is piped into another process or into a file, {{PROJECT_EXECUTABLE}} will act as a drop-in
-replacement for cat(1) and fall back to printing the plain file contents.
+replacement for cat(1) and fall back to printing the plain file contents,
+unless an explicit style is requested.
 
 .SH "OPTIONS"
 General remarks: Command-line options like '-l'/'--language' that take values can be specified as
@@ -131,7 +132,8 @@ always, *never*.
 \fB\-\-decorations\fR <when>
 .IP
 Specify when to use the decorations that have been specified via '\-\-style'. The
-automatic mode only enables decorations if an interactive terminal is detected. Possible
+automatic mode only enables decorations if an interactive terminal is detected. The
+always mode will show decorations even when piping output. Possible
 values: *auto*, never, always.
 .HP
 \fB\-f\fR, \fB\-\-force\-colorization\fR
--- a/doc/long-help.txt
+++ b/doc/long-help.txt
@@ -87,8 +87,8 @@ Options:
 
       --decorations <when>
           Specify when to use the decorations that have been specified via '--style'. The automatic
-          mode only enables decorations if an interactive terminal is detected. Possible values:
-          *auto*, never, always.
+          mode only enables decorations if an interactive terminal is detected. The always mode will
+          show decorations even when piping output. Possible values: *auto*, never, always.
 
   -f, --force-colorization
           Alias for '--decorations=always --color=always'. This is useful if the output of bat is
--- a/src/bin/bat/clap_app.rs
+++ b/src/bin/bat/clap_app.rs
@@ -299,9 +299,10 @@ pub fn build_app(interactive_output: bool) -> Command {
                 .help("When to show the decorations (*auto*, never, always).")
                 .long_help(
                     "Specify when to use the decorations that have been specified \
-                     via '--style'. The automatic mode only enables decorations if \
-                     an interactive terminal is detected. Possible values: *auto*, never, always.",
-                ),
+                    via '--style'. The automatic mode only enables decorations if \
+                    an interactive terminal is detected. The always mode will show \
+                    decorations even when piping output. Possible values: *auto*, never, always.",
+                )
         )
         .arg(
             Arg::new("force-colorization")


lmmx:
I've disabled these two tests on Windows as they are failing the CI (snapshots have one fewer terminal width char?) on 2 of the 3 Windows architectures (so if I changed it to per platform it would still not work) :man_facepalming: 

```
failures:
    piped_output_with_default_style_flag
    piped_output_with_line_numbers_with_header_grid_style_flag
```

- Log https://github.com/sharkdp/bat/actions/runs/18563295925/job/52917493767?pr=3438
- :red_circle: [i686-pc-windows-msvc (windows-2025)](https://github.com/sharkdp/bat/actions/runs/18563295925/job/52917493767?pr=3438#logs)
- :red_circle: x86_64-pc-windows-msvc (windows-2025)
- :green_circle: aarch64-pc-windows-msvc (windows-11-arm)

<details><summary>Click to show full log</summary>

```

failures:

---- piped_output_with_default_style_flag stdout ----

thread 'piped_output_with_default_style_flag' panicked at /rustc/1159e78c4747b02ef996e55082b704c09b970588\library\core\src\ops\function.rs:253:5:
Unexpected stdout, failed diff original var
├── original: ─────┬──────────────────────────────────────────────────────────────────────────
│        │ STDIN
│   ─────┼──────────────────────────────────────────────────────────────────────────
│      1 │ hello
│      2 │ world
│   ─────┴──────────────────────────────────────────────────────────────────────────
├── diff: 
│   --- 	orig
│   +++ 	var
│   @@ -1 +1 @@
│   -─────┬──────────────────────────────────────────────────────────────────────────
│   +─────┬─────────────────────────────────────────────────────────────────────────
│   @@ -3 +3 @@
│   -─────┼──────────────────────────────────────────────────────────────────────────
│   +─────┼─────────────────────────────────────────────────────────────────────────
│   @@ -6 +6 @@
│   -─────┴──────────────────────────────────────────────────────────────────────────
│   +─────┴─────────────────────────────────────────────────────────────────────────
└── var as str: ─────┬─────────────────────────────────────────────────────────────────────────
         │ STDIN
    ─────┼─────────────────────────────────────────────────────────────────────────
       1 │ hello
       2 │ world
    ─────┴─────────────────────────────────────────────────────────────────────────

command=`"D:\\a\\bat\\bat\\target\\x86_64-pc-windows-msvc\\debug\\bat.exe" "--no-config" "--style=default"`
stdin=````
hello
world
```
`
code=0
stdout=```
─────┬─────────────────────────────────────────────────────────────────────────
     │ STDIN
─────┼─────────────────────────────────────────────────────────────────────────
   1 │ hello
   2 │ world
─────┴─────────────────────────────────────────────────────────────────────────
```

stderr=""

---- piped_output_with_line_numbers_with_header_grid_style_flag stdout ----

thread 'piped_output_with_line_numbers_with_header_grid_style_flag' panicked at /rustc/1159e78c4747b02ef996e55082b704c09b970588\library\core\src\ops\function.rs:253:5:
Unexpected stdout, failed diff original var
├── original: ─────┬──────────────────────────────────────────────────────────────────────────
│        │ STDIN
│   ─────┼──────────────────────────────────────────────────────────────────────────
│      1 │ hello
│      2 │ world
│   ─────┴──────────────────────────────────────────────────────────────────────────
├── diff: 
│   --- 	orig
│   +++ 	var
│   @@ -1 +1 @@
│   -─────┬──────────────────────────────────────────────────────────────────────────
│   +─────┬─────────────────────────────────────────────────────────────────────────
│   @@ -3 +3 @@
│   -─────┼──────────────────────────────────────────────────────────────────────────
│   +─────┼─────────────────────────────────────────────────────────────────────────
│   @@ -6 +6 @@
│   -─────┴──────────────────────────────────────────────────────────────────────────
│   +─────┴─────────────────────────────────────────────────────────────────────────
└── var as str: ─────┬─────────────────────────────────────────────────────────────────────────
         │ STDIN
    ─────┼─────────────────────────────────────────────────────────────────────────
       1 │ hello
       2 │ world
    ─────┴─────────────────────────────────────────────────────────────────────────

command=`"D:\\a\\bat\\bat\\target\\x86_64-pc-windows-msvc\\debug\\bat.exe" "--no-config" "--style=header,grid,numbers"`
stdin=````
hello
world
```
`
code=0
stdout=```
─────┬─────────────────────────────────────────────────────────────────────────
     │ STDIN
─────┼─────────────────────────────────────────────────────────────────────────
   1 │ hello
   2 │ world
─────┴─────────────────────────────────────────────────────────────────────────
```

stderr=""
```

</details>

LangLangBart:
> I recently noticed I cannot pipe `bat -n` output anywhere

Isn't that what the `--decorations` flag has been used for ? 

```bash
bat --decorations=always -nr :4 Cargo.toml | cat
   1 [package]
   2 authors = ["David Peter <mail@david-peter.de>"]
   3 categories = ["command-line-utilities"]
   4 description = "A cat(1) clone with wings."
```

lmmx:
> > I recently noticed I cannot pipe `bat -n` output anywhere
> 
> Isn't that what the `--decorations` flag has been used for ?

I meant `bat -n` _with no further flags_ cannot pipe, as discussed in #3316 it is documented as the design that these should pipe, but it wasn't implemented:

> The description on `--style` is misleading. It says "auto: same as 'full', unless the output is piped." and "full: enables all available components (default)".

This is ultimately about nicer _default_ behaviour for piping, as #2983 was, and keeping to the original design/better ergonomics.

LangLangBart:
https://github.com/sharkdp/bat/blob/a730eaae0a17de2c2fd8099471a49e7419db1b54/assets/manual/bat.1.in#L15-L17

https://github.com/sharkdp/bat/blob/a730eaae0a17de2c2fd8099471a49e7419db1b54/assets/manual/bat.1.in#L131-L135

It makes bat's behavior more complex. The rule is no longer "always act like cat
when piped." Instead, it becomes: "act like cat when piped, unless you use
--number (which makes it behave like cat -n) or explicitly set a --style that
isn't plain or auto." While this is arguably more intuitive, it's a departure
from the simple, documented rule.

lmmx:
Ah I hadn’t seen those man page sections before.

> It makes bat's behavior more complex.

I'd say it changes the behaviour rather than making it more complex.

> The rule is no longer "always act like cat when piped."

Yes, that's exactly the point - this was the requested change in #2935 (agreed to be worked on in 2024, did not land in #2983).

> Instead, it becomes: "act like cat when piped, unless you use --number (which makes it behave like cat -n) or explicitly set a --style that isn't plain or auto."

I think this phrasing is unnecessarily verbose. The simpler way to describe it is:

> act like cat when piped, unless you explicitly set a non-`auto` style.

Two clarifications:

- There's no need to mention `plain` in the exceptions, piping with plain style looks the same as cat anyway
- There's no need to single out `--number`, it's just a style alias and is already covered by "explicitly set a style"

Based on the man page you quoted, I thought this was already the intended functionality: that the auto style adapts to piping, while explicit styles like default or numbers would work regardless.

> While this is arguably more intuitive, it's a departure from the simple, documented rule.

To me, intuitive = simple. If someone explicitly requests a style with `-n` or `--style=numbers`, they expect that style when piping.

I find it undesirable for bat to behave like cat when piped, and my impression from the issue tracker is that others agree.

I understood from #2935 and the discussion on the previous PR that this behavior change was desired. If there's still disagreement about whether this should change at all, I'm not sure how to proceed. I took it I was implementing a requested feature.

The key principle: when someone explicitly requests a style, honour it. Default behaviour (no flags) is preserved for backwards compatibility.

I'll update the man page sections to reflect this behavior change:

```diff
@@ -1,3 +1,4 @@
 Whenever the output of {{PROJECT_EXECUTABLE}} goes to a non-interactive terminal, i.e. when the 
 output is piped into another process or into a file, {{PROJECT_EXECUTABLE}} will act as a drop-in 
-replacement for cat(1) and fall back to printing the plain file contents.
+replacement for cat(1) and fall back to printing the plain file contents,
+unless an explicit style is requested.
```

```diff

@@ -2,5 +2,6 @@
 .IP 
 Specify when to use the decorations that have been specified via '\-\-style'. The 
-automatic mode only enables decorations if an interactive terminal is detected. Possible 
-values: *auto*, never, always.
+automatic mode only enables decorations if an interactive terminal is detected. The 
+always mode will show decorations even when piping output. Possible 
+values: *auto*, never, always.
```

lmmx:
- [x] Man page amended accordingly

I then looked for other docs with the same substring to edit likewise

```sh
louis 🌟 ~/dev/bat $ rg -l 'Specify when to use the dec'
src/bin/bat/clap_app.rs
assets/manual/bat.1.in
doc/long-help.txt
tests/syntax-tests/highlighted/Manpage/bat-0.16.man
tests/syntax-tests/source/Manpage/bat-0.16.man
```

- [x] Other docs reviewed:
  - [x] `src/bin/bat/clap_app.rs` amended to produce the same as `docs/long-help.txt`
  - [x] `docs/long-help.txt` amended to match the man page
  - [x] tests/syntax-tests/highlighted/Manpage/bat-0.16.man - man pages for published version 0.16, left as is
  - [x] tests/syntax-tests/source/Manpage/bat-0.16.man - man pages for published version 0.16, left as is

- [x] Docs tests are passing again

The README also mentions cat-like behaviour in two places ([file concatenation](https://github.com/sharkdp/bat/blob/a730eaae0a17de2c2fd8099471a49e7419db1b54/README.md?plain=1#L69-L72) and [xclip](https://github.com/sharkdp/bat/blob/a730eaae0a17de2c2fd8099471a49e7419db1b54/README.md?plain=1#L188-L196) examples), but I'd leave these as is since they still work as shown. The modifier "except if a style is set" could be added, but I think this is sufficiently intuitive/already documented that it would be unnecessary.

:+1: Ready for re-review, taking back out of draft

LangLangBart:
Until now, bat behaved predictably: it showed decorations only in interactive
terminals, stripping them in non‑interactive output unless explicitly forced
(e.g., --decorations=always, --color=always, or --force-colorization).

The proposed change breaks this for some flags. Users would see decorations in
non-interactive output for certain flags but not others. For example, --diff
produces decorations that don't appear in non-interactive terminals, undermining
intuitive flag usage.

I got accustomed to interactive viewing and have learned over the years that
decorations don't carry over unless forced. For the command below, I wouldn't
expect the style to appear in the output:

```bash
BAT_OPTS='--style=header,grid' bat README.md | wc -l
```

If line numbers in piped output is the major reason, users can set the
decorations flag in their config file or use an alias. This solves the issue
without breaking consistency.

- Ref: https://github.com/sharkdp/bat#customization

lmmx:
Thanks for the feedback, to address your points:

The change only affects behaviour when users explicitly pass style flags like `-n` or `--style=numbers` (including when set in `BAT_OPTS`). Default behaviour (no flags/env vars) stays the same, so backward compatibility is preserved. This aligns with the documented behaviour of the `auto` style.

The principle: if someone explicitly asks for a style, honour it. This matches how other Unix tools work. `cat -n` preserves numbers when piped, and `tail -n` with multiple files preserves header decorations. See demos below.

**On `--diff`:** I tried it and since diff isn't a style component, it already uses the default `auto` style behaviour when piped, so no change needed there (see Diff Demo below).

**On `BAT_OPTS`:** If someone explicitly sets a style via `BAT_OPTS`, it's equivalent to setting it on the command line. They can add `--decorations=always` if they want to ensure decorations pipe.

The logic and compatibility were validated in previous reviews. The implementation is consistent with both the documentation and existing `auto` behaviour. I'm confident this is the right approach, but I'm open to addressing specific technical concerns if there are any.

> [!NOTE]  
> Rebased on master branch without changes

# Demos

## Diff Demo

Behaviour is unchanged:

```sh
louis 🌟 ~/dev/bat $ bat --diff README.md
───────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: README.md
───────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 911   │ 
 912   │ ## Project goals and alternatives
 913 ~ │ hello
 914   │ `bat` tries to achieve the following goals:
 915   │ 
───────┴────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
louis 🌟 ~/dev/bat $ bat --diff README.md | wc -l
5
louis 🌟 ~/dev/bat $ ./target/release/bat --diff README.md | wc -l
5
```

## Cat Demo

Comparing `bat -n` behavior with `cat -n` when piping:

```sh
louis 🌟 ~/dev/bat $ echo "hello" | cat | wc -c
6
louis 🌟 ~/dev/bat $ echo "hello" | cat -n | wc -c
13
louis 🌟 ~/dev/bat $ echo "hello" | bat | wc -c
6
louis 🌟 ~/dev/bat $ echo "hello" | bat -n | wc -c
6
```

With this PR, `bat -n` behaves like `cat -n` (preserves line numbers when piped):

```sh
louis 🌟 ~/dev/bat $ echo "hello" | ./target/release/bat -n | wc -c
11
```

## Tail Demo

Headers are preserved when piped (similar to bat's header/grid decorations):

```sh
louis 🌟 ~/dev/bat $ tail -n999 <(echo "foo") <(echo "bar")
==> /dev/fd/63 <==
foo

==> /dev/fd/62 <==
bar
louis 🌟 ~/dev/bat $ tail -n999 <(echo "foo") <(echo "bar") | cat
==> /dev/fd/63 <==
foo

==> /dev/fd/62 <==
bar
```

keith-hall:
@LangLangBart thanks for sharing your concerns. I understand that these changes may break behavior you are used to. But, as we have had a number of people open issues about it, and the wording in the documentation generally suggesting it should work the way it does in this PR, along with the `cat -n` piping behavior, I'm going to merge it. If we get lots of complaints, we can certainly revisit it again. Thanks for your understanding.

keith-hall:
Just an FYI, apparently some workflows (aliasing `cat` with specific arguments) have been disrupted due to this change: https://github.com/sharkdp/bat/issues/3445
Maybe you can help think of the best way forward please @lmmx. Ideally we want to make everyone happy :sweat_smile:

lmmx:
Sorry @keith-hall I was reviewing bug reports in here this morning and did not spot this 🤦‍♂️ Taking a look now

lmmx:
> Maybe you can help think of the best way forward please @lmmx.

That issue seems to be addressed already, not sure what the ask is here @keith-hall but happy to think about any further accomodations. Were there any other reports or just this one so far?

That `cat <&…` pattern can be seen elsewhere ([source](https://grep.app/search?f.lang=Shell&q=cat+%3C%26)) and so can `command cat <&…` ([source](https://grep.app/search?f.lang=Shell&q=command+cat+%3C%26))

I personally think aliasing cat **with a style** is a bit radical, and I think it’s fair to expect that to not be smooth sailing!

It’s much more common if you look at people’s shared dotfiles for them to do it with `--style=plain` or just `bat` alone([source](https://grep.app/search?f.lang=Shell&q=alias+cat%3D%27bat+-)), neither of which would be affected.

keith-hall:
> That issue seems to be addressed already, not sure what the ask is here @keith-hall but happy to think about any further accomodations. Were there any other reports or just this one so far?

Just this one so far that I have seen. I guess I just want us to be prepared in case more come in. I completely agree about aliasing `cat` 😉 but as it is mentioned in the Readme, we kinda have to support it and try to behave in a way that doesn't break too much...

I was wondering whether it would be safer to only fix `-n` specifically when it is on the command line (ignoring env vars and config) and in other cases, update the documentation to refer to `--decorations`. But that also complicates things from a user's perspective - having some arguments work differently when on the command line would be highly unintuitive. So I don't really like that solution 😅 

Probably we can leave things as they are for now. Thanks for checking common usage 👍

lokesh-balla:
> I personally think aliasing cat with a style is a bit radical, and I think it’s fair to expect that to not be smooth sailing!

I do understand that aliasing cat is bound to break sometimes but in this case I don't see why it is considered radical though. 

The way I see it the decorations behave in a confusing way now.

```
# displays with decorations but does not carry the decorations when piped to something else
bat --paging="never" filename.txt
```

```
# displays with decorations and carries the decorations when piped to something else
export BAT_STYLE="header-filename,header-filesize,grid"
bat --paging="never" filename.txt
or
bat --style="header-filename,header-filesize,grid" --paging="never" filename.txt
```

I always assumed not carrying decorations was deliberate choice to avoid issues when piped to something else but if that is not the case then may be we should unify behaviour in the above case as well.

LangLangBart:
> Probably we can leave things as they are for now.

is @sharkdp still around ?

lmmx:
> > I personally think aliasing cat with a style is a bit radical, and I think it's fair to expect that to not be smooth sailing!
> 
> I do understand that aliasing cat is bound to break sometimes but in this case I don't see why it is considered radical though.

Perhaps "radical" was too strong - I meant that using bat as a "drop-in replacement" (as quoted in the man page) refers to unconfigured behaviour, whilst custom style configurations are a different use case. The PR preserves default (unconfigured) behaviour; extending that guarantee to all possible style configurations would require different design choices.

Issue #2944, opened 2024-04-23
Allow --line-range to count from the back of the file to act similar to "tail"


PlasmaHH:
For python users it would probably best if -r acts somewhat like the python syntax...

So for negative numbers it could count from backwards, so "tail -n 5" would be "bat -r -5:" or so and then one could display the first and last portion of a file like "bat -r :5 -r -5:"

ajesipow:
I would also welcome such a feature and would be happy to contribute a PR for it.

But I think it's not a trivial change and would likely have performance implications that may not be desired.

I see two options:
1. Doing a first quick pass on the input to determine the max line count, then calculating the absolute values in `LineRange` for the negative relative line number accordingly. A second pass would handle lines just as today. No buffering needed.

2. Buffering the input for multiple lines in `Controller::print_file_ranges` instead of the single-line `line_buffer`. The buffer size would need to be equivalent to how far from the end of the input the user wants the output (which could be very large of course - resulting in potentially unexpected memory usage). `LineRange` would also need adapting to represent the relative negative value somehow and this would most likely be a breaking change. 

Also, today a range with a relative negative value such as `10:-3` is already valid and translated into an absolute line range of `7:10`. With the feature discussed here, new users may expect this to translate to "from line 10 to the 3rd to last line" instead.

Alternatively, to not break the existing logic, relative negative numbers may only occur by in an open-ended range, i.e. `-3:` or `:-3`.

atimofeev:
I made this wrapper script, which utilizes relative negative value functionality mentioned above:
```bash
#!/usr/bin/env bash
file="$1"
show="${2:-10}"
lines=$(wc -l < "$file")
range="$lines:-$((show - 1))"

exec bat "$file" --color=always --style=plain --paging=never --line-range "$range"
```
The second argument is optional and by default shows 10 lines.
It even works if a file has fewer lines than a show value.

PlasmaHH:
> > But I think it's not a trivial change and would likely have performance implications that may not be desired.
> 
> Well, there is tail(1), so whatever it does should be applicable here :-)
> 
> > I see two options:
> 
> Note that these options are not equivalent in presence of non-regular files. IOWs, they are not alternatives, but rather both should be implemented with the decision made depending on whether the target file is seekable.

I am not quite sure what being seekable or not has to do with anything, after all to determine the number of lines in usual text files one has to go through them.


[...]

PR #3442, opened 2025-10-19
fix: allow hyphen values to `-r`/`--line-range`

lmmx:
- via #2944

This PR adds `.allow_leading_hyphen(true)` to the line range argument parser to support the `-r -10:` syntax (with space-separated arguments).

While the functionality for negative line ranges was already implemented to satisfy #2944, the argument parser wasn't configured to accept them when passed as separate arguments like `-r -10:`. The `=` syntax (`-r=-10:` and `--line-range=-10:`) worked fine, which is why the existing tests didn't catch this: they all use the equals format.

This change allows both formats to work as expected. Added tests using the space-separated syntax via `.arg("-r").arg("-10:")` to cover this case.

- [x] Tests
  - [x] Case where `-r` and `-2:` are passed separately (matching an existing `=`-separated test case)
- [x] Docs: `docs/long-help.txt`
  - Let’s add 1 example of using negative N `-r -10:` to the help text
  - <strike>Simple docstring edit to say both/neither can be negative?</strike> No need to change short help https://github.com/sharkdp/bat/blob/c734087e1d15fb730d7653389c841fd9e3016556/doc/short-help.txt#L57-L58
- [x] Changelog entry
    `- Fix negative values of N not being parsed in <N:M> line ranges without `=` flag value separator, see #3442 (@lmmx)`

## Demo

Before:

```sh
louis 🚶 ~/dev/bat $ echo -e "hello\nworld" | bat -r -1:
error: unexpected argument '-1' found

  tip: to pass '-1' as a value, use '-- -1'

Usage: bat [OPTIONS] [FILE]...
       bat <COMMAND>

For more information, try '--help'.
```

After:

```sh
louis 🚶 ~/dev/bat $ echo -e "hello\nworld" | ./target/debug/bat -r -1: | cat
world
```

**Status:** :shipit: Ready to merge

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ## Bugfixes
 
+- Fix negative values of N not being parsed in <N:M> line ranges without `=` flag value separator, see #3442 (@lmmx)
+
 ## Other
 
 ## Syntaxes
--- a/doc/long-help.txt
+++ b/doc/long-help.txt
@@ -194,6 +194,7 @@ Options:
             '--line-range :40' prints lines 1 to 40
             '--line-range 40:' prints lines 40 to the end of the file
             '--line-range 40' only prints line 40
+            '--line-range -10:' prints the last 10 lines
             '--line-range 30:+10' prints lines 30 to 40
             '--line-range 35::5' prints lines 30 to 40 (line 35 with 5 lines of context)
             '--line-range 30:40:2' prints lines 28 to 42 (range 30-40 with 2 lines of context)
--- a/src/bin/bat/clap_app.rs
+++ b/src/bin/bat/clap_app.rs
@@ -519,6 +519,7 @@ pub fn build_app(interactive_output: bool) -> Command {
                 .short('r')
                 .action(ArgAction::Append)
                 .value_name("N:M")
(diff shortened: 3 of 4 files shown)


CHANGELOG.md
```diff
@@ -37,6 +37,7 @@
 - Add missing shell completions, see #3411 (@keith-hall)
 - Execute help/version/diagnostic commands even with invalid config/arguments present, see #3414 (@keith-hall)
 - Fixed line numbers (`-n`) and style components not printing when piping output, see issue #2935 and PR #3438 (@lmmx)
+- Fix negative values of N not being parsed in <N:M> line ranges without `=` flag value separator, see #3442 (@lmmx)
```
keith-hall: I have made a release just as you started work on this. Could I ask you to move this to the unreleased section please?

lmmx: Sure thing!

lmmx: Fixed in 52a792d

keith-hall (approved):

keith-hall:
Thanks! This may be one of those little usability improvements which will be worth making a new patch release for. Let's see if we get any bugs reported in the near future and we can decide then 🙂

PR #3443, opened 2025-10-19
docs: Improve README on less pager options

injust:
Adding the long names for less options makes the README easier to skim, without having to jump to the first paragraph of https://github.com/sharkdp/bat/blob/master/README.md#using-less-as-a-pager to see what the equivalent long option is called.

Also clarify that `-X` is not added on modern versions of less.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Fix negative values of N not being parsed in <N:M> line ranges without `=` flag value separator, see #3442 (@lmmx)
 
 ## Other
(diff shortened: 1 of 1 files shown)


lmmx:
LGTM, the CI wants a changelog entry added [here](https://github.com/sharkdp/bat/blob/593841a40dadde77ffa194ba4ecb6466e2d6a02e/CHANGELOG.md), something like

```diff
 # unreleased
 
 ## Features
 
 ## Bugfixes
 
 ## Other
+- Updated documentation for pager options long-form names, see #3443 (@injust)
 
```

PR #3447, opened 2025-10-20
fix: prevent Ctrl+C from killing bat when paging without `-K`

### LINKED ISSUE #3444: ctrl-c still exits less, even when bat isn't passing `--quit-on-intr`

1. `BAT_PAGER='less -R' bat foo`
2. <kbd>ctrl</kbd><kbd>c</kbd>

**What happens?**

less exits.

less shouldn't exit because setting $BAT_PAGER should prevent bat from passing `--quit-on-intr` to less.

Homebrew

---

#### Software version

bat 0.25.0

#### Operating system

- OS: macOS 15.7.1 Sequoia
- Kernel: 24.6.0

#### Command-line

```bash
bat --diagnostic foo
```

#### Environment variables

```bash
BAT_CACHE_PATH=<not set>
BAT_CONFIG_PATH=<not set>
BAT_OPTS=<not set>
BAT_PAGER='less -R'
BAT_PAGING=<not set>
BAT_STYLE=<not set>
BAT_TABS=<not set>
BAT_THEME=<not set>
COLORTERM=truecolor
LANG=en_CA.UTF-8
LC_ALL=<not set>
LESS='--quit-if-one-screen --ignore-case --jump-target=.1 --LONG-PROMPT --quiet --RAW-CONTROL-CHARS --incsearch --no-vbell'
MANPAGER='sh -c '\''awk '\''\'\'''\''{ gsub(/\x1B\[[0-9;]*m/, "", $0); gsub(/.\x08/, "", $0); print }'\''\'\'''\'' | bat -p -lman'\'''
NO_COLOR=<not set>
PAGER=<not set>
SHELL=/usr/local/bin/fish
TERM=xterm-ghostty
XDG_CACHE_HOME=<not set>
XDG_CONFIG_HOME=<not set>
```

#### System Config file

#### Config file

```
--theme-light="Catppuccin Latte"
--theme-dark="Catppuccin Mocha"
```

#### Custom assets metadata

```
bat_version: 0.25.0
creation_time:
  secs_since_epoch: 1760872544
  nanos_since_epoch: 104422000
```

#### Custom assets

- metadata.yaml, 97 bytes
- syntaxes.bin, 969578 bytes
- themes.bin, 58252 bytes

#### Compile time information

- Profile: release
- Family: unix
- OS: macos
- Architecture: x86_64
- Pointer width: 64
- Endian: little
- Host: x86_64-apple-darwin

#### Less version

```
> less --version 
less 685 (PCRE2 regular expressions)
Copyright (C) 1984-2025  Mark Nudelman

```

lmmx:
- Fixes #3444 

## Bug

When `BAT_PAGER='less -R'` is set (without `-K`), pressing Ctrl-C still exits `less` immediately instead of canceling the current operation and staying open. This happens because bat receives SIGINT and begins teardown, which causes `less` to lose access to the TTY and exit with EIO.

## Solution
Use `signal-hook` to temporarily ignore SIGINT in bat while the pager is active. This allows `less` to handle Ctrl-C independently. When the pager exits, the signal handler is automatically restored via RAII.

Specifically I registered a dummy flag that's never checked. This prevents the default SIGINT handler from terminating the process, effectively ignoring the signal.

## Changes
- Added `IgnoreSigint` RAII guard that ignores SIGINT while alive
- Wrapped `Child` in `PagerProc` struct to carry the signal guard
- Signal guard is created when spawning external pagers and dropped when pager exits

Now Ctrl-C behaves correctly in `less`:
- Without `-K`: cancels search/operation, stays open
- With `-K`: exits immediately (existing behavior preserved)

## Demo

**Before:** Ctrl C closes the pager but neither does the terminal fully regain control. You have to press Ctrl + C again

```sh
for i in {0..100}; do echo $i; done | BAT_PAGER='less -R' bat
```

**After:** Ctrl + C sends SIGINT and bat actively ignores it, hence does not kill the pager process via teardown

```sh
for i in {0..100}; do echo $i; done | BAT_PAGER='less -R' ./target/debug/bat
```

## Status

- [x] Changelog entry applied
- [x] No docs, this is a fulfillment of already documented behaviour in the pager section of the README
- [x] Ready for review: yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ## Bugfixes
 
 - Fix negative values of N not being parsed in <N:M> line ranges without `=` flag value separator, see #3442 (@lmmx)
+- Ignore SIGINT during paging to fix bug where less exited without `-K` arg, see issue #3444 and PR #3447 (@lmmx)
 
 ## Other
 - Improve README documentation on pager options passed to less, see #3443 (@injust)
--- a/src/output.rs
+++ b/src/output.rs
@@ -13,6 +13,36 @@ use crate::paging::PagingMode;
 #[cfg(feature = "paging")]
 use crate::wrapping::WrappingMode;
 
+#[cfg(all(unix, feature = "paging"))]
+use std::sync::atomic::AtomicBool;
+#[cfg(all(unix, feature = "paging"))]
+use std::sync::Arc;
+#[cfg(all(unix, feature = "paging"))]
+struct IgnoreSigint {
+    _handle: signal_hook::SigId,
+}
+
+#[cfg(all(unix, feature = "paging"))]
+impl IgnoreSigint {
+    fn new() -> Self {
+        // No-op SIGINT handler prevents pager process termination
+        let handle = signal_hook::flag::register(
+            signal_hook::consts::SIGINT,
+            Arc::new(AtomicBool::new(false)),
+        )
+        .expect("failed to ignore SIGINT");
+
+        Self { _handle: handle }
+    }
+}
+
+#[cfg(feature = "paging")]
+pub struct PagerProc {
+    child: Child,
+    #[cfg(unix)]
+    _sigint_guard: IgnoreSigint,
+}
+
 #[cfg(feature = "paging")]
 pub struct BuiltinPager {
     pager: minus::Pager,
@@ -50,10 +80,9 @@ enum SingleScreenAction {
     Nothing,
 }
 
-#[derive(Debug)]
 pub enum OutputType {
     #[cfg(feature = "paging")]
-    Pager(Child),
+    Pager(PagerProc),
     #[cfg(feature = "paging")]
     BuiltinPager(BuiltinPager),
     Stdout(io::Stdout),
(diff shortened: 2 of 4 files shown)


src/output.rs
```diff
@@ -50,10 +80,9 @@ enum SingleScreenAction {
     Nothing,
 }
 
-#[derive(Debug)]
 pub enum OutputType {
     #[cfg(feature = "paging")]
-    Pager(Child),
+    Pager(PagerProc),
```
keith-hall, line 85: Note: this changes the public API (for bat as a library), so to be semver compatible, the next release including this cannot be a patch release

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ## Bugfixes
 
 - Fix negative values of N not being parsed in <N:M> line ranges without `=` flag value separator, see #3442 (@lmmx)
+- Fix bug where less exitted without `-K` arg by ignoring SIGINT during paging, see issue #3444 and PR #3447 (@lmmx)
```
keith-hall: I think this entry would be easier to understand if we rephrase it slightly :slightly_smiling_face: 
```suggestion
- Ignore SIGINT during paging to fix bug where less exited without `-K` arg, see issue #3444 and PR #3447 (@lmmx)
```

src/output.rs
lmmx: hmm, ok. Let me know if I can alleviate any of the friction to do with that. Also: if it’s not something you’re wholly behind don’t feel obliged to merge it.

I feel like it solves the bug in terms of UX but not correct way (i.e. bat stays alive). Whether there are implications to bat staying alive I’m not quite sure though. Unlike most programs bat isn’t going to be producing a stream of data (if I follow it correctly?)

I am not fully certain how to think about the case of `bat` reading a large file into a pager, I wouldn’t imagine that to be a runaway process you might want to Ctrl+C to stop and then stay in the pager to read the partial result of. But I might be wrong.

If that view is right though, bat living on and just ignoring SIGINT seems “basically fine”

PR #3448, opened 2025-10-20
Use `-o fullquote` and `-o noquote` for more robust escaping in Bash completions

akinomyoga:
This PR adds improvements to the shell function `__bat_escape_completions()`, which is used to quote the non-filename completion candidates.

- The current approach of `__bat_escape_completions` still gets unwanted quoting when *the quoted result* accidentally matches a filename visible from the current working directory. To suppress this behavior of Readline, we should actually specify `compopt -o noquote` (which is available in Bash >= 4.3).
- Furthermore, the feature `__bat_escape_completions` tries to provide is actually implemented in Bash 5.3 as `compopt -o fullquote`, so in Bash >= 5.3, we can simply use it instead of quoting the candidates ourselves.
- This PR also includes the fixes for indentation. The indentation is basically made with tabs in this file, but several lines are indented by a random number of spaces. The first commit fixes this.

I've tested the behavior with Bash 5.2 (where `noquote` is used) and Bash 5.3 (where `fullquote` is used).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ## Other
 
+- Use more robust approach to escaping in Bash completions, see #3448 (@akinomyoga)
+
 ## Syntaxes
 
 ## Themes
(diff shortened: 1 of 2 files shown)


Jai-JAP:
Tested, works well with the modifications modified.

PR #3453, opened 2025-10-23
Add syntax mapping for gradle.kts file extension

nicolaic:
Gradle has for quite a while now supported .kts build scripts, but bat does not highlight those files as expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
 
 ## Syntaxes
 
+- Add syntax mapping for `gradle.kts` build scripts #3453 (@nicolaic)
+
 ## Themes
 
 ## `bat` as a library


assets/syntaxes/02_Extra/Kotlin.sublime-syntax
```diff
@@ -5,6 +5,7 @@ name: Kotlin
 file_extensions:
   - kt
   - kts
+  - gradle.kts
```
keith-hall, line 8: I'm a little confused. The `kts` extension is already mapped to Kotlin. And taking an example from https://docs.gradle.org/current/userguide/kotlin_dsl.html, calling it `build.gradle.kts`  and using `bat` (without the change in this PR) to display it renders the same as `bat -l Kotlin build.gradle.kts` so I'm not sure what this PR solves?

nicolaic: Hmm, it seems to work fine in this case. I apologize, I was using bat integration in another tool, and my .kt files were displaying fine, but not my .gradle.kts files and thought I was a missing mapping here, but I now see my .kts files doesn't format either. This is not an upstream issue it appears, and I'll file an issue with the integrating tool.

PR #3457, opened 2025-10-28
fix: `--list-themes` pagination

abhinavcool42:
### Bug
`bat --list-themes --pager=builtin` and `bat --list-themes --pager=less` hang after printing the first line (e.g., “Theme: 1337”). In iTerm2 the “:” prompt is hidden until resize/CR, input is in cooked mode, and display appears off-by-one.
### Repro:
`bat --list-themes --pager=builtin`
`bat --list-themes --pager=less`
### Root cause
- controller was creating its own pager even when an OutputHandle was already provided by the caller which led to nested pagers and TTY state confusion, manifesting as a hang and terminal not restoring properly.
- also treated stdout as “safe” when a pager was already attached, allowing clircle’s stdout identifier to be created in a pager context, which can block and contribute to the bad terminal state.
### Changes
Only create `OutputType` when `output_handle.is_none()`, preventing a second pager from being started.
added `attached_to_pager` that is true when:

- an external `OutputHandle` is provided, or
- an internally created `OutputType` is a pager.

Route errors to the writer when `attached_to_pager`; otherwise use stderr.
Set `stdout_identifier = None` when `attached_to_pager` (and on Windows); otherwise use `clircle::Identifier::stdout()`.


keith-hall:
Thanks!

In my testing, when using this branch and doing `--list-themes` with `less` as the pager (i.e. the default setup), pressing <kbd>Page Down</kbd> gives some weird output (`^[[6~`) instead of scrolling down... Any ideas please?
image

abhinavdiwakar42:
It was caused by launching less with “quit-if-one-screen” (-F) while streaming the list-themes output in chunks. Early on, less may think the content fits on one screen and won’t enter interactive mode yet, so Page Down ends up echoing the escape sequence `^[[6~`. This was confirmed by running the command `bat --list-themes --pager="less -R"` which avoids `-F` and works correctly. Changed the `config.paging_mode` when spawning pager to `PagingMode::Always` so it always stays interactive.
Also we were still adding the `--no-init` when we were not using `--quit-if-one-screen`. changed it to only add `--no-init` when actually using it.

keith-hall:
I think we should respect the user's choices when they have  `--paging=never` etc configured in the config file/env-var or specified on the command line

abhinavdiwakar42:
Agreed. The updated change buffers `list-themes` first with the existing config file, so we don’t force paging and user's choices are honored. Verified with `default`, `--pager=less`, and `--paging=never`.

Issue #1053, opened 2020-06-05
Brain storming: a builtin pager for 'bat'


sharkdp:
Let's use this ticket to discuss some ideas and possible upsides/downsides of writing our own custom pager library that we could use instead of calling an external `$PAGER` process (or, essentially, `less`).

A few advantages if we would use a builtin pager in `bat`:
- Have one consistent pager experience across all operating systems. No need to call external programs that might not be available. No special handling / bug fixing for different versions of `less`. Consistent "quit if one screen" handling, and much more (#860, #1000, #717, #710, #681, #520, #415, #312, #271, #519, #376, #887, #943, #838, …)
- Handling of terminal resize events. This would be my personal main motivation. (#294, #322, #921, #1009)
- Open editor with a key press  (#358, #464, #528, #697, #1051)
- Possibility to always draw the header at the top of the screen, even when scrolling down (#908)
- Builtin `--follow` support (#457, #808)
- Special handling for multiple files? (press 'n' to go to the next file)
- Error handling: we could print errors/warnings in the footer of the pager, instead of in the background (#946)
- Better 'tab' handling (#980, #894)
- Handling of ANSI hyperlinks (#961)
- Disable style components via keypress. For example: switch off the side bar with a keypress - for easy copy&paste.

General advantages of a "pager" as a library (not just for `bat`):

- Make everything easily configurable: custom header and footer messages, keybindings
- Bidirectional communication: send events like "terminal resize" or key presses back to the main application to trigger a redraw or other actions.

The obvious downsides are: (1) this will be a lot of work :smile: and (2) I'm sure there will be TONS of bug reports regarding obscure terminal emulators and operating systems where things are not working as expected. Which results in even more work.

One way to circumvent (2) would be to make the builtin pager an opt-in feature, at least in the beginning.

I have actually started to implement a cross-platform pager-as-a-library like this in a private repository. The first results look promising :smile: 
![out](https://user-images.githubusercontent.com/4209276/83922826-552f3b80-a781-11ea-9da5-ef6616f9ae80.gif)

I'd be happy to hear about your thoughts and ideas!

eth-p:
I'm 100% in support for all the advantages this would provide.

> The obvious downsides are:
> * this will be a lot of work :smile:

It would be, but the payoff would be amazing. All those useful features would make `bat` even better than it currently is.

> One way to circumvent the [volume of bug reports] would be to make this an opt-in feature, at least in the beginning.

[...]

PR #3478, opened 2025-11-05
feat: add paging to '-h' and '--help'

### LINKED ISSUE #1587: Automatically use pager on bat (--list-themes|-h|--help)

MuntasirSZN:
Fixes #1587

continuation of #3432


MuntasirSZN:
@keith-hall your review

MuntasirSZN:
@keith-hall a weird thing I discovered, help uses pager regardless if the help takes full term screen or not. I think it should just print it as its smaller then term.

MuntasirSZN:
(I use --pager=builtin in bat config)

keith-hall:
This is a known bug with the minus integration at the moment: https://github.com/sharkdp/bat/issues/1053#issuecomment-3469835425

PR #3484, opened 2025-11-13
Update quadlet syntax mapping to include `*.{build,pod}` files

cyqsimon:
Some time between #2866 and now, two new kinds of quadlets `*.{build,pod}` have been added. See [`man quadlet`](https://manned.org/man/quadlet).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
 
 ## Syntaxes
 
+- Update quadlet syntax mapping to include *.{build,pod} files #3484 (@cyqsimon)
+
 ## Themes
 
 ## `bat` as a library


xduugu:
Since podman 5.7.0 (released two days ago), there is also `.artifact`.

https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html#artifact-units-artifact

cyqsimon:
> Since podman 5.7.0 (released two days ago), there is also `.artifact`.
> 
> https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html#artifact-units-artifact

That's cool; I didn't know that. Thanks for sharing.

I see that it's still marked as experimental though, so perhaps we should wait for stabilization?

xduugu:
Yes, it's true that it is marked as experimental, but, as I understand it, this only applies to the options available to this unit type. The file syntax will stay the same that it is used for the other quadlet units.

See this note on the [man page](https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html#artifact-units-artifact:~:text=This%20unit,subject%20to%20change.):

> This unit is considered experimental and still in development. Inputs, options, and outputs are all subject to change.

PR #3494, opened 2025-11-23
Add colors to --help

starsep:
I used [clap-cargo](https://github.com/crate-ci/clap-cargo) as it's easy to use and provides sensible default styles.
User can disable colors via `NO_COLOR=1` environment variable

| before | after | `NO_COLOR=1` |
| - | - | - |
| bat-before | bat-after | bat-no-color |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Improve README documentation on pager options passed to less, see #3443 (@injust)
 
 - Use more robust approach to escaping in Bash completions, see #3448 (@akinomyoga)
+- Add colors to `--help` command, see #3494 (@starsep)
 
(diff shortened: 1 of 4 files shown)


keith-hall:
Thanks for your contribution. I think we will get more consistent colors (which match configured themes, and the highlighting used elsewhere) if we go with this approach instead though... https://github.com/sharkdp/bat/pull/3478/files#diff-ed435662f3e5638b7130fc480d825fdc8b6f85945b63c96c96cd52f026d40042R138

starsep:
Makes sense! I have seen that PR but I haven't realized it also adds color to `--help`

Closing this PR as it will be handled anyway :sweat_smile:

PR #3495, opened 2025-11-27
Update PowerShell completions for compatibility with PowerShell v5.1

### LINKED ISSUE #3491: Powershell 5.1 Autocomplete Support
Consider this line isn't supported in 5.1, namely the `??`. Consider reimplementing for 5.1.
https://github.com/sharkdp/bat/blob/1e4a4b765f4a1cad74f3a4204112268f3b303f2f/assets/completions/_bat.ps1.in#L52

keith-hall:
Modified the PowerShell completion script to replace the null-coalescing operator (`??` - a v7 feature) with a conditional statement compatible with PowerShell v5.1. The script now uses an if-else check for null descriptions in the language completion section.

Fixes #3491

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 
 ## Other
 - Improve README documentation on pager options passed to less, see #3443 (@injust)
-
+- Make PowerShell completions compatible with PowerShell v5.1, see #3495 (@keith-hall)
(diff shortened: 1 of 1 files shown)


michaelblyons:
Do you folks want to switch to `SublimeText/PowerShell` from `EditorSyntax`?

Issue #3053, opened 2024-07-30
bat as manpager


ukiyocode:
This is on Linux Mint Willma, with bat installed through homebrew.
When I use the: MANPAGER="sh -c 'col -b | bat -p -l man'" option as described in readme i get weird color codes like: 1mSYNOPSIS0m, but when I simply use MANPAGER="bat -p -l man" everything works fine.

ukiyocode:
well mostly fine it seems to add [0m22m or [0m24m in front of left bracket (, but other than that formatting and colors are preserved

anurag-roy:
Check out [this comment](https://github.com/sharkdp/bat/issues/3025#issuecomment-2212452233) and the whole issue, that might solve your issue.

eth-p:
Additional context for why this is happening:

https://github.com/sharkdp/bat/issues/2568#issuecomment-1938074324
https://github.com/sharkdp/bat/issues/2568#issuecomment-1938232553

In summary, newer versions of `groff` changed to emit ANSI escape sequences by default, and `col` doesn't handle them properly. The following will work for both older and newer versions of `groff`:

```bash
export MANPAGER="sh -c 'sed -u -e \"s/\\x1B\[[0-9;]*m//g; s/.\\x08//g\" | bat -p -lman'"
```

ukiyocode:
That MANROFFOPT='-c' thing doesn't work for me, but eth-p's suggestion works perfectly, thanks :)

smprather:
@eth-p worked for me too. FWIW, I'm on an RHEL8 system with my own terminfo created specifically for MS Terminal (undercurl support, etc).

Issue #3025, opened 2024-07-01
Syntax highlighting doesn't work properly for fish man pages


injust:
**What steps will reproduce the bug?**

1. `set -Ux MANPAGER "sh -c 'col -bx | bat -l man -p'"`
2. `type --help`, `fish_add_path --help`, etc.

**What happens?**
Anything that isn't indented is syntax highlighted orange, ignoring any actual formatting.

**What did you expect to happen instead?**
I couldn't find anything suggesting that man page sections have to be indented, so the syntax highlighting should be able to tolerate and correctly colorize fish's non-indented man page format.

**How did you install `bat`?**

Homebrew

---

#### Software version

bat 0.24.0

#### Operating system

macOS 14.5 (Darwin 23.5.0)

#### Command-line

```bash
bat --diagnostic 
```

#### Environment variables

```bash
SHELL=/usr/local/bin/fish
PAGER=<not set>
LESS=<not set>
LANG=en_CA.UTF-8
LC_ALL=<not set>
BAT_PAGER=<not set>
BAT_PAGING=<not set>
BAT_CACHE_PATH=<not set>
BAT_CONFIG_PATH=<not set>
BAT_OPTS=<not set>
BAT_STYLE=<not set>
BAT_TABS=<not set>
BAT_THEME=<not set>
XDG_CONFIG_HOME=<not set>
XDG_CACHE_HOME=<not set>
COLORTERM=truecolor
NO_COLOR=<not set>
MANPAGER='sh -c '\''col -bx | bat --language=man --style=plain'\'''
```

#### System Config file

Could not read contents of '/etc/bat/config': No such file or directory (os error 2).

#### Config file

Could not read contents of '/Users/jsu/.config/bat/config': No such file or directory (os error 2).

#### Custom assets metadata

```
bat_version: 0.24.0
creation_time:
  secs_since_epoch: 1719638264
  nanos_since_epoch: 856140000
```

#### Custom assets

- metadata.yaml, 97 bytes
- syntaxes.bin, 915557 bytes
- themes.bin, 57307 bytes

#### Compile time information

- Profile: release
- Target triple: x86_64-apple-darwin
- Family: unix
- OS: macos
- Architecture: x86_64
- Pointer width: 64
- Endian: little
- CPU features: cmpxchg16b,fxsr,popcnt,sse,sse2,sse3,sse4.1,sse4.2,ssse3
- Host: x86_64-apple-darwin

#### Less version

```
> less --version 
[...]

Issue #2568, opened 2023-05-12
Similar problem for me on Kitty (Fedora 36 on Sway).


od-cea:
Similar problem for me on Kitty (Fedora 36 on Sway).

```
1mNAME0m
       ls - list directory contents

1mSYNOPSIS0m
       1mls 22m[4mOPTION24m]... [4mFILE24m]...

1mDESCRIPTION0m
       List  information about the FILEs (the current directory by default).  Sort entries alphabetically if none of 1m-cftuvSUX 22mnor 1m--sort 22mis
       specified.
```

_Originally posted by @SaElAh in https://github.com/sharkdp/bat/issues/2479#issuecomment-1439959164_

od-cea:
For the record, I had this exact problem and I've been able to mitigate it with [colorzed-logs](https://github.com/kilobyte/colorized-logs) packages and it's command ansi2txt: `export MANPAGER="sh -c 'ansi2txt | col -bx | bat -l man -p'"`.

It would be great if bat could have an option to pre-filter out all color ANSI codes.

eth-p:
> It would be great if bat could have an option to pre-filter out all color ANSI codes.

I could actually make a PR for this once #2544 is merged. We would have the machinery needed to find (and thus discard) all the types of escape sequences except C0.

eth-p:
> It would be great if bat could have an option to pre-filter out all color ANSI codes.

I could actually make a PR for this once #2544 is merged. We would have the machinery needed to find (and thus discard) all the types of escape sequences except C0.

aus-hawk:
Running with `MANROFFOPT="-c"` [as suggested by the README](https://github.com/sharkdp/bat#man) solved this issue for me.

<details><summary>With <code>MANPAGER="sh -c 'col -bx | bat -l man -p --theme=Nord'"</code></summary>
<p>

![image](https://github.com/sharkdp/bat/assets/14896328/56ad5686-004a-49e7-93eb-5721425c2798)

</p>
</details>

<details><summary>With the previous MANPAGER and <code>MANROFFOPT="-c"</code></summary>
<p>

![image](https://github.com/sharkdp/bat/assets/14896328/6e33d34a-a326-4568-8e9d-16a6a6c91cc4)

</p>
</details>

I think this is a man problem, not a bat problem.

eth-p:
[...]

Issue #2479, opened 2023-02-18
Some utf8 output from man appears as escaped bytes


grodin:
**What steps will reproduce the bug?**

1. execute `man bat` (or other manpages) with `MANPAGER="sh -c 'col -bx  | bat -l man -p'"` on a terminal with a width small enough that man hyphenates some words.

**What happens?**

man output such as 

```
It also communicates with git(1) to show  modifications  with  re\xe2\x80\x90
       spect  to  the  git  index
```

**What did you expect to happen instead?**

The output should be:

```
It also communicates with git(1) to show  modifications  with  re‐
spect  to  the  git  index
```

**How did you install `bat`?**

Occurs with bat v0.22.1 installed by brew on Ubuntu 22.04 and v0.19 installed on the same system via apt.

**bat version and environment**

`> bat --diagnostic`

#### Software version

bat 0.22.1

#### Operating system

Linux 5.15.0-60-generic

#### Command-line

```bash
bat --diagnostic 
```

#### Environment variables

```bash
SHELL=/usr/bin/zsh
PAGER=less
LESS=-R
LANG=en_GB.UTF-8
LC_ALL=<not set>
BAT_PAGER=<not set>
BAT_CACHE_PATH=<not set>
BAT_CONFIG_PATH=<not set>
BAT_OPTS=<not set>
BAT_STYLE=<not set>
BAT_TABS=<not set>
BAT_THEME=<not set>
XDG_CONFIG_HOME=<not set>
XDG_CACHE_HOME=<not set>
COLORTERM=truecolor
NO_COLOR=<not set>
MANPAGER='sh -c '\''col -bx | bat -l man -p'\'''
```

#### System Config file

Could not read contents of '/etc/bat/config': No such file or directory (os error 2).

#### Config file

Could not read contents of '/home/jscdev/.config/bat/config': No such file or directory (os error 2).

#### Custom assets metadata

Could not read contents of '/home/jscdev/.cache/bat/metadata.yaml': No such file or directory (os error 2).

#### Custom assets

'/home/jscdev/.cache/bat' not found

#### Compile time information

- Profile: release
- Target triple: x86_64-unknown-linux-gnu
- Family: unix
- OS: linux
- Architecture: x86_64
- Pointer width: 64
- Endian: little
- CPU features: fxsr,sse,sse2
- Host: x86_64-unknown-linux-gnu

#### Less version

```
> less --version 
less 590 (GNU regular expressions)
[...]

PR #3517, opened 2025-12-05
Improve native man pages and command help syntax highlighting by stripping overstriking

### LINKED ISSUE #652: `man` syntax doesn't highlight bold functions correctly
Terminals tested: alacritty, mate-terminal, urxvt

`bat --version`: 0.12.0 (Installed via `cargo install bat`)

`$MANPAGER`: bat --paging=never -pl man [1] [2]

 [1]: I disabled paging to make sure it's not a problem with `less(1)`.
 [2]: The documentation suggests setting MANPAGER to `sh -c "col -b | bat -pl man"` however I found using `col` actually just garbled the output even more, see screenshot further down.

Output with `MANPAGER='bat -pl man'`
image

Output with `MANPAGER=''` or `MANPAGER='less'`
image

The issue seems to be with highlighting functions / page references (foo(...)) when bold output is used.

When using `col -b` as suggested, it becomes even worse:

Output with `MANPAGER='sh -c "col -b | bat -pl man"'`
image

akirk:
Potentially fixes #652, #2479, #2568, #3025, #3053

I noticed that running `git commit --help | bat -plhelp` displayed text with visible `^H` characters and doubled letters. I realized that this happens because man pages use something called overstriking where (similar to a typewriter) a character is typed twice over each other. `bat` was styling the page and passing through the `^H` which lead to this mixed output.

Screenshot 2025-12-05 at 11 36 56

So as a solution I introduced a `strip_overstrike()` function to simply remove the backspace formatting to allow syntax highlighters to modify the output undisturbed and applies this stripping in both `SimplePrinter` and `InteractivePrinter`. `Cow<str>` is used to avoid allocation when no backspaces are present. 

I am not an overly heavy user of `bat` but in my testing of this fix, I tried to verify that:
-  Now `MANPAGER="bat -p -l man" man ls` displays cleanly.
- `git commit --help | bat -plhelp` now works correctly.
- `--binary=as-text` still preserves raw bytes.

I am not sure if there are other scenarios that I missed where stripping the overstriking might be unintentional.

After all, the above renders like this with this patch:

Screenshot 2025-12-05 at 14 25 40

On as side note: I realized that `git commit --help | bat -plhelp` displays the text not full-width but this is coming from `git commit --help` when it detects that its output is being piped. One workaround is to run `MANWIDTH=$(tput cols) git commit --help | bat -plhelp`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Features
 
-- Strip overstriking before applying syntax highlighting to better support man pages, see #3517 (@akirk)
+- Improve native man pages and command help syntax highlighting by stripping overstriking, see #3517 (@akirk)
 
 ## Bugfixes
 - `--help` now correctly honors `--pager=builtin`. See #3516 (@keith-hall)


src/preprocessor.rs
```diff
@@ -149,6 +150,31 @@ pub fn strip_ansi(line: &str) -> String {
...
+
+    let mut output = String::with_capacity(line.len());
+
+    for c in line.chars() {
+        if c == '\x08' {
+            output.pop();
+        } else {
+            output.push(c);
+        }
+    }
```
keith-hall: is this efficient enough, to go char by char? Perhaps there is a way to seek for the backspace?

akirk: Good point! I have updated the implementation to use `find('\x08')` to move along in chunks.

akirk:
I have added some integration tests which showed me that a safer approach would be (now implemented) to only strip the overstriking when syntax highlighting will be done (which actually more closely matches my original idea).

src/printer.rs
```diff
@@ -622,6 +634,10 @@ impl Printer for InteractivePrinter<'_> {
                 }
             };
 
+            if self.strip_overstrike && line.contains('\x08') {
```
keith-hall: I wonder how necessary the `line.contains` check is, as the first thing `strip_overstrike` does is perform a find - essentially double work? Would it be worth benchmarking with vs without this check to see which is really better for performance? Or to consider doing a find here and passing the position of it into strip_overstrike? Maybe an unnecessary micro-optimization, but I'm a bit worried because this will run on every line of input passing through bat...

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ## Bugfixes
 
+- Strip overstriking before applying syntax highlighting to better support man pages, see #3517 (@akirk)
```
keith-hall: I would argue that it is more of a feature than a bugfix... Maybe doesn't matter too much, but I would be interested to hear other opinions :slightly_smiling_face:

src/printer.rs
```diff
@@ -297,6 +305,9 @@ impl<'a> InteractivePrinter<'a> {
             _ => false,
         };
 
+        // Strip overstrike only when we have syntax highlighting (not plain text).
+        let strip_overstrike = !is_plain_text;
```
keith-hall: I wonder how necessary it is on other inputs apart from the manpages. I guess it shouldn't cause any confusing output when syntax highlighting is enabled, so maybe this approach is fine :+1:

keith-hall (commented):
In general the implementation looks clean to me. I have a couple of minor comments, and I still think we should remove the complicated `MANPAGER` `awk` script from the readme now we have `strip_overstrike` enabled by default.

Could we also have a test to prove that `--show-all` still shows backspace characters as expected, or is it already covered by existing tests? :slightly_smiling_face:

src/printer.rs
akirk: Ok, I had considered that but didn't like the readability too much. But I have changed my mind because it allows us to avoid the whole copy on write which makes it cleaner overall.

While it transforms `strip_overstrike` from a function that helps you strip backspaces in any case to a function where you already need to know the first backspace position, we are not writing a library, so I guess that's fine.

CHANGELOG.md
akirk: Upgraded it :)

src/printer.rs
akirk: Indeed, my thinking was that overstriking will interfere with any (ansi) syntax highlighting, because I would consider overstriking markup / highlighting already.

But on the other hand, it feels more robust to enable it only for man pages and command help. These were meant to be output to a terminal.
Considering we might have source code that tries to output a string with backstriking, that information won't be lost when batting it. Thus I added an integration test for this as well.

README.md
```diff
@@ -195,18 +195,13 @@ bat main.cpp | xclip
 `MANPAGER` environment variable:
 
 ```bash
-export MANPAGER="sh -c 'awk '\''{ gsub(/\x1B\[[0-9;]*m/, \"\", \$0); gsub(/.\x08/, \"\", \$0); print }'\'' | bat -p -lman'"
+export MANPAGER="bat -plman"
 man 2 select
 ```
 (replace `bat` with `batcat` if you are on Debian or Ubuntu)
 
 If you prefer to have this bundled in a new command, you can also use [`batman`](https://github.com/eth-p/bat-extras/blob/master/doc/batman.md).
```
akirk, line 203: Not sure if this also should be removed, even to the point where it can be removed from bat-extras.

keith-hall: Good question 👍 let's leave it for now, we can always remove it later

danrneal:
I still see some messiness:
```
TMUX(1)                                         General Commands Manual                                       TMUX(1)

NAME
       tmux — terminal multiplexer

SYNOPSIS
       tmux   [-2CDlNuVv]   [-c   shell-command]   [-f   file]   [-L  socket-name]  [-S  socket-path]  [-T  features]
            [command [flags]]

DESCRIPTION
       tmux is a terminal multiplexer: it enables a number of terminals to be created, accessed, and controlled  from
       a  single screen.  tmux may be detached from a screen and continue running in the background, then later reat‐
       tached.

       When tmux is started, it creates a new session with a single window and displays it on screen.  A status  line
       at  the  bottom  of  the screen shows information on the current session and is used to enter interactive com‐
       mands.

       A session is a single collection of pseudo terminals under the management of tmux.  Each session  has  one  or
       more  windows linked to it.  A window occupies the entire screen and may be split into rectangular panes, each
       of which is a separate pseudo terminal (the pty[38;2;122;162;247m24m(4) manual page documents the technical details of pseudo  ter‐
       minals).   Any  number  of  tmux  instances  may connect to the same session, and any number of windows may be
       present in the same session.  Once all sessions are killed, tmux exits.
```

Not sure it it can be avoided though.

keith-hall:
Does `--strip-ansi` help @danrneal ?

danrneal:
> Does `--strip-ansi` help @danrneal ?

Sadly same result for man tmux

akirk:
I'll take a look, thanks for the pointer for a case where it doesn't work well!

README.md
eth-p: If ANSI escape sequence and overstriking are both automatically stripped from the input, I would say yeah, we should also remove the link to `batman`.

`batman` was created to handle stripping formatting from the rendered manpages before sending it to `bat`. It doesn't provide any benefit if `bat` can do that out of the box by itself.

akirk:
@danrneal I took a look and for me it renders fine, i.e. like this: `MANPAGER="./target/release/bat -p -l man" man tmux`

Screenshot 2025-12-22 at 13 23 14

Do you have more info about your environment? I run this in macOS with tmux installed via homebrew. Maybe you can also share your raw tmux man file via a gist?

danrneal:
I'm on Debian. I'm on mobile at the moment. But will get a dump in the morning. Tmux either came pre installed or I downloaded it with apt. I'm also running `export MANPAGER=bat --plain --language=man`

danrneal:
I will also confirm that I'm using the correct release in the morning. I'm pretty certain I am but I will confirm before any gooses are chased.

danrneal:
I have confirmed I'm on head. Here is the gist: https://gist.github.com/danrneal/11b5279340cb120a53beeee4536b3fd2

Issue #3526, opened 2025-12-14
'bat -p --terminal-width nn': terminal width not honoured with -p


jghub:
bat version 0.26.0 on OSX:
```
bat --terminal-width nn
```
with `nn` some desired number of columns works just fine but
```
bat -p --terminal-width nn
```
seemingly just ignores the latter option.  (use case: want to use bat as pager in `fzf` preview in plain mode to show colorized  `grep -nC` output where the default line numbering of `bat` is undesirable and I need line wrapping to show the `grep` hits in long lines that are otherwise truncated in the preview pane of `fzf`).

I do hope I am not missing something and this is a bug and no noise :).

PR #3529, opened 2025-12-19
fix issue #3526

H4k1l:
# Description
This change resolves the request made in issue #3526. 
I added a condition to an if statement in the file "*src/bin/bat/app.rs*". 
To test the fix, you can run the command:
```
cargo run -- FILENAME --terminal-width SIZE -p
```


keith-hall:
Thanks. Is it possible to make an integration test for this to ensure it doesn't accidentally get regressed in future? 🙂

H4k1l:
Sure, I’ll add an integration test for this in the next commit.

keith-hall:
Looks like we need to run `cargo fmt` and add an entry in changelog.md, then should be good to merge 👍

PR #3531, opened 2025-12-25
Fix pager spawn under QEMU user-mode emulation

OctopusET:
QEMU user-mode emulation with glibc 2.34+ may return success from posix_spawn even when the child process fails to execute and exits immediately.

It's caused by QEMU's current implementation limitation.

Use try_wait() after spawn to detect this and fall back to stdout.

Related PR: https://github.com/BurntSushi/ripgrep/pull/3248
Failed CI without this PR: https://github.com/OctopusET/bat/actions/runs/18910014240/job/53977874904


OctopusET:
Test success: https://github.com/sharkdp/bat/actions/runs/20509082865/job/58927855111
Test fails: https://github.com/sharkdp/bat/actions/runs/20509291088/job/58928309074?pr=3532

PR #3533, opened 2025-12-25
ci: Use git version of cross for better target support

OctopusET:
riscv64gc-unknown-linux-musl isn't released yet, and maintainer aims for more things for new release. So I don't think new release will be released soon.

https://github.com/cross-rs/cross/pull/1664
https://github.com/cross-rs/cross/issues/1718

Also, I think it's not easy to change version of taiki-e/install-action@v2 upstream.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -195,9 +195,7 @@ jobs:
 
     - name: Install cross
       if: matrix.job.use-cross
-      uses: taiki-e/install-action@v2
-      with:
-        tool: cross
+      run: cargo install cross --git https://github.com/cross-rs/cross --rev 588b3c99db52b5a9c5906fab96cfadcf1bde7863
 
     - name: Overwrite build command env variable
       if: matrix.job.use-cross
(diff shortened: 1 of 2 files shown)


OctopusET:
https://github.com/sharkdp/bat/actions/runs/20510747241/job/58931482644?pr=3532

after that it success

.github/workflows/CICD.yml
```diff
@@ -195,9 +195,7 @@ jobs:
 
     - name: Install cross
       if: matrix.job.use-cross
-      uses: taiki-e/install-action@v2
-      with:
-        tool: cross
+      run: cargo install cross --git https://github.com/cross-rs/cross
```
Enselic: actually let's lock this to a specific SHA1, otherwise CI will be flaky

OctopusET: this must be better

OctopusET: I will try to add the hash soon

OctopusET: added

PR #3557, opened 2026-01-23
Add Just syntax highlighting

### LINKED ISSUE #3256: Syntax coloring for just Justfiles
Syntax coloring for just tool Justfiles. As the files are very much the same syntax as Makefiles you could most likely use the same syntax coloring configs.

an0nfunc:
Adds syntax highlighting support for [Just](https://github.com/casey/just) command runner files (`justfile`, `Justfile`, `.justfile`).

Fixes #3256

The syntax file is adapted from [nk9/just_sublime](https://github.com/nk9/just_sublime) (MIT license) with modifications to use bat's built-in Python and Bash scopes for shebang recipe bodies, since syntect doesn't support the `extends:` feature.


alloc33:
It's time to merge? Came here for `just` syntax and find out this PR on the next day after it's creation

keith-hall:
Thanks for your contribution. I hesitate to merge this for two reasons:
- having the sublime-syntax directly in `bat` and not in a submodule puts the onus on us to maintain it
- I'm not convinced the MIT license has been respected - we need to keep the copyright notice and add attribution i.e. link to the original repository
- it would be nice to improve syntect with extends support so we don't need any workarounds

an0nfunc:
> Thanks for your contribution. I hesitate to merge this for two reasons:
> 
>     * having the sublime-syntax directly in `bat` and not in a submodule puts the onus on us to maintain it
> 

You prefer a submodule then? Besides the syntax thing it should be possible ofc.

>     * I'm not convinced the MIT license has been respected - we need to keep the copyright notice and add attribution i.e. link to the original repository
> 

Will check that.

>     * it would be nice to improve syntect with extends support so we don't need any workarounds

Sure would, but thats kind of out of the scope of this PR. I can have a look at what that entails, but that's certainly not what I set out do to.

PR #3561, opened 2026-01-28
Small typo fixes

cerdelen:
Encountered minor missspellings.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -105,7 +105,7 @@ pub struct Config<'a> {
     /// The maximum number of consecutive empty lines to display
     pub squeeze_lines: Option<usize>,
 
-    // Whether or not to strip ANSI escape codes from the input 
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -99,13 +99,13 @@ pub struct Config<'a> {
...
 
-    // Weather or not to set terminal title when using a pager
+    // Whether or not to set terminal title when using a pager
     pub set_terminal_title: bool,
 
     /// The maximum number of consecutive empty lines to display
     pub squeeze_lines: Option<usize>,
 
-    // Weather or not to set terminal title when using a pager
+    // Whether or not to set terminal title when using a pager
```
keith-hall: This looks like the wrong documentation actually, can we fix it while we are here please?

```suggestion
    // Whether or not to strip ANSI escape codes from the input 
```

PR #3562, opened 2026-01-31
Add --no-header flag and improve theme errors

MaximIgitov:
## Summary
- add a --no-header flag to suppress file headers
- emit a clear error when an unknown theme is requested
- add integration coverage for both behaviors

## Test plan
- not run (Rust toolchain not installed on server)

### DIFF
--- a/src/bin/bat/app.rs
+++ b/src/bin/bat/app.rs
@@ -613,6 +613,11 @@ impl App {
             )),
         };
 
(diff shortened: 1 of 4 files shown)


keith-hall:
Hi, do we need a new flag for not showing the header, considering we already have ways to remove style components?
And I thought we already have a fairly clear error when the theme can't be found? Please correct me :)

PR #3563, opened 2026-01-31
feat: add --quiet-empty flag to suppress output on empty input

### LINKED ISSUE #1936: Feature request: Add flag for null output on null input
### Description

I would love to see a flag that causes `bat` to return null output when it receives null input.

### Use case

This would make bat more clean when using it in conjunction with other commands that may produce null output in some circumstances. EG:

```
# k-bat() ( set -eo pipefail ; kubectl "$@" -o yaml | bat -l yaml ; )
$ k-bat get jklsjdfklsjfklsdf
error: the server doesn't have a resource type "jklsjdfklsjfklsdf"
───────┬───────────────────────────────────────────
       │ STDIN   <EMPTY>
───────┴───────────────────────────────────────────
```

NORMAL-EX:
## Summary

This PR adds a `--quiet-empty` (`-E`) flag that suppresses all output when the input is empty.

## Motivation

Closes #1936

When using `bat` in pipelines with commands that may produce empty output (like `git diff` when there are no changes), the current behavior shows:

```
───────┬──────────────────
       │ STDIN   <EMPTY>
───────┴──────────────────
```

This can be noisy and unexpected. With `--quiet-empty`, users can opt for no output at all when input is empty.

## Changes

- Added `quiet_empty: bool` field to `Config`
- Added `--quiet-empty` / `-E` CLI flag in `clap_app.rs`
- Wired up the config in `app.rs`
- Implemented the behavior in `printer.rs` by checking `content_type.is_none()` in `print_header` and `print_footer`
- Added CHANGELOG entry
- Added integration tests

## Usage

```bash
# No output when input is empty
echo "" | bat --quiet-empty

# Useful with git diff
git diff | bat --quiet-empty

# Can be set via BAT_OPTS environment variable
export BAT_OPTS="--quiet-empty"
```

## Testing

Added 3 integration tests:
- `quiet_empty_suppresses_output_on_empty_stdin`
- `quiet_empty_does_not_affect_non_empty_input`
- `quiet_empty_suppresses_output_on_empty_file`

### DIFF
--- a/assets/completions/_bat.ps1.in
+++ b/assets/completions/_bat.ps1.in
@@ -186,6 +186,7 @@ Register-ArgumentCompleter -Native -CommandName '{{PROJECT_EXECUTABLE}}' -Script
             [CompletionResult]::new('--acknowledgements'      , 'acknowledgements'      , [CompletionResultType]::ParameterName, 'Show acknowledgements.')
             [CompletionResult]::new('--set-terminal-title'    , 'set-terminal-title'    , [CompletionResultType]::ParameterName, 'Sets terminal title to filenames when using a pager.')
             [CompletionResult]::new('--diagnostic'            , 'diagnostic'            , [CompletionResultType]::ParameterName, 'Show diagnostic information for bug reports.')
+            [CompletionResult]::new('--quiet-empty'           , 'quiet-empty'           , [CompletionResultType]::ParameterName, 'Do not produce any output when the input is empty.')
         #   [CompletionResult]::new('-h'                      , 'h'                     , [CompletionResultType]::ParameterName, 'Print this help message.')
             [CompletionResult]::new('--help'                  , 'help'                  , [CompletionResultType]::ParameterName, 'Print this help message.')
         #   [CompletionResult]::new('-V'                      , 'V'                     , [CompletionResultType]::ParameterName, 'Show version information.')


keith-hall:
Hi, if you need any help with this PR please let us know :)

NORMAL-EX:
Thanks for the offer! I think I've got the CI issues sorted out now - turns out I put the `--quiet-empty` flag in the wrong position in the help files. clap generates help text based on argument definition order, so I had to move it after `--diagnostic` to match. Let me know if anything else needs fixing.

keith-hall (approved):
Looks great, thank you! Typically when adding new command line options, we also like to keep the man page and shell completions up to date. Would you be willing to do that please? No worries if not, we can always merge this as it is and do it separately later. (I personally don't have much experience in the various shells or completions and was new to man pages also before working on `bat`, but have been able to copy/paste/adapt other parts to suit. These days probably also AI can help)

NORMAL-EX:
Done! Added `--quiet-empty` to the man page and all shell completions (bash, zsh, fish, PowerShell).

NORMAL-EX:
Funny you should mention AI - I actually gave an AI my GitHub API token and let it handle everything: finding issues, writing code, running into CI failures, fixing them, and now updating the docs. This whole PR is basically an experiment to see how far AI can go with open source contributions 😄

PR #3573, opened 2026-02-01
fix(cache): allow --help flag for cache subcommand

### LINKED ISSUE #3560: "bat cache --help" error: unexpected argument '--help' found
```
Arch Linux

```

`bat --help` shows this at end:

```
....
You can use 'bat cache' to customize syntaxes and themes. See 'bat cache --help' for more
information

```

But:

```
$ bat cache --help
error: unexpected argument '--help' found

Usage: bat cache [OPTIONS] <--build|--clear>
```

NORMAL-EX:
## Summary

Fixes #3560

## Problem

`bat cache --help` currently fails with:
```
error: unexpected argument '--help' found

Usage: bat cache [OPTIONS] <--build|--clear>
```

This happens because the `cache-actions` ArgGroup is set to `required(true)`, which causes clap to validate required arguments before processing the `--help` flag.

## Solution

1. Add `.arg_required_else_help(true)` to the cache subcommand - this shows help when no arguments are provided
2. Remove `.required(true)` from the `cache-actions` ArgGroup - the help behavior now handles the "no args" case

## Testing

After this change:
- `bat cache --help` shows the help message
- `bat cache` (no args) also shows the help message
- `bat cache --build` and `bat cache --clear` work as before

### DIFF
--- a/src/bin/bat/clap_app.rs
+++ b/src/bin/bat/clap_app.rs
@@ -694,11 +694,13 @@ pub fn build_app(interactive_output: bool) -> Command {
             Command::new("cache")
                 .hide(true)
                 .about("Modify the syntax-definition and theme cache")
+                .arg_required_else_help(true)
                 .arg(
                     Arg::new("build")
                         .long("build")
                         .short('b')
                         .action(ArgAction::SetTrue)
+                        .conflicts_with("clear")
                         .help("Initialize (or update) the syntax/theme cache.")
                         .long_help(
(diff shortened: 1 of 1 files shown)


keith-hall (approved):
Nice, thanks! Do you think it is possible to write an integration test to prove `bat cache --help` works to prevent future regressions or not worth it?

NORMAL-EX:
Good idea, added an integration test for `bat cache --help` to prevent regressions.

PR #3576, opened 2026-02-01
feat: Map BUILD to Python (Starlark) for Bazel (fixes #3575)

### LINKED ISSUE #3575: Render Bazel BUILD as (Starlark which is) Python, like it already does for *.bzl
**Syntax:**

It might be neat if `bat` would render files named `BUILD` as the https://bazel.build kind that they typically are? Those are (Starlark which is) Python. Particularly because `bat` already does do this for `*.bzl`, as `bat --list-languages` shows us:

```
Python                    py, py3, pyw, pyi, pyx, pyx.in, pxd, pxd.in, pxi, pxi.in, rpy, cpy, SConstruct, Sconstruct, sconstruct, SConscript, gyp,
                          gypi, Snakefile, vpy, wscript, bazel, bzl, *.xsh, *.xonshrc
```

At least it's worth having an issue here to discuss it, even if it's just to record that's not really desired (possibly with a local workaround)...

https://github.com/search?q=path%3A%2F%28%3F-i%29%28%5E%7C%5C%2F%29BUILD%24%2F&type=code&query=path%3ABUILD shows us that a lot of files named `BUILD` typically are declarations in that syntax. But not all of them neccessarily are, of course.

It's also possible to name such files `BUILD.bazel`, but they still often are not.

Reading https://github.com/sharkdp/bat/?tab=readme-ov-file#adding-or-changing-file-type-associations, as a work-around, `bat --map-syntax='BUILD:Python' java/dev/enola/common/name/BUILD` works.

Reading https://github.com/sharkdp/bat/?tab=readme-ov-file#configuration-file, doing `mkdir ~/.config/bat && echo "--map-syntax='BUILD:Python'" >>~/.config/bat/config` makes that permanent.

Would a PR to add this to Bat be welcome? If this is not desired, perhaps _"because `BUILD` is too generic a filename"_ then just close this?

Reading https://github.com/sharkdp/bat/?tab=readme-ov-file#adding-new-syntaxes--language-definitions, I'm not sure if there is a way, with (?) https://github.com/trishume/syntect/, to make it smart enough to only render `BUILD` _"if it looks like it's a Python (Starlark)" (but not e.g. just a shell script)_ ?

vorburger:
@keith-hall like this, for #3575 ?

Full disclosure: I have NOT actually tested this... as in, for #3575 I HAVE tested `bat --map-syntax='BUILD:Python' BUILD` locally on a pre-built binary, and looking around https://github.com/sharkdp/bat/tree/master/src/syntax_mapping/builtins/common I kinda gathered that's what you meant with your comment in #3575 - but I haven't actually rebuilt `bat` with this... but it looks about right (I hope).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 ## Features
 
+- Preserve `--diff` change markers and snip separators when `--plain` is set. Closes #3630, see #3643 (@mvanhorn)
 - Added support for `hidden_file_extensions` from `.sublime-syntax` files, see #3613 (@Matei02355)
 - Add word wrapping mode via `--wrap=word`, see #3597 (@veeceey)
 - Implement `--unbuffered` mode for streaming input, allowing partial lines to display immediately (e.g. `tail -f | bat -u`). Closes #3555, see #3583 (@mainnebula)
--- a/src/bin/bat/app.rs
+++ b/src/bin/bat/app.rs
@@ -590,7 +590,17 @@ impl App {
 
         // Plain if `--plain` is specified at least once.
         if self.matches.get_count("plain") > 0 {
-            return Some(StyleComponents(HashSet::from([StyleComponent::Plain])));
+            let mut components = HashSet::from([StyleComponent::Plain]);
+            // When --diff is active, preserve change markers and snip separators
+            // so that diff output remains visually useful.
+            if self.matches.try_contains_id("diff").unwrap_or_default()
+                && self.matches.get_flag("diff")
+            {
+                #[cfg(feature = "git")]
+                components.insert(StyleComponent::Changes);
+                components.insert(StyleComponent::Snip);
+            }
+            return Some(StyleComponents(components));
         }
 
         // Default behavior.
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -3984,3 +3984,151 @@ fn word_wrap_short_line_no_wrap() {
         .success()
         .stdout("Single Line\n");
 }
+
+#[cfg(unix)]
+#[cfg(feature = "git")]
+fn setup_diff_test_repo() -> tempfile::TempDir {
+    use std::process::Command;
+
+    let dir = tempfile::tempdir().expect("can create temporary directory");
+    let repo = dir.path();
+
+    // Initialize a git repo and commit a file
+    Command::new("git")
+        .args(["init"])
+        .current_dir(repo)
+        .output()
+        .expect("git init");
+
+    Command::new("git")
+        .args(["config", "user.email", "test@test.com"])
+        .current_dir(repo)
+        .output()
+        .expect("git config email");
+
+    Command::new("git")
+        .args(["config", "user.name", "Test"])
+        .current_dir(repo)
+        .output()
+        .expect("git config name");
+
+    std::fs::write(repo.join("test.txt"), "line 1\nline 2\nline 3\n").expect("can write test file");
+
+    Command::new("git")
+        .args(["add", "test.txt"])
+        .current_dir(repo)
+        .output()
+        .expect("git add");
+
+    Command::new("git")
+        .args(["commit", "-m", "initial"])
+        .current_dir(repo)
+        .output()
+        .expect("git commit");
+
+    // Modify the file so --diff has something to show
+    std::fs::write(
+        repo.join("test.txt"),
+        "line 1\nline 2 modified\nline 3\nline 4 added\n",
+    )
+    .expect("can write modified test file");
+
+    dir
+}
+
+#[cfg(unix)]
+#[cfg(feature = "git")]
+#[test]
+fn diff_plain_preserves_change_markers() {
+    let repo = setup_diff_test_repo();
+
+    // With --diff --plain, output should contain the change marker column
+    // but not other decorations like line numbers or grid
+    let output = bat()
+        .current_dir(repo.path())
+        .arg("--diff")
+        .arg("--plain")
+        .arg("--color=never")
+        .arg("--decorations=always")
+        .arg("test.txt")
+        .assert()
+        .success()
+        .get_output()
+        .stdout
+        .clone();
+
+    let stdout = std::str::from_utf8(&output).expect("valid utf-8");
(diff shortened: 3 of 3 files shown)


vorburger:
Re. 2 failing checks, let me add a line to `CHANGELOG.md` (coming up).

But that "[Minimum supported rust version" failure must be unrelated to this PR.

vorburger:
@keith-hall hm, is Copilot right about the case sensitivity? Given the test failures, I guess this would need more work.

My time is limited and my Rust is too rusty to implement that... BTW see #3577.

Also, would you want to see a test, to accept merging this?

keith-hall:
Yep, looks like Copilot is correct about the case-insensitivity, which is indeed why the tests are failing. Hmm, yeah I guess we need to introduce a way to map a filename case sensitively to proceed further. As for tests, a simple test case like Copilot suggested at https://github.com/sharkdp/bat/pull/3576#discussion_r2751166918 would suffice, if all the existing tests continue to pass unmodified.

src/syntax_mapping/builtins/common/50-bazel.toml
```diff
@@ -0,0 +1,2 @@
+[mappings]
+"Python" = ["BUILD"]
```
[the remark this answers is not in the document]
keith-hall: fixed in https://github.com/sharkdp/bat/pull/3576/commits/56fe0fa2260675c2603607ac35c4c367e5d8d797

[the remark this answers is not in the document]
keith-hall: added test in https://github.com/sharkdp/bat/pull/3576/commits/56fe0fa2260675c2603607ac35c4c367e5d8d797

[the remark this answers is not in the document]
keith-hall: no longer relevant after case sensitive matching was added in https://github.com/sharkdp/bat/pull/3576/commits/56fe0fa2260675c2603607ac35c4c367e5d8d797

keith-hall:
@cyqsimon giving you a heads up about these proposed changes to the syntax mapping code, as you are the original author, I would appreciate your thoughts/review when you get chance please 🙂

cyqsimon:
> @cyqsimon giving you a heads up about these proposed changes to the syntax mapping code, as you are the original author, I would appreciate your thoughts/review when you get chance please 🙂

Ack that. Will take a look tomorrow.

My free time is somewhat constrained these couple of days due to minor personal health issues, so my apologies in advance for any slow responses.

cyqsimon:
I was a little worried after posting my initial opinion that TOML may not be too happy with mixed-type arrays but [it appears that's not a problem since toml 1.0](https://toml.io/en/v1.1.0#array).

keith-hall:
Thanks for the excellent feedback @cyqsimon, I really appreciate it. I completely agree with all your points - I will make the necessary changes 👍

src/syntax_mapping/builtins/README.md
```diff
@@ -20,9 +20,9 @@ syntax mappings defined by all TOML files, and embed them into the binary.
 
 ## File syntax
 
-Each TOML file should contain a single section named `mappings`, with each of
-its keys being a language identifier (first column of `bat -L`; also referred to
-as "target").
+Each TOML file should contain a single section named `mappings` and/or a single
+section named `case_sensitive_mappings`, with each of its keys being a language
+identifier (first column of `bat -L`; also referred to as "target").
```
cyqsimon: This needs updating.

build/syntax_mapping.rs
```diff
@@ -116,18 +140,58 @@ impl FromStr for Matcher {
             bail!(r#"Invalid matcher: "{s}""#);
         }
 
-        Ok(Self(non_empty_segments))
+        Ok(Self {
+            segments: non_empty_segments,
+            case: Case::Insensitive,
+        })
```
cyqsimon, line 148: I would prefer to use `Default::default` because it's semantically clearer why we've chosen to use a particular value:

```suggestion
        Ok(Self {
            segments: non_empty_segments,
            ..Default::default()
        })
```

And then add `#[derive(Default)]` on `Case`.

But admittedly this is a matter of personal taste so I'll leave it up to you.

build/syntax_mapping.rs
```diff
@@ -116,18 +140,58 @@ impl FromStr for Matcher {
...
+            } => {
+                let mut matcher = Matcher::from_str(&glob).map_err(serde::de::Error::custom)?;
+                matcher.case = if case_sensitive {
+                    Case::Sensitive
+                } else {
+                    Case::Insensitive
+                };
+                Ok(matcher)
+            }
+        }
```
cyqsimon, line 183: I think this is the exact situation [`#[serde(from = "FromType")]`](https://serde.rs/container-attrs.html#from) is designed for no? Then you're just writing a `impl From<RawMatcher> for Matcher` which should be much less arcane than a manual serialization impl.

keith-hall: agreed, nice spot - thanks, done.

build/syntax_mapping.rs
keith-hall: I like the idea, can't argue with semantic clarity, done.

src/syntax_mapping/builtins/README.md
keith-hall: indeed, thanks! done

build/syntax_mapping.rs
cyqsimon: I believe you missed this?

keith-hall: Huh, weird, I definitely remember making that change but I don't see it here... I guess I forgot to commit it somehow, thanks for pointing it out. I will make a separate PR for it when I get chance.

cyqsimon: Actually may I do this PR? I've actually changed my mind on having this `Default` impl at all. It has to do with the lack of clarity in the hierarchy involving `Matcher` and `RawMatcher`. You'll see what I mean in code.

keith-hall: Of course, it would be most welcome, thanks 🙂

PR #3578, opened 2026-02-01
Add initial flake.nix; just for develop, for now (fixes#3577)

### LINKED ISSUE #3577: Add flake.nix to auto-enable development environment with a recent Rust version
For me, trying to locally build by following https://github.com/sharkdp/bat#from-source, e.g. just to check out if Copilot is right about its feedback on #3576, doesn't quite work:

```
cargo build --locked
error: failed to parse lock file at: /home/vorburger/git/github.com/sharkdp/bat/Cargo.lock

Caused by:
  lock file version `4` was found, but this version of Cargo does not understand this lock file, perhaps Cargo needs to be updated?
``` 

I understand that this is just because I'm probably on some ancient Cargo / Rust version:

```
$ cargo --version
cargo 1.75.0 (1d8b05cdd 2023-11-20)

$ rustup show
Default host: x86_64-unknown-linux-gnu
rustup home:  /home/vorburger/.rustup

stable-x86_64-unknown-linux-gnu (default)
rustc 1.75.0 (82e1608df 2023-12-21)
```

But my Rust is too rusty to sort this out, just to locally validate a contribution adding a configuration file.

In a perfect world, it would be neat if checking out and locally building the project would "just work".

One way that I know of how to make this possible is using [Nix](https://nixos.org), like e.g. on https://docs.enola.dev/dev/setup. The `flake.nix` required for this would be trivial, and just a few lines. (Just for this; NOT for actually building the Nix package - that's another story.)

Would a PR contributing this be welcome?

vorburger:
Fixes #3577.

With this, one can `rustup self uninstall` and instead https://zero-to-nix.com/start/install/, and possibly https://direnv.net (optional; or just `nix develop`), then `direnv allow`, one gets an up-to-date Cargo version to work on this project. Any other than Cargo of [the 120000 Nix packages](https://search.nixos.org/packages) required by Bat could be added to `flake.nix`.

@keith-hall I'm not sure you'd welcome this?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ## Features
 
-- Added an initial `flake.nix` for a ready made development environment; see #3577 (@vorburger)
+- Added an initial `flake.nix` for a ready made development environment; see #3578 (@vorburger)
 - Add `--quiet-empty` (`-E`) flag to suppress output when input is empty. Closes #1936, see #3563 (@NORMAL-EX)
 - Improve native man pages and command help syntax highlighting by stripping overstriking, see #3517 (@akirk)
 


flake.nix
```diff
@@ -0,0 +1,40 @@
...
+  inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+
+  outputs =
+    { self, ... }@inputs:
+    let
+      supportedSystems = [
+        "x86_64-linux" # 64-bit Intel/AMD Linux
+        "aarch64-linux" # 64-bit ARM Linux
+        "aarch64-darwin" # 64-bit ARM macOS
+      ];
```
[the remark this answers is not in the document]
vorburger: Sure thing Copilot; done!

keith-hall:
I guess whether we would be willing to accept this PR could depend on the answer to a couple of questions - thanks for initiating this idea and discussion. I'm not familiar with `nix` other than knowing what it is and why people would want to use it, so apologies if the questions are a little basic. Also, the decision is not mine alone, we'd want the other maintainers to weigh in.

- Is it possible that someone would use `direnv` but not be a nix flakes user, and have trouble with the `.envrc` specifying to `use flakes`?
- Are the `bat` maintainers going to be expected to maintain this, like will it need any changes when `bat`'s dependencies get changed/bumped etc.? (i.e. will it require more work, and make releases harder? I'm guessing that we will get some PRs to add the other required dependencies from interested users (or issues complaining about them missing!)...)
  - Is there some way to validate in CI that the flake/nix/lock files are valid etc?
  - Is there some way to perhaps automate the specification of the nix dependencies based on what we need to install in CI to get `bat` built etc.? (Side note: if we had a Dockerfile for building `bat`, would that help or be orthogonal?)
- (Do you happen to know) whether there is a precedent that other Rust projects also have nix files in their Git repositories?

vorburger:
> I guess whether we would be willing to accept this PR could depend on the answer to a couple of questions - thanks for initiating this idea and discussion. I'm not familiar with `nix` other than knowing what it is and why people would want to use it, so apologies if the questions are a little basic. Also, the decision is not mine alone, we'd want the other maintainers to weigh in.

That's totally fair! To be explicit: I take absolutely no offense if this ends up in a "Thanks, but No Thanks" and gets closed.

> * Is it possible that someone would use `direnv` but not be a nix flakes user, and have trouble with the `.envrc` specifying to `use flakes`?

My currrent understanding is that this is technically theoretically possible, but unlikely in practice. As far as I can tell, https://direnv.net, while technically unrelated to Nix per se, is particularly popular in the Nix ecosystem. Its original author and [still main contributor](https://github.com/direnv/direnv/graphs/contributors) is an active Nix contributor (zimbatm). I'm not entirely sure what DirEnv would o when it sees a `use flakes` but there is no `nix` available... but I wouldn't worry too much about this combination, as I suspect it's unlikely to affect users, in practice.

> * Are the `bat` maintainers going to be expected to maintain this, like will it need any changes when `bat`'s dependencies get changed/bumped etc.? (i.e. will it require more work, and make releases harder? I'm guessing that we will get some PRs to add the other required dependencies from interested users (or issues complaining about them missing!)...)

Unless the project's release process actually starts to somehow rely on Nix, this would be just "for convenience of Nix users", I guess. This PR, as-is currently, wouldn't require anything from `bat` maintainers. The `cargo` version used could be bumped by anyone wanting to by running `nix flake update nixpkgs`, which would update the `flake.lock`. But you could argue that it's "extra files", of course.... and I totally respect that. This discussion is a bit related to the next question as well.

>   * Is there some way to validate in CI that the flake/nix/lock files are valid etc?

The way that I understand / think about this, is that if this project wanted to go all in on using Nix to configure its Rust Toolchain, and have full consistency between CI and local develop environments (which is a decision its maintainers need to make, not me) then you might want to replace https://github.com/dtolnay/rust-toolchain which I see being used [around here](https://github.com/sharkdp/bat/blob/master/.github/workflows/CICD.yml) with https://github.com/cachix/install-nix-action, perhaps with https://github.com/DeterminateSystems/magic-nix-cache-action. This could use the `flake.nix` introduced in this PR. Whether that's "worth it" (cognitive overhead?) really depends on the maintainers of this repo.

>   * Is there some way to perhaps automate the specification of the nix dependencies based on what we need to install in CI to get `bat` built etc.? 

Personally, I would go about this "the other way around" - make Nix the Source of Truth (SoT), and just make the CI use Nix as well.

> (Side note: if we had a Dockerfile for building `bat`, would that help or be orthogonal?)

To me that's a bit orthogonal. It's a longer discussion, but I view containers (Dockerfile) and Nix as 2 separate solutions for a similar or at least related problem space. What one COULD do, and I do in some of my personal projects, IFF this is a requirement / need (but is it?) is to also build a (Docker) container image, e.g. for https://containers.dev, but instead of repeating stuff in a `Dockerfile`, one could use e.g. https://github.com/xtruder/nix-devcontainer which can also re-use the `flake.nix` this PR introduces. (All this is related to having a container with all required tools for a development environment; a container to distribute `bat` for users is an entirely different discussion, of course.)

Perhaps I should create an example repo and blog about all this space on my https://github.com/vorburger/vorburger.ch-Notes some day?

> * (Do you happen to know) whether there is a precedent that other Rust projects also have nix files in their Git repositories?

IDK, sorry.

flake.nix
```diff
@@ -0,0 +1,40 @@
...
+          }
+        );
+    in
+    {
+      devShells = forEachSupportedSystem (
+        { pkgs, system }:
+        {
+          default = pkgs.mkShellNoCC {
+            packages = with pkgs; [
+              cargo
```
Enselic, line 34: If your original problem was that your cargo was too old, how does it help to not specify a minimal cargo version here?

.envrc
```diff
@@ -0,0 +1 @@
+use flake
```
Enselic, line 1: Can we skip this file? It does not seem like a valid https://direnv.net/ file.

Enselic (changes requested):
As long as we don't affect anyone other than nix users I'm fine with merging.

flake.nix
MCSeekeri: > If your original problem was that your cargo was too old, how does it help to not specify a minimal cargo version here?

This is related to how Nix Flakes work. The specific version of Cargo is determined by the nixpkgs revision pointed to in `flake.lock`, so there’s no need to specify a minimum version in `flake.nix`.
As of now, the version of Cargo is [1.92.0](https://search.nixos.org/packages?channel=unstable&show=cargo&query=cargo).

.envrc
MCSeekeri: > Can we skip this file? It does not seem like a valid https://direnv.net/ file.

https://determinate.systems/blog/nix-direnv/#direnv-and-nix-environments

eth-p:
Apologies for being a bit late to the reply. The changes look good to me. It should be very little effort to maintain and it improves developer experience.

I can also handle any maintenance and improvements related to the flake. I use home-manager on my Mac and NixOS on my PC, so I'm already familiar with the ecosystem.

vorburger:
Awesome! I love open source.

PR #3580, opened 2026-02-06
fix(cache): allow --help flag for cache subcommand

### LINKED ISSUE #3560: "bat cache --help" error: unexpected argument '--help' found

NORMAL-EX:
## Summary

Fixes #3560

## Problem

`bat cache --help` currently fails with:

```
error: unexpected argument '--help' found

Usage: bat cache [OPTIONS] <--build|--clear>
```

This happens because the `cache-actions` ArgGroup is set to `required(true)`, which causes clap to validate required arguments before processing the `--help` flag.

## Solution

Remove `.required(true)` from the `cache-actions` ArgGroup. The behavior remains the same (users must specify either `--build` or `--clear` when running the cache subcommand) because clap will still show an error if neither is provided. However, now `--help` is processed first, allowing users to view the help message.

## Testing

```bash
# Before: error
$ bat cache --help
error: unexpected argument '--help' found

# After: shows help
$ bat cache --help
Modify the syntax-definition and theme cache
...
```

---

*This is a resubmission of #3573 with `cargo fmt` applied.*

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -440,9 +440,9 @@ dependencies = [
 
 [[package]]
 name = "deranged"
-version = "0.3.11"
+version = "0.5.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b42b6fa04a440b495c8b04d0e71b707c585f83cb9cb28cf8cd0d976c315e31b4"
+checksum = "ececcb659e7ba858fb4f10388c250a7252eb0a27373f1a72b8748afdd248e587"
 dependencies = [
  "powerfmt",
 ]
@@ -1068,9 +1068,9 @@ dependencies = [
 
 [[package]]
 name = "num-conv"
-version = "0.1.0"
+version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "51d515d32fb182ee37cda2ccdcb92950d6a3c2893aa280e540671c2cd0f3b1d9"
+checksum = "cf97ec579c3c42f953ef76dbf8d55ac91fb219dde70e49aa4a6b7d74e9919050"
 
 [[package]]
 name = "num-traits"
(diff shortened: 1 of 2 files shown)


NORMAL-EX:
The `cargo audit` failure is unrelated to this PR - it's caused by a newly published security advisory [RUSTSEC-2026-0009](https://rustsec.org/advisories/RUSTSEC-2026-0009) (published 2026-02-05) affecting the `time` crate dependency.

This affects all PRs until the upstream dependency is updated in the main branch.

CHANGELOG.md
keith-hall, line 1: erm, the changelog entry being repeated for every version of `bat` ever released is a bit... odd... we can't go back in time / backport the fix and republish old versions with the same version number it originally had... I suggest just to keep the entry in the unreleased section...

NORMAL-EX:
Fixed both issues:

1. Reverted CHANGELOG.md to only have the entry in the unreleased section — sorry about that!
2. Added the `cache_help_shows_help_message` integration test from the original PR

Thanks for catching these!

NORMAL-EX:
Updated the fix to match the approach from #3573:

- Added `.arg_required_else_help(true)` to the cache subcommand
- Replaced the `ArgGroup` with `.conflicts_with()` on each arg
- Removed unused `ArgGroup` import

Just removing `.required(true)` from the group wasn't enough — `--help` was still rejected. The full removal of the `ArgGroup` + `arg_required_else_help` is needed for clap to properly handle `--help`.

NORMAL-EX:
Found the root cause of the test failure: the parent command sets `.disable_help_flag(true)` (line 37), which propagates to the `cache` subcommand. This means `--help` is not recognized on `bat cache` regardless of the ArgGroup changes.

Added `.disable_help_flag(false)` to the cache subcommand to override the parent setting. Also fixed `cargo fmt` (import line formatting).

keith-hall (approved):

Issue #2668, opened 2023-09-13
broken highlighting of manpage references `page(section)`


LunarLambda:
**How did you install `bat`?**

Archlinux repositories

---

![image](https://github.com/sharkdp/bat/assets/38919842/5bc0b747-944a-4089-ae09-30c826570402)
![image](https://github.com/sharkdp/bat/assets/38919842/9057256c-b609-48a1-ab73-0de606bf50c3)

**bat version and environment**

#### Software version

bat 0.23.0 (871abd27)

#### Operating system

Linux 6.4.11-arch2-1

#### Command-line

```bash
bat --diagnostic 
```

#### Environment variables

```bash
SHELL=/bin/zsh
PAGER=less
LESS=<not set>
LANG=de_DE.UTF-8
LC_ALL=<not set>
BAT_PAGER=<not set>
BAT_CACHE_PATH=<not set>
BAT_CONFIG_PATH=<not set>
BAT_OPTS=<not set>
BAT_STYLE=<not set>
BAT_TABS=<not set>
BAT_THEME=<not set>
XDG_CONFIG_HOME=/home/lambda/cfg
XDG_CACHE_HOME=/home/lambda/.cache
COLORTERM=truecolor
NO_COLOR=<not set>
MANPAGER=bat -plman
```

#### System Config file

Could not read contents of '/etc/bat/config': No such file or directory (os error 2).

#### Config file

Could not read contents of '/home/lambda/cfg/bat/config': No such file or directory (os error 2).

#### Custom assets metadata

Could not read contents of '/home/lambda/.cache/bat/metadata.yaml': No such file or directory (os error 2).

#### Custom assets

'/home/lambda/.cache/bat' not found

#### Compile time information

- Profile: release
- Target triple: x86_64-unknown-linux-gnu
- Family: unix
- OS: linux
- Architecture: x86_64
- Pointer width: 64
- Endian: little
- CPU features: fxsr,sse,sse2
- Host: x86_64-unknown-linux-gnu

#### Less version

```
> less --version 
less 643 (PCRE2 regular expressions)
Copyright (C) 1984-2023  Mark Nudelman

less comes with NO WARRANTY, to the extent permitted by law.
For information about the terms of redistribution,
see the file named README in the less distribution.
Home page: https://greenwoodsoftware.com/less
```

Terminal is xfce4-terminal (`VTE_VERSION` 7202, `TERM` xterm-256color)

LevitatingBusinessMan:
[...]

Issue #3549, opened 2026-01-05
Man page highlighting causes text artifacts in front of left parens


BlueElectivire:
**What steps will reproduce the bug?**

1. Set bat as manpager using the method in the docs (export MANPAGER="bat -plman")
2. Open a manpage which contains a left paren character (which is virtually all of them)

**What happens?**
Left parens get highlighted, but in front of them appear the characters "22m" in a different color.

Screenshot:
<img width="736" height="123" alt="Image" src="https://github.com/user-attachments/assets/c8e083db-998f-4c8f-b935-2ec47d6b6ca5" />

This also happens if there isn't a space before the parens, where it is more evident the "22m" doesn't belong:
<img width="879" height="25" alt="Image" src="https://github.com/user-attachments/assets/0a0d013e-d1af-4b3c-935e-36d871c6fe9d" />

Notice that this doesn't appear on every single open parens, only on the ones highlighted.

**What did you expect to happen instead?**
The relevant characters should be highlighted without adding the artifacts.

**How did you install `bat`?**
pacman

---

**bat version and environment**

#### Software version

bat 0.26.1 (v0.25.0-402-g979ba226)

#### Operating system

- OS: Linux (Arch Linux rolling)
- Kernel: 6.18.3-arch1-1

#### Command-line

```bash
bat -plman --diagnostic
```

#### Environment variables

```bash
BAT_CACHE_PATH=<not set>
BAT_CONFIG_PATH=<not set>
BAT_OPTS=<not set>
BAT_PAGER=<not set>
BAT_PAGING=<not set>
BAT_STYLE=<not set>
BAT_TABS=<not set>
BAT_THEME=<not set>
COLORTERM=truecolor
LANG=en_IL
LC_ALL=<not set>
LESS=<not set>
MANPAGER='bat -plman'
NO_COLOR=<not set>
PAGER=<not set>
SHELL=/usr/bin/zsh
TERM=xterm-ghostty
XDG_CACHE_HOME=/home/blue/.cache
XDG_CONFIG_HOME=/home/blue/.config
```

#### System Config file

Could not read contents of '/etc/bat/config': No such file or directory (os error 2).

#### Config file

```
--theme=ansi
```

#### Custom assets metadata

Could not read contents of '/home/blue/.cache/bat/metadata.yaml': No such file or directory (os error 2).

#### Custom assets

'/home/blue/.cache/bat' not found

#### Compile time information

- Profile: release
[...]

PR #3586, opened 2026-02-17
Fixed manpage syntax to remove ANSI artifacts

BlueElectivire:
Fixed an issue with the function-call regex in Manpage which caused some ANSI escape sequences injected by `man` to be incorrectly interpreted as a part of a function name and consequently break (See issue #3549).
The fix involves directly searching for ANSI escape sequences in non-capturing groups, both before the function name and before the parens, and changing the `\b` to `\s` (because `\b` would trigger on `[` characters).

### DIFF
--- a/tests/syntax-tests/highlighted/Manpage/uwsm-0.26.3.man
+++ b/tests/syntax-tests/highlighted/Manpage/uwsm-0.26.3.man
@@ -1,399 +1,399 @@
-UWSM[35m([0m[32m1[0m[35m)[0m                                 [34mGeneral[0m[34m [0m[34mCommands[0m[34m [0m[34mManual[0m                                UWSM[35m([0m[32m1[0m[35m)[0m
-
-[34mNAME[0m
-       UWSM - Universal Wayland Session Manager.
-
-[34mSYNOPSIS[0m
-       uwsm [[33m-h[0m[35m|[0m[33m-v[0m] {subcommand} [options ...]
-
-[34mDESCRIPTION[0m
-       Launches arbitrary wayland compositor via a set of systemd user units to provide graphical user
-       session  with  environment  management, XDG autostart support, clean shutdown. Provides helpers
-       for launching applications as scopes or services.
-
-[34mSUBCOMMANDS[0m
-       select     Select default compositor Entry.
-       start      Start compositor and graphical session.
-       finalize   Send compositor-set variables and unit startup notification to systemd user manager.
-       stop       Stop graphical session and compositor.
-       app        Application unit launcher (with Desktop Entry support).
-       check      Perform state checks (for scripting and info).
-       aux        Technical functions for use inside units.
-
-       See corresponding SUBCOMMANDS subsections below for further info.
-
-       Help for each subcommand is accessible by running "uwsm {subcommand} [33m-h[0m".
-
-[34mCONFIGURATION[0m
-   Files
-       In XDG config hierarchy:
-       uwsm/env
-       uwsm/env.d/*
-       uwsm/env-${compositor}
-       uwsm/env-${compositor}.d/*   Environment (shell) to  be  sourced  for  the  graphical  session.
-                                    Sourced from directories of increasing priority, in each directory
-                                    common  file is sourced first, then suffixed files in the order of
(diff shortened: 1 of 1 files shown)


assets/syntaxes/02_Extra/Manpage.sublime-syntax
```diff
@@ -69,7 +69,7 @@ contexts:
       escape: '(?={{section_heading}})'
 
   function-call:
-    - match: '\b([A-Za-z0-9_\-]+\.)?([A-Za-z0-9_\-]+)(\()([^)]*)(\))'
+    - match: '(?:\s+)(?:\e\[[\?=]?(?:\d+;?)+[A-Za-z])?([A-Za-z0-9_\-^\e]+\.)?([A-Za-z0-9_\-]+)(?:\e\[[\?=]?(?:\d+;?)+[A-Za-z])?(\()([^)]*)(\))'
```
keith-hall: Let's extract `(?:\e\[[\?=]?(?:\d+;?)+[A-Za-z])` into a variable, so we can reuse it, and make this pattern easier to read

keith-hall (commented):
Nice, thanks. It would be great to add a syntax-test for this - i.e. an example manpage file in /`tests/syntax-tests/source/Manpage/` which contains ANSI escapes around a function call, and a corresponding `highlighted` version to prove the fix works as expected and won't get accidentally regressed in future.

assets/syntaxes/02_Extra/Manpage.sublime-syntax
keith-hall: Also we probably don't need the non-capturing group around the `\s+` at the beginning of the pattern.

keith-hall: Judging from the CI failures, changing `\b` to `\s` causes some function calls to no longer be highlighted as such. Maybe we should consider a lookbehind or something instead to work around `\b` triggering on `[` characters?

rivenirvana:
Will this also close #2668?

assets/syntaxes/02_Extra/Manpage.sublime-syntax
BlueElectivire: Changed some things according to the review, added a test.
Cargo test passes on my machine so I hope it passes the CI this time :)

BlueElectivire:
> Will this also close #2668?

It should, in hindsight it seems my issue (the one I referenced above) may be a duplicate of yours.

BlueElectivire:
@keith-hall just to make sure I got the licensing on the syntax tests right - If I took the test file from an MIT licensed project (UWSM in this case) then I shouldn't need to add anything other than the file itself, right?

BlueElectivire:
Seems the CI failure is in the test I added - Apparently the file I committed kept the coloring from my terminal, although I used bat to generate it and I use the 16-color theme.
Should I have done something else to generate the correct highlighted file?
Would appreciate any help here :)

keith-hall:
> @keith-hall just to make sure I got the licensing on the syntax tests right - If I took the test file from an MIT licensed project (UWSM in this case) then I shouldn't need to add anything other than the file itself, right?

As long as we meet the attribution requirements of the licence then I believe it should be fine.

I believe you would need to use the --no-config argument to generate the highlighted file, because in CI it runs with default config and therefore the default color scheme, which is Monokai. Maybe we need to document that

BlueElectivire:
> > @keith-hall just to make sure I got the licensing on the syntax tests right - If I took the test file from an MIT licensed project (UWSM in this case) then I shouldn't need to add anything other than the file itself, right?
> 
> As long as we meet the attribution requirements of the licence then I believe it should be fine.

Going over it, it says the license and the copyright notice should be present when redistributing "substantial portions" of the program. Not a license expert by any means, but I wouldn't call the manpage a "substantial portion". I could just put the attribution anyway to be safe about it, but I the way I read it that's not necessary.

> I believe you would need to use the --no-config argument to generate the highlighted file, because in CI it runs with default config and therefore the default color scheme, which is Monokai. Maybe we need to document that

Did that, seems to generate the correct colors, so hopefully everything passes.

PR #3588, opened 2026-02-19
warn when pager is missing instead of silently falling back

### LINKED ISSUE #2904: `bat` silently quits, rather than reporting an error, if `less` is missing

1. Install `bat` on a machine without `less` (a minimal Docker image in my case with 22.04.4 LTS (Jammy Jellyfish) x86_64, where of course `bat` was the first thing I wanted to install):  
 ```sudo apt install bat fd-find…```
2. Run it:
``` 
batcat  --generate-config-file
batcat  /root/.config/bat/config
```

**What happens?**
Nothing, but the file exists:
```
root@kromodeor:~/Documents/Code/compiler-tester/llvm# batcat  /root/.config/bat/config
root@kromodeor:~/Documents/Code/compiler-tester/llvm# ls -l  /root/.config/bat/config
-rw-r--r-- 1 root root 1114 Mar 19 15:58 /root/.config/bat/config
```

Error message stating that my poor pathetic installation lacked `less`.

```
time sudo apt install bat fd-find silversearcher-ag tldr  nano
```

---

```
batcat /root/.config/bat/config --diagnostic
bat has been built without the 'bugreport' feature. The '--diagnostic' option is not available.

batcat --version
bat 0.19.0

apt info bat
Package: bat
Version: 0.19.0-1ubuntu0.1
Built-Using: rust-ansi-colours (= 1.0.1-2), rustc (= 1.59.0+dfsg1-1~ubuntu2~22.04.1)
…
```

IMaloney:
fixes #2904

bat was silently falling back to stdout when the pager wasn't found, which confused users who expected output to be paged

now shows a clear warning message when the pager is missing but still works by falling back to stdout

tested with nonexistent pager, real pager, and all existing tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 ## Bugfixes
 - Report error when pager is missing instead of silently falling back, see #3588 (@IMaloney)
+- Fix `--wrap=never` and `-S` flags being ignored when piping to pager, see #3592 (@IMaloney)
 - Fix crash with BusyBox `less` on Windows, see #3527 (@Anchal-T)
 - Fix `bat cache --help` failing with 'unexpected argument' error, see #3580 and #3560 (@NORMAL-EX)
 - `--help` now correctly honors `--pager=builtin`. See #3516 (@keith-hall)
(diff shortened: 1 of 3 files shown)


keith-hall:
Thanks for the recent fixes btw :confetti_ball:

Issue #3587, opened 2026-02-18
bat -S or bat --wrap=never is not working


ciscohack:
Hi Team,

i am running Bat version 0.26.1 on macOS 26.3 and used homebrew to install the bat tool. I started facing issue with bat where bat with option bat -S or bat --wrap=never is not working and in bat long line still getting wrap. request you to check if this is bug in 0.26.1 code or something else. i tried reset cache and reinstall bat but issue still persist. please help me on this 

<img width="1767" height="193" alt="Image" src="https://github.com/user-attachments/assets/be7b8287-62a0-4939-acce-e0c180eee12d" />

#### Environment variables

```bash
BAT_CACHE_PATH=<not set>
BAT_CONFIG_PATH='~/.config/bat/bat.conf'
BAT_OPTS=<not set>
BAT_PAGER='less -iRFK'
BAT_PAGING=<not set>
BAT_STYLE=header,grid
BAT_TABS=<not set>
BAT_THEME=Coldark-Cold
COLORTERM=truecolor
LANG=en_US.UTF-8
LC_ALL=en_US.UTF-8
LESS=-iRFXMx4
MANPAGER=moor
NO_COLOR=<not set>
PAGER=moor
SHELL=/opt/homebrew/Cellar/zsh/5.9/bin/zsh
TERM=tmux-256color
XDG_CACHE_HOME=<not set>
XDG_CONFIG_HOME=/Users/praverai/.config
```

#### System Config file

Could not read contents of '/etc/bat/config': No such file or directory (os error 2).

#### Config file

Could not read contents of '~/.config/bat/bat.conf': No such file or directory (os error 2).

#### Custom assets metadata

Could not read contents of '/Users/praverai/.cache/bat/metadata.yaml': No such file or directory (os error 2).

#### Custom assets

'/Users/praverai/.cache/bat' is empty

#### Compile time information

- Profile: release
- Target triple: aarch64-apple-darwin
- Family: unix
- OS: macos
- Architecture: aarch64
- Pointer width: 64
- Endian: little
- CPU features: aes,crc,dit,dotprod,dpb,dpb2,fcma,fhm,flagm,fp16,frintts,jsconv,lor,lse,neon,paca,pacg,pan,pmuv3,ras,rcpc,rcpc2,rdm,sb,sha2,sha3,ssbs,vh
- Host: aarch64-apple-darwin

#### Less version

```
> less --version
less 692 (PCRE2 regular expressions)
[...]

PR #3590, opened 2026-02-20
fix: respect --wrap=never flag when paging is enabled

IMaloney:
## Description

Fixes issue #3587 where the `--wrap=never` (or `-S`) flag was being ignored when bat was piping output to a pager.

## Problem

When output is piped to a pager (e.g., `less`), the wrapping_mode logic was not checking the explicit `--wrap=never` flag. Instead, it would always default to `NoWrapping(false)`, which meant the `-S` flag was not passed to the pager. This caused lines to wrap in the pager despite the user's explicit request to disable wrapping.

## Solution

The fix restructures the wrapping_mode assignment logic to:
1. First check for explicit `--chop-long-lines` flag (always respected)
2. Then check for explicit `--wrap=*` flags (`character`, `never`) and respect them unconditionally
3. Only fall back to interactive_output-based logic for `auto` mode or default cases

This ensures that when the user explicitly passes `--wrap=never`, it returns `WrappingMode::NoWrapping(true)`, which causes the pager to receive the `-S` flag to disable wrapping.

## Testing

Added two regression tests:
- `wrap_never_with_paging`: Verifies `--wrap=never` is respected when `--paging=always` is used
- `s_flag_with_paging`: Verifies the `-S` flag is respected when `--paging=always` is used

All existing tests continue to pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Improve native man pages and command help syntax highlighting by stripping overstriking, see #3517 (@akirk)
 
 ## Bugfixes
+- Fix `--wrap=never` and `-S` flags being ignored when piping to pager, see #3590 (@IMaloney)
 - Report error when pager is missing instead of silently falling back, see #3588 (@IMaloney)
 - Fix crash with BusyBox `less` on Windows, see #3527 (@Anchal-T)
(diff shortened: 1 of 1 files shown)


IMaloney:
Closing to recreate from clean master branch - accidentally included changes from other PRs

PR #3597, opened 2026-02-23
Add word wrapping mode

### LINKED ISSUE #2751: wrapping by word would be awesome
Been using bat for two years, huge fan :pray: 

An additional `word` option to the `--wrap` flag would be great to avoid splitting the words in two when wrapping long lines. Like in most code editors offer.

veeceey:
Adds a `word` option to `--wrap` that breaks lines at word boundaries (whitespace) instead of mid-character. When a single word exceeds the terminal width, it falls back to character wrapping so nothing gets lost.

Usage: `bat --wrap=word file.rs`

Closes #2751

### DIFF
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -3827,8 +3827,7 @@ fn word_wrap_with_line_numbers() {
         .assert()
         .success()
         .stdout(
-            "\
-   1 The quick brown fox jumps over the
+            "   1 The quick brown fox jumps over the
      lazy dog and then runs away
    2 superlongwordthatdefinitelyexceedst
      heterminalwidthandshouldfallbacktoc


veeceey:
**Test results** (macOS, local build):

```
cargo test
test result: ok. 214 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

Manual test with a long line:
```
$ echo "The quick brown fox jumps over the lazy dog and keeps running across the terminal width boundary to show word wrapping" | bat --wrap=word --terminal-width=40
```
Line breaks at word boundaries as expected. When a single word exceeds width (e.g. a long URL), it falls back to character wrapping.

src/printer.rs
```diff
@@ -807,13 +817,32 @@ impl Printer for InteractivePrinter<'_> {
...
+                                        let rest_start = ws_idx
+                                            + line_buf[ws_idx..]
+                                                .chars()
+                                                .next()
+                                                .map(|ch| ch.len_utf8())
+                                                .unwrap_or(0);
+                                        let remainder = line_buf[rest_start..].to_string();
+                                        (emit, Some(remainder))
+                                    } else {
+                                        (line_buf.clone(), None)
```
keith-hall: It would be great if we can find a way to avoid cloning in the common case of word wrap being disabled

Keksoj:
This is wonderful, thank you so much!

Issue #2376, opened 2022-10-22
Docs: Make syntax tests manual more concreete


BANOnotIT:
Right now manual says that "bat should be in path" in Syntax tests section.

Inside update.sh there's a call of python script that generates highlighted files. For me that wasn't obvious why script generated un highlighted file firstly.

I'd recommend rewrite python script in such a way that it'll use `/target/debug/bat` version if it's presented or somehow ask what program to use. Or adding `export PATH=...` lines in the manual to ease friction of setting up tests.

sharkdp:
> For me that wasn't obvious why script generated un highlighted file firstly.

So why did that happen?



> I'd recommend rewrite python script in such a way that it'll use `/target/debug/bat` version if it's presented

Hm. I'd rather use the version from `PATH` by default. But we should definitely improve the documentation, I agree.

BANOnotIT:
> So why did that happen?

Because python script used os installed version, not built from source one.

KittyBorgX:
Can i take this up if no one has taken it up yet?

foxfromworld:
I’ve opened PR #3610 to address this issue.
I’m new to contributing, but I wanted to give it a try. Please let me know if there’s anything I should improve. Thanks!

PR #3610, opened 2026-03-03
Docs: clarify PATH requirement and explain highlighted outputs (Fixes…

foxfromworld:
This PR improves the documentation for running syntax tests.

- Clarifies that the syntax test script (tests/syntax-tests/update.sh) calls `bat` from the PATH.
- Explains that the script regenerates the highlighted output files under tests/syntax-tests/highlighted/, which are used to verify syntax highlighting.
- Adds explicit instructions for adding the debug build to PATH when only `cargo build --bins` is used.
- Provides example error message when `bat` is not found on PATH.
- Notes that `cargo install --path . --locked` places `bat` in ~/.cargo/bin, usually already in PATH.

This change addresses Issue #2376 by making the syntax tests documentation more concrete and beginner-friendly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 ## Features
 
+- Add word wrapping mode via `--wrap=word`, see #3597 (@veeceey)
 - Implement `--unbuffered` mode for streaming input, allowing partial lines to display immediately (e.g. `tail -f | bat -u`). Closes #3555, see #3583 (@mainnebula)
 - Added an initial `flake.nix` for a ready made development environment; see #3578 (@vorburger)
 - Add `--quiet-empty` (`-E`) flag to suppress output when input is empty. Closes #1936, see #3563 (@NORMAL-EX)
--- a/assets/completions/_bat.ps1.in
+++ b/assets/completions/_bat.ps1.in
@@ -9,7 +9,7 @@ Register-ArgumentCompleter -Native -CommandName '{{PROJECT_EXECUTABLE}}' -Script
     $ArrayCompletion = @('bash', 'fish', 'zsh', 'ps1')
     $ArrayWhen       = @('auto', 'never', 'always')
     $ArrayYesNo      = @('never', 'always')
-    $ArrayWrap       = @('always', 'never', 'character')
+    $ArrayWrap       = @('always', 'never', 'character', 'word')
     $ArrayBinary     = @('no-printing', 'as-text')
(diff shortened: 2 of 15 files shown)


README.md
```diff
@@ -889,6 +889,30 @@ features as a library, check out the [the API documentation](https://docs.rs/bat
 Note that you have to use either `regex-onig` or `regex-fancy` as a feature
 when you depend on `bat` as a library.
 
+### Running syntax tests
```
keith-hall: I personally feel like running syntax tests isn't something most users will want/need to do, therefore I don't think it should be in the readme - it is more of interest only to contributors, so I would suggest to move it to CONTRIBUTING.md instead.

README.md
```diff
@@ -889,6 +889,30 @@ features as a library, check out the [the API documentation](https://docs.rs/bat
...
+`tests/syntax-tests/highlighted/`. These files are used to verify that syntax highlighting works as expected.
+
+- If you only built the binaries with:
+  ```bash
+  cargo build --bins
+  ```
+
+  you need to add the debug build to your PATH before running the tests:
+  ```bash
+  export PATH=$PATH:/path/to/bat/target/debug
```
keith-hall: generally it is recommended to quote the assignment in case `PATH` already contains directories with spaces etc.

```suggestion
  export PATH="$PATH:/path/to/bat/target/debug"
```
Also, we could make it easy (not need adapting), and assume it is being run from `bat`'s project root and suggest something like:

```suggestion
  export PATH="$PATH:$(pwd)/target/debug"
```

keith-hall (changes requested):
Hi, thanks for your contribution. I have a few minor comments, but feel free to disagree.

README.md
foxfromworld: Thanks for the feedback! That makes sense — I’ll move the syntax test instructions from README.md to CONTRIBUTING.md instead.

keith-hall (approved):
Thanks. I wonder if we should also cross reference this from step 5 on https://github.com/sharkdp/bat/blob/master/doc/assets.md#syntax-tests, what do you think?

foxfromworld:
> Thanks. I wonder if we should also cross reference this from step 5 on https://github.com/sharkdp/bat/blob/master/doc/assets.md#syntax-tests, what do you think?

Thanks for the suggestion! I’ve added a cross reference to step 5 in doc/assets.md#syntax-tests as recommended.

PR #3613, opened 2026-03-06
feat(syntax): add support for hidden_file_extensions

### LINKED ISSUE #2227: Read from `.sublime-syntax`hidden_file_extensions

Matei02355:
Closes #2227

This PR adds support for `hidden_file_extensions` from `.sublime-syntax` files. 

The fix strips the leading `.` from hidden dotfiles before looking up the syntax, which allows `syntect`'s extension table to match the hidden files correctly. Verified locally that all tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,17 +5,7 @@
 
 ## Features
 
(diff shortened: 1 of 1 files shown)


keith-hall (approved):
Thanks!

PR #3617, opened 2026-03-07
Add --fallback-syntax for undetected files

Xavrir:
## Summary
- add a new `--fallback-syntax` CLI option (with alias `--fallback-language`) and wire it into config parsing
- apply the fallback only when syntax auto-detection fails (after path + first-line detection), without changing `--language` behavior
- add integration tests for fallback behavior/precedence and update short/long help expectation docs

## Validation
- `cargo build`
- `cargo test`


Xavrir:
no problem man happy to help!

PR #3621, opened 2026-03-08
Fix syntax highlighting for symlinked files by resolving target path

### LINKED ISSUE #1001: no syntax highlighting for symlinked file
See https://github.com/sharkdp/bat/issues/984#issuecomment-627998442. 

bat 0.15.1

`bat ~/.ssh/config` is showing the contents of my ssh_config in all white. This might be because ~/.ssh/config is symlinked (`ln -s`) to a file in my dotfiles repo. 

`bat ~/.ssh/config` should highlight the contents of the file according to https://github.com/robballou/sublimetext-sshconfig. 

Homebrew

---

system
------

**$ uname -srm**
Darwin 19.5.0 x86_64  

**$ sw_vers**
ProductName:	Mac OS X  
ProductVersion:	10.15.5  
BuildVersion:	19F83c  

bat
---

**$ bat --version**
bat 0.15.1  

**$ env**

bat_config
----------

bat_wrapper
-----------

No wrapper script for 'bat'.

bat_wrapper_function
--------------------

tool
----

**$ less --version**

Xavrir:
## Summary

When a symlink name has no recognizable extension (e.g. `Aliases/0install` → `../Formula/zero-install.rb`), syntax detection fails because the symlink path doesn't match any syntax. The target file's `.rb` extension would match Ruby, but was never consulted.

This adds a canonicalization fallback in `get_syntax()`: when path-based syntax detection fails with `UndetectedSyntax`, resolve symlinks via `fs::canonicalize()` and retry detection with the target path.

### Priority chain (preserved):
1. Try syntax by given path (existing — covers cases like `.ssh/config` symlink)
2. **If that fails and the canonicalized path differs, try syntax by resolved target path (NEW)**
3. Try first-line syntax detection (existing fallback)
4. Try `--fallback-syntax` (existing fallback)

## Changes

- `src/assets.rs`: Add symlink resolution fallback in `get_syntax()`, properly propagating `UnknownSyntax` errors from resolved-path detection while only retrying on `UndetectedSyntax`
- New test `syntax_detection_for_symlinked_file_by_target_extension` — mirrors the homebrew-core `Aliases/0install` scenario from the issue
- CHANGELOG entry

## Testing

- 378 tests pass (`cargo test`)
- Existing `syntax_detection_for_symlinked_file` test still passes (`.ssh/config` symlink matched by symlink path, canonicalize fallback never reached)
- New test: `Aliases/0install` (no extension) → `Formula/zero-install.rb` → correctly detects Ruby syntax

## Related

- Closes #1001
- Based on @keith-hall's [experimental fix suggestion](https://github.com/sharkdp/bat/issues/1001#issuecomment-1227847073) using `canonicalize()`
- PR #1019 partially fixed the original report but didn't handle the case where the symlink name has no extension

### DIFF
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -223,18 +223,40 @@ impl HighlightingAssets {
         }
 
         let path = input.path();
-        let path_syntax = if let Some(path) = path {
-            self.get_syntax_for_path(
-                PathAbs::new(path).map_or_else(|_| path.to_owned(), |p| p.as_path().to_path_buf()),
-                mapping,
-            )
+        let absolute_path = path.and_then(|p| {
+            PathAbs::new(p)
+                .ok()
+                .map(|abs| abs.as_path().to_path_buf())
+                .or_else(|| Some(p.to_owned()))
+        });
+
+        let path_syntax = if let Some(ref path) = absolute_path {
+            self.get_syntax_for_path(path, mapping).or_else(|e| {
+                // If syntax detection failed on the given path, retry with the
+                // canonicalized path (which resolves symlinks). This handles
+                // cases like `Aliases/0install -> ../Formula/zero-install.rb`
+                // where the symlink name has no extension but the target does.
+                // See #1001.
+                if matches!(e, Error::UndetectedSyntax(_)) {
+                    if let Ok(resolved) = fs::canonicalize(path) {
+                        if resolved != *path {
+                            return match self.get_syntax_for_path(&resolved, mapping) {
+                                Ok(syntax) => Ok(syntax),
+                                Err(Error::UndetectedSyntax(_)) => Err(e),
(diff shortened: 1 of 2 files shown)


src/assets.rs
```diff
@@ -223,18 +223,38 @@ impl HighlightingAssets {
...
+                }
+                Err(e)
+            })
         } else {
             Err(Error::UndetectedSyntax("[unknown]".into()))
         };
 
         match path_syntax {
-            // If a path wasn't provided, or if path based syntax detection
-            // above failed, we fall back to first-line syntax detection.
```
keith-hall, line 237: maybe this comment is still relevant?

Xavrir: Good catch, restored it.

PR #3636, opened 2026-03-18
Add mapping for ignore files

victor-gp:
This would catch any dotfiles that end in "ignore", but not `.ignore`

For instance, `.dockerignore`, `.npmignore`, `.geminiignore`, etc.

Set as a lower priority rule as @keith-hall suggested in #3635.

I also added a link in the Contribution doc to the Readme for adding syntax mappings.


victor-gp:
Thank you!

PR #3640, opened 2026-03-19
Fix line wrapping for files with control characters

### LINKED ISSUE #3631: bat incorrectly computes text width when option --binary as-text is used

1. have some long enough file which is mostly text but contains unprintable characters(as an example, `/proc/$pid/cmdline`)
2. try to view it with bat, it will complain "binary file"
3. view it with `--binary as-text` flag

**What happens?**

text will wrap incorrectly breaking ASCII art of table view on lines with nonprintable characters, due to `bat` assuming they are width 1 instead of 2(for example, a single character NUL is displayed using 2 chars, '^@').

nice rendering of ASCII table with text inside it.

pacman

---

#### Software version

#### Operating system

- OS: Linux (EndeavourOS rolling)
- Kernel: 6.19.6-arch1-1

#### Command-line

```bash
bat --binary as-text /proc/6071/cmdline --diagnostic 
```

#### Environment variables

```bash
BAT_CACHE_PATH=<not set>
BAT_CONFIG_PATH=<not set>
BAT_OPTS=<not set>
BAT_PAGER=<not set>
BAT_PAGING=<not set>
BAT_STYLE=<not set>
BAT_TABS=<not set>
BAT_THEME=<not set>
COLORTERM=truecolor
LANG=en_US.UTF-8
LC_ALL=<not set>
LESS=<not set>
MANPAGER='sh -c '\''sed "s/\x1B\[[0-9;]*m//g; s/.\x08//g" | bat -p -lman'\'''
NO_COLOR=<not set>
PAGER=<not set>
SHELL=/bin/bash
TERM=xterm-kitty
XDG_CACHE_HOME=<not set>
XDG_CONFIG_HOME=<not set>
```

#### System Config file

#### Config file

Could not read contents of '/home/pavel/.config/bat/config': No such file or directory (os error 2).

#### Custom assets metadata

Could not read contents of '/home/pavel/.cache/bat/metadata.yaml': No such file or directory (os error 2).

#### Custom assets

'/home/pavel/.cache/bat' not found

#### Compile time information

- Profile: release
- Family: unix
- OS: linux
- Architecture: x86_64
- Pointer width: 64
- Endian: little
- CPU features: fxsr,sse,sse2

#### Less version

```
> less --version 

```

eyupcanakman:
Control characters (0x00-0x1F, 0x7F) render as two-column caret notation (^@, ^A, ..., ^?) in the terminal, but `UnicodeWidthChar::width()` returns `None` for them. The wrapping code defaulted to width 0, so wrapping was off for any line with control chars.

Replace the `c.width().unwrap_or(0)` calls with a `char_width()` helper that returns 2 for control characters.

Fixes #3631

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 - Add `--fallback-syntax`/`--fallback-language` to apply syntax highlighting only when auto-detection fails, see #1341 (@Xavrir)
 
 ## Bugfixes
-- Fix incorrect text width computation when using `--binary=as-text` with non-printable characters in caret notation, see #3631 (@eyupcanakman)
+- Fix incorrect text width computation when using `--binary=as-text` with non-printable characters in caret notation, see #3640 and #3631 (@eyupcanakman)
 - Fix `BAT_CONFIG_DIR` pointing at system config directory causing duplicate flag errors. Closes #3589, see #3620 (@Xavrir)
 - Fix syntax highlighting for symlinked files when the symlink name has no extension but the target does. Closes #1001, see #3621 (@Xavrir)
 - Report error when pager is missing instead of silently falling back, see #3588 (@IMaloney)


eyupcanakman:
@keith-hall These are separate paths. `char_width()` is only used with `--binary=as-text`, where control chars go through raw and the terminal shows them as caret notation regardless (2 cols). With `--show-nonprintable`, `replace_nonprintable()` converts them before we get to width calculation, so `char_width()` won't hit any control chars there.

PR #3643, opened 2026-03-20
feat: preserve change markers when combining --diff with --plain

### LINKED ISSUE #3630: Supporting --diff when --plain is set in config
Hello,

I like `bat --plain` as it stays easily copy-pastable but add nice highlights. However I just discovered that `--diff` feature and the fact that it is incompatible with `--plain`.

Here is an example of a diff of a C file with default config:

```
$ bat --diff escseq_on_tty.c
───────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: escseq_on_tty.c
───────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   6   │  *
   7   │  * Compilation:
   8 ~ │  *   cc -O2 escseq_on_tty.c -o escseq_on_tty
   9   │  */
  10   │ #include <ctype.h>
  11   │ #include <stdio.h>
  12 _ │ #include <stdlib.h>
  13   │ #include <unistd.h>
  14   │ 
 ...   │ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ 8< ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─
  18   │ int main(void) {
  19   │   if (isatty(STDOUT_FILENO)) {
  20 + │     // This might be better as an OS level redirect
  21 + │ 
  22   │     int n = 0;
  23   │     char buf[4096];
  24 ~ │     while ((n = fread(buf, 1, 4096, stdin)) > 0) {
  25   │       fwrite(buf, 1, n, stdout);
  26   │     }
 ...   │ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ 8< ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─
  69   │           state = START;
  70   │         } else {
  71 + │           // Regular character
  72   │           return c;
  73   │         }
 ...   │ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ 8< ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─
 148   │   }
 149   │ }
 150 + │ 
 151 + │ // vim: ts=2 sw=2 et
───────┴────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

and here is the result with `--plain`:

```
$ bat --diff --plain escseq_on_tty.c
 *
 * Compilation:
 *   cc -O2 escseq_on_tty.c -o escseq_on_tty
 */
#include <ctype.h>
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>

int main(void) {
  if (isatty(STDOUT_FILENO)) {
    // This might be better as an OS level redirect

    int n = 0;
    char buf[4096];
    while ((n = fread(buf, 1, 4096, stdin)) > 0) {
      fwrite(buf, 1, n, stdout);
    }
          state = START;
        } else {
          // Regular character
          return c;
        }
  }
}

// vim: ts=2 sw=2 et
```

As you can see, as both the side markers and the ellipsis lines are removed, the output become completly useless.

One solution would be to have a plain compatible output, maybe something more similar to `git --diff`: line numbers before the hunks and +/-/~ in the first column.

An other solution would be to automatically or explicitly disable `--plain`.

I though about just passing `--style=full` along with `--diff`, but it seems that ~/.config/bat/config` has precedence over CLI.

Maybe related to https://github.com/sharkdp/bat/issues/1523.

mvanhorn:
## Summary

When `--plain` is set (via CLI or config file), `--diff` output loses all visual markers - no `~`/`+`/`-` change indicators and no `Snip` separators between hunks. The diff line filtering still works (only changed regions are shown), but the output is indistinguishable from plain text.

The root cause is `forced_style_components()` which returns early with only `StyleComponent::Plain` (resolving to an empty decoration set) before `--style` arguments can be parsed. The user's workaround of adding `--style=full` also fails because forced components take priority.

### Fix

When `--plain` is set AND `--diff` is also active, automatically include `Changes` and `Snip` style components. This preserves the `--plain` intent (no grid, no header, no line numbers) while keeping diff output readable.

**Note:** This is a behavior change. I'd welcome feedback on whether this automatic approach is preferred, or if an alternative like a warning message or a new dedicated style component would be better.

## Test plan

- [ ] `bat --diff --plain file.c` shows change markers and snip separators
- [ ] `bat --plain file.c` (without --diff) behavior is unchanged
- [ ] `bat --diff file.c` (without --plain) behavior is unchanged
- [ ] Works when `--plain` is set via `BAT_CONFIG_FILE`
- [ ] `cargo test` passes
- [ ] `cargo clippy` clean

Closes #3630

This contribution was developed with AI assistance (Claude Code).


src/bin/bat/app.rs
```diff
@@ -590,7 +590,16 @@ impl App {
...
         // Plain if `--plain` is specified at least once.
         if self.matches.get_count("plain") > 0 {
-            return Some(StyleComponents(HashSet::from([StyleComponent::Plain])));
+            let mut components = HashSet::from([StyleComponent::Plain]);
+            // When --diff is active, preserve change markers and snip separators
+            // so that diff output remains visually useful.
+            if self.matches.try_contains_id("diff").unwrap_or_default()
+                && self.matches.get_flag("diff")
+            {
+                components.insert(StyleComponent::Changes);
```
keith-hall, line 600: it looks like tests are failing when `bat` is built without the `git` feature, maybe something like this would help:
```suggestion
                #[cfg(feature = "git")]
                components.insert(StyleComponent::Changes);
```

mvanhorn: Fixed in e60875a - added `#[cfg(feature = "git")]` guard on the `Changes` insert. Verified it compiles in both default and `--no-default-features --features minimal-application` configurations.

mvanhorn:
Thanks for the review and merge! The edge case with `--diff --decorations=never` was a good catch.

PR #3652, opened 2026-03-25
Builtin syntax mapping: cleanup matcher glob parsing logic

cyqsimon:
This is a followup on https://github.com/sharkdp/bat/pull/3576#discussion_r2985931494.

- Data flow is now strictly linear from `RawMatcher` to `Matcher`
  - I've also hoisted `RawMatcher` in front of `Matcher` to signal this
- Re-type `RawMatcher.case_sensitive` from `bool` to `Option<bool>`
  - This moves all parser logic away from `RawMatcher`, making it a more faithful representation of the data
- Favour default consts in `Matcher::try_from<RawMatcher>` to `Default` impl on `Case`
  - Because the default choice of casing is a design decision of the logic, not an intrinsic property of the type

I'll admit that all this is just my unnecessary perfectionism, which is possibly becoming increasingly irrelevant these days. But well, I'm old-school like that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@
 - Use git version of cross. See #3533 (@OctopusET)
 - Bump MSRV to 1.88, update `time` crate to 0.3.47 to fix RUSTSEC-2026-0009, see #3581 (@NORMAL-EX)
 - Allow home and end keys to be used with builtin pager, see #3651 (@keith-hall)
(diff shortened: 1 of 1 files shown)


cyqsimon:
Unrelated to this, I've noticed that we can probably eliminate `once_cell` as a dependency now that everything we use from it is included in the standard library well below our MSRV. I'll make a PR for it, queued behind this one.

PR #3653, opened 2026-03-25
Remove dependency on `once_cell`

cyqsimon:
Queued behind #3652. Will rebase once (or if) that one is merged.

`std` now has ~~all~~ almost all the features from `once_cell` we need, and they have been stable well before our current MSRV. I think it's now okay to remove this dependency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -122,6 +122,7 @@
 - Update onig_sys dependency to 69.9.1 to fix a gcc build failure #3400 (@CosmicHorrorDev)
 - Add a cargo feature (`vendored-libgit2`) to build with vendored libgit2 version without depending on the system's one #3426 (@0x61nas)
 - Update syntect dependency to v5.3.0 to fix a few minor bugs, see #3410 (@keith-hall)
+- Remove dependency on once_cell #3653 (@cyqsimon)
 
 ## Syntaxes
 


src/assets.rs
```diff
@@ -90,8 +89,8 @@ impl HighlightingAssets {
 
     /// Return the collection of syntect syntax definitions.
     pub fn get_syntax_set(&self) -> Result<&SyntaxSet> {
-        self.syntax_set_cell
-            .get_or_try_init(|| self.serialized_syntax_set.deserialize())
+        let syntax_set = self.serialized_syntax_set.deserialize()?;
+        Ok(self.syntax_set_cell.get_or_init(|| syntax_set))
```
cyqsimon: Hmm I just realised this is not quite equivalent. But okay the fix should be easy enough.

src/assets.rs
```diff
@@ -90,8 +89,14 @@ impl HighlightingAssets {
...
-        self.syntax_set_cell
-            .get_or_try_init(|| self.serialized_syntax_set.deserialize())
+        // IMPRV: OnceCell::get_mut_or_init is preferable once stabalized
+        match self.syntax_set_cell.get() {
+            Some(set) => Ok(set),
+            None => {
+                let syntax_set = self.serialized_syntax_set.deserialize()?;
+                Ok(self.syntax_set_cell.get_or_init(|| syntax_set))
+            }
+        }
```
cyqsimon, line 99: This works but obviously is less than ideal.

I'll leave this decision up to maintains, whether the benefit of removing an external dependency outweighs the drawback of having this workaround.

PR #3654, opened 2026-03-26
fix: correct $LESSCLOSE exit-status warning

cuiweixie:
## Summary

Fix inverted logic when reporting `$LESSCLOSE` cleanup failures: bat warned on successful exit and stayed silent when the command failed.

## Test plan

- [ ] `cargo test` (lessopen module tests cover env handling)


keith-hall (commented):
Thanks! Do you think it is possible to add an assertion for this in the integration tests, to prove the warning appears only when it should?

PR #3655, opened 2026-03-26
docs: add instructions for removing fish help abbreviations

### LINKED ISSUE #3536: [FR/docs] Add documentation on how to delete fish abbreviation for `bat -plhelp`
I seem to be unable to delete the abbreviation created [based on this](https://github.com/sharkdp/bat?tab=readme-ov-file#highlighting---help-messages). Something like `abbr --erase --help` or `abbr --erase -h` does not work (it will show the help message for abbr!). `abbr --erase \-h` does not work.  `abbr --erase (abbr --list)` does not work. Etc. I'm sure there is probably a way, but currently, I am unaware of it. As I imagine someone else might run into this in the future, I think that adding it to this section would be useful!

Thanks.

claw-explorer:
## Summary

Add documentation showing how to remove the fish abbreviations for `--help` and `-h` that are suggested in the README's "Highlighting `--help` messages" section.

## Problem

As reported in #3536, once a user creates these abbreviations, removing them is non-obvious because the abbreviation names start with dashes. Running `abbr --erase --help` doesn't work — fish interprets `--help` as a flag to `abbr` itself, showing help text instead of erasing the abbreviation.

## Solution

Add a small section showing the correct removal syntax:

```fish
abbr -e -- --help
abbr -e -- -h
```

The `--` (end-of-options marker) before the abbreviation name prevents fish from interpreting the dash-prefixed name as a flag.

Closes #3536

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,16 +244,15 @@ abbr -a --position anywhere -- --help '--help | bat -plhelp'
 abbr -a --position anywhere -- -h '-h | bat -plhelp'
 ```
 
-To remove these abbreviations later:
-
-```fish
-abbr -e -- --help
-abbr -e -- -h
-```
-
-The `--` before the abbreviation name is required because `--help` and `-h` start with dashes, which would otherwise be interpreted as flags to `abbr` itself.
-
 This way, you can keep on using `cp --help`, but get colorized help pages.
+
+> [!TIP]
+> To remove these abbreviations later, run:
+> ```fish
+> abbr -e -- --help
+> abbr -e -- -h
+> ```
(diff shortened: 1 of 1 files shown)


keith-hall (commented):
Maybe it is worth putting these instructions in a callout or something to separate it from the main flow, otherwise it doesn't quite feel like a natural placement for the new instructions, to me at least, it reads like "to remove the abbreviations later, do this. This way, you can keep on using `cp --help`, but get colorized help pages." (but removing abbreviations wouldn't allow this.)

Issue #1145, opened 2020-08-19
MANPAGER won't work on Mandoc's man implementation


Bumbadawg:
**What version of `bat` are you using?**
0.15.4

**Describe the bug you encountered:**
Using `export MANPAGER="sh -c 'col -bx | bat -l man -p'"`
results in `-bx: 1: Syntax error: Unterminated quoted string`

**Describe what you expected to happen?**
paging with bat

**How did you install `bat`?**
**Void Linux** package manager

** The bug **
It was already reported here for **Termux**, and it coudl well be the case for **Alpine** and other Mandoc's man implementations cases.
See https://github.com/termux/termux-packages/issues/4781#issuecomment-573110911

** Lazy solution **
Instead of recompiling man from GNU (https://www.nongnu.org/man-db/development.html)
one can be using `export MANPAGER='nvim +Man!'`

sharkdp:
Thank you for reporting this.

This is not a bug in `bat`. This implementation of man is currently not supported. This should probably be documented.

lahwaacz:
It does not work because mandoc does not do shell tokenization of the `$MANPAGER` string, it just splits it into a list of words and passes them to `execvp`. So the error comes from `sh` which complains about an unterminated quoted string from the word `'col`.

`bat` itself does not have any problems displaying the output of mandoc's `man` command, you just have to configure the pager appropriately. This can be done by putting the following in a wrapper script and then `export MANPAGER="your_wrapper"`:
```
#!/bin/sh
cat "$1" | col -bx | bat --language man --style plain
```

eth-p:
> ```
> #!/bin/sh
> cat "$1" | col -bx | bat --language man --style plain
> ```

Not that performance matters much for a shell script like this, but you can even cut out the cat invocation:

```shell
col -bx < "$1" | bat --language man -p
```

ju1ius:
> col -bx < "$1" | bat --language man -p

Or:
```sh
#!/bin/sh
exec col -bx | bat -pl man

lahwaacz:
> > col -bx < "$1" | bat --language man -p
> 
> Or:
> 
> ```shell
> #!/bin/sh
> exec col -bx | bat -pl man
> ```

[...]

Issue #667, opened 2019-09-20
README suggests suboptimal MANPAGER config


qstrahl:
Hello! Thank you for the lovely software.

 I noticed that when setting `MANPAGER="sh -c 'col -b | bat -l man'"` as recommended by the README, there were some minor formatting issues - I'm still investigating possible solutions to most of them, but one of them at least has a very simple solution, which is to pass the `-x` switch to `col` (in addition to the recommended `-b`).

I found that the default behaviour of `col` was to take some dubious artistic liberties with whitespace, particularly noticeable in the very first and last lines of most (all?) man pages, the header and footer. The `-x` switch tells `col` to output multiple spaces instead of tab characters, which ameliorated the alignment issues for me.

sharkdp:
Thank you for reporting this. This is a duplicate of (or related to) #652. I should have already updated the README with the `col -bx` part (did that now in 4102175d981b495c377fd621eb46498b1b77a9f6).

If you find more things which should be corrected, please feel free to open a new ticket or comment here. Thanks!

PR #3684, opened 2026-04-08
docs: fix README instructions on how to use bat with man

chances190:
Update the documentation for using `bat` as a colorizing pager for `man`. The current setup sometimes causes man to break the highlighting and print raw escape characters into bat.

print of the error

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,10 +192,11 @@ bat main.cpp | xclip
 #### `man`
 
 `bat` can be used as a colorizing pager for `man`, by setting the
-`MANPAGER` environment variable:
+`MANPAGER` and `MANROFFOPT` environment variables:
 
 ```bash
-export MANPAGER="bat -plman"
+export MANPAGER="sh -c 'col -bx | bat -plman'"
+export MANROFFOPT="-c"
 man 2 select
 ```
 (on some older Debian or Ubuntu releases, the executable is named `batcat` instead of `bat`)


keith-hall:
Thanks for your contribution. I'm pretty sure we used to suggest exactly this but had to change it due to incompatibilities...

chances190:
Sorry for not checking the history. Apparently it was this way at first

45f3c733 Extend 'man' integration section

````diff
--- a/README.md
+++ b/README.md
@@ -143,11 +143,17 @@
 #### `man`

-`bat` can be used as a pager for `man`. You can use [`batman`](https://github.com/eth-p/bat-extras/blob/master/doc/batman.md), or like this in your shell:
+`bat` can be used as a colorizing pager for `man`, by setting the
+`MANPAGER` environment variable:

 ```bash
 export MANPAGER="sh -c 'col -bx | bat -l man -p'"
 man 2 select
 ```

-The [Manpage syntax](assets/syntaxes/Manpage.sublime-syntax) is developed in this repository and still needs some work.
+It might also be necessary to set `MANROFFOPT="-c"` if you experience
+formatting problems.
+
+If you prefer to have this bundled in a new command, you can also use [`batman`](https://github.com/eth-p/bat-extras/blob/master/doc/batman.md).
+
+Note that the [Manpage syntax](assets/syntaxes/Manpage.sublime-syntax) is developed in this repository and still needs some work.

4102175d Add '-x' flag for 'col', see #652, see #667
````


Then, someone changed it for an one-liner to strip the escape chacaters

8453bc93 Update README steps for using bat as MANPAGER
````diff
--- a/README.md
+++ b/README.md
@@ -201,20 +201,20 @@
 #### `man`

 `bat` can be used as a colorizing pager for `man`, by setting the

 ```bash
-export MANPAGER="sh -c 'col -bx | bat -l man -p'"
+export MANPAGER="sh -c 'sed -u -e \"s/\\x1B\[[0-9;]*m//g; s/.\\x08//g\" | bat -p -lman'"
 man 2 select
 ```
 (replace `bat` with `batcat` if you are on Debian or Ubuntu)

-It might also be necessary to set `MANROFFOPT="-c"` if you experience
-formatting problems.
-
 If you prefer to have this bundled in a new command, you can also use [`batman`](https://github.com/eth-p/bat-extras/blob/master/doc/batman.md).

-Note that the [Manpage syntax](assets/syntaxes/02_Extra/Manpage.sublime-syntax) is developed in this repository and still needs some work.
+> [!WARNING]
+> This will [not work](https://github.com/sharkdp/bat/issues/1145) out of the box with Mandoc's `man` implementation.
+>
+> Please either use `batman`, or convert the shell script to a [shebang executable](https://en.wikipedia.org/wiki/Shebang_(Unix)) and point `MANPAGER` to that.

-Also, note that this will [not work](https://github.com/sharkdp/bat/issues/1145) with Mandocs `man` implementation.
+Note that the [Manpage syntax](assets/syntaxes/02_Extra/Manpage.sublime-syntax) is developed in this repository and still needs some work.

5d92a7c5 docs: fix up README
 ````
 

 This script was changed a bit a few times and ended up getting removed when #3517 introduced code to remove escape characters:
 
59c58969 Simplify MANPAGER guide
 ````diff
--- a/README.md
+++ b/README.md
@@ -192,20 +192,15 @@
 #### `man`

 `bat` can be used as a colorizing pager for `man`, by setting the

 ```bash
-export MANPAGER="sh -c 'awk '\''{ gsub(/\x1B\[[0-9;]*m/, \"\", \$0); gsub(/.\x08/, \"\", \$0); print }'\'' | bat -p -lman'"
+export MANPAGER="bat -plman"
 man 2 select
 ```
 (replace `bat` with `batcat` if you are on Debian or Ubuntu)

 If you prefer to have this bundled in a new command, you can also use [`batman`](https://github.com/eth-p/bat-extras/blob/master/doc/batman.md).

-> [!WARNING]
-> This will [not work](https://github.com/sharkdp/bat/issues/1145) out of the box with Mandoc's `man` implementation.
->
-> Please either use `batman`, or convert the shell script to a [shebang executable](https://en.wikipedia.org/wiki/Shebang_(Unix)) and point `MANPAGER` to that.
-
 Note that the [Manpage syntax](assets/syntaxes/02_Extra/Manpage.sublime-syntax) is developed in this repository and still needs some work.

 ````

I'm now unsure of what's happening. That PR should have fixed this, but as you can see in my screenshot, it didn't.
The two ways it displays the correct output for me are:
1. `MANPAGER="sh -c 'col -bx | bat -plman'" MANROFFOPT='-c' man 2 select`
2. `man 2 select | bat -plman`

keith-hall:
The latest changes haven't been released, could that be the reason for the mismatch? How did you install `bat`? The most reliable way to check current behavior would be to install from source. FYI: There are plans to release a new version soon, once a new syntect (the syntax highlighting crate `bat` uses) release is published, which seems like it could happen reasonably soon.

chances190:
Built from source: bat 0.26.1 (v0.25.0-628-gfa2b9a50)

I found the problem! man only generates overstrikes when `MANROFFOPT='-c'` is set. Otherwise, it generates ANSI output. While that PR is correct, it's an edge case caused by setting an specific env var.

Here's a comparison with the select 2 manpage:
Screenshot From 2026-04-09 12-24-06
Screenshot From 2026-04-09 12-29-53

Should I update the docs to `MANROFFOPT='-c'` or to `MANPAGER='bat -plman --strip-ansi=always'` ?

keith-hall:
Thanks for investigating! It feels like we have the code in `bat` to support either case, so perhaps we can offer both in the README - for those users whom like to have `MANROFFOPT='-c'` for other reasons, or because it is the default on their system, they could choose to use the first approach, while others can use the `--strip-ansi` way. What do you think?

PR #3687, opened 2026-04-10
Support BAT_WIDTH for terminal width

### LINKED ISSUE #3679: Support `BAT_WIDTH` to set `--terminal-width` size as an alternative of `COLUMNS`
Ideally this would override `COLUMNS` as it's more specific.

Loosely related to #3660

officialasishkumar:
## Summary
- add `BAT_WIDTH` as an environment-backed alias for `--terminal-width`
- surface `BAT_WIDTH` in `--diagnostic` output and the terminal-width help text
- add integration coverage for env-driven width selection and normal CLI override precedence

## Testing
- `cargo test --lib`
- `cargo test --test integration_tests`

Closes #3679

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 - Preserve `--diff` change markers and snip separators when `--plain` is set. Closes #3630, see #3643 (@mvanhorn)
 - Added support for `hidden_file_extensions` from `.sublime-syntax` files, see #3613 (@Matei02355)
 - Add word wrapping mode via `--wrap=word`, see #3597 (@veeceey)
-- Support configuring `--terminal-width` via `BAT_WIDTH`, see #0000 (@officialasishkumar)
+- Support configuring `--terminal-width` via `BAT_WIDTH`, see #3679 (@officialasishkumar)
 - Implement `--unbuffered` mode for streaming input, allowing partial lines to display immediately (e.g. `tail -f | bat -u`). Closes #3555, see #3583 (@mainnebula)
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 - Map `BUILD` case sensitively to Python (Starlark) for Bazel, see #3576 (@vorburger)
 
 ## Bugfixes
+- Support configuring `--terminal-width` via `BAT_WIDTH`, see #3687 (@officialasishkumar)
```
keith-hall: I would say that this isn't a bugfix, so the changelog entry should be under a different heading

keith-hall (commented):
In general looks fine, but I do have one question: is there a good reason for using a new dedicated environment variable as opposed to just using `BAT_OPTS`?

CHANGELOG.md
officialasishkumar: Moved the changelog entry under `Features` while rebasing onto current `master`.

officialasishkumar:
On the `BAT_OPTS` question: I kept a dedicated `BAT_WIDTH` var because it matches the existing option-specific env vars (`BAT_TABS`, `BAT_THEME`, `BAT_PAGER`, `BAT_STYLE`) and makes terminal width discoverable in the same place. `BAT_OPTS` still works when someone wants to bundle multiple defaults together.

keith-hall (approved):

5HT2:
Thank you!! :)

PR #3688, opened 2026-04-12
fix: sanitize subprocess call in generate_snapshots.py

orbisai0security:
## Summary
Fix critical severity security issue in `tests/snapshots/generate_snapshots.py`.

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | V-001 |
| **Severity** | CRITICAL |
| **Scanner** | multi_agent_ai |
| **Rule** | `V-001` |
| **File** | `tests/snapshots/generate_snapshots.py:32` |

**Description**: The Python test script uses subprocess.call() with shell=True, which executes commands through the system shell and interprets shell metacharacters. At line 32, a 'command' variable is passed directly to the shell without sanitization. If this variable contains or is influenced by external data (environment variables, configuration files, command-line arguments), attackers can inject arbitrary shell commands using operators like semicolons, pipes, or command substitution.

## Changes
- `tests/snapshots/generate_snapshots.py`

## Verification
- [x] Build passes
- [x] Scanner re-scan confirms fix
- [x] LLM code review passed

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*

### DIFF
--- a/tests/snapshots/generate_snapshots.py
+++ b/tests/snapshots/generate_snapshots.py
@@ -4,6 +4,7 @@ import itertools
 import subprocess
 import pathlib
 import shutil
+from typing import Iterable
 
 
 def generate_snapshots():
@@ -19,14 +20,14 @@ def generate_snapshots():
 
 
 def generate_style_snapshot(style):
-    generate_snapshot(style.replace(",", "_"), "--style={}".format(style))
+    generate_snapshot(style.replace(",", "_"), ["--style={}".format(style)])
 
 
-def generate_snapshot(name, arguments):
+def generate_snapshot(name: str, arguments: Iterable[str]):
     output_file = "output/{name}.snapshot.txt".format(name=name)
     command = [
         "cargo", "run", "--", "--paging=never", "--color=never",
-        "--decorations=always", arguments, "sample.rs"
+        "--decorations=always", *arguments, "sample.rs"
     ]
     print("generating snapshot for {}".format(name))
     with open(output_file, "w") as f:
--- a/tests/snapshots/generate_snapshots.py
+++ b/tests/snapshots/generate_snapshots.py
@@ -23,18 +23,19 @@ def generate_style_snapshot(style):
 
 
 def generate_snapshot(name, arguments):
-    command = "cargo run -- --paging=never --color=never --decorations=always "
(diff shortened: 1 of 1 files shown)


tests/snapshots/generate_snapshots.py
```diff
@@ -23,18 +23,19 @@ def generate_style_snapshot(style):
...
 def generate_snapshot(name, arguments):
-    command = "cargo run -- --paging=never --color=never --decorations=always "
-    command += "{args} sample.rs > output/{name}.snapshot.txt".format(
-        name=name,
-        args=arguments
-    )
+    output_file = "output/{name}.snapshot.txt".format(name=name)
+    command = [
+        "cargo", "run", "--", "--paging=never", "--color=never",
+        "--decorations=always", arguments, "sample.rs"
```
keith-hall: If `arguments` is a string and contains a space, this may not work as expected. Maybe we need to provide type hinting and ensure it is an iterable of strings and expand it like:
```suggestion
        "--decorations=always", *arguments, "sample.rs"
```
And update the call site(s) accordingly.

keith-hall (commented):
Thanks for your contribution. One small concern before merging, see the inline comment.

tests/snapshots/generate_snapshots.py
anupamme: @orbisai0security can you address code review comments?

orbisai0security:
✅ **Changes Applied**

I've updated the code based on your feedback:

Addressing @keith-hall's review comment: the `arguments` parameter in `generate_snapshot` is currently a plain string inserted as a single list element. If it contained a space (e.g., a flag with a value like `--some-flag value`), it would be passed as one token to the subprocess instead of two separate arguments, causing incorrect behavior.

The fix:
1. Add `from typing import Iterable` and add type hints `(name: str, arguments: Iterable[str])` to `generate_snapshot`.
2. Use `*arguments` to unpack the iterable into the command list.
3. Update the call site in `generate_style_snapshot` to pass `["--style={}".format(style)]` (a list) instead of a bare string.

**Files modified:**
- `tests/snapshots/generate_snapshots.py`

The changes have been pushed to this PR branch. Please review!

keith-hall (approved):

PR #3689, opened 2026-04-13
[Python] Support uv as script runner in shebang

janlarres:
This is a backport from upstream: https://github.com/sublimehq/Packages/blob/0fb98d39560bc4806ccf18da3c16e4a39ab805c0/Python/Python.sublime-syntax#L4131


keith-hall (approved):
Thanks. Note that work is being done on syntect, such that we may not need these backports for long

PR #3691, opened 2026-04-18
sanitize filenames

curious-rabbit:
- File header banner (src/printer.rs); a filename containing ANSI/OSC bytes (e.g. $'\e[2Jpwned.txt') was painted directly to the terminal when bat displayed the file's header in interactive mode (the default). It could change colors, clear the screen, or emit further escape sequences.
- Terminal title (src/bin/bat/main.rs); with --set-terminal-title, the filename was wrapped in an OSC \x1b]0;…\x07 sequence. A \x07 (BEL) inside the filename terminated the OSC early; a nested \x1b]0; let the attacker start a new one.
- Binary-file warning (src/printer.rs); the "[bat warning] Binary content from …" message echoed description.summary() (the filename) verbatim to the terminal.
- File-open errors (src/input.rs); four error branches interpolated the raw path into the message written to stderr.

Note: please consider adding a security policy to enable githubs feature to report vulnerabilities in private


keith-hall (approved):
Nice, thanks.

PR #3693, opened 2026-04-19
fix(zsh): fix word-splitting on language names with spaces in tab completion

### LINKED ISSUE #2897: Zsh tab complete with --language flag is broken

1. Open a zsh shell with tab completion enabled
2. Type `bat -l` and press the tab key three times to get into the completion menu

**What happens?**

The actual text in the prompt is incorrect, as it displays a string of various other characters, rather than the actual result expected

For tab complete to function normally, selecting the text in the menu and placing it into the prompt, looking more something akin to `$ bat -lbash`

pacman

#### Software version

bat 0.24.0 (fc954685)

#### Operating system

Linux 6.7.9-arch1-1

#### Command-line

```bash
bat --diagnostic 
```

#### Environment variables

```bash
SHELL=/bin/zsh
PAGER=<not set>
LESS=<not set>
LANG=en_US.UTF-8
LC_ALL=<not set>
BAT_PAGER=<not set>
BAT_PAGING=<not set>
BAT_CACHE_PATH=<not set>
BAT_CONFIG_PATH=<not set>
BAT_OPTS=<not set>
BAT_STYLE=<not set>
BAT_TABS=<not set>
BAT_THEME=<not set>
XDG_CONFIG_HOME=/home/jade/.config
XDG_CACHE_HOME=/home/jade/.cache
COLORTERM=truecolor
NO_COLOR=<not set>
```

#### System Config file

#### Config file

```
# This is `bat`s configuration file. Each line either contains a comment or
# a command-line option that you want to pass to `bat` by default. You can
# run `bat --help` to get a list of all possible configuration options.

# Specify desired highlighting theme (e.g. "TwoDark"). Run `bat --list-themes`
# for a list of all available themes
--theme="tokyonight_night"

# Enable this to use italic text on the terminal. This is not supported on all
# terminal emulators (like tmux, by default):
--italic-text=always

# Uncomment the following line to disable automatic paging:
#--paging=never

# Uncomment the following line if you are using less version >= 551 and want to
# enable mouse scrolling support in `bat` when running inside tmux. This might
# disable text selection, unless you press shift.
#--pager="less --RAW-CONTROL-CHARS --quit-if-one-screen --mouse"

# Syntax mappings: map a certain filename pattern to a language.
#   Example 1: use the C++ syntax for .ino files
#   Example 2: Use ".gitignore"-style highlighting for ".ignore" files
#--map-syntax "*.ino:C++"
#--map-syntax ".ignore:Git Ignore"
--style=full
--color=always
--decorations=always
```

#### Custom assets metadata

```
bat_version: 0.24.0
creation_time:
  secs_since_epoch: 1709359655
  nanos_since_epoch: 606552614
```

#### Custom assets

- metadata.yaml, 97 bytes
- syntaxes.bin, 915557 bytes
- themes.bin, 44563 bytes

#### Compile time information

- Profile: release
- Family: unix
- OS: linux
- Architecture: x86_64
- Pointer width: 64
- Endian: little
- CPU features: fxsr,sse,sse2

#### Less version

```
> less --version 
less 643 (PCRE2 regular expressions)
Copyright (C) 1984-2023  Mark Nudelman

```

YoshKoz:
## Problem

Fixes #2897

When pressing Tab after `bat -l` (or `bat --language=`), language names containing spaces—such as `HTML (Jinja2)`, `Apache Conf`, `ARM Assembly`—are split into separate tokens by the shell's word-splitting on `$()` command substitution. This produces garbage completion entries like `(Jinja2)`, `Conf`, `Assembly`, etc.

Root cause: line 93 of `bat.zsh.in` uses `$()` which word-splits on `$IFS` (includes spaces) even when `local IFS=$'\n'` is set, because IFS only affects word-splitting of _unquoted variable expansions_, not command substitutions assigned directly to an array.

## Fix

Replace `$()` with `${(f)"$(...)"}` — the zsh flag `(f)` splits the result on newlines only. This is the exact pattern already used for theme completions on line 100:

```zsh
# Before (line 93):
languages=( $(bat --list-languages | awk ...) )

# After:
languages=( ${(f)"$(bat --list-languages | awk ...)"} )
```

One-character-class change, consistent with the existing theme completion pattern.

## Test

```zsh
bat -l <Tab>
# Should now complete "HTML (Jinja2)", "Apache Conf", etc. as single entries
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@
 - Fixed test compatibility with future Cargo build directory changes, see #3550 (@nmacl)
 - Fixed bug caused by using `--plain` and `--terminal-width=N` flags simultaneously, see #3529 (@H4k1l)
 - Fixed syntax tests path, see #3610 (@foxfromworld)
+- Fix zsh tab completion word-splitting language names containing spaces (e.g. `HTML (Jinja2)`, `Apache Conf`), see #3693 (@YoshKoz)
 
 ## Other
 - Use git version of cross. See #3533 (@OctopusET)
--- a/assets/completions/bat.zsh.in
+++ b/assets/completions/bat.zsh.in
@@ -90,7 +90,7 @@ _{{PROJECT_EXECUTABLE}}_main() {
         languages)
             local IFS=$'\n'
             local -a languages
-            languages=( $({{PROJECT_EXECUTABLE}} --list-languages | awk -F':|,' '{ for (i = 1; i <= NF; ++i) printf("%s:%s\n", $i, $1) }') )
+            languages=( ${(f)"$({{PROJECT_EXECUTABLE}} --list-languages | awk -F':|,' '{ for (i = 1; i <= NF; ++i) printf("%s:%s\n", $i, $1) }')"} )
 
             _describe 'language' languages && ret=0
         ;;


keith-hall (approved):
Nice, thanks. Feel free to add a note in the bugfix section of the unreleased area of the changelog.

PR #3700, opened 2026-04-25
Fixed hardcoded terminal probing when `--list-themes` is called

optimistiCli:
# Fix: `--list-themes` ignores `--theme` and always probes the terminal

## Problem

`bat --list-themes` unconditionally queried the terminal for its background
color via OSC 10/11 escape sequences, regardless of what `--theme` was set to.
This caused display corruption in terminals and multiplexers that do not
support OSC 10/11 (e.g. GNU Screen), even when the user had explicitly chosen
a fixed theme such as `--theme=dark` or `--theme=Dracula`.

## Root cause

This was a regression introduced in commit bc42149a ("Merge color scheme
options into theme / BAT\_THEME"). Before that commit, `list_themes` was
called with the user's actual color-scheme preference
(`app.color_scheme_preference()`). During the refactor that merged
`--color-scheme` into `--theme`, the call site was silently changed to the
hardcoded value `ColorSchemePreference::default()` (equivalent to
`DetectColorScheme::Auto`), which always attempts a terminal probe when stdout
is a tty — completely ignoring the user's `--theme` setting.

## Fix

`list_themes` now receives the full `ThemeOptions` (the same struct used
everywhere else in bat) and computes the `(default)` marker via the existing
`theme()` function. Since `theme()` already handles all `ThemePreference`
variants correctly — only probing for `Auto`, never for `Dark`, `Light`, or
`Fixed` — this restores the pre-regression behaviour and makes the listing
consistent with how bat resolves the theme for normal output.

Changes:

- `App::theme_options` visibility raised to `pub(crate)` so `main.rs` can
  call it at the `--list-themes` site.
- `list_themes` parameter changed from `DetectColorScheme` to `ThemeOptions`.
- Default-theme resolution changed from
  `default_theme(color_scheme(detect_color_scheme).unwrap_or_default())` to
  `theme(theme_options).to_string()`.
- Call site updated to pass `app.theme_options()` instead of the hardcoded
  `DetectColorScheme::default()`.

## Testing

```
# Should print the theme list with no terminal probe (no OSC 11 sent):
bat --list-themes --theme=dark
bat --list-themes --theme=Dracula
bat --list-themes --theme=light

# Should still probe and mark the correct default:
bat --list-themes
bat --list-themes --theme=auto
bat --list-themes --theme=auto:always
```

To verify that no probe is sent, run inside GNU Screen or capture terminal
traffic; with `--theme=dark` or a named theme, no `\e]11;?\a` sequence should
appear.

**PS:** Human here (-: The fix and the text above are courtesy of an LLM. I inspected the diffs, did some test runs and to the best of my knowledge the fix is valid.


keith-hall:
Thanks for your contribution. I'm a little unsure about this change from the perspective that I would consider the default theme to be whatever `bat` would use with default/no custom configuration, not whatever the user has configured... Curious to know what others think / what their expectations would be? I can see the benefit of showing which is the currently configured theme (and I would expect it to be marked as such, as opposed to being referred to as the default), so this change definitely has some merit...

optimistiCli:
If I may, pertaining to the behaviour in question, by default `bat` afaik queries the terminal to figure out whether the user has a light or dark theme activated, thus the default is to follow user's preference. Setting the `--theme` is imho a strong indication user's preference.

keith-hall (approved):
Okay, let's do it, thanks - if users really want to see the actual default they can presumably run `--list-themes --no-config` 🙂

PR #3703, opened 2026-04-26
Replace libgit2 with gitoxide (#2)

blinxen:
This is a continuation of https://github.com/sharkdp/bat/pull/3235

This time all tests pass without any changes :D Looks promising!

**Context**: This PR replaces the usage of `libgit2` with a full Rust implementation of `git`. This was tried [before](https://github.com/sharkdp/bat/pull/3235) but a couple of issues were found in the Rust implementation of `git`. These issues seem to be fixed now. Therefore, I am opening this again and would like to see what the maintainers think.  

CC: @byron

### DIFF
--- a/src/diff.rs
+++ b/src/diff.rs
@@ -1,10 +1,13 @@
 #![cfg(feature = "git")]
 
+use gix::diff::blob::pipeline::{Mode, WorktreeRoots};
+use gix::diff::blob::{Algorithm, HunkIter, ResourceKind};
+use gix::index::hash::Kind;
+use gix::object::tree::EntryKind;
+use gix::{self, ObjectId};
+use path_abs::PathInfo;
 use std::collections::HashMap;
-use std::fs;
 use std::path::Path;
-
-use git2::{DiffOptions, IntoCString, Repository};
 
 #[derive(Copy, Clone, Debug)]
 pub enum LineChange {
@@ -16,68 +19,70 @@ pub enum LineChange {
 
 pub type LineChanges = HashMap<u32, LineChange>;
 
-pub fn get_git_diff(filename: &Path) -> Option<LineChanges> {
-    let repo = Repository::discover(filename).ok()?;
-
-    let repo_path_absolute = fs::canonicalize(repo.workdir()?).ok()?;
-
-    let filepath_absolute = fs::canonicalize(filename).ok()?;
-    let filepath_relative_to_repo = filepath_absolute.strip_prefix(&repo_path_absolute).ok()?;
-
-    let mut diff_options = DiffOptions::new();
-    let pathspec = filepath_relative_to_repo.into_c_string().ok()?;
-    diff_options.pathspec(pathspec);
-    diff_options.context_lines(0);
-
-    let diff = repo
-        .diff_index_to_workdir(None, Some(&mut diff_options))
-        .ok()?;
-
-    let mut line_changes: LineChanges = HashMap::new();
-
-    let mark_section =
-        |line_changes: &mut LineChanges, start: u32, end: i32, change: LineChange| {
-            for line in start..=end as u32 {
-                line_changes.insert(line, change);
+fn collect_changes_from_hunks(hunks: HunkIter) -> Option<LineChanges> {
+    let mut changes: LineChanges = HashMap::new();
+    for hunk in hunks {
+        if hunk.before.is_empty() && !hunk.after.is_empty() {
+            for line in hunk.after {
+                changes.insert(line + 1, LineChange::Added);
             }
-        };
-
-    let _ = diff.foreach(
-        &mut |_, _| true,
-        None,
-        Some(&mut |delta, hunk| {
-            let path = delta.new_file().path().unwrap_or_else(|| Path::new(""));
-
-            if filepath_relative_to_repo != path {
-                return false;
-            }
-
-            let old_lines = hunk.old_lines();
-            let new_start = hunk.new_start();
-            let new_lines = hunk.new_lines();
-            let new_end = (new_start + new_lines) as i32 - 1;
-
-            if old_lines == 0 && new_lines > 0 {
-                mark_section(&mut line_changes, new_start, new_end, LineChange::Added);
-            } else if new_lines == 0 && old_lines > 0 {
-                if new_start == 0 {
-                    mark_section(&mut line_changes, 1, 1, LineChange::RemovedAbove);
-                } else {
-                    mark_section(
-                        &mut line_changes,
-                        new_start,
-                        new_start as i32,
-                        LineChange::RemovedBelow,
-                    );
-                }
+        } else if hunk.after.is_empty() && !hunk.before.is_empty() {
+            if hunk.after.start == 0 {
+                changes.insert(1, LineChange::RemovedAbove);
             } else {
-                mark_section(&mut line_changes, new_start, new_end, LineChange::Modified);
+                changes.insert(hunk.after.start, LineChange::RemovedBelow);
             }
+        } else {
+            for line in hunk.after {
+                changes.insert(line + 1, LineChange::Modified);
+            }
+        }
+    }
 
-            true
-        }),
-        None,
+    Some(changes)
+}
+
+pub fn get_git_diff(filename: &Path) -> Option<LineChanges> {
+    let filepath_absolute = filename.canonicalize().ok()?;
+    let repository = gix::discover(filepath_absolute.parent().ok()?).unwrap();
+    let repo_path_absolute = repository.workdir()?.canonicalize().ok()?;
+    let filepath_relative_to_repo = filepath_absolute.strip_prefix(&repo_path_absolute).ok()?;
+    let mut cache = repository
+        .diff_resource_cache(
+            Mode::ToGit,
+            WorktreeRoots {
+                old_root: None,
+                new_root: repository.workdir().map(Path::to_path_buf),
+            },
+        )
+        .ok()?;
+    cache
+        .set_resource(
+            repository
+                .head_tree()
+                .ok()?
+                .lookup_entry_by_path(filepath_relative_to_repo.to_str()?)
+                .ok()??
+                .object_id(),
+            EntryKind::Blob,
+            filepath_relative_to_repo.to_str()?.into(),
+            ResourceKind::OldOrSource,
+            &repository,
+        )
+        .ok()?;
+    cache
+        .set_resource(
+            ObjectId::null(Kind::Sha1),
+            EntryKind::Blob,
+            filepath_relative_to_repo.to_str()?.into(),
+            ResourceKind::NewOrDestination,
+            &repository,
+        )
+        .ok()?;
+    let diff = gix::diff::blob::diff_with_slider_heuristics(
+        Algorithm::Myers,
+        &cache.prepare_diff().ok()?.interned_input(),
     );
 
-    Some(line_changes)
+    collect_changes_from_hunks(diff.hunks())
 }
--- a/src/diff.rs
+++ b/src/diff.rs
@@ -44,9 +44,9 @@ pub fn get_git_diff(filename: &Path) -> Option<LineChanges> {
     let filepath_absolute = filename.canonicalize().ok()?;
     let repository = gix::discover(filepath_absolute.parent().ok()?).ok()?;
     let repo_path_absolute = repository.workdir()?.canonicalize().ok()?;
-    let filepath_relative_to_repo = gix::path::to_unix_separators_on_windows(
-        gix::path::into_bstr(filepath_absolute.strip_prefix(&repo_path_absolute).ok()?)
-    );
+    let filepath_relative_to_repo = gix::path::to_unix_separators_on_windows(gix::path::into_bstr(
+        filepath_absolute.strip_prefix(&repo_path_absolute).ok()?,
+    ));
     let index = repository.index_or_load_from_head_or_empty().ok()?;
     let index_entry = index.entry_by_path(filepath_relative_to_repo.as_ref())?;
     let mut cache = repository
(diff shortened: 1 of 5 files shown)


Byron:
This is great to see, thanks so much 🙏.

The one thing that seems critical here is to have an example of a diff that (now) looks the same in `bat` as it looks with `git`, and ideally different with the previous implementation based on an older `gix`.

Unfortunately, whenever I try to find such an example, I come up empty 🤦‍♂️.

And before I forget please do set the default `diff.algorithm` to histogram, via `gix::open::Options::isolated().config_overrides(Some("diff.algorithm=histogram"))`. It is very good and most importantly doesn't seem to be attackable, so cannot be tricked into long run times like `myers`. For context, the `gix- merge` crate was fuzzed and the fuzzer was going after the diff step which picked up Myers as default, causing diff-times of 1s which ultimately timed out certain fuzzer runs. `histogram` took 50ms for the same input.

blinxen:
> The one thing that seems critical here is to have an example of a diff that (now) looks the same in bat as it looks with git

This should already be proven by the tests, even though it is not that big of a diff.

> ideally different with the previous implementation based on an older gix

Previous PR should show this difference since it required me to change some test files.

> And before I forget please do set the default diff.algorithm to histogram, via gix::open::Options::isolated().config_overrides(Some("diff.algorithm=histogram")). It is very good and most importantly doesn't seem to be attackable, so cannot be tricked into long run times like myers. For context, the gix- merge crate was fuzzed and the fuzzer was going after the diff step which picked up Myers as default, causing diff-times of 1s which ultimately timed out certain fuzzer runs. histogram took 50ms for the same input.

:+1:

blinxen:
Is there actually any need to call `gix::open::Options::isolated().config_overrides(Some("diff.algorithm=histogram"))`? We are calling `diff_with_slider_heuristics` directly and we pass the algorithm there.

Byron:
Then there is no need, it can be set there.

My clanker came up with something that seems to be worth taking a look at.

> - Restored the documented index-vs-worktree semantics by loading the old diff resource from the Git index entry rather than `HEAD`.

This might be wrong, but if so I think it makes sense to document the current behaviour more heavily. The new regression tests seem to cover the altered behaviour though.

I only took a quick look, but enough to hand it off for more diligence.

#### Review by GPT 5.5

Resolved the review findings in `src/diff.rs`.

- Replaced `gix::discover(...).unwrap()` with fallible handling so `--diff`/change markers outside a Git repository return `None` instead of crashing.
- Restored the documented index-vs-worktree semantics by loading the old diff resource from the Git index entry rather than `HEAD`.
- Kept the worktree file as the new diff resource.
- Added regression coverage for:
  - ordinary files outside Git repositories,
  - staged changes producing no worktree-vs-index markers,
  - newly added indexed files with further unstaged edits.

Validation run:

```sh
cargo fmt --check
cargo test diff::tests --features git
cargo check --features git
cargo test diff_plain --features git
```

#### Patch

```patch
diff --git a/src/diff.rs b/src/diff.rs
index 07b79dc4..cbb2d8d4 100644
--- a/src/diff.rs
+++ b/src/diff.rs
@@ -2,14 +2,12 @@
 
 use gix::diff::blob::pipeline::{Mode, WorktreeRoots};
 use gix::diff::blob::{Algorithm, HunkIter, ResourceKind};
-use gix::index::hash::Kind;
 use gix::object::tree::EntryKind;
-use gix::{self, ObjectId};
 use path_abs::PathInfo;
 use std::collections::HashMap;
 use std::path::Path;
 
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub enum LineChange {
     Added,
     RemovedAbove,
@@ -44,9 +42,13 @@ fn collect_changes_from_hunks(hunks: HunkIter) -> Option<LineChanges> {
 
 pub fn get_git_diff(filename: &Path) -> Option<LineChanges> {
     let filepath_absolute = filename.canonicalize().ok()?;
-    let repository = gix::discover(filepath_absolute.parent().ok()?).unwrap();
+    let repository = gix::discover(filepath_absolute.parent().ok()?).ok()?;
     let repo_path_absolute = repository.workdir()?.canonicalize().ok()?;
     let filepath_relative_to_repo = filepath_absolute.strip_prefix(&repo_path_absolute).ok()?;
+    let filepath_relative_to_repo =
+        gix::path::to_unix_separators_on_windows(gix::path::into_bstr(filepath_relative_to_repo));
+    let index = repository.index_or_load_from_head_or_empty().ok()?;
+    let index_entry = index.entry_by_path(filepath_relative_to_repo.as_ref())?;
     let mut cache = repository
         .diff_resource_cache(
             Mode::ToGit,
@@ -58,23 +60,18 @@ pub fn get_git_diff(filename: &Path) -> Option<LineChanges> {
         .ok()?;
     cache
         .set_resource(
-            repository
-                .head_tree()
-                .ok()?
-                .lookup_entry_by_path(filepath_relative_to_repo.to_str()?)
-                .ok()??
-                .object_id(),
-            EntryKind::Blob,
-            filepath_relative_to_repo.to_str()?.into(),
+            index_entry.id,
+            index_entry.mode.to_tree_entry_mode()?.kind(),
+            filepath_relative_to_repo.as_ref(),
             ResourceKind::OldOrSource,
             &repository,
         )
         .ok()?;
     cache
         .set_resource(
-            ObjectId::null(Kind::Sha1),
+            repository.object_hash().null(),
             EntryKind::Blob,
-            filepath_relative_to_repo.to_str()?.into(),
+            filepath_relative_to_repo.as_ref(),
             ResourceKind::NewOrDestination,
             &repository,
         )
@@ -86,3 +83,72 @@ pub fn get_git_diff(filename: &Path) -> Option<LineChanges> {
 
     collect_changes_from_hunks(diff.hunks())
 }
+
+#[cfg(test)]
+mod tests {
+    use super::{get_git_diff, LineChange};
+    use std::path::Path;
+    use std::process::Command;
+
+    fn git(repo: &Path, args: &[&str]) {
+        let output = Command::new("git")
+            .args(args)
+            .current_dir(repo)
+            .output()
+            .expect("git command can run");
+        assert!(
+            output.status.success(),
+            "git {args:?} failed\nstdout: {}\nstderr: {}",
+            String::from_utf8_lossy(&output.stdout),
+            String::from_utf8_lossy(&output.stderr)
+        );
+    }
+
+    fn setup_repo() -> tempfile::TempDir {
+        let dir = tempfile::tempdir().expect("can create temporary directory");
+        let repo = dir.path();
+        git(repo, &["init"]);
+        git(repo, &["config", "user.email", "test@test.test"]);
+        git(repo, &["config", "user.name", "Test"]);
+        dir
+    }
+
+    #[test]
+    fn no_git_diff_outside_repository() {
+        let dir = tempfile::tempdir().expect("can create temporary directory");
+        let file = dir.path().join("file.txt");
+        std::fs::write(&file, "line 1\n").expect("can write file");
+
+        assert_eq!(get_git_diff(&file), None);
+    }
+
+    #[test]
+    fn git_diff_compares_worktree_to_index() {
+        let repo = setup_repo();
+        let file = repo.path().join("file.txt");
+        std::fs::write(&file, "line 1\nline 2\n").expect("can write file");
+        git(repo.path(), &["add", "file.txt"]);
+        git(repo.path(), &["commit", "-m", "initial"]);
+
+        std::fs::write(&file, "line 1\nline 2 modified\n").expect("can write file");
+        git(repo.path(), &["add", "file.txt"]);
+
+        assert_eq!(get_git_diff(&file), Some(Default::default()));
+    }
+
+    #[test]
+    fn git_diff_uses_index_for_newly_added_files() {
+        let repo = setup_repo();
+        std::fs::write(repo.path().join("committed.txt"), "committed\n").expect("can write file");
+        git(repo.path(), &["add", "committed.txt"]);
+        git(repo.path(), &["commit", "-m", "initial"]);
+
+        let file = repo.path().join("new.txt");
+        std::fs::write(&file, "line 1\nline 2\n").expect("can write file");
+        git(repo.path(), &["add", "new.txt"]);
+        std::fs::write(&file, "line 1\nline 2 modified\n").expect("can write file");
+
+        let changes = get_git_diff(&file).expect("newly added file has an index entry");
+        assert_eq!(changes.get(&2), Some(&LineChange::Modified));
+    }
+}
```

blinxen:
Not sure if we need to care about windows paths (I expected `gix` to handle it) but I applied the suggestion from GPT. I also liked the tests it added, it was a good start. I think I covered most standard cases.

src/diff.rs
```diff
@@ -16,68 +17,165 @@ pub enum LineChange {
...
+
+pub fn get_git_diff(filename: &Path) -> Option<LineChanges> {
+    let filepath_absolute = filename.canonicalize().ok()?;
+    let repository = gix::discover(filepath_absolute.parent().ok()?).ok()?;
+    let repo_path_absolute = repository.workdir()?.canonicalize().ok()?;
+    let filepath_relative_to_repo = gix::path::to_unix_separators_on_windows(gix::path::into_bstr(
+        filepath_absolute.strip_prefix(&repo_path_absolute).ok()?,
+    ));
+    let index = repository.index_or_load_from_head_or_empty().ok()?;
+    let index_entry = index.entry_by_path(filepath_relative_to_repo.as_ref())?;
```
Byron, line 51: It is true that these `relative_path: &BStr` parameters expect to use slashes, so on Windows some processing is necessary.
`gix-index` is very plumbing and unfortunately that's directly exposed right now.

blinxen: Can you add a warning in the documentation so it is more clear for anyone using it? Or maybe `gix` should wrap it and handle the windows case automatically :thinking: Not sure which makes more sense for `gix`.

blinxen:
I think this PR is open for review again.

keith-hall (approved):
Looks good to me, thanks! 🚀

Byron:
Do we have performance comparisons of the before/after states?
A quick hyperfine run would probably do just to be sure.

blinxen:
I rebased this PR again to master and ran hyperfine. Not sure though how representative it is.


Current release:

```bash
hyperfine -N --warmup 3 "bat README.md"
Benchmark 1: bat README.md
  Time (mean ± σ):       3.4 ms ±   0.3 ms    [User: 1.6 ms, System: 2.5 ms]
  Range (min … max):     2.9 ms …   4.4 ms    844 runs
```

This PR:

```bash
hyperfine -N --warmup 3 "target/release/bat README.md"
Benchmark 1: target/release/bat README.md
  Time (mean ± σ):       3.2 ms ±   0.3 ms    [User: 1.6 ms, System: 2.3 ms]
  Range (min … max):     2.7 ms …   4.5 ms    880 runs
```

Byron:
For future runs, you could also do `hyperfine -N --warmup 3 "target/release/bat README.md" "bat README.md"` to get a built-in comparison.

But from what I can tell, it's basically the same. This isn't necessarily surprising unless the work done for diffing becomes substantial.

This reminds me that you probably also want a test that sets the `repositoryVersion` in the configuration to 42 and asserts that `bat` falls back to a version without Git without failure. While the code should do that, it feels important enough to have insurance.

blinxen:
> This reminds me that you probably also want a test that sets the repositoryVersion in the configuration to 42 and asserts that bat falls back to a version without Git without failure. While the code should do that, it feels important enough to have insurance.

This does not seem to do anything.

I added this test

```rust

    #[test]
    fn faulty_git_version_does_not_panic() {
        let repo = setup_repo();
        let file = create_and_track_file(&repo, "file.txt");
        std::fs::write(&file, "line 1\nline 2 modified\n").expect("can write file");
        assert_eq!(get_git_diff(&file).expect("one change").len(), 2);

        git(repo.path(), &["config", "core.repositoryformatversion", "42"]);
        println!("{}", std::fs::read_to_string(repo.path().to_string_lossy().to_string() + "/.git/config").unwrap());

        assert_eq!(get_git_diff(&file), None);
    }
```

and get this output:

```
failures:

---- diff::tests::faulty_git_version_does_not_panic stdout ----
[core]
        repositoryformatversion = 42
        filemode = true
        bare = false
        logallrefupdates = true
[user]
        email = test@test.test
        name = Test


thread 'diff::tests::faulty_git_version_does_not_panic' (51388) panicked at src/diff.rs:192:9:
assertion `left == right` failed
  left: Some({2: Modified, 1: Modified})
 right: None
```

Should `gix` have behaved differently?

Byron:
I see, thanks for trying it out!

After looking at the code, I noticed that [it's very lenient](https://github.com/GitoxideLabs/gitoxide/blob/3b4e4363693b61dc47f1de3ca6fe7ecabed6a894/gix/src/config/cache/incubate.rs#L43-L49) and notably different from Git (see below the fold).

However, you should be able to fail it with

```gitconfig
[core]
        repositoryformatversion = one
```

**Then the original intention would be fulfilled: what happens if the Git repository won't open for whichever reason.**

Thanks again.

<details><summary>Codex</summary>
<p>

I confirmed with the local Git binary:

```text
core.repositoryformatversion = 0
extensions.objectformat = sha1
```

fails with:

```text
fatal: repo version is 0, but v1-only extension found:
	objectformat
```

So Git’s behavior is:

- `version = 0`, no `extensions.objectFormat`: SHA-1, allowed.
- `version = 1`, no `extensions.objectFormat`: SHA-1, allowed.
- `version = 0`, `extensions.objectFormat = sha1` or `sha256`: rejected.
- `version = 1`, valid `extensions.objectFormat`: accepted.

The Gitoxide snippet is more lenient because for `repo_format_version != 1` it returns SHA-1 immediately and does not reject a present `extensions.objectFormat`.

</p>
</details>

blinxen:
This seems to work. Thanks! The PR should be pretty complete now.

blinxen:
I rebased this PR to master again and updated to gix 0.85. Can it get merged soon since it got approved? The failing CI is probably a GitHub Runner issue. The log just says "Error:" and nothing else.

PR #3708, opened 2026-04-30
fix: git diff detection now works in worktree directories

### LINKED ISSUE #1979: git integration lacks worktree awareness
Git has worktrees which allow one clone to have multiple working areas.

Say you have clone to directory "foo" and then you create a worktree.
`git worktree add ../bar`

Move to `../bar` make some changes.  See them with `git diff`.  However `bat` doesn't notice that this directory tree is a git working area and doesn't detect the diff.

e.g. `bat -d <filename>`

`bat` is git aware in the main clone directory, but not in separate worktree directories.

Bat version: 0.18.2
Environment: OpenSUSE Tumbleweed
Installed via distro package: `zypper install bat`

toller892:
Fixes #1979

## Problem
`bat -d` fails to detect git changes in worktree directories. The `get_git_diff()` function uses `repo.workdir()` to compute the relative file path, but for git worktrees this returns the main repository's workdir, not the worktree directory. The file path then fails `strip_prefix()`, causing bat to silently not show diffs.

As noted by @sharkdp in https://github.com/sharkdp/bat/issues/1979#issuecomment-1008174442, the problem is in `src/diff.rs` lines 20-25.

## Fix
Replace `repo.workdir()` with a `find_repo_root()` helper that walks up from the file to locate `.git`. This correctly handles both:
- Regular repos (where `.git` is a directory)
- Worktrees (where `.git` is a file pointing to the main repo)

## Changes
- Added `find_repo_root()` helper function
- Removed dependency on `repo.workdir()` for path computation
- File: `src/diff.rs` (+21/-4 lines)

### DIFF
--- a/src/diff.rs
+++ b/src/diff.rs
@@ -2,7 +2,7 @@
 
 use std::collections::HashMap;
 use std::fs;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 use git2::{DiffOptions, IntoCString, Repository};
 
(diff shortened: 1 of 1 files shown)


keith-hall:
Thanks. Note that [libgit2 may be replaced with a pure Rust alternative soon](https://github.com/sharkdp/bat/pull/3703), so it would be great to have an integration test which proves work trees work, to ensure it doesn't get regressed accidentally

toller892:
@keith-hall good point about regression testing with the libgit2 replacement. An integration test for worktree diff detection could follow the pattern in `tests/tester/mod.rs`:

1. Create a git repo with `Repository::init()`
2. Create a worktree via `repo.worktree("name", path, None)` (git2 provides this)
3. Modify a tracked file in the worktree
4. Run `bat --diff` on the modified file
5. Assert the output contains diff markers

I do not have a Rust toolchain in this environment to add the test, but a maintainer or another contributor could add it following this approach. The code change itself (+21/-4 in `src/diff.rs`) is minimal and mechanically correct — `find_repo_root()` walks up from the file to locate `.git`, which handles both regular repos and worktrees.

Issue #3488, opened 2025-11-15
Document what the different styles do


jonatan-ivanov:
`man bat` (0.26.0) says the following:
```
By default, the following components are enabled: changes, grid, header-filename, numbers, snip

Possible values: *default*, full, auto, plain, changes, header, header-filename, header-filesize, grid, rule, numbers, snip.
```

The `default` style is explained but it is obscure what `auto` or `full` and what's the difference between the three of them (especially `auto`'s behavior is environment-dependent). I guess right now the only way to figure out is reading the code:
https://github.com/sharkdp/bat/blob/1e4a4b765f4a1cad74f3a4204112268f3b303f2f/src/style.rs#L24-L63

I think some of the components are also not necessarily clear without an explanation (e.g.: `rule`, `snip`)

keith-hall:
This is perhaps a duplicate of https://github.com/sharkdp/bat/issues/3228, it would be nice to keep all related discussion in one place 🙂

jonatan-ivanov:
Thank you! I think it is. Let me close it and comment there instead.

PR #3709, opened 2026-04-30
docs(README): list available --style components in Output style section

### LINKED ISSUE #3228: List of potential styles?
The documentation has many examples of style options, but theres no list.  It would be nice if the documentation had a list to know what styles are available to enable or disable.

mvanhorn:
## Summary

The "Output style" section explains how to use `--style` but does not enumerate what the values are, so users either need to know them or read the man page / source. This PR adds two tables to that section listing the four pre-defined styles and the eight individual components, with the descriptions copied verbatim from `bat --help`.

## Why this matters

Issue #3228 (and dup #3488) report the same gap: every other code-shape thing in the README around `--style=numbers,changes` is named, but the components themselves are not introduced anywhere on the page. The man page has them and `bat --help` has them, but neither is discoverable from a quick scroll through the README. The reporter explicitly says "either you know them or you dig through the source code." Adding the lists inline keeps the explanation next to the example that uses them.

## Changes

`README.md` "Output style" section:

- One sentence calling out the default-enabled set (`changes, grid, header-filename, numbers, snip`) so the existing "by default" mention in `bat --help` has a home in the README.
- A 4-row "pre-defined styles" table (`default`, `full`, `auto`, `plain`) with the same one-line descriptions `bat --help` prints.
- An 8-row "individual components" table (`changes`, `header`, `header-filename`, `header-filesize`, `grid`, `rule`, `numbers`, `snip`) with the same descriptions.

The existing tip block about `+`/`-` modifier prefixes is unchanged.

## Testing

Every value and description was verified against `bat --help` output on bat 0.26.1 locally. The README markdown was eyeballed for table rendering. No code paths were touched.

Closes #3228

### DIFF
--- a/README.md
+++ b/README.md
@@ -507,6 +507,30 @@ and line numbers but no grid and no file header. Set the `BAT_STYLE` environment
 variable to make these changes permanent or use `bat`'s
 [configuration file](#configuration-file).
 
+By default, `bat` enables `changes`, `grid`, `header-filename`, `numbers`, and `snip`.
+
+The available pre-defined styles are:
+
+| Style | Description |
+|-------|-------------|
+| `default` | Enables the recommended style components listed above. |
+| `full` | Enables all available components. |
+| `auto` | Same as `default`, unless the output is piped. |
+| `plain` | Disables all available components. |
+
+The available individual components are:
+
+| Component | Description |
+|-----------|-------------|
+| `changes` | Show Git modification markers. |
+| `header` | Alias for `header-filename`. |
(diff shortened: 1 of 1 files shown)


keith-hall (approved):
Thanks

injust:
Mentioning this here in case someone else was also wondering.

Without `--force-colorization`: The `default` and `auto` styles behave the same when piped (no colorization or decorations).

With `--force-colorization`: Both styles colorize, but only the `default` style outputs decorations.

So the difference is that `auto` "forces" `--decorations=never` when piped.

PR #3723, opened 2026-05-04
[bat] Fix ctrl-c exiting less when BAT_PAGER is set

### LINKED ISSUE #3444: ctrl-c still exits less, even when bat isn't passing `--quit-on-intr`

mitre88:
When a user explicitly sets BAT_PAGER environment variable, they expect their pager configuration to be respected. The -K (--quit-on-intr) flag was being added to less even when BAT_PAGER was set, causing ctrl-c to exit less immediately instead of allowing the user to scroll.

This fix ensures -K is only added when the pager comes from the generic PAGER environment variable (which defaults to 'less' on most systems), not when the user explicitly configures BAT_PAGER, --pager, or config.

Fixes #3444

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -151,7 +151,11 @@ impl OutputType {
 
                 // Ensures that 'less' quits together with 'bat'
                 // The BusyBox version of less does not support -K
-                if less_version != Some(LessVersion::BusyBox) {
+                // Only add -K if the pager came from the generic PAGER env var,
+                // not if it came from BAT_PAGER, --pager, or config
+                if less_version != Some(LessVersion::BusyBox)
+                    && pager.source == PagerSource::EnvVarPager
+                {
                     p.arg("-K"); // Short version of '--quit-on-intr'
                 }
 


src/output.rs
```diff
@@ -151,7 +151,11 @@ impl OutputType {
 
                 // Ensures that 'less' quits together with 'bat'
                 // The BusyBox version of less does not support -K
-                if less_version != Some(LessVersion::BusyBox) {
+                // Only add -K if the pager came from the generic PAGER env var,
+                // not if it came from BAT_PAGER, --pager, or config
+                if less_version != Some(LessVersion::BusyBox)
+                    && pager.source == PagerSource::EnvVarPager
```
keith-hall, line 157: Would it make sense to reuse the existing variable here?
```suggestion
                    && replace_arguments_to_less
```

keith-hall (commented):
Thanks for your contribution. I think we might have some integration tests covering `less` argument behavior, it would be great to add some test cases to ensure this doesn't get regressed.

Also please feel free to add an entry in the changelog.

injust:
First, you'd need to update https://github.com/sharkdp/bat/blob/master/README.md#using-less-as-a-pager, because this changes the behaviour to never passing `--quit-on-intr` when $BAT_PAGER is set.

injust:
Second, I don't think this would actually fix https://github.com/sharkdp/bat/issues/3444. Please read through the discussion there. It's been some time, but it seems that less still exits on ctrl-c, even if bat is _not_ passing `--quit-on-intr`.

Could you please confirm that less is not exiting for you on ctrl-c after this change, as long as $BAT_PAGER is set?

injust:
And please retitle the PR to something like "Stop passing `--quit-on-intr` to less when using $BAT_PAGER", because that's actually the change.

mitre88:
Thanks for the detailed review. After rereading #3444 and the prior debugging there, I agree this PR is not a sufficient fix: it only changes when `-K` is added, but the reported issue can still happen when bat is not passing `-K` at all. I also cannot honestly confirm that this keeps `less` usable after Ctrl-C with `BAT_PAGER='less -R'`.

Closing this rather than adding docs/changelog around an incomplete behavior change.

PR #3725, opened 2026-05-05
fix sanitation for --strip-ansi

curious-rabbit:
`--strip-ansi=always` left two classes of ANSI escape sequence in the output:

- **8-bit C1 introducers** (U+0090, U+0098, U+009B, U+009D, U+009E, U+009F). On terminals that interpret 8-bit C1 in UTF-8 (kitty, by default; older xterm and VTE configurations), these are the single-codepoint equivalents of `ESC P`, `ESC X`, `ESC [`, `ESC ]`, `ESC ^`, `ESC _` . They introduce DCS, SOS, CSI, OSC, PM, and APC sequences respectively. Bat's parser treated them as text.
- **DCS / SOS / PM / APC bodies**. Even when introduced by the 7-bit `ESC P/X/^/_`, the body up to the string terminator was emitted as text and survived `strip_ansi`.

This patch teaches `EscapeSequenceOffsetsIterator` to recognise the 8-bit introducers as their 7-bit equivalents, and to consume string-terminated bodies (via either form of introducer) as a single opaque `Unknown` segment. Both then drop out of `strip_ansi` along with the existing CSI/OSC handling.

A small unit test in `src/preprocessor.rs` covers the three new cases (8-bit CSI, 7-bit DCS body, 8-bit DCS body with 8-bit ST). 

Note:
--strip-ansi=always is bypassed entirely when bat selects SimplePrinter (i.e. when stdout is piped or --color=never is set), so bat --strip-ansi=always file | grep … **returns the raw escape sequences**. 
The existing strip_ansi_does_not_affect_simple_printer test seems to lock this in deliberately. is that still the intended scope, or shoudl this be fixed? strip-ansi=always implies it always filters but thats not true

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 - Syntax highlighting for Python files using uv as script runner in shebang #3689 (@janlarres)
 
 ## Bugfixes
+- Strip 8-bit C1 introducers (U+0090, U+0098, U+009B, U+009D, U+009E, U+009F) and DCS/SOS/PM/APC sequence bodies in `--strip-ansi=always`. Previously these passed through as text, see #3725 (@curious-rabbit)
 - Fix `--list-themes` unconditionally probing the terminal via OSC 10/11 even when `--theme` was set to an explicit value, see #3700 (regression introduced in bc42149a). (@optimistiCli)
 - Fix inverted `$LESSCLOSE` warning so bat warns on nonzero exit, not on success. See #3654 (@cuiweixie)
 - Sanitize control characters in filenames before displaying them in the file header, error messages, and the terminal title, preventing ANSI escape injection via crafted filenames. Closes #3054, see #3691 (@curious-rabbit)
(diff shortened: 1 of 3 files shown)


keith-hall:
> Note:
> --strip-ansi=always is bypassed entirely when bat selects SimplePrinter (i.e. when stdout is piped or --color=never is set), so bat --strip-ansi=always file | grep … **returns the raw escape sequences**.
> The existing strip_ansi_does_not_affect_simple_printer test seems to lock this in deliberately. is that still the intended scope, or shoudl this be fixed? strip-ansi=always implies it always filters but thats not true

I believe the intention is to leave the output unchanged when piping in scripts (and `cat` is an alias to `bat` and `strip-ansi` comes from the config file instead of the command line...)
Probably `bat` could benefit from a refactor which would always honor command line arguments, as this is quite a common problem I think, but maybe it makes sense to leave it as it is for this PR, what do you think?

curious-rabbit:
Closed in favor of #3729

PR #3726, opened 2026-05-05
fix command injection in LESS template

curious-rabbit:
`%s` in a $LESSOPEN or $LESSCLOSE template was replaced with the filename and the result was passed to `sh -c`, so a filename like `; rm -rf ~ ;` was executed as a command. Wrap the filename in POSIX single quotes before substituting.

### DIFF
--- a/src/lessopen.rs
+++ b/src/lessopen.rs
@@ -14,6 +14,26 @@ use crate::{
     input::{Input, InputKind, InputReader, OpenedInput, OpenedInputKind},
 };
 
+/// Wrap `s` in POSIX single quotes so it cannot break out of a shell argument.
+fn shell_quote(s: &str) -> String {
+    let mut quoted = String::with_capacity(s.len() + 2);
+    quoted.push('\'');
+    for c in s.chars() {
+        if c == '\'' {
+            quoted.push_str("'\\''");
+        } else {
+            quoted.push(c);
+        }
+    }
+    quoted.push('\'');
+    quoted
+}
+
+/// Substitute the first `%s` in a $LESSOPEN/$LESSCLOSE template with a quoted value.
+fn shell_substitute(template: &str, replacement: &str) -> String {
+    template.replacen("%s", &shell_quote(replacement), 1)
+}
+
 /// Preprocess files and/or stdin using $LESSOPEN and $LESSCLOSE
 pub(crate) struct LessOpenPreprocessor {
     lessopen: String,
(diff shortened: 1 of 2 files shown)


src/lessopen.rs
```diff
@@ -384,4 +404,43 @@ mod tests {
...
+        assert_eq!(super::shell_quote("$(payload)"), "'$(payload)'");
+        assert_eq!(super::shell_quote("`payload`"), "'`payload`'");
+        assert_eq!(super::shell_quote("a|b&c;d"), "'a|b&c;d'");
+        assert_eq!(super::shell_quote("..>/dev/null"), "'..>/dev/null'");
+        assert_eq!(super::shell_quote("\n\t"), "'\n\t'");
+    }
+
+    #[test]
+    fn shell_quote_embedded_single_quote() {
+        assert_eq!(super::shell_quote("it's"), "'it'\\''s'");
```
keith-hall, line 455: Is this really the correct escaping of a single quote for the shell? Why not just
```suggestion
        assert_eq!(super::shell_quote("it's"), r"'it\'s'");
```

Is it because escape characters are ignored inside single quoted strings, so we need to end the string scope, escape the quote and then reapply the string scope afterwards?

src/lessopen.rs
```diff
@@ -181,15 +201,15 @@ impl LessOpenPreprocessor {
                         lessclose: self
                             .lessclose
                             .as_ref()
-                            .map(|s| s.replacen("%s", &path_str, 1).replacen("%s", &stdout, 1)),
+                            .map(|s| shell_substitute(&shell_substitute(s, &path_str), &stdout)),
```
keith-hall: This is perhaps an existing bug, but if `path_str` contains `%s` it would get replaced in the second call instead of the "original" second occurrence of `%s`. Do you think it is worth fixing that at the same time?

keith-hall (commented):
Thanks for your contribution. I have a couple of small questions before I approve 🙂

src/lessopen.rs
```diff
@@ -14,6 +14,26 @@ use crate::{
     input::{Input, InputKind, InputReader, OpenedInput, OpenedInputKind},
 };
 
+/// Wrap `s` in POSIX single quotes so it cannot break out of a shell argument.
+fn shell_quote(s: &str) -> String {
+    let mut quoted = String::with_capacity(s.len() + 2);
+    quoted.push('\'');
+    for c in s.chars() {
```
keith-hall, line 29: Would it be more efficient to search for quote characters instead of checking and pushing one by one? If there are no quotes for example, the whole string can be pushed in one go

curious-rabbit: `shell_quote` now pushes the whole string in one go when there's no `'` in it.

src/lessopen.rs
curious-rabbit: Fixed

src/lessopen.rs
curious-rabbit: Yes, that's exactly why. `'...'` has no escape character, so we close, write `\'`, and reopen

keith-hall (approved):

PR #3728, opened 2026-05-05
Fix README Fedora section links

Daeraxa:
The modular repository has been retired as of Fedora 39 (https://fedoraproject.org/wiki/Changes/RetireModularity). `bat` is now present in `updates` (https://docs.fedoraproject.org/en-US/quick-docs/fedora-repositories/#the-updates-repository).

This PR removes the link to the (now dead/404 page for modularity) and makes it the same as the others. I didn't think it necessary to include the details of it being in `updates` as it isn't something that needs to be enabled by default.

It also updates the link to the Fedora package from the build system page to the proper package page (that still links back to the build system) and feels more unified with the links to other distro packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,7 +307,7 @@ pacman -S bat
 
 ### On Fedora
 
-You can install [the `bat` package](https://koji.fedoraproject.org/koji/packageinfo?packageID=27506) from the official sources:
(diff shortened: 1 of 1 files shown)


keith-hall (approved):
Thanks

PR #3736, opened 2026-05-09
fix: detect mintty/MSYS2 terminals for colored output

### LINKED ISSUE #3034: bat --help not showing color from version 0.23.0-1
I am using MinGW x64
1. Install bat `0.22.1-1` and run command `bat --help`. Color shows as expected
2. Install bat `0.23.0-1` and run command `bat --help`. See just black and white
image
image

Download packages from [msys2-mingw64-repo](https://repo.msys2.org/mingw/mingw64/)
 ```
pacman -U mingw-w64-x86_64-bat-0.22.1-1-any.pkg.tar.zst
pacman -U mingw-w64-x86_64-bat-0.23.0-1-any.pkg.tar.zst
```

---

[diagnostic.txt](https://github.com/user-attachments/files/16325846/diagnostic.txt)

### PR DESCRIPTION
**Pull request by Metbcy**
## Summary

Fixes #3034.

On MinGW/MSYS2 terminals (mintty), `std::io::stdout().is_terminal()` returns `false` because mintty uses POSIX PTYs that Windows's `isatty()` does not recognize. This caused `bat --help` (and regular file display) to lose colored output starting with v0.23.0, when help rendering moved from clap's built-in output to bat's own syntax-highlighting pipeline gated on `interactive_output`.

## Changes

Adds a fallback terminal detection check for mintty/MSYS2 environments on Windows by looking for:

- `TERM_PROGRAM=mintty` (set by mintty itself)
- `MSYSTEM` env var (set by all MSYS2 shells: MINGW64, UCRT64, CLANG64, etc.)

When either is present and `is_terminal()` returns `false`, the output is still treated as interactive.

The check is `#[cfg(windows)]` gated and compiles to a no-op on other platforms.

## Testing

Verified all existing tests pass (232 passed, 0 failed).

To manually test on a MinGW/MSYS2 terminal:
1. Open mintty (e.g. Git Bash or MSYS2 shell)
2. Run `bat --help` and confirm colored output is shown
3. Run `bat --color=never --help` and confirm plain output
4. Run `bat somefile.rs` and confirm syntax highlighting works

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 - Syntax highlighting for Python files using uv as script runner in shebang #3689 (@janlarres)
 
 ## Bugfixes
+- Detect mintty/MSYS2 terminals on Windows where `is_terminal()` incorrectly returns false, restoring colored output for `bat --help` and normal file display. Closes #3034, see #3736 (@Metbcy)
 - Quote filenames before substituting them into `$LESSOPEN` / `$LESSCLOSE` templates, preventing shell injection when a filename contains shell metacharacters, see #3726 (@curious-rabbit)
 - Fix `--list-themes` unconditionally probing the terminal via OSC 10/11 even when `--theme` was set to an explicit value, see #3700 (regression introduced in bc42149a). (@optimistiCli)
 - Fix inverted `$LESSCLOSE` warning so bat warns on nonzero exit, not on success. See #3654 (@cuiweixie)
--- a/src/bin/bat/app.rs
+++ b/src/bin/bat/app.rs
@@ -59,6 +59,13 @@ impl App {
 
         let interactive_output = std::io::stdout().is_terminal();
 
+        // On mintty/MSYS2 terminals, is_terminal() returns false because
+        // mintty uses POSIX PTYs that Windows isatty() does not recognize.
+        // We detect this environment to allow --color=auto to still produce
+        // colored output, but we do NOT override interactive_output itself
+        // so that piping (bat file | grep foo) still works correctly. See #3034.
+        let is_mintty = Self::is_mintty_terminal();
+
         // Check if the -n / --number option was passed on the command line
         // (before merging with config file and environment variables).
         // This is needed to honor the -n flag when piping output, similar to `cat -n`.
@@ -107,7 +114,7 @@ impl App {
             let use_color = match matches.get_one::<String>("color").map(|s| s.as_str()) {
                 Some("always") => true,
                 Some("never") => false,
-                _ => interactive_output, // auto: use color if interactive
+                _ => interactive_output || is_mintty,
             };
 
             let pager = matches.get_one::<String>("pager").map(|s| s.as_str());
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by keith-hall** [comment]
Thanks for your contribution. Doesn't this mean that when using mintty style terminals, it will always be treated as interactive mode? That seems undesirable...What if the user has aliased `cat` and is piping to another program? By default `bat` detects that and adds no decorations or colours etc.

**Comment by Metbcy** [comment]
Good catch @keith-hall. You're right, the original version would have forced interactive mode globally, which would break piping (`bat file | grep foo` would get unwanted colors/decorations).

I did a quick update to the approach: the mintty detection now only affects the `--color=auto` decision, not the `interactive_output` flag itself. So piping still works correctly (no decorations, no colors) while direct terminal use in mintty gets colored output as expected.

The key change:
- `interactive_output` stays purely based on `is_terminal()`
- `use_color` in auto mode uses `interactive_output || is_mintty`

This means mintty users get colored output when viewing files directly, but piped output stays plain.

I'm newer to Rust, so thanks for the call-out, it helps my learning 👍

**Comment by tats-u** [comment]
Did you report it also in https://github.com/rust-lang/rust/issues? I believe it should be handled in Rust's std.

**Comment by wyf027** [comment]
Code review (macOS, non-Windows): the split between `interactive_output` and `use_color` for mintty looks correct after @keith-hall's feedback — piping should stay non-interactive while `--color=auto` can still enable colors on mintty/MSYS2.

CI is green. Would be great to get a Windows/mintty maintainer pass before merge. Thanks @Metbcy!

**Comment by CAOShurong** [comment]
I tested this on native Windows x86_64 at the current head `d65807048a6272ecb1535e336d871395a1c24308`, built with the project's MSRV (Rust 1.88.0, MSVC). The remaining pipe case is specific to the help path: `MSYSTEM` is present in Git Bash whether stdout is a terminal or a pipe, so it cannot distinguish those states by itself.

Using byte-level checks for ANSI ESC characters:

```text
                                      base f3d0773   head d658070
MSYSTEM=MINGW64, redirected auto help       0             656
MSYSTEM=MINGW64, redirected color=never     0               0
MSYSTEM=MINGW64, redirected auto file       0               0
MSYSTEM=MINGW64, redirected color=always   32              32
```

The real Git Bash pipeline reproduces the same result:

```text
bat --paging=never --color=auto --help | cat   # exit 0, 656 ESC bytes
bat --paging=never --style=plain --color=auto sample.rs | cat
                                                # exit 0, 0 ESC bytes
```

The existing targeted pipe regressions also pass (`5 passed`), so they do not cover this `MSYSTEM` + help combination. This confirms that the earlier piping concern remains for `--help`, even though ordinary file piping is still plain.

One other scope detail: `is_mintty` is currently consumed only by the help branch's `use_color`; normal file colorization later in `app.rs` still uses `self.interactive_output`. So the current diff does not appear to implement the PR description's “regular file display” part.

Could the fallback be kept inactive when stdout is redirected/piped, with a Windows regression for `MSYSTEM=MINGW64` plus captured stdout? The PR scope or implementation should also be aligned regarding regular file display. I made no changes to the PR; these results are against the pinned base/head above.

PR #3746, opened 2026-05-16
fix: --wrap=never breaks less --quit-if-one-screen (#3738)

### LINKED ISSUE #3738: `bat --wrap=never` breaks `less --quit-if-one-screen`

1. Find a file `foo` that a long line (i.e. longer than terminal width) but fits on the screen (i.e. shorter than terminal height)
2. `LESS=--quit-if-one-screen bat --wrap=never foo`

**What happens?**

less doesn't immediately quit, even though the entire file fits on the screen.

less immediately quits, like it does if you ran less directly (i.e. `LESS=--quit-if-one-screen less foo`).

Homebrew

---

#### Software version

bat 0.26.1

#### Operating system

- OS: macOS 15.7.5 Sequoia
- Kernel: 24.6.0

#### Command-line

```bash
bat --wrap=never authorized_keys --diagnostic 
```

#### Environment variables

```bash
BAT_CACHE_PATH=<not set>
BAT_CONFIG_PATH=<not set>
BAT_OPTS=<not set>
BAT_PAGER=<not set>
BAT_PAGING=<not set>
BAT_STYLE=<not set>
BAT_TABS=<not set>
BAT_THEME=dark
COLORTERM=truecolor
LANG=en_CA.UTF-8
LC_ALL=<not set>
LESS=--quit-if-one-screen
MANPAGER=<not set>
NO_COLOR=<not set>
PAGER=<not set>
SHELL=/usr/local/bin/fish
TERM=ghostty
XDG_CACHE_HOME=<not set>
XDG_CONFIG_HOME=''
```

#### System Config file

#### Config file

Could not read contents of '/var/folders/jj/8r7yjy_522771gyb6zhff4jc0000gn/T/tmp.jKTCQCt09E/.config/bat/config': No such file or directory (os error 2).

#### Custom assets metadata

Could not read contents of '/var/folders/jj/8r7yjy_522771gyb6zhff4jc0000gn/T/tmp.jKTCQCt09E/.cache/bat/metadata.yaml': No such file or directory (os error 2).

#### Custom assets

'/var/folders/jj/8r7yjy_522771gyb6zhff4jc0000gn/T/tmp.jKTCQCt09E/.cache/bat' not found

#### Compile time information

- Profile: release
- Family: unix
- OS: macos
- Architecture: x86_64
- Pointer width: 64
- Endian: little
- Host: x86_64-apple-darwin

#### Less version

```
> less --version 

```

mvanhorn:
## Summary
Fixes `bat --wrap=never` defeating less's `--quit-if-one-screen` when content fits on one screen.

## Why this matters
`LESS=--quit-if-one-screen` is a popular less configuration. Today, this works for `less foo` directly but breaks under `bat --wrap=never foo`. Reproducer from #3738:

```
LESS=--quit-if-one-screen bat --wrap=never <wide-line-file-fits-vertically>
```

less doesn't auto-quit even though the file fits, forcing the user to press q.

Root cause: with `--wrap=never`, bat passes `-S` (chop long lines) to less. When less also receives `-F` (either from bat's own quit-if-one-screen path or from `$LESS`), the `-S`/`-F` combination keeps less from auto-quitting on one-screen content. The patch skips `-S` whenever `-F` is in play, so less can quit as expected.

Tradeoff to be aware of: when `$LESS` contains `-F`, long lines will now wrap in less instead of being chopped. If preserving `-S` matters more than auto-quit in that case, the approach can be reworked.

## Testing
New integration tests in `tests/integration_tests.rs` install a mock `less` that captures the arguments it receives, and assert the exact flags bat passes per wrapping mode: `--wrap=never` with and without quit-if-one-screen, plus regression coverage for `--wrap=auto` and `--wrap=character`.

Fixes #3738

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@
 - Fix i686 `.deb` package using incorrect architecture name (`i686` instead of `i386`), preventing installation on Debian. Closes #3611, see #3650 (@Sim-hu)
 - Fix inconsistent `.deb` MUSL package names (aarch64-musl used `arm64` instead of `musl-linux-arm64`, and `musleabihf` target missed `bat-musl` prefix). Closes #3482, see #3642 (@mvanhorn)
 - Fix incorrect text width computation when using `--binary=as-text` with non-printable characters in caret notation, see #3640 and #3631 (@eyupcanakman)
+- Fix `--wrap=never` defeating `less --quit-if-one-screen`. Closes #3738, see #3746 (@mvanhorn)
 - Fix `BAT_CONFIG_DIR` pointing at system config directory causing duplicate flag errors. Closes #3589, see #3620 (@Xavrir)
 - Fix syntax highlighting for symlinked files when the symlink name has no extension but the target does. Closes #1001, see #3621 (@Xavrir)
 - Report error when pager is missing instead of silently falling back, see #3588 (@IMaloney)
--- a/src/output.rs
+++ b/src/output.rs
@@ -1,3 +1,5 @@
+#[cfg(feature = "paging")]
+use std::env;
 use std::fmt;
 use std::io;
 #[cfg(feature = "paging")]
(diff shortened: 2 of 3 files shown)


injust:
This PR seems like AI slop.

mvanhorn:
Fair callout - the PR description didn't accurately describe what the patch does, and I've rewritten it. What the change actually does: when less will receive -F (from bat's quit-if-one-screen path or from $LESS), bat stops adding -S for --wrap=never, since the -S/-F combination is what keeps less from auto-quitting on one-screen content (#3738). The test installs a mock less and asserts the exact flags bat passes per wrapping mode. There is a real tradeoff here: with $LESS containing -F, long lines will wrap in less instead of being chopped. If preserving -S matters more than auto-quit in that case, happy to rework the approach.

PR #3750, opened 2026-05-17
docs: note Debian 12+ installs the bat binary

### LINKED ISSUE #2455: `batcat` has been renamed to `bat` in Debian
Hi, the bat utility has been finally moved back to `/usr/bin/bat` from `/usr/bin/batcat`, as the package that used to "own" `/usr/bin/bat` hasn't been part of Debian for a couple of years. See <https://bugs.debian.org/1029096> for the relevant Debian bug report.

This means that most likely the `batcat` workaround won't be needed in Debian 12 and Ubuntu 23.04 (not released yet), and it might make sense to mention this already in the README.

Related to #982

wyf027:
## Summary
- Clarify that Debian 12 (Bookworm) and newer ship the executable as `bat`
- Note that `batcat` only applies to older Debian/Ubuntu releases
- Update fish completion comment accordingly

Fixes #2455

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 
 ## Other
 
+- Note that Debian 12+ installs the executable as `bat` instead of `batcat`, see #3750 (@leno23)
 - Add instructions for removing fish help abbreviations to README, see #3655 (@claw-explorer). Closes #3536
 - Add .NET slnx extension, see #3682 (@ltrzesniewski)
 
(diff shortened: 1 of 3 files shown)


ded-furby (changes requested):
The Debian 12/Bookworm claim in this patch looks factually wrong today. Debian’s current Bookworm package page for `bat` still says: the executable and manpage are renamed from `bat` to `batcat` because of the filename clash. That means the new README sentence and the updated fish completion comment would both mislead Debian 12 users.

I think this needs to be reframed before merge, either by:
- avoiding a hard-coded `Debian 12+` claim entirely, or
- citing a distro/version combination that is actually shipping `bat` under that name right now.

The relevant package page is: https://packages.debian.org/bookworm/bat

PR #3779, opened 2026-05-31
fix: force paging=never for --list-themes when BAT_OPTS has paging=always

### LINKED ISSUE #1618: --list-themes + BAT_OPTS can be problematic
Thank you so much for creating bat! Very happy with the app.

I'm customizing things a bit and my BAT_OPTS is breaking `--list-themes`. It's paging one theme at a time and the theme names are not shown. That makes it unusable, unfortunately. Maybe we should ignore `--paging=always` (or perhaps all config) when running that command?

`BAT_OPTS="--theme=TwoDark --paging=always --pager='less -j3 -i -J -M -R -S -w -z-4'"`

#### Software version

bat 0.18.0

#### Operating system

Darwin 19.4.0

#### Command-line

```bash
bat --diagnostic 
```

#### Environment variables

```bash
SHELL=/bin/zsh
PAGER=less
BAT_PAGER=<not set>
BAT_CACHE_PATH=<not set>
BAT_CONFIG_PATH=<not set>
BAT_OPTS="--theme=TwoDark --paging=always --pager='less -j3 -i -J -M -R -S -w -z-4'"
BAT_STYLE=<not set>
BAT_TABS=<not set>
BAT_THEME=<not set>
XDG_CONFIG_HOME=<not set>
XDG_CACHE_HOME=<not set>
COLORTERM=truecolor
NO_COLOR=<not set>
MANPAGER=<not set>
```

#### Config file

Could not read contents of '/Users/amd/.config/bat/config': No such file or directory (os error 2).

#### Compile time information

- Profile: release
- Family: unix
- OS: macos
- Architecture: x86_64
- Pointer width: 64
- Endian: little
- CPU features: fxsr,sse,sse2,sse3,ssse3
- Host: x86_64-apple-darwin

#### Less version

```
> less --version 
Copyright (C) 1984-2016  Mark Nudelman

Homepage: http://www.greenwoodsoftware.com/less
```

sridhar-3009:
## Problem

When `BAT_OPTS` contains `--paging=always` (a common personal config), running `bat --list-themes` opens a separate pager session per theme, making theme names invisible and the output entirely unusable.

## Root cause

The `paging=always` branch in `app.rs` had no guard for `--list-themes`, so the pager ran unconditionally.

## Fix

Check `list-themes` flag before applying `paging=always`. Override to `PagingMode::Never` when listing themes, consistent with the existing `--list-themes` guard already present in the `auto` branch (line 306).

## Testing

Manual: set `BAT_OPTS="--paging=always"` and run `bat --list-themes` — all themes now display correctly in a single pass.

Fixes #1618

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 - Syntax highlighting for Python files using uv as script runner in shebang #3689 (@janlarres)
 
 ## Bugfixes
+- Force `paging=never` for `--list-themes` when `BAT_OPTS` contains `--paging=always`, see #3779 (@sridhar-3009). Closes #1618
 - Pass `--no-paging` to `bat` invocations inside the bash / zsh / fish / PowerShell shell completion scripts so that shell-level pager wiring (e.g. `LESSOPEN='|-bat -f -pp %s'`) cannot inject ANSI escape sequences into the completion candidates. Closes #3760 (@mvanhorn)
 - Quote filenames before substituting them into `$LESSOPEN` / `$LESSCLOSE` templates, preventing shell injection when a filename contains shell metacharacters, see #3726 (@curious-rabbit)
 - Fix `--list-themes` unconditionally probing the terminal via OSC 10/11 even when `--theme` was set to an explicit value, see #3700 (regression introduced in bc42149a). (@optimistiCli)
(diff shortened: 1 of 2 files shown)


ded-furby (commented):
The paging guard in `src/bin/bat/app.rs` looks directionally right, but I don't see a regression test in the diff for the reported case. `bat`'s CONTRIBUTING guide explicitly calls out functional regressions in `tests/integration_tests.rs`, and this path has a couple of precedence edges (`BAT_OPTS=--paging=always`, `--list-themes`, and the later `-pp` override). Adding one integration test for the original report would make this much safer to merge and harder to regress later.

PR #3802, opened 2026-06-12
Improved manpage prettifier in README.md

DrHyde:
In particular this version works on OpenBSD which likes to invoke $MANPAGER with a filename command line argument instead of piping to it, and copes better with ^H characters in the input


DrHyde:
no changelog entry required as this is just documentation

PR #3804, opened 2026-06-16
Fix capacity-overflow panic in snip separator at --terminal-width 1

leeewee:
Fixes #3803.

`InteractivePrinter::print_snip` computes the snip-separator width as `term_width - panel_count - snip_left_count - title_count`. At `--terminal-width 1` the panel is disabled (`panel_count == 0`, `snip_left_count == 0`) and the `"8<"` title gives `title_count == 2`, so the subtraction underflows `usize` (release has no overflow-checks, so it wraps to ~`usize::MAX`); `str::repeat` then aborts with `capacity overflow` whenever a snip separator is emitted — i.e. two or more disjoint line ranges (or a diff gap), which the default style does:

```console
$ printf 'l1\nl2\nl3\nl4\nl5\nl6\nl7\nl8\n' > t.txt
$ bat --color=always --terminal-width 1 --line-range 1:2 --line-range 7:8 t.txt
thread 'main' panicked at library/alloc/src/slice.rs:523:50:
capacity overflow
```

This uses `saturating_sub` for the `repeat` counts so they clamp to `0` instead of underflowing. Normal-width snip rendering is unchanged (the separator dashes are simply omitted at width 1).

Added an integration test (`snip_at_terminal_width_one_does_not_panic`) and a CHANGELOG entry.

Issue #3803, opened 2026-06-16
bat panics (capacity overflow) on `--terminal-width 1` with multiple line ranges


leeewee:
**What steps will reproduce the bug?**

```console
$ printf 'l1\nl2\nl3\nl4\nl5\nl6\nl7\nl8\n' > t.txt

# explicit snip style:
$ bat --no-config --paging=never --color=always --style=snip \
      --terminal-width 1 --line-range 1:2 --line-range 7:8 t.txt
thread 'main' panicked at library/alloc/src/slice.rs:523:50:
capacity overflow
$ echo $?
101

# also without any --style (the default style includes snip):
$ bat --no-config --paging=never --color=always \
      --terminal-width 1 --line-range 1:2 --line-range 7:8 t.txt
... capacity overflow ... (exit 101)
```

**What happens?**

`bat --terminal-width 1` panics with `capacity overflow` (exit 101) when the output contains a "snip" separator — i.e. two or more disjoint `--line-range`s (or a diff gap) with a style that includes `snip` (the default).

**What did you expect to happen instead?**

A tiny terminal width should clamp the separator to empty (as the code already does for the panel when `term_width` is too small), not underflow and abort.

**Root Cause**

`InteractivePrinter::print_snip` (`src/printer.rs:606-610`):

```rust
let snip_left = "─ ".repeat((self.config.term_width - panel_count - (title_count / 2)) / 4);
let snip_left_count = snip_left.chars().count();

let snip_right =
    " ─".repeat((self.config.term_width - panel_count - snip_left_count - title_count) / 2);
```

At `term_width == 1` the panel is disabled (`panel_count == 0`, `snip_left_count == 0`) and `title_count == 2` (`"8<"`), so the `snip_right` operand `1 - 0 - 0 - 2` underflows `usize` and wraps to ~`usize::MAX`; `/ 2` then
`" ─".repeat(huge)` aborts with `capacity overflow`. (`snip_left` divides by 4 and stays 0 at width 1, so only `snip_right` faults; width 2 gives `2-0-0-2 == 0`.)

**How did you install `bat`?**

Built from source (`cargo build --release`) at commit `af1f53d9` (current `master`, reporting as 0.26.1).

---

**bat version and environment**

#### Software version

bat 0.26.1 (v0.25.0-726-gaf1f53d9)

#### Operating system

- OS: Linux (Ubuntu 20.04)
- Kernel: 5.4.0-216-generic

_Found by our static analysis tooling._

PR #3812, opened 2026-06-26
fix: avoid usize underflow in --list-languages at narrow terminal width

mahirhir:
Mirrors #3804 to its remaining sibling site.

#3804 hardened `print_snip` with `saturating_sub` after #3803 reported a `capacity overflow` panic at `--terminal-width 1` (the `term_width - panel_count - title_count` subtraction underflowing `usize`). That fix covered the snip renderer. The `--list-languages` renderer has the identical `term_width - X - Y` shape and was left unhardened:

`src/bin/bat/main.rs:155`
```rust
let desired_width = config.term_width - longest - separator.len();
```

Unlike the `term_width -` sites in `printer.rs` (356/418/731/810), this one sits outside the panel-disable guard at `printer.rs:261`, so nothing clamps it. `bat --list-languages --terminal-width 1` underflows.

Honest scope: in debug / overflow-checked builds this panics (exit 101). In release the subtraction wraps, and since `desired_width` only feeds a `>=` comparison it silently yields wrong column layout rather than crashing. Same class as #3804, narrowed to the language list, latent since the code was introduced.

Verified both directions on a standalone replica of the arithmetic: the old form panics (debug) / wraps to wrong output (release), the patched form is clean, and the two are byte-identical at a normal width (100), so there is zero behavior change away from the edge case. `saturating_sub` is the minimal change and matches the precedent set by #3804.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@
 - Fixed syntax tests path, see #3610 (@foxfromworld)
 - Fix zsh tab completion word-splitting language names containing spaces (e.g. `HTML (Jinja2)`, `Apache Conf`), see #3693 (@YoshKoz)
 - Fix zsh tab completion offering invalid `-l` arguments (file globs, paths, hidden filenames) sourced from the second column of `--list-languages`. Closes #3735, see #3737 (@truffle-dev)
+- Fix `usize` underflow in `--list-languages` when `--terminal-width` is smaller than the longest language name, see #3812 (@greymoth-jp)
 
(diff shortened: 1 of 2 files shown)


mahirhir:
Added the changelog entry. The two CI failures are pre-existing flakes unrelated to this change: a crates.io SSL/curl download failure on the syntaxes-test job, and a `cross` metadata fetch failure on i686-musl. Could you re-run those jobs?

keith-hall (approved):
Thanks

Issue #2745, opened 2023-10-25
Bash scripts with .txt are not colored


Tonus1:
**What steps will reproduce the bug?**

1. change the extension of your script to nn script.nn, it works
2. change to txt, script.txt, it no longer displays colors

Would have espected the syntax is detected even when the file ends in .txt

**How did you install `bat`?**

Used the AlackBuild provided at slackbuilds.org

bat 0.24.0 on Slackware64-15.0

#### Software version

bat 0.24.0

#### Operating system

Linux 5.15.117

#### Command-line

```bash
/usr/bin/bat --diagnostic mutt-colors.txt 
```

#### Environment variables

```bash
SHELL=/bin/bash
PAGER=<not set>
LESS=<not set>
LANG=fr_FR.utf8
LC_ALL=<not set>
BAT_PAGER=<not set>
BAT_PAGING=<not set>
BAT_CACHE_PATH=<not set>
BAT_CONFIG_PATH=<not set>
BAT_OPTS=<not set>
BAT_STYLE=<not set>
BAT_TABS=<not set>
BAT_THEME=<not set>
XDG_CONFIG_HOME=<not set>
XDG_CACHE_HOME=<not set>
COLORTERM=truecolor
NO_COLOR=<not set>
MANPAGER='sh -c '\''col -bx | bat -l man -p --paging always'\'''
```

#### System Config file

Could not read contents of '/etc/bat/config': No such file or directory (os error 2).

#### Config file

```
# This is `bat`s configuration file. Each line either contains a comment or
# a command-line option that you want to pass to `bat` by default. You can
# run `bat --help` to get a list of all possible configuration options.

# Specify desired highlighting theme (e.g. "TwoDark"). Run `bat --list-themes`
# for a list of all available themes
#--theme="TwoDark"
--theme="Solarized (dark)"

# Enable this to use italic text on the terminal. This is not supported on all
# terminal emulators (like tmux, by default):
--italic-text=always

# Uncomment the following line to disable automatic paging:
#--paging=never

# Uncomment the following line if you are using less version >= 551 and want to
# enable mouse scrolling support in `bat` when running inside tmux. This might
# disable text selection, unless you press shift.
#--pager="less --RAW-CONTROL-CHARS --quit-if-one-screen --mouse"

# Syntax mappings: map a certain filename pattern to a language.
#   Example 1: use the C++ syntax for Arduino .ino files
#   Example 2: Use ".gitignore"-style highlighting for ".ignore" files
[...]

PR #3816, opened 2026-06-30
Fix --ignored-suffix to fall back to first-line detection

adnrivera:
When an ignored suffix is also a registered extension (e.g. `.txt` maps to "Plain Text"), `get_syntax_for_file_extension` matched the raw extension before attempting the suffix strip. As a result `--ignored-suffix .txt` had no effect on such files and first-line/shebang detection was never reached, so a shell script saved as `*.txt` stayed unhighlighted.

Strip ignored suffixes first and detect on the remainder; only fall back to the raw extension when no ignored suffix applies. When a suffix is stripped but the remainder has no syntax, return `None` so the caller can use first-line detection. Default behavior (no `--ignored-suffix`) is unchanged.

See #2745.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 - Syntax highlighting for Python files using uv as script runner in shebang #3689 (@janlarres)
 
 ## Bugfixes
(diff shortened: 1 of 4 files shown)


keith-hall (approved):
Thanks

PR #3832, opened 2026-07-07
Fix --wrap=word leaking a leading space onto the continuation line

upuddu:
When a word ends exactly at the terminal width, the separating whitespace is the character that triggers the wrap, and it was being pushed onto the next visual line as a spurious leading space. This consumes that whitespace when it triggers the wrap.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - Syntax highlighting for Python files using uv as script runner in shebang #3689 (@janlarres)
 
 ## Bugfixes
+- Fix `--wrap=word` leaking a leading space onto the continuation line when a word ends exactly at the terminal width, so the separating space triggered the wrap (@upuddu)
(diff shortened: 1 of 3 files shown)


keith-hall (approved):

upuddu:
Heads-up: the two red checks here (`cargo audit` and the `all-jobs` gate that depends on it) are **not caused by this PR**. `cargo audit` is flagging newly-published RustSec advisories in transitive deps — `crossbeam-epoch` 0.9.18 (RUSTSEC-2026-0204), `quick-xml` 0.39.4 (RUSTSEC-2026-0194/0195), etc. Those versions are pinned in `master`'s `Cargo.lock`, so `master` fails the same audit today. This PR only changes `src/printer.rs`, a test, and the changelog — no dependency or `Cargo.lock` changes. Fixing it means a repo-wide dependency bump or an `.cargo/audit.toml` ignore update, which seems better handled separately. Happy to rebase once that lands.

keith-hall:
A lock file update has now landed btw 🙂

PR #3745, opened 2026-05-16
fix: omit less -S when quit-if-one-screen and --wrap=never (fixes #3738)

wyf027:
## Summary

`bat --wrap=never` passes `-S` (chop long lines) to `less`. Combined with quit-if-one-screen (`-F`, including via `$LESS=--quit-if-one-screen`), wide lines make `less` think the file does not fit on one screen, so it no longer quits immediately (#3738).

Skip `-S` when bat enables quit-if-one-screen.

## Test plan

- [x] `cargo test --test integration_tests wrap_never_does_not_pass_chop_long_lines_with_quit_if_one_screen`
- [x] Mock `less` asserts `-F` and `-S` are not passed together for `LESS=--quit-if-one-screen` + `--wrap=never`

Fixes #3738

PR #3834, opened 2026-07-08
fix: don't hard-wrap with --wrap=auto when --paging=always is set

### LINKED ISSUE #1081: --wrap unpaged
I love it when `bat` wraps lines in a single page of output. When output needs to be sent to the pager, though, the wrapped lines are actively unhelpful. I'm very used to issuing `-S` to `less` once a log file is open to change the line wrapping live.

I'd like a `--wrap` mode that behaves like `auto` for unpaged output, but `never` when piping to `less`.

soy-chrislo:
I hit this while scripting bat with `--paging=always` piped into `less` — the output came out double-wrapped and unreadable, and it turned out to be exactly what #1081 describes.

`--paging=always` unconditionally pipes bat's output through a pager (e.g. `less`), which already has its own line-wrapping controls. `--wrap=auto` was still hard-wrapping at the detected terminal width in that case, since `interactive_output` reflects whether bat's own stdout is a real terminal — which it still is even when piped to a pager — producing unreadable, doubly-wrapped output inside the pager.

This defers to the pager instead when `--paging=always` is active and no explicit `--terminal-width` was given (which still takes precedence, as an intentional override). It doesn't require buffering input or output, since `PagingMode::Always` is known at config-resolution time regardless of content — unlike `QuitIfOneScreen`, whose pager invocation depends on the rendered content length, which isn't known yet at that point. Scoped to `PagingMode::Always` only for that reason; `QuitIfOneScreen` is left untouched and documented as a known limitation rather than guessed at. This doesn't change whether/when bat spawns a pager, only whether bat itself hard-wraps before handing off — no new pager-invocation logic added.

Before writing this I checked what had already been tried: #1081 has been open since 2020, and the only prior attempt at it (#3777) took a different, opt-in approach (a new `--wrap=unpaged` flag) and was closed without merge — it's unclear whether that was a design rejection or just abandoned, so I want to flag that alternative rather than assume mine is preferred. Changing `auto`'s default behavior (this PR) vs. adding an opt-in flag (#3777) is a real tradeoff worth a maintainer call: the opt-in path is strictly additive and zero-risk, while this PR fixes what the original issue describes as broken behavior without requiring a new flag. Separately, `QuitIfOneScreen` has its own open, active thread (#3738, with PRs #3745/#3746 still unresolved) that this PR intentionally doesn't touch, for the reason above.

Verified manually with a real pty (`script`) showing the before/after (hard-wrapped vs. intact line), then covered by a new integration test (`wrap_auto_does_not_hard_wrap_when_pager_is_always_active`) using this repo's existing `openpty` pattern — confirmed it fails without the fix, not a placebo — plus a regression guard (`wrap_auto_still_wraps_with_explicit_terminal_width_and_paging_always`) confirming explicit `--terminal-width` still forces wrapping. Full suite passes: 247 tests, 0 failures.

This is my first contribution to bat — open to feedback on scope or approach.

Closes #1081

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@
 - Syntax highlighting for Python files using uv as script runner in shebang #3689 (@janlarres)
 
 ## Bugfixes
-- Fix `--wrap=auto` hard-wrapping lines even when `--paging=always` sends output through a pager, producing unreadable double-wrapped output. `auto` now defers to the pager in that case, matching explicit `--terminal-width` (which still takes precedence). Closes #1081, see #TODO (@soy-chrislo)
+- Fix `--wrap=auto` hard-wrapping lines even when `--paging=always` sends output through a pager, producing unreadable double-wrapped output. `auto` now defers to the pager in that case, matching explicit `--terminal-width` (which still takes precedence). Closes #1081, see #3834 (@soy-chrislo)
 - `--strip-ansi`: also strip 8-bit C1 introducers (U+0090, U+0098, U+009B, U+009D, U+009E, U+009F) and DCS/SOS/PM/APC sequence bodies, which previously passed through. See #3729 (@curious-rabbit)
 - Fix `--ignored-suffix` not falling back to first-line/shebang detection when the ignored suffix is also a registered extension (e.g. `--ignored-suffix .txt` on a shebang script), see #2745 and #3816 (@adnrivera)
 - Fix `capacity overflow` panic when printing a snip separator at `--terminal-width=1` with multiple line ranges. Closes #3803, see #3804 (@leeewee)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 - Syntax highlighting for Python files using uv as script runner in shebang #3689 (@janlarres)
 
 ## Bugfixes
+- Fix `--wrap=auto` hard-wrapping lines even when `--paging=always` sends output through a pager, producing unreadable double-wrapped output. `auto` now defers to the pager in that case, matching explicit `--terminal-width` (which still takes precedence). Closes #1081, see #TODO (@soy-chrislo)
 - `--strip-ansi`: also strip 8-bit C1 introducers (U+0090, U+0098, U+009B, U+009D, U+009E, U+009F) and DCS/SOS/PM/APC sequence bodies, which previously passed through. See #3729 (@curious-rabbit)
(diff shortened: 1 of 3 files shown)


soy-chrislo:
**Update:** the new pty-based test (`wrap_auto_does_not_hard_wrap_when_pager_is_always_active`) was initially flaky on macOS — it failed only on the `x86_64-apple-darwin (macos-15-intel)` CI runner, passing everywhere else (Linux, Windows, macOS arm64).

Root cause: on macOS, the kernel can discard data still queued on a pty master once the last slave-side fd closes — which happens as soon as the child (bat + its pager) exits — unlike Linux, which retains it (see [Apple Developer Forums](https://developer.apple.com/forums/thread/663632), and [pexpect#662](https://github.com/pexpect/pexpect/issues/662) for the same issue in another project). The test read the master only once, right after waiting for the child to exit, without holding any fd of its own to the slave side. Fixed by having the test keep its own clone of the slave fd alive across the wait+read (the same `try_clone()` pattern already used by `no_args_doesnt_break` in this file), asserting the child's exit status before reading, and no longer silencing `read()` errors.

Just pushed the fix, and verified it green on a real `x86_64-apple-darwin (macos-15-intel)` GitHub Actions runner (not just locally) before pushing.

---

**Separately, unrelated to this PR:** CI's `cargo audit` check is currently failing on `master` too (pre-existing, not introduced by this diff, which only touches `CHANGELOG.md`, `src/bin/bat/app.rs`, and `tests/integration_tests.rs` — no `Cargo.toml`/`Cargo.lock` changes). Three RUSTSEC advisories are hard failures: `crossbeam-epoch` (RUSTSEC-2026-0204, via `bugreport → sysinfo → rayon → crossbeam-deque`) and `quick-xml` (RUSTSEC-2026-0194 / RUSTSEC-2026-0195, via `plist`).

I looked into it out of curiosity and confirmed a fix is possible without touching any `Cargo.toml`: `cargo update -p crossbeam-epoch -p plist -p anyhow` bumps exactly those 4 packages (quick-xml comes along transitively through plist 1.10.0) with zero manifest changes, since the version ranges already declared by bat and its intermediate dependents (`crossbeam-deque ^0.9.17`, `plist ^1.9.0`/`syntect ^1.3`, `anyhow ^1.0.97`) already permit the patched versions. Verified locally (`cargo audit` goes from 3 vulnerabilities to 0) and confirmed green on a real CI run on my fork. Happy to open that as a separate PR if useful — didn't want to bundle an unrelated dependency bump into this fix.

PR #3849, opened 2026-07-18
Highlight Rust async and await as keywords

### LINKED ISSUE #1534: Rust's async await is not highlighted correctly
**Syntax:**

async / await in Rust. For example:

```rust
async fn foo() {
  async_foo().await.await.await;
}
```

Currently it highlights as following:

image

It should be highlighting `async` and `await` as keywords.

Manually installing RustEnhanced can improve the situation:

image

I understand this is completely up to upstream (https://github.com/sublimehq/Packages), but given the upstream issue https://github.com/sublimehq/Packages/pull/2305 is hinged. I wonder if it makes sense to have bat bundle https://github.com/rust-lang/rust-enhanced/ directly.

In #906 the same idea was rejected with the reasoning that const generic was not stablized hence the breakage was not considered as a bug. However, async await in Rust have been stabilized for more than a year now. It is surprizing to see bat still can't highlight it correctly. Can we re-consider the decision?

Thanks.

----

Quick instructions for anyone searching this in the future:

```
mkdir -p "$(bat --config-dir)/syntaxes"
cd "$(bat --config-dir)/syntaxes"
git clone https://github.com/rust-lang/rust-enhanced.git
bat cache --build
```

coyaSONG:
## Summary

- extend bat's existing Rust syntax patch so `async` and `await` use the control-keyword scope
- add the issue's async function and chained `.await` expression to the Rust syntax fixture
- commit the regenerated highlighted fixture while leaving generated binary asset caches out of the pull request

Closes #1534.

## Root cause

bat's pinned Sublime Packages revision predates Rust async/await syntax, so both tokens fell through to ordinary identifier styling.

## Validation

- verified the updated patch applies cleanly to the exact pinned syntax source
- ran the full repository asset builder successfully
- rebuilt bat and confirmed `async`/`await` render with the Rust control-keyword color while `async_foo` remains a function identifier
- confirmed the committed Rust fixture is byte-identical to a fresh focused render
- `cargo fmt --check`
- `cargo test --locked`
- `git diff --check`

## AI assistance

OpenAI Codex was used to inspect the issue and repository guidance, implement the change, and run the checks listed above.


JMak-Security:
Since this has been open since 2026-07-18 with no review activity, I've carried the fix forward in #3892 — same patch, plus verification that it applies cleanly against the exact pinned submodule commit (759d6eed) rather than current upstream, which has since restructured this section entirely. Thanks for the original diagnosis and test case @coyaSONG, they made this straightforward to port over.

coyaSONG:
Closing this draft because #3892 has carried the fix forward and is ready for maintainer review. I rechecked both PRs against the same current base: the syntax patch and both regression-fixture blobs are identical, with only the changelog PR/author attribution differing, and both CI suites are green. Keeping #3892 as the single active implementation avoids duplicate review work. Thanks @JMak-Security for preserving the diagnosis and test case.

JMak-Security:
@coyaSONG, Thanks for the update! Glad to help get this fixed.

PR #3857, opened 2026-07-24
feat: implement `-b` / `--number-nonblank` for `cat -b` compatibility

### LINKED ISSUE #3856: Support `-b` / `--number-nonblank` for `cat` compatibility
## What would you like to see?

Add a top-level `-b` / `--number-nonblank` option that numbers nonblank output lines.

Expected behavior:

- Number lines containing any content, including spaces or tabs.
- Don't number empty lines containing only a line ending.
- Give `-b` precedence over `-n`, regardless of option order. Both `-bn` and `-nb` should use nonblank numbering.
- Otherwise follow the existing `-n` behavior, including numbering during piped/loop-through output and the current per-input numbering behavior.
- Support files, standard input, and multiple input operands wherever `-n` is supported.

For example, given:

```text
alpha

beta
```

`bat -b` should number `alpha` and `beta`, while leaving the empty line unnumbered.

## Why?

`-b` is supported by widely used GNU and BSD/macOS `cat` implementations. Scripts, aliases, and interactive commands written for those implementations can currently fail immediately when `cat` is replaced with `bat`:

```console
$ printf 'alpha\n\nbeta\n' | bat -b
error: unexpected argument '-b' found
```

On macOS, `/bin/cat -b`, `/bin/cat -bn`, and `/bin/cat -nb` all number only nonblank lines. Supporting the same option and precedence would remove one common incompatibility without requiring broader changes to `bat`'s display model.

This missing option was also noted in the POSIX/drop-in compatibility discussion in [#134](https://github.com/sharkdp/bat/issues/134#issuecomment-2160884143), but there doesn't appear to be a dedicated issue or pull request for it.

## Implementation considerations

Recent `-n` work in [#3496](https://github.com/sharkdp/bat/pull/3496) already ensures explicit numbering works when output is piped and handles combined short flags. A `-b` implementation could reuse that path while selecting a nonblank numbering mode.

The short `-b` used by `bat cache --build` is scoped to the `cache` subcommand, so it shouldn't conflict with a top-level `bat -b` option.

## Suggested acceptance tests

- `bat -b FILE` numbers only nonblank lines.
- Whitespace-only lines are numbered.
- `bat -bn FILE` and `bat -nb FILE` both behave like `bat -b FILE`.
- `printf 'a\n\nb\n' | bat -b` retains nonblank line numbers when piped onward.
- Numbering works with multiple input files consistently with existing `bat -n` behavior.
- Existing `-n`, style, decorations, squeeze, range, and unbuffered tests remain passing.

MeGaurav4:
Implements `cat -b` compatibility via `-b` / `--number-nonblank`.

### Behaviour

```console
$ printf 'alpha\n\nbeta\n' | bat -b
   1 alpha
     
   2 world
```

- Non-blank lines are numbered sequentially
- Blank lines (empty / newline-only) are printed without a number and
  do not advance the line counter
- Lines containing spaces or tabs are *not* blank and **are** numbered
- `-b` takes precedence over `-n`; both `-bn` and `-nb` produce nonblank numbering
- `-bp` disables numbering (plain mode wins)

### Implementation

Core change is in `src/controller.rs`: when `number_nonblank` is set,
the line counter is not incremented for lines whose buffer contains
only `\r` / `\n` bytes.  In `src/printer.rs` the decoration engine
receives `line_number = 0` as sentinel for those lines, and
`LineNumberDecoration::generate` returns empty gutter spaces in that
case.  The raw-arg scanner in `App::new()` was extended to detect `-b`
the same way it detects `-n` (so it works with combined short flags
like `-pb` / `-bp`).

### Testing

- 5 new integration tests: plain `-b`, `-b` in loop-through mode,
  `-bn`/`-nb` precedence, `-bp` override, piped stdin
- All 436 existing tests continue to pass
- Help-text snapshots regenerated

Closes #3856.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ## Features
 
+- Add `-b` / `--number-nonblank` flag to only number non-blank lines, for `cat -b` compatibility. Closes #3856, see #3857 (@MeGaurav4)
 - Add a `--sanitize=<auto|always|never>` flag for safe display of untrusted input. It implies `--strip-ansi` at the same value and additionally substitutes terminal-active control bytes (cursor moves, charset switches, beep, etc.) and Unicode bidi / zero-width formatting characters with the Unicode replacement character (U+FFFD). Mitigates Trojan-Source-style spoofing (CVE-2021-42574). See #3729 (@curious-rabbit)
 - Map justfile, Justfile, .justfile, and *.justfile to Makefile syntax highlighting, see #3623 (@zachvalenta)
 - Preserve `--diff` change markers and snip separators when `--plain` is set. Closes #3630, see #3643 (@mvanhorn)
--- a/doc/long-help.txt
+++ b/doc/long-help.txt
@@ -83,6 +83,16 @@ Options:
   -n, --number
           Only show line numbers, no other decorations. This is an alias for '--style=numbers'
 
+  -b, --number-nonblank
+          Only show line numbers for non-blank lines, no other decorations. This is an alias for
+          '--style=numbers'. Non-blank lines are lines that contain any character before the line
+          ending, including spaces and tabs. When used together with --number (-n),
+          --number-nonblank (-b) takes precedence.
+          
+          Example:
+            printf 'alpha\n\nbeta\n' | bat -b
+            numbers 'alpha' and 'beta', but skips the empty line.
+
       --color <when>
           Specify when to use colored output. The automatic mode only enables colors if an
           interactive terminal is detected - colors are automatically disabled if the output goes to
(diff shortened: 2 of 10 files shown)


TyceHerrman:
@MeGaurav4 are you still working on this? if not, I can start from your work and credit you on a new pr and fix the failing ci checks

MeGaurav4:
Hey Tyce, thanks for checking in, and thanks for the offer either way. Still working on this: I've merged current master into the branch and fixed the failing `cargo fmt` check. The `cargo audit` failure was a stale advisory that's already cleared on master. CI is re-running now, so I'll follow up once it's green.

keith-hall (approved):

PR #3876, opened 2026-08-02
fix(syntax): avoid repeated log level scans

### LINKED ISSUE #3866: log syntax exhibits catastrophic performance on bash xtrace logs

1. download log file
2. print it with bat --no-pager and autodetected syntax

**What happens?**
i save my logs at log-"tool"-"time".log files, for now i save bash xtrace log of one program

size of file: 564K
cat time: 0.009s
bat: 37s
max line size: 7095

I ran a few tests and found that if a filename contains "log" it is automatically parsed using the "log" syntax, which incredible reduces performance, -l sh or -l bash works fast

also i remove pluses from log it still running slow

https://streamable.com/1f2bsa

nixpkgs

bat 0.26.1

#### Software version

bat 0.26.1

#### Operating system

- OS: Linux (NixOS 26.05)
- Kernel: 6.18.40

#### Command-line

```bash
/run/current-system/sw/bin/bat --diagnostic --no-pager log-bash-20260801-032836.log 
```

#### Environment variables

```bash
BAT_CACHE_PATH=<not set>
BAT_CONFIG_PATH=<not set>
BAT_OPTS=<not set>
BAT_PAGER=<not set>
BAT_PAGING=<not set>
BAT_STYLE=<not set>
BAT_TABS=<not set>
BAT_THEME=<not set>
COLORTERM=truecolor
LANG=en_US.UTF-8
LC_ALL=<not set>
LESS=<not set>
NO_COLOR=<not set>
PAGER=less
SHELL=/run/current-system/sw/bin/zsh
TERM=xterm-ghostty
XDG_CACHE_HOME=/home/ozumr/.cache
XDG_CONFIG_HOME=/home/ozumr/.config
```

#### System Config file

#### Config file

Could not read contents of '/home/ozumr/.config/bat/config': No such file or directory (os error 2).

#### Custom assets metadata

Could not read contents of '/home/ozumr/.cache/bat/metadata.yaml': No such file or directory (os error 2).

#### Custom assets

'/home/ozumr/.cache/bat' not found

#### Compile time information

- Profile: release
- Family: unix
- OS: linux
- Architecture: x86_64
- Pointer width: 64
- Endian: little
- CPU features: fxsr,sse,sse2

#### Less version

```
> less --version 

```

[log-bash-20260801-032836.log](https://github.com/user-attachments/files/30609108/log-bash-20260801-032836.log)

Matei02355:
## Summary

- prevent log-level lookaheads from being retried at every character
- use `^` for normal Log lines and classify embedded Syslog messages once
- keep Sublime Text on its custom regex engine while preserving existing highlighting
- fix the severe slowdown reported in #3866

## Root cause

The Log syntax uses zero-width lookaheads to apply a meta scope to an entire line when it contains an error, warning, info, or debug marker. When a lookahead failed, the regex engine retried it at every later character, repeatedly scanning the remainder of long lines.

For normal Log files, anchoring each lookahead with `^` prevents those futile retries. Syslog enters the Log syntax after consuming its prefix, so it invokes a dedicated unanchored context exactly once at the start of the message. This preserves the existing line-wide scopes without using `\G`, which would make Sublime Text fall back to Oniguruma.

## Performance

Using the reporter's 573,835-byte, 3,193-line attachment with the v0.26.1 release binary:

- before: 50.923 seconds
- after: 0.169 seconds

That is approximately a 300× speedup.

## Validation

- `cargo test --locked`
- `cargo fmt --all -- --check`
- `cargo clippy --locked --all-targets --all-features -- -D warnings`
- rebuilt a syntax cache; existing Log and Syslog highlighted fixtures are byte-identical to the baseline
- verified Syslog error, warning, info, debug, structured-data, and `CMD` precedence cases against the original syntax

Fixes #3866.

### DIFF
--- a/assets/syntaxes/02_Extra/log.sublime-syntax
+++ b/assets/syntaxes/02_Extra/log.sublime-syntax
@@ -55,6 +55,24 @@ contexts:
       push:
         - debug_line_meta
         - main_pop_at_eol
+  # For syntaxes that embed Log after consuming a prefix. Call this context once.
+  log_level_lines_at_current_position:
+    - match: (?=.*{{error}})
+      set:
+        - error_line_meta
+        - main_pop_at_eol
+    - match: (?=.*{{warning}})
+      set:
+        - warning_line_meta
+        - main_pop_at_eol
+    - match: (?=.*{{info}})
+      set:
+        - info_line_meta
+        - main_pop_at_eol
+    - match: (?=.*{{debug}})
+      set:
+        - debug_line_meta
+        - main_pop_at_eol
   log_levels:
     - match: '{{error}}'
       scope: markup.error.log
--- a/assets/syntaxes/02_Extra/log.sublime-syntax
+++ b/assets/syntaxes/02_Extra/log.sublime-syntax
@@ -38,19 +38,20 @@ contexts:
       scope: markup.underline.link.scheme.log
       push: url-host
   log_level_lines:
-    - match: (?=.*{{error}})
+    # Prevent retrying each lookahead at every position.
+    - match: \G(?=.*{{error}})
       push:
         - error_line_meta
         - main_pop_at_eol
-    - match: (?=.*{{warning}})
+    - match: \G(?=.*{{warning}})
       push:
         - warning_line_meta
         - main_pop_at_eol
-    - match: (?=.*{{info}})
+    - match: \G(?=.*{{info}})
       push:
         - info_line_meta
         - main_pop_at_eol
-    - match: (?=.*{{debug}})
+    - match: \G(?=.*{{debug}})
       push:
         - debug_line_meta
         - main_pop_at_eol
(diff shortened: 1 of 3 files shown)


assets/syntaxes/02_Extra/log.sublime-syntax
```diff
@@ -38,19 +38,20 @@ contexts:
       scope: markup.underline.link.scheme.log
       push: url-host
   log_level_lines:
-    - match: (?=.*{{error}})
+    # Prevent retrying each lookahead at every position.
+    - match: \G(?=.*{{error}})
```
keith-hall: If this syntax definition were to be used in Sublime Text, using `\G` would slow it down there significantly, because it would fallback to Oniguruma instead of Sublime's custom regex engine. Perhaps it makes sense to use `^` instead of `\G`? I like that the assertion is outside the lookahead in case it helps the syntax engine do less work 👍


```suggestion
    - match: ^(?=.*{{error}})
```

keith-hall:
Also, this PR passes in CI, perhaps it doesn't need to be a draft any more? 🙂

keith-hall (approved):

PR #3877, opened 2026-08-02
Detect binary content beyond the first line

### LINKED ISSUE #3554: encrypted file not recognized as binary

1. Try to use bat in a gpg encrypted file.

```
tar -cvzf uncrypted_folder.tar.gz uncrypted_folder
gpg --symmetric --batch --yes --passphrase-file keyfile.key --cipher-algo AES256 --output encrypted_filename.tar.enc unencrypted_folder.tar.gz
```

**What happens?**
If I try to use the tool to read a gpg encrypted file, it has apper glyphs issues and does not recognize as binary as it should. It didnt happen in all enc files, but it did in most. 

I tested in 3 different terms, all same issue.

Show nothing or just be recognized as binary file.

---

bat 0.26.1 (979ba22)

```
No fonts contain glyphs for these codepoints: \u{7b4}.
Placeholder glyphs are being displayed instead.
You may wish to install additional fonts, or adjust
your configuration so that it can find them.
Set warn_about_missing_glyphs=false to suppress this
message. 
```

Image

Matei02355:
## Summary

- inspect up to the first 1024 already-buffered bytes before splitting out the first line
- preserve the reader's bytes, line boundaries, UTF-16 handling, and streaming behavior
- add regressions for cross-line binary detection, the 1024-byte boundary, single-read/EOF inputs, and the CLI binary header

Fixes #3554.

## Root cause

`content_inspector` checks up to 1024 bytes for a NUL byte, but bat passed only the first line. Random or encrypted data can contain a newline before its first NUL byte, so the shortened sample was classified as UTF-8 and binary bytes were sent to the terminal.

The fix snapshots the already-buffered prefix with `BufRead::fill_buf()` before `read_until` splits out the first line. The snapshot is non-consuming, so no bytes are lost or reordered, and it does not add a post-line blocking read. Empty input is handled without requesting a second EOF event.

## Compatibility

BOM detection still takes precedence, so UTF-16 input handling is unchanged. ZIP signature detection is also unchanged.

PR #3763 changes the same input initialization path for a distinct issue (#2262: bounding reads for newline-free binary files). If it merges first, this PR will need a small rebase, but the behaviors are complementary.

## Validation

- `cargo test --all-targets --all-features` — clean run: 455 passed, 5 ignored
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo fmt --check`
- `git diff --check`


keith-hall (approved):
Thanks, looks good

PR #3883, opened 2026-08-07
fix(log-syntax): anchor log_level_lines lookaheads with \G to prevent O(n²) backtracking

### LINKED ISSUE #3866: log syntax exhibits catastrophic performance on bash xtrace logs

RinZ27:
Fixes #3866.

The `log` syntax's four `log_level_lines` patterns used unanchored lookaheads:

```yaml
- match: (?=.*{{error}})
- match: (?=.*{{warning}})
- match: (?=.*{{info}})
- match: (?=.*{{debug}})
```

Syntect/Oniguruma retries unanchored patterns at every character position in a line. On a line with no log-level keyword, each of the four lookaheads scans to the end of the line from every starting offset — O(n²) work. With lines up to 7,096 chars (as in the attached bash xtrace log), this adds up fast: a 564 KB file I tested took ~37 s to render while `cat` handles it in 0.009 s.

The fix is a single `\G` in each pattern:

```yaml
- match: (?=\G.*{{error}})
- match: (?=\G.*{{warning}})
- match: (?=\G.*{{info}})
- match: (?=\G.*{{debug}})
```

`\G` anchors the match to the current parser position (the start of the line when these contexts are entered), so each pattern is tried exactly once per line rather than at every character. The same 564 KB file now renders in 0.016 s.

Highlighting behaviour on normal log files (ERROR/WARN/INFO/DEBUG lines) is unchanged — verified locally with both typical structured logs and the xtrace file from the issue.

**Changelog:** Added an entry to the `## Bugfixes` section of the unreleased changelog.

### DIFF
--- a/assets/syntaxes/02_Extra/log.sublime-syntax
+++ b/assets/syntaxes/02_Extra/log.sublime-syntax
@@ -38,19 +38,23 @@ contexts:
       scope: markup.underline.link.scheme.log
       push: url-host
   log_level_lines:
-    - match: (?=.*{{error}})
+    # \G anchors to the current parser position (start of line) so these
+    # lookaheads are only attempted once per line instead of at every character
+    # position, preventing catastrophic O(n²) backtracking on long lines that
+    # contain no log-level keyword (e.g. bash xtrace output).
+    - match: (?=\G.*{{error}})
       push:
         - error_line_meta
         - main_pop_at_eol
-    - match: (?=.*{{warning}})
+    - match: (?=\G.*{{warning}})
       push:
         - warning_line_meta
         - main_pop_at_eol
-    - match: (?=.*{{info}})
+    - match: (?=\G.*{{info}})
       push:
         - info_line_meta
         - main_pop_at_eol
-    - match: (?=.*{{debug}})
+    - match: (?=\G.*{{debug}})
       push:
         - debug_line_meta
         - main_pop_at_eol
@@ -94,7 +98,7 @@ contexts:
     - match: \b(?={{hours_minutes_seconds}})
       push: time
   time:
-    - match: (?:{{hours_minutes_seconds}})(?:(\.)\d{3})?\b
+    - match: (?:{{hours_minutes_seconds}})(?:(\.}\d{3})?\b
       scope: meta.time.log meta.number.integer.decimal.log constant.numeric.value.log
       captures:
         1: punctuation.separator.decimal.log
@@ -129,7 +133,7 @@ contexts:
         2: constant.numeric.value.log
         3: constant.numeric.value.log punctuation.separator.decimal.log
         4: constant.numeric.value.log
-    - match: \b\d+(\.)\d+\b
+    - match: \b\d+(\.}\d+\b
       scope: meta.number.float.log constant.numeric.value.log
       captures:
         1: punctuation.separator.decimal.log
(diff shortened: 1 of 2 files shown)


keith-hall:
> Syntect/Oniguruma retries unanchored patterns at every character position in a line.

Not exactly - these patterns contain no dynamic elements, so syntect runs the regex match once per line and remembers the resulting position instead of running the search again.

> The fix is a single `\G` in each pattern:
> 
> ```yaml
> - match: (?=\G.*{{error}})
> - match: (?=\G.*{{warning}})
> - match: (?=\G.*{{info}})
> - match: (?=\G.*{{debug}})
> ```
> 
> `\G` anchors the match to the current parser position (the start of the line when these contexts are entered), so each pattern is tried exactly once per line rather than at every character.

If this syntax definition were to be used in Sublime Text, using `\G` would slow it down there significantly, because it would fallback to Oniguruma instead of Sublime's custom regex engine. As we know it is the start of the line, perhaps it makes sense to use `^` instead of `\G`, and perhaps even put it outside the lookahead as it is an assertion in case that helps the syntax engine do less work?

```yaml
 - match: ^(?=.*{{error}})
```

assets/syntaxes/02_Extra/log.sublime-syntax
```diff
@@ -94,7 +98,7 @@ contexts:
     - match: \b(?={{hours_minutes_seconds}})
       push: time
   time:
-    - match: (?:{{hours_minutes_seconds}})(?:(\.)\d{3})?\b
+    - match: (?:{{hours_minutes_seconds}})(?:(\.}\d{3})?\b
```
keith-hall, line 101: why were these other lines touched and broken?

keith-hall:
Thanks for your contribution. This got fixed by an earlier PR: https://github.com/sharkdp/bat/pull/3876

PR #3897, opened 2026-08-28
fix: use last occurrence in combined short flags

cuishuang:
## Summary

Use the last occurrence of `-p`, `-n`, and `-b` when determining the effective option in combined short arguments.

The raw argument scanner currently uses `Iterator::position`, which finds the first occurrence of each flag. This works for combinations such as `-pn` and `-np`, but produces incorrect results when a flag occurs more than once.

For example, `-npn` ends with `-n`, so numbering should be enabled. The scanner currently compares the first `n` with `p` and incorrectly allows the loop-through printer to be selected, which drops the requested line numbers.

The same issue affects `-b` combinations:

- `-pbp` should use plain output.
- `-bpb` should number non-blank lines.

`-pnp` already produces plain output through clap's option handling, but the raw scan still selects the wrong internal printing path.

Switch the scans to `rposition()` and add regression coverage for all four combinations. The existing rule that `-b` takes precedence over `-n` remains unchanged.

## Tests

- `cargo test --test integration_tests`
- `cargo fmt --all -- --check`


cuishuang:
@keith-hall Would you mind taking a look at this follow-up to the combined short-flag handling?

The raw argument scanner currently uses the first occurrence of -p, -n, and -b; repeated flags such as -npn, -pbp, and -bpb therefore select the wrong internal printing path. This switches the scan to the last occurrence and adds regression coverage for those combinations.

Thanks!

keith-hall (approved):
Thanks

PR #3912, opened 2026-09-04
Allow boolean flags to be repeated

logarithmone1128:
bat prepends the arguments from its config file to the command line. So a user who keeps `--show-all` in `~/.config/bat/config` and then types `bat -A file` passes the flag twice, and clap rejects it:

```console
$ printf -- '--show-all\n' > ~/.config/bat/config
$ bat -A src/main.rs
error: the argument '--show-all' cannot be used multiple times
$ echo $?
2
```

The file is never displayed. `--plain` and `--number` do not have this problem, because they already override themselves; `--show-all` and seven others do not.

Measured on `master` and this branch, once through the config file and once by passing the flag twice directly:

| flag | master: config + CLI | master: CLI twice | this branch |
|---|---|---|---|
| `--show-all` | error | error | ok |
| `--diff` | error | error | ok |
| `--chop-long-lines` | error | error | ok |
| `--squeeze-blank` | error | error | ok |
| `--unbuffered` | error | error | ok |
| `--quiet-empty` | error | error | ok |
| `--no-custom-assets` | error | error | ok |
| `--set-terminal-title` | error | error | ok |

The fix is `.overrides_with("<self>")`, which is what `plain`, `number`, `number-nonblank`, `force-colorization` and `no-paging` already do.

Two things I checked rather than assumed, since `overrides_with` on a `SetTrue` flag could plausibly make the second occurrence turn the first one off, or interfere with the existing conflict rules:

- Giving a flag twice produces **byte-identical output** to giving it once, for `-A`, `-s`, `-d` and `--chop-long-lines`. The flag still takes effect.
- `-A -l rust` still exits 2, so `--show-all`'s `conflicts_with("language")` is unaffected.

Single-use output is byte-identical to `master` for every flag touched.

The two added tests fail on `master` — I checked by reverting only the `clap_app.rs` change and keeping them — and pass with it.

`cargo test` passes: 146 lib + 258 integration + the rest, 0 failures.

I scoped this to the flags that change how a file is displayed. The same is true of `--list-themes`, `--list-languages`, `--acknowledgements`, `--config-dir` and `--cache-dir`, but those print something and exit, so nobody would have them in a config file. Happy to include them if you would rather have it uniform.

One thing I could not verify locally: `cargo fmt --check`, because the rustfmt component will not download here. The change is eight copies of the same single line, placed to match the lines around them.


keith-hall (approved):
Thanks

